PR #27822, opened 2024-06-07
Use data file interface

### LINKED ISSUE #27812: Use data file interface in NEML2 action
### Motivation

The NEML2 action loads a NEML2 input file. We're planning to distribute those inputs with our applications. They define constitutive models and won't have to be modified by the average user (similar to a material class source file). It would be nice if we could easily use those NEML2 inputs with level 1 binary installations.

### Design

Utilize the datafile interface to attempt retrieval from the installed app's data directory if a local relative path cannot be resolved.

### Impact

Enable effective use of NEML2 from binary only installs.

dschwen:
Closes #27812


dschwen:
This should not change existing behavior. The new method still prefers local files and only falls back to searching in the data file directories if no other copy is found.

loganharbour (approved):

dschwen:
The cause of the Bison failure is known and unrelated to this PR.

PR #27856, opened 2024-06-10
Improvements to damped nested KKS model in MOOSE

### LINKED ISSUE #27855: Improvements to damped nested KKS model in MOOSE
### Motivation

1. Parsed materials do not have a way to specify upstream dependencies on other material properties. This causes issues when we use `compute=false` materials in nested phase-field models that depend on other `compute=false` materials.
2. Epsilon parameter in function parser is a globally set variable currently, which causes fuzzy comparisons (>,<,=). This results in spurious results from the damped nested solve.
3. Nested KKS multiphase models use a copy material property of the free energies and their derivatives. This turns out to be unnecessary and doubles memory overhead from the nested solves for no reason. 

### Design

1. Add a parameter to [ParsedMaterialHelper](https://github.com/idaholab/moose/blob/next/framework/include/materials/ParsedMaterialHelper.h) for upstream material properties that need to be updated prior to evaluating current material property.
2. Add a parameter in [FunctionParserUtils](https://github.com/idaholab/moose/blob/next/framework/include/utils/FunctionParserUtils.h) that allows setting the epsilon value for a particular material.
3. Remove the copy variables of free energies and free energy derivatives from the multiphase nested KKS material and kernel classes.

### Impact

This will improve the nested KKS solver capability in MOOSE for use in complex multiphase systems with log free energies. It also could have use in other cases where we run into errors from fuzzy fparser comparisons or `compute=false` materials.

chaibhave:
#closes #27855

### DIFF
--- a/framework/include/materials/ParsedMaterialHelper.h
+++ b/framework/include/materials/ParsedMaterialHelper.h
@@ -121,14 +121,17 @@ protected:
    */
   const VariableNameMappingMode _map_mode;
 
-  // Vector to hold list of material names that must be updated prior to evaluating current material
-  // (for compute = false materials)
+  /**Vector to hold list of material names that must be updated prior to evaluating current material
+  * (for compute = false materials)
+  */
   std::vector<MaterialName> _upstream_mat_names;
 
   /// This is true by default, but can be disabled to make non-existing properties default to zero
   const bool _error_on_missing_material_properties;
 
-  // Vector to hold list of materials that must be updated prior to evaluating current material (for
-  // compute = false materials)
+  /**
+   *  Vector to hold list of materials that must be updated prior to evaluating current material (for
+   * compute = false materials)
+   */
   std::vector<MaterialBase *> _upstream_mat;
 };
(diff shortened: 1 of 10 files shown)


framework/include/materials/ParsedMaterialHelper.h
```diff
@@ -83,6 +83,7 @@ class ParsedMaterialHelper : public FunctionMaterialBase<is_ad>, public Function
 
   void initQpStatefulProperties() override;
   void computeQpProperties() override;
+  void initialSetup() override final;
```
lindsayad, line 86: Even though `override` implies `virtual`, I've been adding `virtual` in front lately since it helps when reading the resulting doxygen

framework/include/materials/ParsedMaterialHelper.h
```diff
@@ -120,6 +121,17 @@ class ParsedMaterialHelper : public FunctionMaterialBase<is_ad>, public Function
    */
   const VariableNameMappingMode _map_mode;
 
+  /**Vector to hold list of material names that must be updated prior to evaluating current material
```
lindsayad, line 124: put text on new line

chaibhave: Fixed.

framework/include/materials/ParsedMaterialHelper.h
chaibhave: Done.

lindsayad:
I leave the meaty review for @dschwen

dschwen (approved):

Issue #26718, opened 2024-02-02
Add GPU support for libtorch within MOOSE


grmnptr:
## Reason
MOOSE has been relying on CPUs, but GPUs can considerably accelerate libtorch-based neural network evaluations on both Mac and Linux distributions. This would reduce expensive data-driven material model evaluations.

## Design
The main points:
- Transitioning to conda-based packages on Macs and adding MPS support on ARMs
- Creating a CUDA environment that can be used both through conda and in a container for Linux boxes
- Option to select the used device within the App combined with automatic device-detection routines.

## Impact
Will accelerate the training and evaluation processes of libtorch-base neural networks.

lindsayad:
Need reason for closing

grmnptr:
Closed by:
https://github.com/idaholab/moose/pull/27890
https://github.com/idaholab/moose/pull/28114

PR #27890, opened 2024-06-12
Add apptainer infrastructure to test with CPU/GPU libtorch

grmnptr:
Refs #26718

### DIFF
--- a/apptainer/moose-dev.def
+++ b/apptainer/moose-dev.def
@@ -24,8 +24,6 @@
 {%- set WASP_DIR = '/opt/wasp' -%}
 {#- The installation location for libtorch                                                    -#}
 {%- set LIBTORCH_DEST = '/opt' -%}
-{#- The version of libtorch to be defined                                                     -#}
-{%- set LIBTORCH_VER = '2.1.0' -%}
 
 {#- The script used to install wasp                                                           -#}
 {%- set WASP_BUILD_SCRIPT = 'update_and_rebuild_wasp.sh' -%}
@@ -60,9 +58,8 @@ From: {{ APPTAINER_FROM }}
 
     # Libtorch-related vars
     LIBTORCH_DEST={{ LIBTORCH_DEST }}
-    LIBTORCH_VER={{ LIBTORCH_VER }}
+    LIBTORCH_VER=2.1.0
     LIBTORCH_DISTRIBUTION={{ WITH_LIBTORCH or "cpu" }}
-    CUDA_PREFIX="cu"
 
     # Pinned versions
     MINIFORGE_VERSION=23.3.1-1
@@ -80,7 +77,7 @@ From: {{ APPTAINER_FROM }}
 
     # We install CUDA Toolkit if the user wants cuda-based libtorch.
     # Right now this assumes that cuda-based distributions start with -cu-
-    if [[ ${LIBTORCH_DISTRIBUTION} == $CUDA_PREFIX* ]]; then
+    if [[ ${LIBTORCH_DISTRIBUTION} == cu* ]]; then
         dnf -y install epel-release
         CUDA_RPM=${BUILD_DIR}/cuda.rpm
         curl -L https://developer.download.nvidia.com/compute/cuda/11.4.0/local_installers/cuda-repo-rhel8-11-4-local-11.4.0_470.42.01-1.x86_64.rpm -o ${CUDA_RPM}
--- a/apptainer/moose-dev.def
+++ b/apptainer/moose-dev.def
@@ -83,7 +83,7 @@ From: {{ APPTAINER_FROM }}
     LIBTORCH_DISTRIBUTION={{ WITH_LIBTORCH }}
 
     # Install libtorch from a precompiled package
-    ${ROOT_BUILD_DIR}/setup_libtorch.sh --version=${LIBTORCH_VER} --libtorch-dest=${LIBTORCH_DEST} --libtorch-distribution=${LIBTORCH_DISTRIBUTION}
+    ${ROOT_BUILD_DIR}/setup_libtorch.sh --version=${LIBTORCH_VER} --libtorch-dest=${LIBTORCH_DEST} --libtorch-distribution=${LIBTORCH_DISTRIBUTION} --cleanup
 
 {%- if WITH_LIBTORCH.startswith('cu') %}
     # We install CUDA Toolkit if the user wants cuda-based libtorch.
--- a/apptainer/moose-dev.def
+++ b/apptainer/moose-dev.def
@@ -9,7 +9,12 @@
 {#- CONDA_ALT_CHANNEL: An alternate channel to add to conda                                   -#}
 {#- MOOSE_TOOLS_VERSION: The version to pin moose-tools to (if any)                           -#}
 {#- MOOSE_JOBS: Number of jobs to pass to the builds                                          -#}
-{#- WITH_LIBTORCH: The libtorch distribution to use (e.g. cpu, cu121, etc)                    -#}
+{#- WITH_LIBTORCH: The libtorch distribution to use (cpu, cu121, etc); default: cpu           -#}
+
+{#- Set WITH_LIBTORCH default if not set                                                      -#}
+{%- if WITH_LIBTORCH is not defined %}
+{%- set WITH_LIBTORCH = 'cpu' %}
+{%- endif %}
 
 {#- The within-container build directory to use                                               -#}
 {%- set ROOT_BUILD_DIR = '/root/build' -%}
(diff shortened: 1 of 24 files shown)


loganharbour (changes requested):
- Need to update versioner.yaml to include `setup_libtorch.sh` as a dependency for `moose-dev`.
- Need to update `moose-dev` conda since the versions have now changed

framework/src/base/MooseApp.C
```diff
@@ -321,6 +321,9 @@ MooseApp::validParams()
   params.addParam<bool>(
       "automatic_automatic_scaling", false, "Whether to turn on automatic scaling by default.");
 
+  params.addCommandLineParam<std::string>(
+      "libtorch_device", "--libtorch_device", "cpu", "The device type we want to run libtorch on.");
```
loganharbour, line 325:
```suggestion
      "libtorch_device", "--libtorch-device", "cpu", "The device type we want to run libtorch on.");
```

framework/src/base/MooseApp.C
```diff
@@ -321,6 +321,9 @@ MooseApp::validParams()
   params.addParam<bool>(
       "automatic_automatic_scaling", false, "Whether to turn on automatic scaling by default.");
 
+  params.addCommandLineParam<std::string>(
```
loganharbour, line 324: The type `MooseEnum` is supported for command line parameters now. You should set your enum here so you get a better error when a non-valid choice is provided.

apptainer/moose-dev.def
```diff
@@ -21,6 +22,10 @@
 
 {#- The installation location for wasp                                                        -#}
 {%- set WASP_DIR = '/opt/wasp' -%}
+{#- The installation location for libtorch                                                    -#}
+{%- set LIBTORCH_DEST = '/opt' -%}
+{#- The version of libtorch to be defined                                                     -#}
+{%- set LIBTORCH_VER = '2.1.0' -%}
```
loganharbour, line 28: Put this as a variable in `%post` instead, we only use jinja variables like this for things that need to be used in both `%post` and `%env`

apptainer/moose-dev.def
```diff
@@ -49,6 +58,12 @@ From: {{ APPTAINER_FROM }}
     WASP_GIT_REMOTE={{ WASP_GIT_REMOTE }}
     export MOOSE_JOBS={{ MOOSE_JOBS }}
 
+    # Libtorch-related vars
+    LIBTORCH_DEST={{ LIBTORCH_DEST }}
+    LIBTORCH_VER={{ LIBTORCH_VER }}
+    LIBTORCH_DISTRIBUTION={{ WITH_LIBTORCH or "cpu" }}
+    CUDA_PREFIX="cu"
```
loganharbour, line 65: This is only used once, just go use it in the call so it's easier to follow:

```
if [[ ${LIBTORCH_DISTRIBUTION} == cu* ]]; the
```

framework/include/base/MooseApp.h
```diff
@@ -950,6 +963,16 @@ class MooseApp : public ConsoleStreamInterface,
...
   void removeRelationshipManager(std::shared_ptr<RelationshipManager> relationship_manager);
 
+#ifdef LIBTORCH_ENABLED
+  /**
+   * Function to determine the device which should be used by libtorch on this
+   * application. We use this function to decide what is available on different
+   * builds.
+   * @param device String to describe if a cpu or a gpu should be used.
+   */
+  torch::DeviceType determineLibtorchDeviceType(const std::string & device);
```
loganharbour, line 973:
```suggestion
  torch::DeviceType determineLibtorchDeviceType(const std::string & device) const;
```

Should also be private because it should really only be called within the constructor

framework/include/base/MooseApp.h
```diff
@@ -76,6 +84,11 @@ class MooseApp : public ConsoleStreamInterface,
                  public MooseBase
 {
 public:
+#ifdef LIBTORCH_ENABLED
+  /// Get the device torch is supposed to be running on.
+  torch::DeviceType getLibtorchDevice() { return _libtorch_device; }
```
loganharbour, line 89:
```suggestion
  torch::DeviceType getLibtorchDevice() const { return _libtorch_device; }
```

framework/include/base/MooseApp.h
grmnptr: It is private

framework/src/base/MooseApp.C
grmnptr: noice, thx

python/TestHarness/TestHarness.py
```diff
@@ -581,6 +588,11 @@ def appendRecoverableTests(self, testers):
         testers.extend(new_tests)
         return testers
 
+    def addLibtorchGPUCLIParam(self, testers):
```
loganharbour, line 591: Don't do this here. Just add it to `RunApp.getCommand()`

loganharbour (changes requested):

python/TestHarness/TestHarness.py
grmnptr: Don't you need to remove this then?

loganharbour: removed

grmnptr: noice

framework/src/libtorch/utils/LibtorchArtificialNeuralNet.C
```diff
@@ -174,6 +185,12 @@ dataStore<Moose::LibtorchArtificialNeuralNet>(
 
   dataStore(stream, items, context);
 
+  int8_t device_type = static_cast<std::underlying_type<torch::DeviceType>::type>(nn->deviceType());
+  dataStore(stream, device_type, context);
+
+  int8_t data_type = static_cast<std::underlying_type<torch::ScalarType>::type>(nn->dataType());
```
lindsayad, line 191: It seems odd to do a `static_cast` with generic code and then potentially do implicit conversions to `int8_t`. Can you use `auto` instead?

framework/src/libtorch/utils/LibtorchArtificialNeuralNet.C
```diff
@@ -205,8 +222,16 @@ dataLoad<Moose::LibtorchArtificialNeuralNet>(
   activation_functions.resize(num_activation_items);
   dataLoad(stream, activation_functions, context);
 
+  int8_t device_type;
+  dataLoad(stream, device_type, context);
+  const torch::DeviceType divt(static_cast<torch::DeviceType>(device_type));
+
+  int8_t data_type;
```
lindsayad, line 229: Can you use `std::underlying_type<torch::DeviceType>::type` instead of `int8_t`? Also interesting that you don't need a `typename` in front of `std::underlying_type<torch::DeviceType>::type` 🤔

loganharbour (changes requested):
Still need to bump moose-dev, which will definitely conflict due to https://github.com/idaholab/moose/pull/28046

framework/src/libtorch/utils/LibtorchArtificialNeuralNet.C
grmnptr: Yeah well, this is the reason I used `int8_t`:
```
enum class ScalarType : int8_t {
#define DEFINE_ST_ENUM_VAL_(_1, n) n,
  AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS(DEFINE_ST_ENUM_VAL_)
#undef DEFINE_ENUM_ST_ENUM_VAL_
      Undefined,
  NumOptions
};
```
But yeah this should work with out for sure too. I modified it to auto.

framework/src/libtorch/utils/LibtorchArtificialNeuralNet.C
grmnptr: Added the mods!

loganharbour (approved):
WIP recipes are unrelated and not ran in production. Running recipes have succeeded in the past, with now unrelated changes.

grmnptr (commented):
Nice, thanks for the help @loganharbour ! I approve your changes.

scripts/versioner.yaml
milljm, line 1: 👍  approved

scripts/setup_libtorch.sh
milljm, line 1: 👍  approved

python/TestHarness/testers/RunApp.py
milljm, line 1: 👍  approved

python/TestHarness/TestHarness.py
milljm, line 1: 👍  approved

apptainer/moose-dev.def
milljm, line 1: 👍  approved

conda/moose-dev/meta.yaml
milljm, line 1: 👍  approved

conda/moose/conda_build_config.yaml
milljm, line 1: 👍  approved

framework/doc/packages_config.yml
milljm, line 1: 👍  approved

milljm (approved):
For my part, this looks good!

PR #26348, opened 2023-12-15
Implement RestoringInterface

hugary1995:
This interface allows objects to customize the behavior during a problem restore event.

close #26347

PR #27905, opened 2024-06-14
Allow NEML2 material models to work correctly when an FEM solve fails

### LINKED ISSUE #26347: SolveFailedInterface
## Reason

If an object modifies the "state" of the problem other than stateful properties and adaptivity, there is currently no interface to notify an object about a solve failure event.

## Design

Mimic the `MeshChangedInterface` and add a `SolveFailedInterface`.

## Impact

This will add new API for objects that want to restore the problem state in a solve failure event.

### LINKED ISSUE #27904: NEML2 material model does not handle solve failure event correctly
### Bug Description

When an FEM solve fails, NEML2 material model does not handle the roll-back of solution state correctly.

### Steps to Reproduce

This can be easily seen by manually failing a time step using the Terminator.

### Impact

This prevents us from getting the correct solution when a solve fails (or fingers crossed the solve never fails).

### [Optional] Diagnostics

_No response_

hugary1995:
- Add a SolveFailedInterface which is very similar to MeshChangedInterface.
- Handle the solve failure event in ExecuteNEML2Model by rolling back stateful data.

close #27904 
close #26347


hugary1995:
Since this partly replicates PR #26348, I'll just copy over @loganharbour 's comments:

> Fundamentally I don't think that I agree with this. Restart is when things are allowed to change. There should be minimal conditional logic in restore.

In this PR, I constrained the use case to be only "handling stuff when a solve fails" instead of the restore event.

> Regardless, please use MooseApp::registerInterfaceObject() and MooseApp::getInterfaceObjects() instead of creating a new variable.

I'll remember to do that! Let me first see if this breaks other stuff.

dschwen:
Converting to draft as this still needs to be changed to use the registerInterfaceObject stuff

hugary1995:
Adapted to use MooseApp::registerInterfaceObject() and MooseApp::getInterfaceObjects().

hugary1995:
Closed in favor of a more logical approach: #27966

PR #27910, opened 2024-06-16
Add file socket support to `WebServerControl`

### LINKED ISSUE #27909: Use file sockets for the `WebServerControl`
### Motivation

The `WebServerControl` currently only works with ports. This has a few disadvantages:

- It's hard to find available ports concurrently, which means that testing can currently fail non-deterministically
- Running on a port is a problem with shared resources, because it means technically others can control your simulation

We should also allow the option to provide a unix file socket as the control point.

### Design

Add the option to run via a file socket within tinyhttp, and use this file socket for the majority of testing.

### Impact

Should produce deterministic testing and add separation to the control object.

loganharbour:
Closes https://github.com/idaholab/moose/issues/27909

### DIFF
--- a/framework/contrib/tinyhttp/include/tinyhttp/http.h
+++ b/framework/contrib/tinyhttp/include/tinyhttp/http.h
@@ -54,6 +54,7 @@
 #include <fstream>
 #include <list>
 #include <chrono>
+#include <variant>
 
 #ifdef TINYHTTP_THREADING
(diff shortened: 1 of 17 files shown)


dschwen (approved):
Looks fiiiine

loganharbour:
@dschwen need one more - macs weren't happy with super long file paths for unix sockets it seems

PR #27948, opened 2024-06-19
Give objects the ability to fine-tune petsc options

pbehne:
Currently, petsc options are handled in the `MOOSE::PetscSupport` namespace in `PetscSupport.C`. The functions defined here set petsc options for various general cases. However, these general settings are not applicable to some cases and can result in unused petsc option warmings.

This PR is the first step to address #28053. A capability is introduced such that various moose objects can specify exceptions to the general cases in `PetscSupport`. These exceptions are given as a list of options that are to be ignored, or they would otherwise cause unused option warnings. With this design, special cases that cause warnings can be addressed by ignoring the irrelevant petsc options. An example of how to use this new features is given in `ExplicitTimeIntegrator`. With this change, the unused option warnings that appeared when running `test/tests/time_integrators/actually_explicit_euler_verification/ee-1d-linear.i` disappear.

If we want to go with this proposed design, the test suite should be examined for all occurrences of unused option warnings and these options should be turned off for those tests. This examination can be easily conducted by setting a default value for the `absent_out` parameter in the `RunApp` tester to identify relevant tests that have unused option warnings.

### DIFF
--- a/framework/include/utils/PetscSupport.h
+++ b/framework/include/utils/PetscSupport.h
@@ -38,13 +38,16 @@ namespace PetscSupport
 class PetscOptions
 {
 public:
-  PetscOptions() : flags("", "", true) {}
+  PetscOptions() : flags("", "", true), ignored_flags("", "", true) {}
 
   /// PETSc key-value pairs
   std::vector<std::pair<std::string, std::string>> pairs;
 
   /// Single value PETSc options (flags)
   MultiMooseEnum flags;
+
+  /// Flags to explicitly not set, even if they are specified in flags
+  MultiMooseEnum ignored_flags;
 
   /// Preconditioner description
   std::string pc_description;
@@ -212,6 +215,19 @@ void colorAdjacencyMatrix(PetscScalar * adjacency_matrix,
                           const char * coloring_algorithm);
 
 /**
+ * disable a particular petsc flag
+ */
+void disablePetscFlag(const std::string & flag, PetscOptions & petsc_options);
+
+/**
+ * disable "ignored" petsc flags
+ * @param ignored_petsc_flags Container holding the flags of the petsc options to be ignored
+ * @param petsc_options Data structure which handles petsc options within moose
+ */
+void disableIgnoredPetscFlags(const MultiMooseEnum & ignored_petsc_flags,
+                              PetscOptions & petsc_options);
+
+/**
  * disable printing of the nonlinear convergence reason
  */
 void disableNonlinearConvergedReason(FEProblemBase & fe_problem);
@@ -220,6 +236,16 @@ void disableNonlinearConvergedReason(FEProblemBase & fe_problem);
  * disable printing of the linear convergence reason
  */
 void disableLinearConvergedReason(FEProblemBase & fe_problem);
+
+/**
+ * disable common ksp flags
+ */
+void disableCommonKSPFlags(FEProblemBase & fe_problem);
+
+/**
+ * disable common snes flags
+ */
+void disableCommonSNESFlags(FEProblemBase & fe_problem);
 
 #define SNESGETLINESEARCH SNESGetLineSearch
 }
--- a/framework/include/utils/PetscSupport.h
+++ b/framework/include/utils/PetscSupport.h
@@ -202,7 +202,7 @@ void setSinglePetscOption(const std::string & name,
 
 /**
  * Same as setSinglePetscOption, but does not set the option if it doesn't make sense for the
- * current simulation type.
+ * current simulation type, e.g. if \p name is contained within \p dont_add_these_options
  */
 void setSinglePetscOptionIfAppropriate(const MultiMooseEnum & dont_add_these_options,
                                        const std::string & name,
@@ -234,7 +234,7 @@ void colorAdjacencyMatrix(PetscScalar * adjacency_matrix,
                           const char * coloring_algorithm);
 
 /**
- * Function to ensure that particular petsc option is not added to the PetscOptions
+ * Function to ensure that a particular petsc option is not added to the PetscOptions
  * storage object to be later set unless explicitly specified in input or on the command line.
  */
 void dontAddPetscFlag(const std::string & flag, PetscOptions & petsc_options);
--- a/framework/include/utils/PetscSupport.h
+++ b/framework/include/utils/PetscSupport.h
@@ -107,6 +107,12 @@ void outputNorm(libMesh::Real old_norm, libMesh::Real norm, bool use_color = fal
 PetscErrorCode petscLinearMonitor(KSP /*ksp*/, PetscInt its, PetscReal rnorm, void * void_ptr);
 
 /**
+ * Process some MOOSE-wrapped PETSc options.
+ */
+void processSingletonMooseWrappedOptions(FEProblemBase & fe_problem,
+                                         const InputParameters & params);
+
+/**
  * Stores the PETSc options supplied from the InputParameters with MOOSE
  */
 void storePetscOptions(FEProblemBase & fe_problem, const InputParameters & params);
--- a/framework/include/utils/PetscSupport.h
+++ b/framework/include/utils/PetscSupport.h
@@ -138,7 +138,7 @@ void storePetscOptionsFromParams(PetscOptions & po,
  * @param petsc_flags Container holding the flags of the petsc options
  * @param petsc_options Data structure which handles petsc options within moose
  */
-void AddPetscFlagsToPetscOptions(const MultiMooseEnum & petsc_flags, PetscOptions & petsc_options);
+void addPetscFlagsToPetscOptions(const MultiMooseEnum & petsc_flags, PetscOptions & petsc_options);
 
 /**
  * Populate name and value pairs in a given PetscOptions object using vectors of input arguments
@@ -146,7 +146,7 @@ void AddPetscFlagsToPetscOptions(const MultiMooseEnum & petsc_flags, PetscOption
  * @param mesh_dimension The mesh dimension, needed for multigrid settings
  * @param petsc_options Data structure which handles petsc options within moose
  */
-void AddPetscPairsToPetscOptions(
+void addPetscPairsToPetscOptions(
     const std::vector<std::pair<MooseEnumItem, std::string>> & petsc_pair_options,
     const unsigned int mesh_dimension,
     PetscOptions & petsc_options);
--- a/framework/include/utils/PetscSupport.h
+++ b/framework/include/utils/PetscSupport.h
@@ -107,7 +107,7 @@ void storePetscOptions(FEProblemBase & fe_problem, const InputParameters & param
  * @param petsc_flags Container holding the flags of the petsc options
  * @param petsc_options Data structure which handles petsc options within moose
  */
-void processPetscFlags(const MultiMooseEnum & petsc_flags, PetscOptions & petsc_options);
+void AddPetscFlagsToPetscOptions(const MultiMooseEnum & petsc_flags, PetscOptions & petsc_options);
 
 /**
  * Populate name and value pairs in a given PetscOptions object using vectors of input arguments
@@ -115,10 +115,10 @@ void processPetscFlags(const MultiMooseEnum & petsc_flags, PetscOptions & petsc_
  * @param mesh_dimension The mesh dimension, needed for multigrid settings
  * @param petsc_options Data structure which handles petsc options within moose
  */
-void
-processPetscPairs(const std::vector<std::pair<MooseEnumItem, std::string>> & petsc_pair_options,
-                  const unsigned int mesh_dimension,
-                  PetscOptions & petsc_options);
+void AddPetscPairsToPetscOptions(
+    const std::vector<std::pair<MooseEnumItem, std::string>> & petsc_pair_options,
+    const unsigned int mesh_dimension,
+    PetscOptions & petsc_options);
 
 /**
  * Returns the valid petsc line search options as a set of strings
--- a/framework/include/utils/PetscSupport.h
+++ b/framework/include/utils/PetscSupport.h
@@ -38,7 +38,10 @@ namespace PetscSupport
 class PetscOptions
 {
 public:
-  PetscOptions() : flags("", "", true), ignored_flags("", "", true) {}
+  PetscOptions()
+    : flags("", "", true), ignored_options("", "", true), user_set_options("", "", true)
+  {
+  }
 
   /// PETSc key-value pairs
   std::vector<std::pair<std::string, std::string>> pairs;
@@ -47,7 +50,10 @@ public:
   MultiMooseEnum flags;
 
   /// Flags to explicitly not set, even if they are specified in flags
-  MultiMooseEnum ignored_flags;
+  MultiMooseEnum ignored_options;
+
+  /// Options that are set by the user at the input level
+  MultiMooseEnum user_set_options;
 
   /// Preconditioner description
   std::string pc_description;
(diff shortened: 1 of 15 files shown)


framework/src/parser/CommandLine.C
```diff
@@ -414,6 +416,10 @@ CommandLine::printUsage() const
 
   Moose::out << "Solver Options:\n"
              << "  See PETSc manual for details" << std::endl;
+
+  // If we get here, we are not running a simulation and should silence petsc's unused options
+  // warning
+  Moose::PetscSupport::setSinglePetscOption("-options_left", "0");
```
pbehne, line 422: Changes in this file are unrelated to the new `ignored_options` attribute in `PetscOptions`, but having this here does get rid of some additional unused option warnings.

framework/src/timeintegrators/ExplicitTimeIntegrator.C
```diff
@@ -53,6 +53,9 @@ ExplicitTimeIntegrator::ExplicitTimeIntegrator(const InputParameters & parameter
 
   if (_solve_type == LUMPED || _solve_type == LUMP_PRECONDITIONED)
     _ones = addVector("ones", false, PARALLEL);
+
+  // ignore all common SNES-related petsc options to prevent unused option warnings
+  Moose::PetscSupport::ignoreCommonSNESOptions(_fe_problem);
```
pbehne, line 58: This change shows how the new `ignored_options` attribute in `PetscOptions` is utilized to disable options specific to the `ExplicitTimeIntegrator` that are not used.

pbehne:
@lindsayad, would you mind reviewing this PR to ensure that this new feature is how we want to handle unused petsc options? Confirmation on this would be good to have before I hunt down all occurrences of the warnings in the test suite. If not you, could you tag someone you think would be a good reviewer?

I'm thinking we split issue #28053 up. After merging this PR, we can have another PR that suppresses unused options in the test suite.

framework/include/utils/PetscSupport.h
```diff
@@ -38,14 +38,23 @@ namespace PetscSupport
...
+  }
 
   /// PETSc key-value pairs
   std::vector<std::pair<std::string, std::string>> pairs;
 
   /// Single value PETSc options (flags)
   MultiMooseEnum flags;
 
+  /// Flags to explicitly not set, even if they are specified in flags
+  MultiMooseEnum ignored_options;
```
lindsayad, line 53: I would prefer that we generally replace `ignored` with something like `dont_add_these`. Ignoring something scares me. It sounds like someone has asked for something but we're going to ignore it. "Don't add this" sounds a lot less scary and also reflects more truly what your code is doing: it's making it so that we don't add particular flags that don't make sense for a given simulation as opposed to "we're going to add these flags that don't make sense and then ignore them"

framework/include/utils/PetscSupport.h
```diff
@@ -153,6 +196,15 @@ void setSinglePetscOption(const std::string & name,
                           const std::string & value = "",
                           FEProblemBase * const problem = nullptr);
 
+/**
+Same as setSinglePetscOption, but does not set the option if it belongs to the given list of
+ignored options.
+*/
+void setSinglePetscOptionIfNotIgnored(const MultiMooseEnum & ignored_options,
```
lindsayad, line 203:
```suggestion
void setSinglePetscOptionIfAppropriate(const MultiMooseEnum & ignored_options,
```
or something like that

framework/include/utils/PetscSupport.h
```diff
@@ -153,6 +196,15 @@ void setSinglePetscOption(const std::string & name,
                           const std::string & value = "",
                           FEProblemBase * const problem = nullptr);
 
+/**
+Same as setSinglePetscOption, but does not set the option if it belongs to the given list of
+ignored options.
```
lindsayad, line 201:
```suggestion
Same as setSinglePetscOption, but does not set the option if it doesn't make sense for the current simulation type
```

framework/include/utils/PetscSupport.h
```diff
@@ -187,6 +244,16 @@ void disableNonlinearConvergedReason(FEProblemBase & fe_problem);
  */
 void disableLinearConvergedReason(FEProblemBase & fe_problem);
 
+/**
+ * disable common ksp flags
+ */
+void ignoreCommonKSPOptions(FEProblemBase & fe_problem);
```
lindsayad, line 250:
```suggestion
void dontAddCommonKSPOptions(FEProblemBase & fe_problem);
```

framework/include/utils/PetscSupport.h
```diff
@@ -187,6 +244,16 @@ void disableNonlinearConvergedReason(FEProblemBase & fe_problem);
...
+
+/**
+ * disable common snes flags
+ */
+void ignoreCommonSNESOptions(FEProblemBase & fe_problem);
```
lindsayad, line 255:
```suggestion
void dontAddCommonSNESOptions(FEProblemBase & fe_problem);
```

framework/include/utils/PetscSupport.h
```diff
@@ -177,6 +229,11 @@ void colorAdjacencyMatrix(PetscScalar * adjacency_matrix,
                           std::vector<unsigned int> & vertex_colors,
                           const char * coloring_algorithm);
 
+/**
+ * disable a particular petsc flag
```
lindsayad, line 233: More particular language would be good. Does this remove a flag if it exists? Does it tell the `PetscOptions` object to not add the flag if someone attempts to add it later? Both?

lindsayad: We use the disable language repeatedly in the following doxygen. I think if you give a clear definition of what "disable" means here, you do not have to go and add it to all the following doc-strings

lindsayad: If "disable" just means "dontAdd", then let's change the function name to match those that you added later in the file

pbehne: Once a petsc option is already actually set, I don't like to mess with it. My approach is to modify the `PetscOptions` object before options are set. So I think it makes sense to change `disable` to `dontAdd`. What do you like more, `dontAdd` or `dontSet`?

framework/include/utils/PetscSupport.h
```diff
@@ -102,23 +111,45 @@ PetscErrorCode petscLinearMonitor(KSP /*ksp*/, PetscInt its, PetscReal rnorm, vo
...
+
+/**
+ *  Sets the FE problem's MFFD type from the input params.
+ */
+void setMFFDTypeFromParams(MffdType & fe_problem_mffd_type, const InputParameters & params);
+
+/**
+ * Stores the Petsc flags and pair options fron the input params in the given PetscOptions object.
+ */
+void storePetscOptionsFromParams(PetscOptions & po,
```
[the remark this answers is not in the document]
pbehne: Consider the current definition of `storePetscOptions` (which I haven't pushed just yet with the updated signatures):
```
void
storePetscOptions(FEProblemBase & fe_problem, const InputParameters & params)
{
  setSolveTypeFromParams(fe_problem, params);
  setLineSearchFromParams(fe_problem, params);
  setMFFDTypeFromParams(fe_problem, params);
  storePetscOptionsFromParams(fe_problem, params);
}
```
See that I split `storePetscOptions` into 4 subfunctions. I did this because it felt a bit more organized to me. Note that while `storePetscOptionsFromParams` *stores* options in a `PetscOptions` object, the other 3 subfunctions *set* options directly and don't use the `PetscOptions` object. Now that I think about it more, however, these 3 subfunctions are just storing options, there are just dedicated attributes for the storage of solve type, line search, and mffd type outside of a `PetscOptions` object.

Would you prefer I eliminate the subfunctions and just have one long `storePetscOptions` function?

framework/include/utils/PetscSupport.h
lindsayad: Probably `dontAdd`

framework/include/utils/PetscSupport.h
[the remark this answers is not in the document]
lindsayad: Hmm Im starting to realize that this PR is going to have a lot of conflicts with #29072. One way we could remove conflict is if you adopt https://github.com/idaholab/moose/pull/29072/files#diff-813f6769dc958bcf657d6f3422f2bec7ca9d01f4df25a8fbab2905ffbec8fa7bR116. I believe that function is equivalent to the sum of `setSolveTypeFromParams`, `setLineSearchFromParams`, and `setMFFDTypeFromParams`. Then you can simplify the existing `storePetscOptions` function like I did in #29072. 

It would be nice to avoid adding the `storePetscOptionsFromParams` function because as it stands now a developer may think that it does the same thing as `storePetscOptions` when in reality they differ in setting/not-setting solve-type/line-search/mffd-type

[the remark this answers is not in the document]
lindsayad: I am fine with keeping `setSolveTypeFromParams`, `setLineSearchFromParams`, and `setMFFDTypeFromParams` and having those be called from `Moose::PetscSupport::processSingletonMooseWrappedOptions`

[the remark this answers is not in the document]
pbehne: Whoa. There are a lot of conflicts. Let me know if my latest updates are what you wanted, or if you want more than that.

pbehne:
@lindsayad, pinging you again on this so it stays on your list.

framework/include/utils/PetscSupport.h
```diff
@@ -38,14 +38,23 @@ namespace PetscSupport
...
   std::vector<std::pair<std::string, std::string>> pairs;
 
   /// Single value PETSc options (flags)
   MultiMooseEnum flags;
 
+  /// Flags to explicitly not set, even if they are specified programmatically
+  MultiMooseEnum dont_add_these_options;
+
+  /// Options that are set by the user at the input level
+  MultiMooseEnum user_set_options;
```
zachmprince, line 56: There doesn't seem to be anything testing the fact that a warning gets published on a user-defined option. So I would suggest adding something where 1) a nonsense flag is set and 2) a flag that is usually added programmatically, isn't, but is added via input.

zachmprince (commented):
Implementation looks good. I just had one suggestion for a test or two. I'm not sure if there is something already out there, though.

framework/include/utils/PetscSupport.h
pbehne: See latest changes. I added a test to cover 2). 1) seems to be already be covered by `misc/petsc_option_left.test_non_petsc_options_not_left`.

lindsayad, line 56: To be honest I hate that these are MultiMooseEnums since the PETSc options database is constantly changing and there is no hope of enumerating them all in MOOSE. Additionally, I think a `std::set` or `std::vector` would just be a lot clearer. But I am ok with you adding these new data members, following the existing pattern. Fixing this can be follow-on work by someone

framework/include/utils/PetscSupport.h
```diff
@@ -153,6 +200,15 @@ void setSinglePetscOption(const std::string & name,
                           const std::string & value = "",
                           FEProblemBase * const problem = nullptr);
 
+/**
+ * Same as setSinglePetscOption, but does not set the option if it doesn't make sense for the
+ * current simulation type.
```
lindsayad, line 205:
```suggestion
 * current simulation type, e.g. if \p name is contained within \p dont_add_these_options
```

framework/include/utils/PetscSupport.h
```diff
@@ -178,14 +234,34 @@ void colorAdjacencyMatrix(PetscScalar * adjacency_matrix,
                           const char * coloring_algorithm);
 
 /**
- * disable printing of the nonlinear convergence reason
+ * Function to ensure that particular petsc option is not added to the PetscOptions
```
lindsayad, line 237:
```suggestion
 * Function to ensure that a particular petsc option is not added to the PetscOptions
```

framework/src/utils/PetscSupport.C
```diff
@@ -467,10 +503,14 @@ storePetscOptions(FEProblemBase & fe_problem, const InputParameters & params)
...
   }
+}
 
+void
+setLineSearchFromParams(FEProblemBase & fe_problem, const InputParameters & params)
+{
   if (params.isParamValid("line_search"))
   {
-    MooseEnum line_search = params.get<MooseEnum>("line_search");
+    const MooseEnum line_search = params.get<MooseEnum>("line_search");
```
lindsayad, line 513:
```suggestion
    const auto & line_search = params.get<MooseEnum>("line_search");
```

framework/src/utils/PetscSupport.C
```diff
@@ -494,31 +534,20 @@ storePetscOptions(FEProblemBase & fe_problem, const InputParameters & params)
...
     }
   }
+}
 
+void
+setMFFDTypeFromParams(FEProblemBase & fe_problem, const InputParameters & params)
+{
   if (params.isParamValid("mffd_type"))
   {
     MooseEnum mffd_type = params.get<MooseEnum>("mffd_type");
```
lindsayad, line 544:
```suggestion
    const auto & mffd_type = params.get<MooseEnum>("mffd_type");
```

framework/src/utils/PetscSupport.C
```diff
@@ -782,32 +825,64 @@ getPetscValidParams()
...
+getCommonSNESFlags()
 {
   return MultiMooseEnum(
-      "-dm_moose_print_embedding -dm_view -ksp_converged_reason -ksp_gmres_modifiedgramschmidt "
-      "-ksp_monitor -ksp_monitor_snes_lg-snes_ksp_ew -ksp_snes_ew -snes_converged_reason "
-      "-snes_ksp -snes_ksp_ew -snes_linesearch_monitor -snes_mf -snes_mf_operator -snes_monitor "
-      "-snes_test_display -snes_view",
+      "-ksp_monitor_snes_lg -snes_ksp_ew -ksp_snes_ew -snes_converged_reason "
+      "-snes_ksp -snes_linesearch_monitor -snes_mf -snes_mf_operator -snes_monitor "
+      "-snes_test_display -snes_view -snes_monitor_cancel",
```
lindsayad, line 833: yea attempting to keep track of these options is insanity

lindsayad (approved):

framework/src/utils/PetscSupport.C
pbehne: Is there any petsc API that can provide a list of valid options instead of us maintaining these lists?

lindsayad: from @balay:

> I don't think there is a single database to query for all valid options.
> 
> The options are distributed across petsc code - and when that part of the code is run - it checks against the input-options base. [i.e. - if that part of the code is -not- run,there is no check for that option]. Also prefix usage in library or use code can change the options.
> 
> Hence './binary -h' prints only the options that are used/available in  the code path of the currently running code. [and it can't print all]

pbehne: I'm going to mark this as resolved for now to allow the merge. Maybe we will think of a better management system in the future. All of these options were hardcoded in `PetscSupport` before I ever touched it, I just further split them into the ksp and snes categories.

lindsayad:
@pbehne you've managed to trigger something way down in hypre in the failed next branch testing. It's likely going to take me a decent while to figure it out. In the meantime it's probably best if you rewrite your tests and avoid using hypre

loganharbour:
See https://civet.inl.gov/event/217284/

pbehne:
> In the meantime it's probably best if you rewrite your tests and avoid using hypre

On it.

PR #27966, opened 2024-06-22
Store NEML2 stateful data in MOOSE-managed material properties

### LINKED ISSUE #27904: NEML2 material model does not handle solve failure event correctly

hugary1995:
close #27904

## Reason

Currently, we store stateful data of a NEML2 model inside the NEML2 model itself, and advance the stateful data at `timestepSetup`. This implementation saves an additional copying from MOOSE old material properties. However, it has the following drawbacks:
- Upon solve failure, it is not possible revert back stateful data without additional infrastructure such as the one proposed in #27905
- Upon events that change the batch size, such as mesh change, it is not possible to handle material property reprojection and/or stateful material property initialization
- Does not work with checkpoint/restart. Although this could be easily addressed by adding custom DataIO methods.

## Design

Perform the laborious copy from and to MOOSE material property.

Some implementational details
- `NEML2ToMOOSEMaterialProperty` now also accepts an optional material property as the "initial condition" for the corresponding NEML2 variable. This is significant IMO because there has been no generic way of handling this in native MOOSE materials. We could use this to initialize for example the internal variables such as plastic strain.
- In `ExecuteNEML2Model`, I added a check to prevent retrieval of output (and derivative) views after the construction phase.
- `ExecuteNEML2Model::timestepSetup` and `ExecuteNEML2Model::advanceStep` are removed as they are no longer needed.
- In `ExecuteNEML2Model`, `_in_out_allocated` is removed in favor of a more general approach.
- Minor changes in the gatherer UOs and NEML2-to-MOOSE materials.
- A bug fix in `NEML2Utils::set`.
- Adapted all the regression input files. No regolding needed.

## Impact

This addresses all the drawbacks mentioned above.

### DIFF
--- a/modules/solid_mechanics/include/neml2/materials/NEML2ToMOOSEMaterialProperty.h
+++ b/modules/solid_mechanics/include/neml2/materials/NEML2ToMOOSEMaterialProperty.h
@@ -36,11 +36,16 @@ public:
   virtual void computeProperties() override;
 
 protected:
+  virtual void initQpStatefulProperties() override;
+
   /// User object managing the execution of the NEML2 model
   const ExecuteNEML2Model & _execute_neml2_model;
 
   /// Emitted material property
   MaterialProperty<T> & _prop;
+
+  /// Initial condition
+  const MaterialProperty<T> * _prop0;
 
   /// labled view to the requested output
   const neml2::BatchTensor & _output_view;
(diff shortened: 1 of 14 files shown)


hugary1995:
code.ornl.gov is down over the weekend so I'll have to remember to reactivate the tests.

dschwen (approved):
Looks good. Appreciate the improved error checking.

Issue #27974, opened 2024-06-24
FileMeshGenerator and uniform_refine produce weird results


maxnezdyur:
### Bug Description

I had some weird results with an exodus file of mine when refining it. As I increased uniform_refine I got weird results. I found MWE using an exodus file MOOSE already had. Seems like child elements are not C0?

### Steps to Reproduce

Exodus file exists in MOOSE at `modules/solid_mechanics/test/tests/volumetric_locking_verification/42_node_side.e`
``` # Test for volumetric locking correction

# 2D cook's membrane problem with a trapezoid
# that is fixed at one end and is sheared at
# other end. Poisson's ratio is 0.4999.

# Using Quad4 elements and no volumetric locking,
# vertical displacement at top right corner is 3.78
# due to locking.

# Using Quad4 elements with volumetric locking, vertical
# dispalcement at top right corner is 7.78.

# Results match with Nakshatrala et al., Comp. Mech., 41, 2008.

# Check volumetric locking correction documentation for
# more details about this problem.

[GlobalParams]
  displacements = 'disp_x disp_y'
  # volumetric_locking_correction = true
[]

[Mesh]
  [generate]
    type = FileMeshGenerator
    file = 42_node_side.e
  []
  second_order = true
  uniform_refine = 3
[]

[Variables]
  [disp_x]
    order = SECOND
  []
  [disp_y]
    order = SECOND
  []
[]
[Physics]
  [SolidMechanics]
    [QuasiStatic]
      [all]
        strain = SMALL
        incremental = true
      []
    []
  []
[]

[BCs]
  [no_x]
    type = DirichletBC
    variable = disp_x
    boundary = 1
    value = 0.0
  []
  [no_y]
    type = DirichletBC
    variable = disp_y
    boundary = 1
    value = 0.0
  []
  [Nuem]
    type = NeumannBC
    variable = disp_y
    boundary = 2
    value = 6.25
  []
[]

[Materials]
  [elasticity_tensor]
    type = ComputeIsotropicElasticityTensor
    youngs_modulus = 250.0
    poissons_ratio = 0.4999
  []
  [stress]
    type = ComputeFiniteStrainElasticStress
  []
[]

[Preconditioning]
  [SMP]
    type = SMP
    full = true
  []
[]

[Executioner]
  type = Transient
  solve_type = 'NEWTON'

  petsc_options_iname = '-pc_type'
  petsc_options_value = 'lu'

  num_steps = 1
  nl_abs_tol = 1e-5
[...]

PR #27987, opened 2024-06-25
Use non-shell elements in FV test input file

roystgnr:
## Reason

@GiudGiud reports that we ended up with shell elements instead of quad elements here by accident, because it was a Cubit default for 2D output. The elements solve the same way, *except* that with old libMesh a "full-order" `all_second_order()` update to a 4-node shell would produce an 8-node shell, not a 9-node, and we'd like to change that without breaking MOOSE tests.

This should let us get libMesh/libmesh#3886 passing to make that change.

## Design

I generated the new `expansion_quad.e` by using ncdump, changing `SHELL4` to `QUAD4`, deleting 2 from sideset indices (since Exodus quad shell edge-sides are numbered 3-6 instead of 1-4), and using ncgen.

## Impact

We did expect the csv results to change here, since we're now using a biquadratic basis rather than a serendipity basis when the `second_order` conversion adds the internal nodes.

Refs #27974


GiudGiud (approved):

lindsayad (commented):
I've thought about two shell element meshes in two days when I had not thought about them at all in the last two years

PR #27991, opened 2024-06-25
Adding direct acceleration calculation to central difference

TheGreatCid:
This PR is addressing the first half of #28008, which to implement a method that allows for a solution update calculated by the direct calculation of acceleration from the internal and external forces.

## Reason

This is the first step in improving explicit contact as described in #28008 

## Design

The "direct" solution update, is calculated as the following:

$a_n=M^{-1}(F^{ext}-F^{int})$
$v_{n+1/2}=v_{n-1/2}+\Delta t a_n$
$d_{n+1}=d_n + \Delta t v_{n+1/2}$

These steps can be found in `ExplicitTimeIntegrator.C`

The following changes have been made:

- `NonlinearSystemBase.h`
     - Added `setNodalBCs()` member function. This sets nodal BCs for use with the direct time integrator. It is identical to `setInitialSolution()` but without predictor calls. 
- `TimeIntegrator.h`
     - Added function `isDirect()` that returns a bool describing whether or not the direct time integrator is being used
- `ExplicitTimeIntegrator.C`
     - Added additional option for `is_direct` to compute the solution update from acceleration
- `ActuallyExplicitEuler.C`
     - Added a call to `setNodalBCs()` after solution update if using direct solver
- `CentralDifference.C`
     - Added a direct acceleration calculation if in the nonlinear system. This effectively does nothing as is, but will be useful for future work with explicit contact.
     - Added parameter `use_direct` to allow a user to specify if they want to do use a direct acceleration calculation for time integration
- `InertialForce.C`
     - Modified `computeQpResidual()` to have an option to return 0 if using a direct time integrator
     - Modified `computeQpJacobian()` to return the mass matrix without factor tags needed for the current method 

## Impact

This will allow a user to specific `use_direct=true` to invoke a central difference time integrator that computes the solution update based on a direct calculation of acceleration from the internal and external forces.


GiudGiud:
@TheGreatCid are you working with anyone at the national labs on this? To know who to assign for the review

TheGreatCid:
I'm working with Daniel Schwen. I did not realize drafts were public, this is not ready for review yet

GiudGiud:
no worries. Creating drafts is a good idea to get early feedback

TheGreatCid:
Document still needs to be updated, however any code review would be appreciated

TheGreatCid:
Documentation has been added.

framework/src/timeintegrators/CentralDifference.C
```diff
@@ -24,17 +24,22 @@ CentralDifference::validParams()
...
   : ActuallyExplicitEuler(parameters),
     _du_dotdot_du(_sys.duDotDotDu()),
-    _solution_older(_sys.solutionState(2))
+    _solution_older(_sys.solutionState(2)),
+    _use_direct(getParam<bool>("use_direct"))
 {
   if (_solve_type == LUMPED)
     _is_lumped = true;
+  if (_use_direct)
+    _is_direct = true;
```
dschwen, line 42: Why do you have `_is_direct` and `_use_direct`?

framework/src/timeintegrators/CentralDifference.C
```diff
@@ -72,6 +77,12 @@ CentralDifference::computeTimeDerivatives()
     mooseError("CentralDifference: Time derivative of solution (`u_dotdot`) is not stored. Please "
                "set uDotDotRequested() to true in FEProblemBase before requesting `u_dot`.");
 
+  // Don't update time derivate if using reference configuration
+  if (_sys.name() == "nl0" && _is_direct)
+  {
+    return;
+  }
```
dschwen, line 84:
```suggestion
    return;
```

framework/src/timeintegrators/ActuallyExplicitEuler.C
```diff
@@ -36,6 +37,10 @@ ActuallyExplicitEuler::validParams()
 ActuallyExplicitEuler::ActuallyExplicitEuler(const InputParameters & parameters)
   : ExplicitTimeIntegrator(parameters), _constant_mass(getParam<bool>("use_constant_mass"))
 {
+  if (_is_direct && _solve_type != LUMPED)
+  {
+    mooseError("Direct time integration must be used with a lumped mass matrix!");
+  }
```
dschwen, line 43:
```suggestion
    mooseError("Direct time integration must be used with a lumped mass matrix!");
```

framework/src/timeintegrators/ExplicitTimeIntegrator.C
```diff
@@ -116,16 +116,46 @@ ExplicitTimeIntegrator::performExplicitSolve(SparseMatrix<Number> & mass_matrix)
...
+
+        auto accel_scaled = accel.clone();
+
+        // Scaling the acceleration
+        accel_scaled->scale((_dt + _dt_old) / 2);
+
+        // Adding old vel to new vel
+        auto old_vel = _sys.solutionUDotOld();
+        vel += *old_vel;
+        vel += *accel_scaled;
```
dschwen, line 137: Could you do

```suggestion
        auto accel_scaled = accel.clone();

        // Scaling the acceleration
        accel_scaled->scale((_dt + _dt_old) / 2);

        // Adding old vel to new vel
        auto & vel = *_sys.solutionUDot();
        const auto & old_vel = _sys.solutionUDotOld(); // <- avoid a copy here
        vel = *old_vel; // <- assign the old vel, overwriting vel, no need to zero
        vel += *accel_scaled;
```

framework/src/timeintegrators/CentralDifference.C
TheGreatCid: `_is_direct = true` sets the variable defined in `TimeIntegrator.h` that determines whether or not the time integrator is using a direct accel. calculation. `use_direct` is a parameter in `CentralDifference.C` asking the user if they want to use a direct accel. calculation.

framework/src/timeintegrators/ActuallyExplicitEuler.C
TheGreatCid: Addressed by [8230654](https://github.com/idaholab/moose/pull/27991/commits/8230654ac447a3d2f07439e15dd2398fec640e3f)

framework/src/timeintegrators/CentralDifference.C
dschwen: So you can have `_is_direct == true` and `_use_direct == false`? That's a bit confusing. What would be the behavior in that case?

TheGreatCid: It is a bit confusing. You couldn't have that case because `_is_direct` is not set by the user, it is just a flag to show that that the user specified `_use_direct`. 
aaand, writing that out made me realize how silly that is. I'll fix this.

TheGreatCid: Addressed by [fd5f612](https://github.com/idaholab/moose/pull/27991/commits/fd5f61230a5171580af25121f4b5a7dc55630c62)

TheGreatCid: Let me know if you find this fix acceptable

framework/include/timeintegrators/TimeIntegrator.h
```diff
@@ -138,6 +139,11 @@ class TimeIntegrator : public MooseObject, public Restartable
    */
   virtual const bool & isLumped() const { return _is_lumped; }
 
+  /**
+   * Returns whether or not time integrator is using a direct acceleration calculation
+   */
+  virtual const bool & isDirect() const { return _is_direct; }
```
dschwen, line 145:
```suggestion
  virtual const bool & isDirect() const { return false; }
```
and then override in `CentralDifference`

framework/include/timeintegrators/TimeIntegrator.h
```diff
@@ -196,6 +202,9 @@ class TimeIntegrator : public MooseObject, public Restartable
   /// Boolean flag that is set to true if lumped mass matrix is used
   bool _is_lumped;
 
+  /// Boolean flag that is set to true if direct acceleration calculation is used
+  bool _is_direct;
+
```
dschwen, line 207:
```suggestion
```

move to `CentralDifference`

framework/include/timeintegrators/TimeIntegrator.h
dschwen: ugh, this might be a reference to a temporary. But for a `bool` this should work...

dschwen: nah, it gives a warning. :-(

framework/include/timeintegrators/TimeIntegrator.h
dschwen: let's leave this here then.

framework/src/timeintegrators/TimeIntegrator.C
```diff
@@ -42,6 +42,7 @@ TimeIntegrator::TimeIntegrator(const InputParameters & parameters)
     _n_nonlinear_iterations(0),
     _n_linear_iterations(0),
     _is_lumped(false),
+    _is_direct(isParamValid("use_direct") ? getParam<bool>("use_direct") : false),
```
dschwen, line 45:
```suggestion
    _is_direct(false),
```

move `_is_direct = getParam<bool>("use_direct");` into CentralDifference

framework/include/timeintegrators/TimeIntegrator.h
TheGreatCid: What is the reasoning for having this function return false? It should return `_is_direct` which is a bool that stores whether or not we are using the direct method.

framework/include/timeintegrators/TimeIntegrator.h
TheGreatCid: > let's leave this here then.

Should I ignore the below suggestion to move `_is_direct` into central difference?

framework/include/timeintegrators/TimeIntegrator.h
lindsayad: I don't see any reason to return a reference here.

I agree with the general line of @dschwen's thought. I don't think an `_is_direct` member should be introduced into the inheritance tree until you reach a place where it could always be applied, e.g. I don't want this `isParamValid` check in the `TimeIntegrator` constructor:

```c++
    _is_direct(isParamValid("use_direct") ? getParam<bool>("use_direct") : false)
```

So if you need this public API on the base class then default it to return `false` like @dschwen suggests

lindsayad (changes requested):

framework/include/timeintegrators/TimeIntegrator.h
TheGreatCid: Got it, ill make that change

framework/include/systems/NonlinearSystemBase.h
```diff
@@ -292,6 +292,8 @@ class NonlinearSystemBase : public SolverSystem, public PerfGraphInterface
 
   void setInitialSolution();
 
+  // Used for enforcing nodal BC's when using a direct acceleration calculation
+  void setNodalBCs();
```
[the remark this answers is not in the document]
TheGreatCid: Can you elaborate on what you mean by a "passkey"?

framework/include/timeintegrators/TimeIntegrator.h
TheGreatCid: [36cf76e](https://github.com/idaholab/moose/pull/27991/commits/36cf76e60b00c7e3e4641fd082f78c4a580d7e63)

I believe I have address this here

framework/doc/content/source/timeintegrators/CentralDifference.md
```diff
@@ -4,16 +4,45 @@
...
+        \mathbf{\dot{u}}(t) &=& \frac{\mathbf{u}(t+\Delta t)-\mathbf{u}(t-\Delta t)}{2 \Delta t} \\
+        \mathbf{\ddot{u}}(t) &=& \frac{\mathbf{u}(t+\Delta t)-2\mathbf{u}(t)+\mathbf{u}(t-\Delta t)}{\Delta t^2}
+    \end{aligned}
 \end{equation}
 
-The central difference time integrator derives from the [`ActuallyExplicitEuler`](ActuallyExplicitEuler.md) class and therefore circumvents the nonlinear solver. It can be used with `consistent`, `lumped` or `lump_preconditioned`, `solve_type` options. Information on these solve options can be found on the [`ActuallyExplicitEuler`](ActuallyExplicitEuler.md) page.
+This leads to the following solution update,
+
+\begin{equation}
+    \mathbf{u}(t+\Delta t)=\mathbf{u(t)}+\left[\frac{1}{\Delta t^2}\mathbf{M}+\frac{1}{2\Delta t}\mathbf{C}\right]^{-1}\left[\mathbf{M}\left[\frac{1}{\Delta t^2}(\mathbf{u}(t-\Delta t)\right]-\mathbf{C}\left[\frac{1}{2\Delta t}\left(\mathbf{u}(t-\Delta t)-\mathbf{u}(t)\right)\right]+\mathbf{F}^{\text{ext}}-\mathbf{F}^{\text{int}}(\mathbf{u}(t))\right]
```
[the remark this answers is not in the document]
TheGreatCid: I don't believe it's arbitrary. Part of the residual is the difference between the external and internal forces.

framework/include/timeintegrators/TimeIntegrator.h
dschwen: Yeah, and simply return by value here `virtual bool isDirect() const { return false; }`

framework/include/systems/NonlinearSystemBase.h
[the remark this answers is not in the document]
dschwen: https://arne-mertz.de/2016/10/passkey-idiom/

framework/include/timeintegrators/TimeIntegrator.h
TheGreatCid: [36cf76e](https://github.com/idaholab/moose/pull/27991/commits/36cf76e60b00c7e3e4641fd082f78c4a580d7e63)

Let me know if this is what you had in mind

framework/include/systems/NonlinearSystemBase.h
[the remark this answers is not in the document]
TheGreatCid: Got it, though the thumbs down from @GiudGiud is ominous...

framework/doc/content/source/timeintegrators/CentralDifference.md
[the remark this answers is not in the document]
lindsayad: But forces are vectors with both magnitudes and directions. I would rather a sign difference be implicit in the numerical values and not the documentation. If I push on point x with my hand or the material pushes on point x in the same direction, these sign differences make me stop and wonder whether the result is different

framework/include/systems/NonlinearSystemBase.h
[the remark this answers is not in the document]
lindsayad: Maybe @GiudGiud would prefer that everyone be able to willy-nilly set nodal boundary conditions because our downstream applications have shown repeatedly that what you giveth they shall take. A general friending of all of `NonlinearSystemBase` would be an alternative but feels like overkill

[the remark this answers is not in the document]
TheGreatCid: So for the passkey then, should I follow similar process of that found in `ReporterTransferInferface.c`? 

```
void
ReporterTransferInterface::addReporterTransferMode(const ReporterName & name,
                                                   const ReporterMode & mode,
                                                   FEProblemBase & problem)
{
  checkHasReporterValue(name, problem);
  problem.getReporterData(ReporterData::WriteKey())
      .getReporterStateBase(name)
      .addConsumer(mode, _rti_transfer);
}
```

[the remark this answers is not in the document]
GiudGiud: > A general friending of all of NonlinearSystemBase would be an alternative but feels like overkill

it would be simple and readable.
And it's not like people will start writing crazy calls into CentralDifference without our knowledge

but dont let me distract you

framework/doc/content/source/timeintegrators/CentralDifference.md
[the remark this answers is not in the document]
TheGreatCid: Ah, I see what you are saying. I believe you are right here. I'll make that change.

framework/include/systems/NonlinearSystemBase.h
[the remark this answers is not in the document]
dschwen: Don't make David a pawn in your crusade against the passkey idiom! 😅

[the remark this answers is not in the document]
GiudGiud: David can chose the bright side of code simplicity by himself

framework/doc/content/source/timeintegrators/CentralDifference.md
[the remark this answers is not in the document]
TheGreatCid: #[dde04f5](https://github.com/idaholab/moose/pull/27991/commits/dde04f59f27db8ffffae1453402638896a3d7506)

framework/include/systems/NonlinearSystemBase.h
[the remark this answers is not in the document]
TheGreatCid: I'm inclined to agree with @GiudGiud. In my opinion, the passkey idiom harms readability. However my experience here is very limited.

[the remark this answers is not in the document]
lindsayad: A lot of template code doesn't read very well but we use it anyway because it serves a functional purpose. The passkey gives very precise control and I would argue increases readability as it immediately makes clear exactly what `CentralDifference` needs access to in `NonlinearSystemBase`. Whereas if you just make it a `friend` you do not know what of `NonlinearSystemBase's` multitude of data it needs access to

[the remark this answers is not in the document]
GiudGiud: > as it immediately makes clear exactly what CentralDifference needs access to in NonlinearSystemBase

immediately after asking ChatGPT what's up with this struct key I've never seen before :D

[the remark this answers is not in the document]
TheGreatCid: @lindsayad I took a shot at implementing a key
[a9736d1](https://github.com/idaholab/moose/pull/27991/commits/a9736d14fe42bdec55ae210f101518fb44977de5)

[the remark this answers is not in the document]
lindsayad: I ask ChatGPT C++ questions on the reg. I do not mind

TheGreatCid:
@lindsayad This implementation of CD really only works with solid mechanics due to how the mass matrix needs to be constructed. At least I am not familiar enough with fluid mechanics to know if the call to the mass matrix construction would return the correct matrix.
Should I put in a warning to state this?

framework/include/systems/NonlinearSystemBase.h
```diff
@@ -292,6 +301,8 @@ class NonlinearSystemBase : public SolverSystem, public PerfGraphInterface
 
   void setInitialSolution();
 
+  // Used for enforcing nodal BC's when using a direct acceleration calculation
```
lindsayad, line 304:
```suggestion
  /// Used for enforcing nodal BC's when using a direct acceleration calculation
```

framework/src/systems/NonlinearSystemBase.C
```diff
@@ -986,6 +986,41 @@ NonlinearSystemBase::setInitialSolution()
     setConstraintSecondaryValues(initial_solution, true);
 }
 
+void
+NonlinearSystemBase::setDirichletBCs(const WriteKey)
+{
+  NumericVector<Number> & initial_solution(solution());
```
lindsayad, line 992:
```suggestion
  auto & soln = solution();
```

framework/src/systems/NonlinearSystemBase.C
```diff
@@ -986,6 +986,41 @@ NonlinearSystemBase::setInitialSolution()
...
+    if (node->processor_id() == processor_id())
+    {
+      // reinit variables in nodes
+      _fe_problem.reinitNodeFace(node, boundary_id, 0);
+
+      if (_preset_nodal_bcs.hasActiveBoundaryObjects(boundary_id))
+      {
+        const auto & preset_bcs = _preset_nodal_bcs.getActiveBoundaryObjects(boundary_id);
+        for (const auto & preset_bc : preset_bcs)
+          preset_bc->computeValue(initial_solution);
```
lindsayad, line 1009:
```suggestion
          preset_bc->computeValue(soln);
```

framework/src/systems/NonlinearSystemBase.C
```diff
@@ -986,6 +986,41 @@ NonlinearSystemBase::setInitialSolution()
...
+      }
+      if (_ad_preset_nodal_bcs.hasActiveBoundaryObjects(boundary_id))
+      {
+        const auto & preset_bcs_res = _ad_preset_nodal_bcs.getActiveBoundaryObjects(boundary_id);
+        for (const auto & preset_bc : preset_bcs_res)
+          preset_bc->computeValue(initial_solution);
```
lindsayad, line 1015:
```suggestion
          preset_bc->computeValue(soln);
```

framework/src/systems/NonlinearSystemBase.C
```diff
@@ -986,6 +986,41 @@ NonlinearSystemBase::setInitialSolution()
...
+      }
+    }
+  }
+
+  _sys.solution->close();
```
lindsayad, line 1020:
```suggestion
  soln.close();
```

lindsayad (changes requested):
I'm slowly working through hoping that by doing so I can understand your mass matrix question better 😄 

But I can also just ask: are you running into issues related to mass matrices and fluid mechanics? Or was this a preemptive question?

TheGreatCid:
> But I can also just ask: are you running into issues related to mass matrices and fluid mechanics? Or was this a preemptive question?

It's a preemptive question. My motivation for this was to fix explicit contact, so I did not give much though to other physics

lindsayad:
My number one concern about this PR is whether it's too specific to solid mechanics such that all this work should be there instead of the framework. If this doesn't work for other physics, then maybe we are missing some generality in the framework that would enable you to implement the time integration you want in your module/application. I would ask you and @dschwen to weigh in and tell me whether you feel like you're locked into a box that doesn't fit very well and if that is the case, how we might make it better

TheGreatCid:
We are certainly missing some generality here. You may be right that this would be best suited for the solid mechanics module. 
There may a way to make this more general, but I not familiar enough with the other modules to figure that out quickly.

TheGreatCid:
The lack of generality can be seen in the changes I made to `InertialForce.C`. I need to modify the residual returned and `ComputeQpJacobian` as these had terms from the original CD formulation. I am assuming those exist in the other physics as well. 
It might be overkill to change other time dependent kernels to might this formulation as the point of having this addition CD formulation is for explicit contact

lindsayad:
The changes in `InertialForce` definitely look a bit funny as Jacobians are supposed to be the derivatives of the residual with respect to the solution dofs and you're returning a 0 residual ... yet the Jacobian is not 0. We introduced `LinearSystem` recently. I wonder if that would be a better paradigm for the work you're doing where you fill a RHS and a matrix (which would just be the mass matrix when using explicit time integration), and there is not the same J = dR/du conceptual relationship

TheGreatCid:
Yea, I did know it was a bit of a bastardization of the method, but I wasn't aware of another option as that is how the mass matrix is returned in `ActuallyExplicitEuler`. I was not aware of `LinearSystem`, I'll need to take a look at that. It looks like I need to take another stab at this and restructure things. I appreciate the feedback!

TheGreatCid:
Though the fact that the linear system does not support time integrators is an issue

lindsayad:
Ha, yes that is a big issue. It would be great to have someone work on that, but it would make sense if that's not what you want to focus your attention on

lindsayad:
If it didn't make sense or would take a lot of effort to re-use the current `TimeIntegrator` classes for `LinearSystem`, then this would be a great opportunity to use [PETSc TS](https://petsc.org/release/manual/ts/#ch-ts)

lindsayad:
Perhaps @grmnptr could weigh-in with his thoughts on time integration for `LinearSystem`

grmnptr:
We talked about this in person and the conclusion was that it would be very hard to do this with the `LinearSystem` because at the moment it does not have support for:
1. Finite Elements
2. TimeIntegrators 
So a lot of work would be needed here. I recommended registering a new matrix to the nonlinear system in the time integrator and creating a custom ThreadedLoop to assemble the mass matrix into it. That might run into issues on the other fronts so feel free to share other suggestions!

lindsayad:
If we are going that route, then I really think we should limit the number of changes made in the framework. I don't want to add second class support in the framework because we don't currently have the time to do the work that needs to be done at some point for first class support

TheGreatCid:
The better route might be to move this time integrator into solid_mechanics, and then use the tagging system and `MassMatrix.C` to pass the mass matrix in. `MassMatrix.C` is something I was unaware of until yesterday evening.

TheGreatCid:
@lindsayad I opened a new PR, https://github.com/idaholab/moose/pull/27991, with a different approach to implementing this time integrator

lindsayad:
I won't review the other one yet as you have it marked as draft, but I do like it a lot better. It looks quite clean!

TheGreatCid:
I left it as a draft because the documentation/tests are not yet implemented, but the code is ready for review. Should I mark it as ready now, or wait until I added in doc/tests?

lindsayad:
Might as well wait until you've added the documentation and tests so that a reviewer can look at it in one fell swoop. The input file can influence how I feel about the code design

PR #28047, opened 2024-07-02
Split NEML2 tests for different devices

hugary1995:
This PR duplicates existing NEML2 tests into two almost identical groups -- one for cpu, another for cuda.

The new test specs shall follow #27890

ref #26718


hugary1995:
@grmnptr @loganharbour @dschwen

PR #28061, opened 2024-07-03
Fix linker option for macs for libtorch tests

grmnptr:
refs #26718

### DIFF
--- a/.coverage
+++ b/.coverage
@@ -90,7 +90,7 @@ require_total = 88
 require_total = 84
 
 [thermal_hydraulics]
-require_total = 89
+require_total = 88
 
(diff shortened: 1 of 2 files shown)


loganharbour (approved):
https://civet.inl.gov/job/2312132/ https://civet.inl.gov/job/2312140/ are unrelated failures due to an issue with pulling containers that are on next.

milljm:
The error about the container not being available is odd. I wish I knew more about it. But it looks like to me there was no need to create moose-dev, but its failing to find the one it needs. Does this PR need to be rebased on Next maybe?

loganharbour (approved):

PR #28083, opened 2024-07-08
Explicit Dynamics: Initial Condition for Solution Time Derivatives

### LINKED ISSUE #28073: Explicit Dynamics: Initial Condition for Solution Time Derivatives
### Motivation

I want to be able to assign an initial velocity condition to a dynamic system that has displacement as it's primary variable. This is possible in implicit dynamics because the kernels use "solutionUdot" in their residual calculations. However, because the CentralDifference time integrator calculates solutionUdot and solutionUdotdot (velocity and acceleration) using solution_old and solution_older, there is no way to directly set an initial velocity condition. 

### Design

Add a method of assigning an initial condition to solution time derivatives in explicit dynamics.

### Impact

The enhancement will add something new.

ddm42:
- Added capability of using ICs to affect old states

closes #28073


framework/include/problems/FEProblemBase.h
```diff
@@ -2318,6 +2318,9 @@ class FEProblemBase : public SubProblem, public Restartable
   virtual void setCurrentLowerDElem(const Elem * const lower_d_elem, const THREAD_ID tid) override;
   virtual void setCurrentBoundaryID(BoundaryID bid, const THREAD_ID tid) override;
 
+  // Used for setting previous states
+  int global_current_state;
```
dschwen, line 2322:
```suggestion
  int _global_current_state;
```
we prefix member variables withe an underscore

framework/src/ics/InitialConditionBase.C
```diff
@@ -28,9 +30,12 @@ InitialConditionBase::validParams()
                         "When set to true, a UserObject retrieved "
                         "by this IC will not be executed before the "
                         "this IC");
-
   params.addParamNamesToGroup("ignore_uo_dependency", "Advanced");
 
+  params.addParam<int>("state", 0, "Specify which state being set.");
```
dschwen, line 35:
```suggestion
  params.addParam<int>("state", 0, "Specify which state is being set, with 0 being the current state, 1 the old state, 2 the older state, etc..");
```

framework/src/ics/InitialConditionBase.C-e
```diff
@@ -0,0 +1,100 @@
+//* This file is part of the MOOSE framework
```
dschwen, line 1: What is this file? `framework/src/ics/InitialConditionBase.C-e`

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,45 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
...
-      _block_ics[tid][var.name()].insert(object->blockIDs().begin(), object->blockIDs().end());
+      _block_ics[tid][ic_key].insert(object->blockIDs().begin(), object->blockIDs().end());
   }
 
   // Non-restricted
   else
   {
-    std::map<std::string, std::set<SubdomainID>>::const_iterator iter =
-        _block_ics[tid].find(var.name());
+    std::map<std::tuple<std::string, int>, std::set<SubdomainID>>::const_iterator iter =
```
dschwen, line 78:
```suggestion
    auto iter =
```

dschwen: I encourage more use of `auto`

framework/src/ics/InitialConditionWarehouse.C-e
```diff
@@ -0,0 +1,104 @@
+//* This file is part of the MOOSE framework
```
dschwen, line 1: remove this from the PR, too

framework/src/loops/ComputeInitialConditionThread.C
```diff
@@ -93,8 +93,26 @@ ComputeInitialConditionThread::operator()(const ConstElemRange & range)
...
+          if (ic->_my_state == loop_state)
+          {
+            ic->compute();
+            break_loop = true;
+          }
+        }
+        if (break_loop == true)
+        {
+          break;
+        }
```
dschwen, line 114:
```suggestion
        for (auto ic : vec)
          if (ic->_my_state == loop_state)
          {
            ic->compute();
            break_loop = true;
          }

        if (break_loop == true)
          break;
```

I'm not sure i understand the `break_loop` logic. Wouldn't that skip states for all ICs if one of them affects an older state? Even if they act on different variables?

ddm42: This for-loop logic was implemented inside an existing loop that runs for each relevant variable. So the ics being considered are all for the same variable. Does that make sense? check lines 78-81.

framework/src/ics/InitialConditionWarehouse.C-e
ddm42: will do. Something was wrong with my clang-format and its interaction with Mac.

ddm42 (commented):
Thanks for the comments. I'll commit and test them

framework/src/ics/InitialConditionBase.C
```diff
@@ -28,9 +30,15 @@ InitialConditionBase::validParams()
                         "When set to true, a UserObject retrieved "
                         "by this IC will not be executed before the "
                         "this IC");
-
   params.addParamNamesToGroup("ignore_uo_dependency", "Advanced");
 
+  params.addParam<int>("state",
+                       0,
+                       "Specify which state is being set, with 0 being the current state, 1 the "
+                       "old state, 2 the older state, etc..");
```
dschwen, line 38: Let's try this:

```suggestion
  MooseEnum stateEnum("CURRENT=0 OLD=1 OLDER=3", "CURRENT", /* allow_out_of_range = */ true);
  params.addParam<int>("state", stateEnum, "Specify state that is being set.");
```

framework/src/ics/InitialConditionBase.C
```diff
@@ -52,6 +60,7 @@ InitialConditionBase::InitialConditionBase(const InputParameters & parameters)
     DependencyResolverInterface(),
     Restartable(this, "InitialConditionBases"),
     ElementIDInterface(this),
+    _my_state(getParam<int>("state")),
```
dschwen, line 63:
```suggestion
    _my_state(getParam<MooseEnum>("state")),
```

The enum should automatically cast to int

framework/src/ics/InitialConditionBase.C
dschwen: Let's just set `allow_out_of_range` to false and not support states older than `OLDER` for now. (so, simply remove the `, /* allow_out_of_range = */ true` above

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,44 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
       mooseError("You are trying to set a boundary restricted variable on non-nodal variable. That "
                  "is not allowed.");
 
-    std::map<std::string, std::set<BoundaryID>>::const_iterator iter =
-        _boundary_ics[tid].find(var.name());
+    std::map<std::tuple<std::string, int>, std::set<BoundaryID>>::const_iterator iter =
```
dschwen, line 48:
```suggestion
    auto iter =
```

let's use `auto` where we can

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,44 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
...
-                                            object->boundaryIDs().end());
+      _boundary_ics[tid][ic_key].insert(object->boundaryIDs().begin(), object->boundaryIDs().end());
   }
 
   // Block Restricted
   else if (object->blockRestricted())
   {
-    std::map<std::string, std::set<SubdomainID>>::const_iterator iter =
-        _block_ics[tid].find(var.name());
+    std::map<std::tuple<std::string, int>, std::set<SubdomainID>>::const_iterator iter =
```
dschwen, line 62: same her

python/MooseDocs/test/gold/latex/extensions/include.tex
```diff
@@ -1,9 +1,9 @@
...
+\par We hold these truths to be self-evident, that all men are created equal, that they are endowed by their Creator with certain unalienable Rights, that among these are Life, Liberty and the pursuit of Happiness.
+\par We hold these truths to be self-evident, that all men are created equal, that they are endowed by
 \par Header {\textasciicircum} Right of the People to alter or to abolish it, and to institute new Government\$ Footer
 \begin{quote}
-\par Congress shall make no law respecting an establishment of religion, or prohibiting the free exercise thereof; or abridging the freedom of speech, or of the press; or the right of the people peaceably to assemble, and to petition the Government for a redress of grievances. 
-\par A well regulated Militia, being necessary to the security of a free State, the right of the people to keep and bear Arms, shall not be infringed. 
+\par Congress shall make no law respecting an establishment of religion, or prohibiting the free exercise thereof; or abridging the freedom of speech, or of the press; or the right of the people peaceably to assemble, and to petition the Government for a redress of grievances.
```
dschwen, line 7: You have a bunch of TeX files that just have whitespace changes. Those shoudl be removed from this PR. I can help you with that.

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -35,6 +36,7 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
 {
   // Check that when object is boundary restricted that the variable is nodal
   const MooseVariableFEBase & var = object->variable();
+  const std::tuple<std::string, int> ic_key = std::tuple(var.name(), object->_my_state);
```
dschwen, line 39:
```suggestion
  const auto ic_key = std::tuple(var.name(), object->_my_state);
```

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,42 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
       mooseError("You are trying to set a boundary restricted variable on non-nodal variable. That "
                  "is not allowed.");
 
-    std::map<std::string, std::set<BoundaryID>>::const_iterator iter =
-        _boundary_ics[tid].find(var.name());
+    auto iter = _boundary_ics[tid].find(ic_key);
```
dschwen, line 48:
```suggestion
    const auto iter = _boundary_ics[tid].find(ic_key);
```

Sorry, that should have been const

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,42 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
...
-                                            object->boundaryIDs().end());
+      _boundary_ics[tid][ic_key].insert(object->boundaryIDs().begin(), object->boundaryIDs().end());
   }
 
   // Block Restricted
   else if (object->blockRestricted())
   {
-    std::map<std::string, std::set<SubdomainID>>::const_iterator iter =
-        _block_ics[tid].find(var.name());
+    auto iter = _block_ics[tid].find(ic_key);
```
dschwen, line 61: same

dschwen (commented):
Looks fine to me if those minor comments are addressed.

framework/src/ics/InitialConditionWarehouse.C
ddm42: What do you mean by same? the type is already "auto"?

framework/include/problems/FEProblemBase.h
```diff
@@ -2348,6 +2348,9 @@ class FEProblemBase : public SubProblem, public Restartable
   virtual void setCurrentLowerDElem(const Elem * const lower_d_elem, const THREAD_ID tid) override;
   virtual void setCurrentBoundaryID(BoundaryID bid, const THREAD_ID tid) override;
 
+  // Used for setting previous states
+  int _global_current_state;
```
hugary1995, line 2352: Is this public? It's kinda difficult to check in the web browser, but if so move it to protected at least.

framework/include/timeintegrators/CentralDifference.h
```diff
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "ActuallyExplicitEuler.h"
+#include "libmesh/numeric_vector.h"
```
hugary1995, line 13: why is this?

framework/src/ics/InitialConditionBase.C
```diff
@@ -8,9 +8,11 @@
 //* https://www.gnu.org/licenses/lgpl-2.1.html
 
 #include "InitialConditionBase.h"
+#include "MooseEnum.h"
 #include "SystemBase.h"
 #include "MooseVariableFE.h"
 #include "UserObject.h"
+#include "libmesh/libmesh_common.h"
```
hugary1995, line 15: why this?

framework/src/ics/InitialConditionBase.C
```diff
@@ -28,9 +30,13 @@ InitialConditionBase::validParams()
                         "When set to true, a UserObject retrieved "
                         "by this IC will not be executed before the "
                         "this IC");
-
   params.addParamNamesToGroup("ignore_uo_dependency", "Advanced");
 
+  MooseEnum stateEnum("CURRENT=0 OLD=1 OLDER=2", "CURRENT");
+  params.addParam<MooseEnum>("state", stateEnum, "Specify state that is being set.");
```
hugary1995, line 36: Can you expand the description a bit more?

framework/src/ics/InitialConditionBase.C
```diff
@@ -28,9 +30,13 @@ InitialConditionBase::validParams()
...
+
+  params.addParamNamesToGroup("state", "Advanced");
```
hugary1995, line 38: I feel like this doesn't have to go into the advanced group, but this is apparently debatable.

framework/src/ics/InitialConditionWarehouse.C
hugary1995: hmm I see this conversation is marked as resolved, but the suggested change hasn't been applied. Is github broken?

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,42 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
...
 
-    std::map<std::string, std::set<BoundaryID>>::const_iterator iter =
-        _boundary_ics[tid].find(var.name());
+    auto iter = _boundary_ics[tid].find(ic_key);
     if (iter != _boundary_ics[tid].end() && object->hasBoundary(iter->second))
       mooseError("The initial condition '",
                  object->name(),
-                 "' is being defined on a boundary that already has an initial condition defined.");
+                 "' is being defined on a boundary that already has an initial condition defined "
+                 "(for the variable and state combo).");
```
hugary1995, line 53: This is user-facing so let's make the wording a bit more professional. For "variable and state combo", also tell people which combo it is.

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,42 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
...
-        _block_ics[tid].find(var.name());
+    auto iter = _block_ics[tid].find(ic_key);
     if (iter != _block_ics[tid].end() &&
         (object->hasBlocks(iter->second) ||
          (iter->second.find(Moose::ANY_BLOCK_ID) != iter->second.end())))
       mooseError("The initial condition '",
                  object->name(),
-                 "' is being defined on a block that already has an initial condition defined.");
+                 "' is being defined on a block that already has an initial condition defined (for "
+                 "the variable and state combo).");
```
hugary1995, line 68: Same suggestion as above.

framework/src/ics/InitialConditionWarehouse.C
```diff
@@ -43,43 +45,42 @@ InitialConditionWarehouse::addObject(std::shared_ptr<InitialConditionBase> objec
...
   {
-    std::map<std::string, std::set<SubdomainID>>::const_iterator iter =
-        _block_ics[tid].find(var.name());
+    auto iter = _block_ics[tid].find(ic_key);
     if (iter != _block_ics[tid].end())
       mooseError("The initial condition '",
                  object->name(),
-                 "' is being defined on a block that already has an initial condition defined.");
+                 "' is being defined on a block that already has an initial condition defined (for "
+                 "the variable and state combo).");
```
hugary1995, line 81: Same suggestion on wording.

framework/include/ics/InitialConditionBase.h
```diff
@@ -91,6 +91,9 @@ class InitialConditionBase : public MooseObject,
 
   virtual const std::set<std::string> & getSuppliedItems() override;
 
+  // variable used when applying initial conditions to previous states
+  int _my_state;
```
hugary1995, line 95: Is this public? Move it to protected at least.

framework/src/loops/ComputeInitialConditionThread.C
```diff
@@ -93,8 +93,23 @@ ComputeInitialConditionThread::operator()(const ConstElemRange & range)
...
+        for (auto ic : vec)
+          if (ic->_my_state == loop_state)
+          {
+            ic->compute();
+            break_loop = true;
+          }
+
+        if (break_loop == true)
+          break;
+      }
```
hugary1995, line 112: You could help me understand this better... Why do we immediately break after one IC is computed? What if I have multiple IC objects for the same variable with the same state, just defined on different blocks/boundaries, does this mean only one of them gets computed?

As a side note, directly accessing members with an underscore prefix is not the best practice in MOOSE.

framework/src/problems/FEProblemBase.C
```diff
@@ -3332,64 +3335,86 @@ FEProblemBase::projectSolution()
   FloatingPointExceptionGuard fpe_guard(_app);
 
   ConstElemRange & elem_range = *_mesh.getActiveLocalElementRange();
-  ComputeInitialConditionThread cic(*this);
-  Threads::parallel_reduce(elem_range, cic);
 
-  if (haveFV())
+  // loop over _global_current_state = 2, 1, 0
+  for (_global_current_state = 1; _global_current_state >= 0; _global_current_state--)
```
hugary1995, line 3340: But this is not 2, 1, 0 -- this is 1, 0. Now I wonder if you have test coverage for checking the older IC being set correctly...

framework/src/timeintegrators/CentralDifference.C
```diff
@@ -114,3 +114,30 @@ CentralDifference::computeTimeDerivatives()
...
+  // using the Central Difference method
+  // u_dotdot_old = (first_term - second_term + third_term) / dt / dt
+  //       first_term = u
+  //      second_term = 2 * u_old
+  //       third_term = u_older
+  u_dotdot -= u_old;
+  u_dotdot -= u_old;
+  u_dotdot += u_older;
+  u_dotdot *= 1.0 / (_dt * _dt);
+}
```
hugary1995, line 143: How does this even work not in a header file?

hugary1995 (changes requested):

framework/src/loops/ComputeInitialConditionThread.C
ddm42: We are not breaking after one IC is computed. Lines 104-108 compute for every ic in the vector before it has a chance to break. 

However, I am open to suggestions for better logic. Just remember that vec is a vector of ic's for only ONE variable. So other variables will not be affected.

ddm42: I used a member variable _fe_problem._global_current_state because we needed to transfer the loop state from FEProblemBase::projectSolution(). Is there another location for a loop variable that would be available in both files? That was why I chose to implement it this way. But I am open to suggestions.

framework/src/ics/InitialConditionBase.C
ddm42: I'll remove this. It must have added it automatically.

framework/src/timeintegrators/CentralDifference.C
ddm42: Why would implementation need to be in the header file? I moved it because that seemed wrong to me. But I can move it back if there was a reason.

framework/src/problems/FEProblemBase.C
ddm42: I don't. I shall add some.

framework/src/ics/InitialConditionWarehouse.C
ddm42: No I was marking them resolved because I made the change on my end. Just trying to keep track of the ones I have left.

framework/src/loops/ComputeInitialConditionThread.C
hugary1995: I need to look at this more closely. I suggest other reviewers should also double check this.

ddm42: I would love to implement it a cleaner way. Thanks for checking it out.

framework/src/timeintegrators/CentralDifference.C
hugary1995: See this answer: https://stackoverflow.com/a/495056

ddm42: I see. Then yes I'm surprised it still works. I'll move it back in my next push.

hugary1995: well it works probably because people never instantiate the template types to be anything other than NumericVector (or whatever that is). If that can be guaranteed, you can remove the template and separate the declaration and implementation. Otherwise, I suggest putting it back to the header.

framework/src/problems/FEProblemBase.C
```diff
@@ -1082,8 +1084,9 @@ FEProblemBase::initialSetup()
 
   if (!_app.isRecovering() && !_app.isRestarting())
```
lynnmunday, line 1085:
```suggestion
  if (!_app.isRecovering() && !_app.isRestarting())
```

framework/src/problems/FEProblemBase.C
```diff
@@ -1082,8 +1084,9 @@ FEProblemBase::initialSetup()
...
   {
-    // During initial setup the solution is copied to the older solution states (old, older, etc)
-    copySolutionsBackwards();
+    // This functionality was moved to FEProblemBase.C::projectSolution()
+    // // During initial setup the solution is copied to the older solution states (old, older, etc)
+    // // copySolutionsBackwards();
   }
```
lynnmunday, line 1090:
```suggestion
  
```

framework/src/problems/FEProblemBase.C
lynnmunday, line 1085:
```suggestion
  if (!_app.isRecovering() && !_app.isRestarting())
```

lynnmunday, line 1085: delete if statement

framework/src/problems/FEProblemBase.C
```diff
@@ -3332,64 +3328,86 @@ FEProblemBase::projectSolution()
...
+      for (auto & sys : _solver_systems)
+        sys->copySolutionsBackwards();
+      _aux->copySolutionsBackwards();
+    }
   }
 
-  _aux->solution().close();
-  _aux->solution().localize(*_aux->sys().current_local_solution, _aux->dofMap().get_send_list());
+  // Resetting _global_current_state to 0 in case anyone needs to run ICs again (which they do
+  // sometimes).
```
hugary1995, line 3409: In the comment, can you spell out the situations in which people want to run ICs again?

framework/src/variables/MooseVariableData.C
```diff
@@ -9,6 +9,7 @@
 
 #include "MooseVariableData.h"
 #include "MooseVariableField.h"
+#include "MooseVariableDataBase.h"
```
hugary1995, line 12:
```suggestion
```

hugary1995 (changes requested):

framework/src/loops/ComputeInitialConditionThread.C
```diff
@@ -93,8 +93,23 @@ ComputeInitialConditionThread::operator()(const ConstElemRange & range)
...
+        for (auto ic : vec)
+          if (ic->getMyState() == loop_state)
+          {
+            ic->compute();
+            break_loop = true;
+          }
+
+        if (break_loop == true)
+          break;
+      }
```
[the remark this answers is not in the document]
ddm42: It is called once for each state.

framework/src/problems/FEProblemBase.C
```diff
@@ -3332,64 +3328,86 @@ FEProblemBase::projectSolution()
...
-  Threads::parallel_reduce(elem_range, cic);
 
-  if (haveFV())
+  // loop over _global_current_state = 2, 1, 0
+  for (_global_current_state = 2; _global_current_state >= 0; _global_current_state--)
   {
-    using ElemInfoRange = StoredRange<MooseMesh::const_elem_info_iterator, const ElemInfo *>;
-    ElemInfoRange elem_info_range(_mesh.ownedElemInfoBegin(), _mesh.ownedElemInfoEnd());
+    ComputeInitialConditionThread cic(*this);
+    Threads::parallel_reduce(elem_range, cic);
```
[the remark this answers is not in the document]
ddm42: Yes. This was required to avoid more complicated logic that would be prone to more edge cases. Here are the design requirements:

1. _global_current_state has to exist inside FEProblemBase. This was to respect the ordering of IC computing that happens in FEProblemBase::projectSolution(). In that function, someone decided that there was a particular order that ICs needed to be computed in. However, inside ComputeInitialConditionThread::compute() was the approperiate place to apply the logic of which ICs to compute. 

2. And then, sys->copySolutionsBackwards() needs to be called from FEProblemBase, as far as I can tell. 

This led me to need some way of transferring loop_state information back and forth between FEProblemBase::projectSolution() and ComputeInitialConditionThread::compute().

3. The final requirement was that if someone only specifies initial conditions for the current state, that IC needs to be copied backward to the old and older states. This was the old functionality of MOOSE.
4. Otherwise, if someone specified multiple states, then the proper ICs needed to be copied back to their appropriate states. 

So if there is a better way to accomplish these things, I'm open to suggestions.

The way the loop inside ComputeInitialConditionThread::compute() works is that it checks to see if state=2 was specified for the variable, then state=1, then state=0. If state=2 was specified, then it will abort the loop, and the state=2 initial condition will get copied back to the appropriate state. But if someone only specified state=0 for that state, then the loop won't find any initial conditions for states 1 and 2, so it will choose the state=0 IC, and that will get copied back to the old states.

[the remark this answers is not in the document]
hugary1995: 1 & 2: sure, I didn't complain about the way you transfer state between FEProblem and ICs...

3 & 4 should definitely be documented somewhere. I personally prefer that to be part of the IC system documentation.

> The way the loop inside ComputeInitialConditionThread::compute() works is that it checks to see if state=2 was specified for the variable, then state=1, then state=0. If state=2 was specified, then it will abort the loop, and the state=2 initial condition will get copied back to the appropriate state. But if someone only specified state=0 for that state, then the loop won't find any initial conditions for states 1 and 2, so it will choose the state=0 IC, and that will get copied back to the old states.

I think I understand what you want to achieve now. I'd like to suggest a slightly different implementation because your current implementation computes the IC three times if only state=0 is set for a variable, once for state=0, once for state=1, and once for state=2. If the IC computation is very expensive, you have to evaluate the expensive IC three times!

How about this? Let's first compute IC like we used to, that is, compute IC at state=0, copy the solution vector backwards to the old solution vector (**not the older!**). Then, go to state=1, compute ICs with state=1, and copy the old solution vector to the older solution vector. Finally, go to state=2, compute ICs with state=2. This way you still preserve the old behavior when only state=0 is set for an IC, and you can avoid excessive IC evaluations by simply copying the solution vectors backwards.

The challenge with this new approach is that you need to let the threaded IC loop directly write into old and older solution vectors, instead of what you are currently doing (i.e., always writing to the current solution vector, and copy it backwards).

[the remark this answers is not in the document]
ddm42: Great! Actually that would be better, but in my first attempt at implementing this, I was unable to the write the old and older state solution vectors directly. Because they seem to live inside LibMesh or PetsC. Do you have any templates that I could look at without breaking a lot of Moose things? Last time there seemed to be no way to make a new vector before the old state vector was created.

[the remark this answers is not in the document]
hugary1995: I'm not terribly familiar with that part. Could you perhaps create the old and older vectors earlier?

@dschwen any pointers?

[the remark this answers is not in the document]
ddm42: I worked with him on this, and he seemed to think copying backwards seemed to be the best method. But yes @dschwen please weigh in when you get back from vacation Thursday. I'll attempt to implement this before then. Thank you for the advice @hugary1995

framework/include/ics/InitialConditionBase.h
```diff
@@ -91,13 +91,18 @@ class InitialConditionBase : public MooseObject,
 
   virtual const std::set<std::string> & getSuppliedItems() override;
 
+  int & getMyState();
```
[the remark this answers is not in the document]
ddm42: I tried this, but when looping through the states, it leaves more room for errors. When the unsigned int is reduced each time during the for loops, eventually it wants to become negative, indicating the end of the loop has been reached. But because it is unsigned, the remains positive and causes an endless loop. There are solutions for this of course, but it will create room for mistakes if someone wants to change the logic. I think we should leave all of them as signed ints.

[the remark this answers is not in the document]
ddm42: I'll implement it this way. But I respectfully disagree. Casting to signed ints explicitly just for the comparison operators, or adding 1 and ending the loop at 1 are the two simple ways of fixing the problem. And both seem less perfect than having signed ints. Especially when someone might make the mistake again in the future. We could even save more storage space by making them shorts or bytes in if the data is what you're concerned about.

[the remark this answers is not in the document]
ddm42: Hopefully my concerns won't be important anymore though when I implement your new logic for applying old and older ICs.

framework/src/problems/FEProblemBase.C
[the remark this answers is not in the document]
ddm42: @dschwen and I looked at it together, and there are several layers of APIs to go through to accomplish the "writing to old solution states." I will instead implement some logic to only compute each IC object once.

modules/solid_mechanics/test/tests/old_state_ic/velocity_ic.i
```diff
@@ -0,0 +1,143 @@
...
+    value = 0
+    state = CURRENT
+  []
+  [old]
+    type = ConstantIC
+    variable = disp_x
+    value = -1
+    state = OLD
+  []
+[]
```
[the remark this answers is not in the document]
ddm42: Alright I added some tests to check OLD and OLDER. Required editing CopyValueAux. And I added some tests to make sure it's working properly.

ddm42:
> I just had some minor suggestions. Though we should probably double (triple) check the loop logic as I pointed out in the comments. I feel like your loop is computing certain variable ICs some extra times. Even though that doesn't matter from the correctness standpoint, that's still some extra (wasted) computation...
> 
> Other than that just add one more test for `OLDER` and I'll approve this.

dschwen:
We probably need to move a test from modules into framework.

modules/solid_mechanics/test/tests/old_state_ic/tests
```diff
@@ -0,0 +1,27 @@
+[Tests]
+  issues = '#28073'
+  [old_state_ic]
+    type = CSVDiff
+    input = 'old_state_ic.i'
+    csvdiff = 'old_state_ic_out.csv'
```
dschwen, line 6: Missing `design` documentation link.

framework/include/auxkernels/CopyValueAux.h
```diff
@@ -24,8 +24,13 @@ class CopyValueAux : public AuxKernel
 protected:
   virtual Real computeValue() override;
 
+  // variable used to specify state being copied
+  unsigned short _state;
+
   /// The variable to project from
   const VariableValue & _v;
```
maxnezdyur, line 31: Get rid of all the other VariableValues now. With the change I sent in the .C file you don't need to store all of these anymore.

framework/src/auxkernels/CopyValueAux.C
```diff
@@ -19,11 +19,22 @@ CopyValueAux::validParams()
...
   return params;
 }
 
 CopyValueAux::CopyValueAux(const InputParameters & parameters)
-  : AuxKernel(parameters), _v(coupledValue("source")), _source_variable(*getVar("source", 0))
+  : AuxKernel(parameters),
+    _state(getParam<MooseEnum>("state")),
+    _v(coupledValue("source")),
+    _v_old(_state == 1 ? coupledValueOld("source") : _v),
+    _v_older(_state == 2 ? coupledValueOlder("source") : _v),
```
maxnezdyur, line 36:
```suggestion
 _v(_state == 0   ? coupledValue("source")
       : _state == 1 ? coupledValueOld("source")
                     : coupledValueOlder("source")),
```
If you use the above then you do not need to chang the computeValue method that was originally there.

maxnezdyur (changes requested):
In `framework/doc/content/syntax/ICs/index.md`

add something like 
```
## Old and Older ICs

The ICs system supports the ability to set ICs on old and older states. This can be useful for setting initial velocity or acceleration conditions. It can be set with the [!param](/ICs/ConstantIC/state) parameter and specifying `OLD` or `OLDER`.
```
to tell users about the new feature and how to use it.

framework/doc/content/syntax/ICs/index.md
```diff
@@ -62,6 +62,13 @@ object's variable *must* be defined on the block - otherwise the IC for the
 next lowest block ID for the node is used - and so forth until one has the
 variable defined.
 
+## Old and Older ICs
+
+The ICs system supports the ability to set ICs on old and older states. This can be useful for setting initial velocity or acceleration conditions. It can be set with the [!param](/ICs/ConstantIC/state) parameter and specifying `OLD` or `OLDER`.
```
hugary1995, line 67:
```suggestion
The ICs system supports the ability to set ICs on old and older states. This can be useful for initializing old and older states of variables needed for various time integration schemes. It can be set with the [!param](/ICs/ConstantIC/state) parameter and specifying `OLD` or `OLDER`.
```

framework/include/ics/InitialConditionInterface.h
```diff
@@ -0,0 +1,37 @@
...
+   *
+   * @param parameters The parameters object holding data for the class to use.
+   */
+  InitialConditionInterface(const InputParameters & parameters);
+
+  virtual ~InitialConditionInterface();
+
+  static InputParameters validParams();
+
+  unsigned short & getState();
```
hugary1995, line 32: Add a doxygen comment, especially about the fact that the returned reference is writable.

framework/include/ics/InitialConditionWarehouse.h
```diff
@@ -40,8 +40,13 @@ class InitialConditionWarehouse : public MooseObjectWarehouseBase<InitialConditi
...
   ///@{
-  /// Variable name to block/boundary IDs for error checking
-  std::vector<std::map<std::string, std::set<BoundaryID>>> _boundary_ics;
-  std::vector<std::map<std::string, std::set<SubdomainID>>> _block_ics;
+  /// Maps used to check if multiple ICs define the same variable on the same block/boundary for the same state (e.g CURRENT, OLD, OLDER).
+  /// They are vectors of maps because each map is repeated for each thread.
+  /// Each map relates string-int tuples to a set of block/boundary IDs.
+  /// The string-int tuple is a unique identifier for a specific variable and state. The string-int tuple is renamed to ic_key_type for clarity.
+  /// The algorithm then makes sure that a new IC object does not overlap with a previous IC object (i.e. same block/boundary).
+  typedef std::tuple<std::string, unsigned short> ic_key_type;
```
hugary1995, line 48: I'd prefer `using` over `typedef`, so something like
```cpp
using ic_key_type = std::tuple<VariableName, unsigned short>;
```

framework/include/problems/FEProblemBase.h
```diff
@@ -884,6 +884,8 @@ class FEProblemBase : public SubProblem, public Restartable
 
   void projectSolution();
 
+  unsigned short getCurrentICState();
```
hugary1995, line 887: Add a doxygen comment.

framework/src/problems/FEProblemBase.C
```diff
@@ -1084,6 +1085,65 @@ FEProblemBase::initialSetup()
...
+    {
+      // state 0 copy (we'll overwrite current state when evaluating ICs and need to restore it once
+      // we're done with the old/older state ICs)
+      std::vector<std::unique_ptr<NumericVector<Real>>> state0_sys_buffers(_solver_systems.size());
+      std::unique_ptr<NumericVector<Real>> state0_aux_buffer;
+
+      // save state 0
+      for (const auto i : index_range(_solver_systems))
+      {
+        _solver_systems[i]->system().update(); // not sure if this update is needed
```
hugary1995, line 1115: I don't think this is needed. Same applies to the following system updates. If any test fails without this call, I'd like to understand why.

framework/src/problems/FEProblemBase.C
```diff
@@ -1084,6 +1085,65 @@ FEProblemBase::initialSetup()
...
+      // recover state 0
+      for (const auto i : index_range(_solver_systems))
+      {
+        _solver_systems[i]->solutionState(0) = *state0_sys_buffers[i];
+        _solver_systems[i]->solutionState(0).close();
+        _solver_systems[i]->update();
+      }
+      _aux->solutionState(0) = *state0_aux_buffer;
+      _aux->solutionState(0).close();
+      _aux->update();
```
hugary1995, line 1145: This logic now looks correct to me.

hugary1995 (changes requested):
I have some minor comments. Did you also add a test for state=OLDER?

maxnezdyur:
> I have some minor comments. Did you also add a test for state=OLDER?

The input file`old_state_ic.i` applies an different IC value for all 3 states.

hugary1995 (approved):

dschwen (approved):

Issue #27091, opened 2024-03-14
Add project suffix option to apptainer generator


loganharbour:
### Motivation

We would like to push to something like `app-suffix/app-suffix-x86-64:latest` in apptainer.

### Design

Add a `--project-suffix` option to `ApptainerGenerator`

### Impact

None. New capability.

PR #28088, opened 2024-07-10
Properly use --to-project-suffix in push action

loganharbour:
refs #27091

### DIFF
--- a/scripts/apptainer_generator.py
+++ b/scripts/apptainer_generator.py
@@ -288,13 +288,13 @@ class ApptainerGenerator:
         self.run(command)
         return file
 
-    def apptainer_push(self, project: str, name: str, from_tag: str, to_tag=None):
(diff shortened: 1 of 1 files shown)


loganharbour:
@milljm if you wouldn't mind. 

This satisfies the case where we want to push a variant of an application to another repo. For example, say we want to build apptainer `moose` but with a variant `foo`. We can then do

```
apptainer_generator.py push app . --suffix foo --to-project-suffix foo
```

which will push to

```
oras://mooseharbor.hpc.inl.gov/moose-foo/moose-foo-x86_64:abcd123
```

I tried to implement this in the past, but didn't actually test it. This fixes my mess up, and I tested it this time.

milljm (approved):

milljm (approved):

loganharbour:
Failures are unrelated.

Issue #28112, opened 2024-07-11
Improve apptainer prompts


loganharbour:
### Motivation

The apptainer prompt isn't very good at retaining useful information pertaining to the version.

### Design

Utilize version controlling to provide the correct container name with containers generated by the apptainer generator.

### Impact

New capability.

PR #28113, opened 2024-07-11
Moose dev prompt

loganharbour:
Refs https://github.com/idaholab/moose/issues/28112. Doesn't close it yet because we will eventually move this down to the base container.

### DIFF
--- a/apptainer/moose-dev.def
+++ b/apptainer/moose-dev.def
@@ -33,10 +33,16 @@
 {#- The script used to install wasp                                                           -#}
 {%- set WASP_BUILD_SCRIPT = 'update_and_rebuild_wasp.sh' -%}
 
+{%- set SET_PROMPT_SCRIPT = '/.singularity.d/set_prompt.bash' -%}
+
 BootStrap: {{ APPTAINER_BOOTSTRAP }}
 From: {{ APPTAINER_FROM }}
 
 %environment
+    #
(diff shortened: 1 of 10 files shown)


apptainer/moose-dev.def
```diff
@@ -65,6 +71,21 @@ From: {{ APPTAINER_FROM }}
...
+    # we can. But we don't want to do a full rebuild at the moment,
+    # so this is good for now.
+    mv ${ROOT_BUILD_DIR}/set_prompt.bash  ${SINGULARITY_ROOTFS}{{ SET_PROMPT_SCRIPT }}
+    cat <<'EOF' > ${SINGULARITY_ROOTFS}/.singularity.d/env/99-zzz_prompt.sh
+if [ -n "${CUSTOM_PROMPT}" ]; then
+    PS1=${CUSTOM_PROMPT}
+else
+    if [ -e {{ SET_PROMPT_SCRIPT }} ]; then
+        source {{ SET_PROMPT_SCRIPT }}
+        set_prompt
```
milljm, line 84: I _think_ this needs to #
```suggestion
        PS1="\$(set_prompt)"
```

loganharbour: actually not needed, since set_prompt sets PS1=

milljm (approved):

PR #28155, opened 2024-07-16
Prerequest curls whenever we `_need_curl`

roystgnr:
## Reason
While investigating #28153 I found that, although I couldn't replicate that exact bug in a recent MOOSE, there was a similar bug that could be triggered when using kernels that require curls vs divs.

## Design
Use `_need_curl` where appropriate, don't just hard-code that `NEDELEC_ONE` always needs kernels and other FE types always don't.

While I was at it I also turned those `map`s-to-`true` into `set`s, because that's an isomorphism and the code looks a lot cleaner afterward.

## Impact
If I didn't make any typos (and I tried to use regexps as much as possible to avoid that) and if we don't have any bugs in curl-based kernels, this shouldn't affect any already-working cases' results.  Anyone using `NEDELEC_ONE` with kernels that don't need curl should get a faster runtime.  Anyone using kernels that need curl with other vector FE types should see their code stop failing assertions in `devel` and segfaulting in `opt`.

### DIFF
--- a/framework/include/base/Assembly.h
+++ b/framework/include/base/Assembly.h
@@ -1585,7 +1585,7 @@ public:
   template <typename OutputType>
   const typename OutputTools<OutputType>::VariablePhiSecond & feSecondPhi(FEType type) const
   {
-    _need_second_derivative[type] = true;
+    _need_second_derivative.insert(type);
     buildFE(type);
     return _fe_shape_data[type]->_second_phi;
   }
@@ -1626,7 +1626,7 @@ public:
   template <typename OutputType>
   const typename OutputTools<OutputType>::VariablePhiSecond & feSecondPhiFace(FEType type) const
   {
-    _need_second_derivative[type] = true;
+    _need_second_derivative.insert(type);
     buildFaceFE(type);
     return _fe_shape_data_face[type]->_second_phi;
   }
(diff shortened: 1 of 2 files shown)


lindsayad (approved):

GiudGiud (dismissed):

roystgnr:
It looks like interface kernels always lead to reinit() for both the elem and it's neighbor.  So then when `VectorFEWave` calls `curlPhi()` it sets `_need_curl` to true, but it doesn't call `curlPhiFaceNeighbor()`, so the only time when `buildVectorFaceNeighborFE()` is reached (via `_phi_face = _phi_face_assembly_method(..)` calling `fePhiFaceNeighbor()`), we're still in the `MooseVariableBase` constructor and we haven't seen a request for curls and it dutifully does not request a curl ahead of time, but when we `reinitFEFaceNeighbor()` later *then* we've noticed that we need curls and we request them too late.

I'll see if I can come up with a fix that's not too inefficient/intrusive.

roystgnr:
Could I get a re-approval reviewing that last commit?  It seems to me like that's the best fix here, and it should certainly be fine for correctness, but it's possible that there's a more efficient alternative I'm missing.

GiudGiud (approved):

PR #28165, opened 2024-07-17
Enable Quadratic Elements in Patterned(Hex|Cartesian) Mesh Generators

### LINKED ISSUE #28164: Enable Quadratic Elements in Patterned(Hex|Cartesian) Mesh Generators
### Motivation

Current Reactor Module tools are only capable of creating meshes with linear elements, which limits its applications. This enhancement follows up on #26986.

### Design

Patterned(Hex|Cartesian) mesh generators optionally generate boundary elements around a stitched pattern. This enhancement will allow the use of quadratic elements (QUAD8/QUAD9) in the boundary region.

### Impact

aaronoaks:
closes #28164


modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
```diff
@@ -831,6 +838,32 @@ PatternedCartesianMeshGenerator::generate()
...
+        // only adjust elements on outer boundary
+        if (!(*elem)->on_boundary())
+          continue;
+
+        // adjust point 5
+        Point pt_5 = ((*elem)->point(1) + (*elem)->point(2)) / 2.0;
+        out_mesh->add_point(pt_5, (*elem)->node_ptr(5)->id());
+
+        // adjoin point 7
+        Point pt_7 = ((*elem)->point(0) + (*elem)->point(3)) / 2.0;
```
GiudGiud, line 856: why are we moving them?

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
```diff
@@ -1213,31 +1215,47 @@ PolygonMeshGeneratorBase::buildSimplePeripheral(
...
+
+      new_elem = std::make_unique<Quad4>();
+      new_elem->set_node(0) = mesh->node_ptr(j + (num_sectors_per_side + 1) * (i));
+      new_elem->set_node(1) = mesh->node_ptr(j + 1 + (num_sectors_per_side + 1) * (i));
+      new_elem->set_node(2) = mesh->node_ptr(j + 1 + (num_sectors_per_side + 1) * (i + 1));
+      new_elem->set_node(3) = mesh->node_ptr(j + (num_sectors_per_side + 1) * (i + 1));
+
+      Elem * elem = mesh->add_elem(std::move(new_elem));
+
+      // add subdoamin/boundary IDs
```
GiudGiud, line 1228:
```suggestion
      // add subdomain and boundary IDs
```

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
```diff
@@ -1213,31 +1215,47 @@ PolygonMeshGeneratorBase::buildSimplePeripheral(
...
       if (j == 0)
-        boundary_info.add_side(elem_Quad4, 3, OUTER_SIDESET_ID);
+        boundary_info.add_side(elem, 3, OUTER_SIDESET_ID);
       if (j == num_sectors_per_side - 1)
-        boundary_info.add_side(elem_Quad4, 1, OUTER_SIDESET_ID);
+        boundary_info.add_side(elem, 1, OUTER_SIDESET_ID);
     }
   }
 
+  mesh->remove_orphaned_nodes();
```
GiudGiud, line 1249: add a comment for why we need this

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
```diff
@@ -1213,31 +1215,47 @@ PolygonMeshGeneratorBase::buildSimplePeripheral(
...
+
+  // convert element to second order if needed
```
GiudGiud, line 1251: is this converting only the boundary or boundary + pattern?

modules/reactor/test/tests/meshgenerators/patterned_hex_mesh_generator/tests
```diff
@@ -18,6 +18,46 @@
...
+                Mesh/pattern/duct_sizes=17
+                Mesh/pattern/duct_intervals=1
+                Mesh/pattern/duct_block_ids=30
+                Mesh/pattern/duct_block_names="duct_block"
+                Mesh/pattern/uniform_mesh_on_sides=true
+                Mesh/pattern/boundary_element_type=QUAD8
+                Mesh/hex_1/tri_element_type=TRI6
+                Mesh/hex_1/quad_element_type=QUAD8
+                --mesh-only "patterned_hex_quad8.e"'
+    requirement ='The system shall stitch hexagon meshes to form a large mesh with hexagon boundary using QUAD8 elements.'
```
GiudGiud, line 38:
```suggestion
    requirement ='The system shall be able to stitch hexagonal meshes to form a larger patterned mesh with a hexagonal boundary using QUAD8 elements.'
```

GiudGiud: same for the others please
no need to adapt others

GiudGiud (changes requested):
thanks for the contribution!

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
```diff
@@ -831,6 +838,32 @@ PatternedCartesianMeshGenerator::generate()
...
       }
     }
+
+    // adjust quadratic mid-edge nodes after boundary node redistribution
+    if (_boundary_quad_elem_type != QUAD_ELEM_TYPE::QUAD4)
+    {
+      for (auto elem = out_mesh->elements_begin(); elem != out_mesh->elements_end(); ++elem)
+      {
+        // only adjust elements on outer boundary
+        if (!(*elem)->on_boundary())
```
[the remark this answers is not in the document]
aaronoaks: My understanding is the pattern needs to be full at this stage. Any holes in the desired pattern would need to be labeled and then deleted in a later MG.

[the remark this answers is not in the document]
GiudGiud: even if the pattern is full one of the meshes in the pattern could have a hole though?
if the pattern is made of 3D pin meshes we would hit this too

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
```diff
@@ -831,6 +838,32 @@ PatternedCartesianMeshGenerator::generate()
...
+
+        // adjust point 5
+        Point pt_5 = ((*elem)->point(1) + (*elem)->point(2)) / 2.0;
+        out_mesh->add_point(pt_5, (*elem)->node_ptr(5)->id());
+
+        // adjoin point 7
+        Point pt_7 = ((*elem)->point(0) + (*elem)->point(3)) / 2.0;
+        out_mesh->add_point(pt_7, (*elem)->node_ptr(7)->id());
+
+        // adjust point 8 for QUAD9
```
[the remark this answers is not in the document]
aaronoaks: reworded

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
aaronoaks: reworded

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
aaronoaks: This was left in from an earlier attempt by mistake, it has been removed.

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
aaronoaks: This should convert only the boundary. This mesh should be stitched to the pattern after returning from this function.

modules/reactor/test/tests/meshgenerators/patterned_hex_mesh_generator/tests
aaronoaks: reworded

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
aaronoaks: The existing algorithm generates entire boundary region, and then if the `uniform_mesh_on_sides` option is used then the MG redistributes the boundary nodes on each side so that they are uniformly spaced. When the elements are QUAD4 this is fine, but for QUAD8/9 the mid-edge nodes staying in their original locations results in weird looking elements. This section moves the mid-edge nodes to midpoints based on the new boundary node locations so the element shape is better.

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
```diff
@@ -118,6 +118,11 @@ PatternedCartesianMeshGenerator::validParams()
       "interface_boundary_id_shift_pattern",
       "User-defined shift values for each pattern cell. A double-indexed array starting with the "
       "upper-left corner.");
+  MooseEnum quad_elem_type("QUAD4 QUAD8 QUAD9", "QUAD4");
+  params.addParam<MooseEnum>(
+      "boundary_element_type",
```
miaoyinb, line 123: "boundary" could be a confusing term here as it is used with "element". Maybe something like "boundary_region_element_type" or "quad_element_type_in_boundary_region"?

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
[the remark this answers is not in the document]
miaoyinb: As we allow users to use customized meshes as input here (need meta data added), I agree we should consider that these meshes have holes.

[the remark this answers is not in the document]
aaronoaks: True, I don't think that is the intended use for the pattern input meshes but could probably be done. We have the outer sideset ID at this point in the algorithm. Is there a function to filter down to the elements with a side on that sideset?

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
```diff
@@ -831,6 +838,32 @@ PatternedCartesianMeshGenerator::generate()
...
+    // adjust quadratic mid-edge nodes after boundary node redistribution
+    if (_boundary_quad_elem_type != QUAD_ELEM_TYPE::QUAD4)
+    {
+      for (auto elem = out_mesh->elements_begin(); elem != out_mesh->elements_end(); ++elem)
+      {
+        // only adjust elements on outer boundary
+        if (!(*elem)->on_boundary())
+          continue;
+
+        // adjust point 5
```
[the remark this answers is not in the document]
aaronoaks: reworded

modules/reactor/src/meshgenerators/PatternedHexMeshGenerator.C
```diff
@@ -861,6 +868,32 @@ PatternedHexMeshGenerator::generate()
         out_mesh->add_point(p_tmp, node_azi_list[i * side_intervals + j - 1].second);
       }
     }
+
+    // adjust quadratic mid-edge nodes after boundary node redistribution
```
miaoyinb, line 872: I think this can be a method that can be called by both Cart and Hex.

modules/reactor/doc/content/source/meshgenerators/PatternedCartesianMeshGenerator.md
```diff
@@ -21,6 +21,9 @@ In order to generate the square patterned mesh, three fundamental parameters are
...
 
+ By default, the generated boundary mesh consists of linear elements. If quadratic elements are preferred, users can set [!param](/Mesh/PatternedCartesianMeshGenerator/boundary_element_type) as `QUAD8` or `QUAD9`.
+
+
```
miaoyinb, line 26:
```suggestion
```

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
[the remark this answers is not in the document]
GiudGiud: from a boundary_info (which we can get from the mesh) you can get the tuples (sideset id, element id, side index)

miaoyinb (commented):
just some minor comments.

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
GiudGiud: let's have some of this explanation as a comment right before we do it in both MGs?

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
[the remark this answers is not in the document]
miaoyinb: You might use [`libMesh::BoundaryInfo::boundary_ids()`](https://mooseframework.inl.gov/docs/doxygen/libmesh/classlibMesh_1_1BoundaryInfo.html#a022bf188585426a4660dc4f4d2e722f7).

modules/reactor/src/meshgenerators/PatternedHexMeshGenerator.C
aaronoaks: extracted to common base class

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
aaronoaks: renamed to `boundary_region_element_type`

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
aaronoaks: added

modules/reactor/src/meshgenerators/PatternedCartesianMeshGenerator.C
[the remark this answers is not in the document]
aaronoaks: limited using boundary_info

modules/reactor/src/meshgenerators/PatternedHexMeshGenerator.C
```diff
@@ -120,6 +120,11 @@ PatternedHexMeshGenerator::validParams()
       "interface_boundary_id_shift_pattern",
       "User-defined shift values for each pattern cell. A double-indexed array starting with the "
       "upper-left corner.");
+  MooseEnum quad_elem_type("QUAD4 QUAD8 QUAD9", "QUAD4");
+  params.addParam<MooseEnum>(
+      "boundary_region_element_type",
```
GiudGiud, line 125: let's add this to the external boundary parameter group (see existing addParamNamesToGroup below)

modules/reactor/src/meshgenerators/PatternedHexMeshGenerator.C
```diff
@@ -861,6 +868,14 @@ PatternedHexMeshGenerator::generate()
...
       }
     }
+
+    // if quadratic elements are used, additional nodes need to be adjusted based on the new
+    // boundary node locations. adjust side mid-edge nodes to the midpoints of the corner
+    // points, and if QUAD9, adjust center point to new centroid.
+    if (_boundary_quad_elem_type != QUAD_ELEM_TYPE::QUAD4)
+    {
+      adjustPeripheralQuadraticElements(*out_mesh, _boundary_quad_elem_type);
+    }
```
GiudGiud, line 878:
```suggestion
      adjustPeripheralQuadraticElements(*out_mesh, _boundary_quad_elem_type);
```

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
```diff
@@ -1213,34 +1215,87 @@ PolygonMeshGeneratorBase::buildSimplePeripheral(
...
+  {
+    boundary_id_type boundary_id = std::get<2>(side_item);
+    dof_id_type elem_id = std::get<0>(side_item);
+
+    if (boundary_id == OUTER_SIDESET_ID)
+      elem_set.insert(elem_id);
+  }
+
+  // adjust nodes for outer boundary elements
+  for (dof_id_type elem_id : elem_set)
```
GiudGiud, line 1278:
```suggestion
  for (const auto elem_id : elem_set)
```

GiudGiud (commented):
looks good to me after those

GiudGiud (approved):

GiudGiud:
Test failures (timeouts in solid mech, zapdos and a threaded NS)

PR #28171, opened 2024-07-18
Quadratic elements for peripheral ring

### LINKED ISSUE #28169: Enable Quadratic Elements in PeripheralRingMeshGenerator
### Motivation

Original Reactor Module tools are only capable of creating meshes with linear elements, which limits its applications. 

Efforts have been made to enable quadratic element capabilities. This enhancement follows up on #26986 #28164

### Design

Provide quadratic element (QUAD9) as an option in `PeripheralRingMeshGenerator`.

### Impact

miaoyinb:
closes #28169


miaoyinb:
@GiudGiud FYI. Currently it can only do QUAD4 and QUAD9, I can add QUAD8 if needed.

miaoyinb:
@eshemon FYI

modules/reactor/test/tests/meshgenerators/peripheral_ring_mesh_generator/tests
```diff
@@ -77,6 +77,15 @@
     recover = false
     mesh_mode = 'REPLICATED'
   []
+  [peripheral_ring_boundary_layers_quadratic]
+    type = 'Exodiff'
+    input = 'ring_quadratic.i'
+    exodiff = 'ring_boundary_layers_quadratic_in.e'
+    cli_args = '--mesh-only ring_boundary_layers_quadratic_in.e'
+    requirement ='The system shall be able to add a peripheral ring consistin of quandratic with boundary layers to the input mesh and preserve volumes even with polygonization.'
```
GiudGiud, line 85:
```suggestion
    requirement ='The system shall be able to add a peripheral ring consisting of quadratic elements with boundary layers to the input mesh and preserve volumes despite polygonization.'
```

framework/include/utils/FillBetweenPointVectorsTools.h
```diff
@@ -296,6 +296,31 @@ bool isCurveOpenSingleSegment(MeshBase & mesh,
  * @param mesh input mesh that contains the node series to be examined
  * @param max_node_radius the maximum radius of the nodes in the series
  * @param ordered_node_list the ordered node ids on the given curve
+ * @param node_assm the assembly of the node id pairs that correspond vertices of a series of sides
```
GiudGiud, line 299:
```suggestion
 * @param node_assm the assembly of the node id pairs that correspond to vertices of the sides on the closed loop
```

framework/include/utils/FillBetweenPointVectorsTools.h
```diff
@@ -296,6 +296,31 @@ bool isCurveOpenSingleSegment(MeshBase & mesh,
...
+ * @param midpoint_node_list the node ids of the midpoints of the sides for quadratic sides
+ * @param origin_pt origin position of the given mesh (used for azimuthal angle calculation)
+ * @param is_closed_loop whether the series of nodes form a closed loop with consecutive nodes's
+ * azimuthal angles change monotonically
```
GiudGiud, line 303:
```suggestion
 * azimuthal angles changing monotonically
```

framework/include/utils/FillBetweenPointVectorsTools.h
GiudGiud: if it s correct

GiudGiud (commented):
will look at the code latr

modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
```diff
@@ -150,46 +150,88 @@ PeripheralRingMeshGenerator::generate()
...
+  {
+    if (std::get<2>(side) == _input_mesh_external_bid)
+    {
+      const auto etype = input_mesh->elem_ptr(std::get<0>(side))->side_type(std::get<1>(side));
+      if (etype == EDGE2)
+        linear_side = true;
+      else if (etype == EDGE3)
+        quadratic_side = true;
+      else
+        paramError("input", "Input contains elements that are not supported by this generator.");
```
pbehne, line 180: Please add a `RunException` test for this.

modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
```diff
@@ -150,46 +150,88 @@ PeripheralRingMeshGenerator::generate()
...
+    }
+  }
+  if (linear_side && quadratic_side)
+    paramError("input", "Input contains mixed element types on the external boundary.");
```
pbehne, line 184: Please add a `RunException` test for this.

modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
```diff
@@ -440,14 +503,32 @@ PeripheralRingMeshGenerator::innerBdryLayerNodesDefiner(
     // To mitigate this flipping issue, we check the node flipping here using the cross product of
     // neighboring node-to-origin vectors. Flipped nodes are marked and excluded during the
     // follow-up interpolation.
-    if (!MooseUtils::absoluteFuzzyEqual(input_bdry_angles[i], M_PI / 2.0))
+    if (!MooseUtils::absoluteFuzzyEqual(input_bdry_angles[i], M_PI / 2.0, 1.0e-4))
```
pbehne, line 506: Why the tolerance change?

modules/reactor/test/tests/meshgenerators/peripheral_ring_mesh_generator/ring_quadratic.i
```diff
@@ -0,0 +1,33 @@
...
+      type = PolygonConcentricCircleMeshGenerator
+      num_sides = 8
+      num_sectors_per_side = '4 4 4 4 4 4 4 4'
+      background_intervals = 2
+      polygon_size = 5.0
+
+      preserve_volumes = on
+
+      tri_element_type = TRI6
+      quad_element_type = QUAD8
```
pbehne, line 12: Why `QUAD8` instead of `QUAD9`? Just curious.

pbehne (changes requested):
Please put a short description of what this PR does in the header of the PR, even if it is just copying and pasting the issue.

While reviewing, I cleaned up some of the changed files (not really related to your PR). I'd like to get these changes in with this PR. May I push to your branch (you may have to add me as a collaborator to your fork of MOOSE). If not, I can also make a PR with my changes to your fork (after you add me as a collaborator).

modules/reactor/test/tests/meshgenerators/peripheral_ring_mesh_generator/ring_quadratic.i
GiudGiud: this PR adds quad9 for the periphery. for the mesh inside, quad8 will use less memory and is just as compatible with the boundary

miaoyinb:
> Please put a short description of what this PR does in the header of the PR, even if it is just copying and pasting the issue.
> 
> While reviewing, I cleaned up some of the changed files (not really related to your PR). I'd like to get these changes in with this PR. May I push to your branch (you may have to add me as a collaborator to your fork of MOOSE). If not, I can also make a PR with my changes to your fork (after you add me as a collaborator).

I think the PR template says it does not need any description if the issue # is linked in the header. Anyway I will add you as a collaborator, please do a PR there so that I could take a look. Thanks.

modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
miaoyinb: The previous `PeripheralRingMeshGenerator::innerBdryLayerNodesDefiner()` has a bug. It would have issue in some cases when the inner boundary layer is too thick. This, along with the following additions, is part of the fix.

modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
miaoyinb: Does current MOOSE have an easy way to load/create a third-order 2D mesh to trigger this error?

pbehne:
> > Please put a short description of what this PR does in the header of the PR, even if it is just copying and pasting the issue.
> > While reviewing, I cleaned up some of the changed files (not really related to your PR). I'd like to get these changes in with this PR. May I push to your branch (you may have to add me as a collaborator to your fork of MOOSE). If not, I can also make a PR with my changes to your fork (after you add me as a collaborator).
> 
> I think the PR template says it does not need any description if the issue # is linked in the header. Anyway I will add you as a collaborator, please do a PR there so that I could take a look. Thanks.

Thanks for pointing out the template instructions. Since you have an issue number, could you go ahead and delete the empty headers (Reason, Design, Impact)?

PR is up here: https://github.com/miaoyinb/moose/pull/5

modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
pbehne: Tagging @roystgnr on this one since I'm not sure. He will likely know off the top of his head.

pbehne: I see that libmesh has a 1D `Edge4` element. I'll see if I can figure out what kind of element uses this.

pbehne: As far as I can tell from looking through libmesh source, `Edge4` is not used as a side in any element. I'll let Roy have the final word, though.

miaoyinb: Right, I do not see any complete third-order 2D elements there.

roystgnr: We currently don't have any 3rd-order elements outside of 1D, so no way to add test coverage for this.

We might be adding some in the next year, so it's good that we're catching that possibility here, but not soon enough that I expect I'll remember this PR when we're finally able to add a test for it.

pbehne (approved):
Thanks for adding this feature! I'm ready to merge if @GiudGiud is.

miaoyinb:
@GiudGiud Thank you for reviewing/merging my other PR. Just a friendly reminder that this PR is still pending.

GiudGiud:
sorry for the delay. I am setting up a new machine the other one died by IT

modules/reactor/test/tests/meshgenerators/peripheral_ring_mesh_generator/ring_quadratic.i
```diff
@@ -0,0 +1,33 @@
...
+  [pccmg]
+      type = PolygonConcentricCircleMeshGenerator
+      num_sides = 8
+      num_sectors_per_side = '4 4 4 4 4 4 4 4'
+      background_intervals = 2
+      polygon_size = 5.0
+
+      preserve_volumes = on
+
+      tri_element_type = TRI6
```
GiudGiud, line 11: If I  comment these two out to make the mesh linear, I get a different volumes for the external ring

GiudGiud (commented):
there's something I dont understand with the test

modules/reactor/test/tests/meshgenerators/peripheral_ring_mesh_generator/ring_quadratic.i
miaoyinb: I will look into this, thanks

miaoyinb: Interesting, paraview says the areas are the same...(706.858 total which is the same as the linear elements values reported by MOOSE)

Maybe we need to take a look at the source code that calculates these printed volume values...

miaoyinb:
```
[Mesh]
  [dir1]
    type = ElementGenerator
    nodal_positions = '0 0 0
                       1 0 0
                       1 1 0
                       0 1 0
                       0.5 -0.5 0
                       1 0.5 0
                       0.5 1 0
                       0 0.5 0
                       0.5 0.5 0'
    element_connectivity = '0 1 2 3 4 5 6 7 8'
    elem_type = 'QUAD9'
  []
[]
```
The element area is 1.25 but MOOSE says 1.333 (or quadratic element area has a different definition?)

GiudGiud: @roystgnr thoughts?

GiudGiud: I think when you say 1.25 you have a triangle between the three slopes at the bottom
shouldnt it be quadratic?

miaoyinb: I just tried to give an extreme case. There might be some rules that I am not aware of. Let's see if @roystgnr has any insights.

miaoyinb: > I think when you say 1.25 you have a triangle between the three nodes at the bottom of the element shouldnt it be quadratic?

Ah, so the edge should be parabolic for QUAD9 element. In that case the radius correction function needs to be modified.

GiudGiud: I think so.
Irrelevant here but I think it might be the same for quad8?

miaoyinb: fixed. sorry for the batch regolding...

modules/reactor/test/tests/meshgenerators/patterned_hex_peripheral_modifier/gold/single_hex_quadratic_in.e
GiudGiud, line 1: the nodes should not have moved for those? the modifying of the periphery was sqauree right?

modules/reactor/test/tests/meshgenerators/peripheral_ring_mesh_generator/gold/ring_boundary_layers_quadratic_in.e
GiudGiud, line 1: let's remove that gold file from commit #1 to avoid the regold

modules/reactor/test/tests/meshgenerators/patterned_hex_peripheral_modifier/gold/single_hex_quadratic_in.e
GiudGiud: true true

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
```diff
@@ -1138,23 +1138,68 @@ PolygonMeshGeneratorBase::quadElemDef(ReplicatedMesh & mesh,
...
+    Real tmp_acc = 0.0;
+    Real tmp_acc_azi = 0.0;
+    for (const auto i : index_range(azi_interval_list))
+    {
+      tmp_acc += std::sin(azi_interval_list[i]);
+      tmp_acc_azi += azi_interval_list[i];
+    }
+    return std::sqrt(tmp_acc_azi / tmp_acc);
+  }
+  else // order == 2
```
GiudGiud, line 1165: @roystgnr for a quad9, does the position of a middle node influence the shape of the outer boundary?

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
```diff
@@ -1138,23 +1138,68 @@ PolygonMeshGeneratorBase::quadElemDef(ReplicatedMesh & mesh,
...
+    // Circle area S_c = pi * r_0 * r_0
+    Real tmp_acc = 0.0;
+    Real tmp_acc_azi = 0.0;
+    for (const auto i : index_range(azi_interval_list))
+    {
+      tmp_acc += std::sin(azi_interval_list[i].first + azi_interval_list[i].second);
+      tmp_acc +=
+          4.0 / 3.0 * std::sin((azi_interval_list[i].first + azi_interval_list[i].second) / 2) *
+          std::sin((azi_interval_list[i].first + azi_interval_list[i].second) / 2) *
+          std::sin((azi_interval_list[i].first + azi_interval_list[i].second) / 2) /
```
GiudGiud, line 1197: MathUtils
```suggestion
          4.0 / 3.0 * Utility::pow<3>(std::sin((azi_interval_list[i].first + azi_interval_list[i].second) / 2)) /
```

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
```diff
@@ -1138,23 +1138,68 @@ PolygonMeshGeneratorBase::quadElemDef(ReplicatedMesh & mesh,
...
+          std::cos(azi_interval_list[i].first / 2) / std::cos(azi_interval_list[i].second / 2);
+      tmp_acc_azi += azi_interval_list[i].first + azi_interval_list[i].second;
```
GiudGiud, line 1199: why both for this case?

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
roystgnr: It does not.  Not for Lagrange mappings, not for Rational Bezier Bernstein mappings.

There are mappings with less locality than the ones we use, but we'll probably never support any directly, only indirectly the way we support IsoGeometric Analysis NURBS mappings - via extraction operators that give more locally-defined elements to integrate on.

GiudGiud: thanks

how exact is the volume computation for second order elements in libmesh? I still see tiny diffs (1e-5) with the correction we have here

miaoyinb: I think that summation of small $\sin^3(\theta)$ compromises the errors...

GiudGiud: could be

are you summing two separate quadratic bands, separated at the mid point of the edge?
I think the whole edge is covered by a single quadratic band?

GiudGiud: > But are you diffing against the exact biquadratic volume, or against a perfect circle? Biquadratics (in Lagrange; I'm working on RBB mesh generator tools) aren't enough to exactly mesh a circle; no polynomial mapping is. The difference between a unit circle and its biquadratic approximation is probably more like 1e-5.

I am diffing against the corrected 1st order mesh, which should match the perfect circle volume exactly already.
While the bi-quadratics do not exactly mesh the circle, we are correcting them here so that the volume is the exact same. This is what this routine is intended for. So yes it's only correcting a very small error already

miaoyinb: > could be
> 
> are you summing two separate quadratic bands, separated at the mid point of the edge? I think the whole edge is covered by a single quadratic band?

Yes, it's a single band determined by the three nodes.

miaoyinb: If we really want to eliminate that 1e-6 relative error, one possible way would be to make a dummy TRI6 mesh with the same external nodes and use its area as the reference to derive a correction factor.

framework/src/utils/FillBetweenPointVectorsTools.C
```diff
@@ -466,7 +467,7 @@ surrogateGenerator(std::vector<Real> & weighted_surrogate_index,
       weighted_surrogate_index[j] += (*it_0 - unweighted_surrogate_index[j - 1]) * wt[it_start - 1];
       weighted_surrogate_index[j] +=
           (unweighted_surrogate_index[j] - *(it_1 - 1)) * wt[it_start + it_dist - 1];
-      for (unsigned int k = 1; k < it_dist; k++)
+      for (const auto k : make_range(long(1), it_dist))
```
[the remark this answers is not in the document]
pbehne: Here is the compiler error when I don't cast to long:
`In file included from /Users/behnpa/projects/moose_quadratic_ring/framework/build/unity_src/utils_Unity.C:20:
/Users/behnpa/projects/moose_quadratic_ring/framework/src/utils/FillBetweenPointVectorsTools.C:470:27: error: no matching function for call to 'make_range'
      for (const auto k : make_range(1, it_dist))`
I tried using auto instead of long, and got:
`/Users/behnpa/projects/moose_quadratic_ring/framework/src/utils/FillBetweenPointVectorsTools.C:470:38: error: 'auto' not allowed here
      for (const auto k : make_range(auto(1), it_dist))
`

modules/reactor/test/tests/meshgenerators/peripheral_ring_mesh_generator/ring_quadratic.i
roystgnr: > Ah, so the edge should be parabolic for QUAD9 element

This is correct.  In the `ElementGenerator` case in particular, 4/3 is the correct area.

framework/src/utils/FillBetweenPointVectorsTools.C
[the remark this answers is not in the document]
pbehne: For loops that start from 1, should we just use the other style of for loop?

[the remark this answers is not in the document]
GiudGiud: unless you find a way to get the right data type for the 1 with various compiling options I would say yes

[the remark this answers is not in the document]
miaoyinb: could be overkill but this would work...

`make_range(castIndexForRange(foo,1),foo)`

```
template <typename T, typename P>
const T
castIndexForRange(const T &, //type_input
                  const P & value_input)
{
  return (T)(value_input);
}
```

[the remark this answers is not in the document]
roystgnr: `make_range()` is just some syntactic sugar.  It's just to make it easier to avoid the typical C-style mistake of `for (foo i=begin(); i != expensive_call_that_shouldnt_be_here(); ++i)`, not because there's anything else we could fix with range-for vs C-style-for loops.  (though I'll bet we have a lot of both kinds of loops that should be `std::foreach`, and I'm thinking of wrapping that to play with different `std::execution::` options...)

`castIndexForRange()` is syntactic Carolina Reaper powder.  Maybe it's an acquired taste?

I think C-style is fine; otherwise `1l` is more terse than `long(1)`

modules/reactor/src/meshgenerators/PolygonMeshGeneratorBase.C
GiudGiud: It s more that if we expect the error to be 1e-6 already with quad-elements (as Roy mentioned), then why do we still see that after correction

I m yet to run this again and make up my mind on this

GiudGiud: it might just be that the error is a larger than expected pre-correction

miaoyinb: I will also run more cases to get a clearer idea what is going on.

GiudGiud:
saw your changes, I ll take a look tomorrow

GiudGiud:
It looks like the background size in the square lattice, when deformation of non-circular regions is set to false, is different between quadratic and linear elements. It affects the regolded files for patterned_cartesian_mg (pattern_sq_) and I think that's it.

Which volume is most important to preserve in that case?

GiudGiud:
Otherwise we are good on all fronts here

miaoyinb:
> It looks like the background size in the square lattice, when deformation of non-circular regions is set to false, is different between quadratic and linear elements. It affects the regolded files for patterned_cartesian_mg (pattern_sq_) and I think that's it.
> 
> Which volume is most important to preserve in that case?

If the non-circular region deformation is activated, the background region is deformed to allow small hexagon/square size that cuts the original patterned unit cells. In that case, we do not care about the volume preserve as they are deformed anyway.

But if the deformation is not activated, the midpoint should be colinear with the vertices and the quandratic elements and linear elements should have the same sizes(?)

GiudGiud:
I think the square size should be the same for both linear and quadractic though? We lose the volume conservation of the background but we impose a square size right?

the diff is in the case with the deformation

miaoyinb:
> I think the square size should be the same for both linear and quadractic though? We lose the volume conservation of the background but we impose a square size right?
> 
> the diff is in the case with the deformation

Right.

GiudGiud:
ok so it's not tied to this PR but seems there's a problem in the PatternedCartesian w/ deformation & second order.
pls let me know if you want to address it here with the regold or if we should create an issue

miaoyinb:
Ah, I figured out the source of the issue. Let's address it here.

miaoyinb:
That should be fixed now.

GiudGiud:
this didnt fix it. If you zoom in on the corner you can see that pattern_sq.e and pattern_sq_quad8.e for the patterned_cartesian_mg dont line up at the pin background / duct background interface

miaoyinb:
> this didnt fix it. If you zoom in on the corner you can see that pattern_sq.e and pattern_sq_quad8.e for the patterned_cartesian_mg dont line up at the pin background / duct background interface

Let me check it again. It seems the hex and cartesian have different issues...

miaoyinb:
The boundary of the pin background and assembly background after deformation is okay to be different between linear and quadratic elements. The size of the outer boundary of the assembly (i.e., inner size of the innermost duct) is important and must be consistent.

GiudGiud (approved):
thanks for the contribution!

Issue #28008, opened 2024-06-27
Improvements for Explicit Contact


TheGreatCid:
### Motivation

The current implementation of explicit contact is based on the work by Heinsteinn "[Contact-impact modeling in explicit transient dynamics](https://www.sciencedirect.com/science/article/abs/pii/S0045782599003424)"

The formulation proposed by Heinsteinn relies on a central difference time integrator where the solution updates comes from a direct calculation of the acceleration from the external and internal forces. 

The current central difference time integrator in MOOSE does not handle the solution update in this way, and thus some concessions had to be made to implement the contact model proposed by Heinsteinn.

It would be pertinent to update the central difference formulation in MOOSE to support a solution update from the direct calculation of acceleration, and then implement the exact form of the aforementioned contact model.

### Design

The change to central difference would add an additional solver option rather than an override of the current method
For the direct acceleration calculation method several changes/additions need to be made
 - A calculation of the mass matrix without the additional "factor tags" used for the current formulation
 - An additional variant of the explicit time integrator solution update to support direct acceleration calculation
 - A way to enforce nodal BCs for the direct calculation method
 - An update the `ComputeTimeDerivatives` member function in `CentralDifference.C`
 - The ability to not include the residual contribution from the inertial forces when using the direct method

For contact, the following changes need to be made
 - Change the initial guess to be based on the mass matrix rather than on the "mass proxy"
 - Change the method to overwrite $v_{n+1/2}$ rather than $d_{n+1}$
 - Add support to multi body contact
 - (Probably other aspects that I am not recalling at the moment)

### Impact

The goal is to create a more robust explicit dynamics formulation that supports multi-body contact. Additionally, the direct acceleration calculation method is more straight forward than the current method, so hopefully it is a bit more user friendly. However, I do not expect to see improvements with this method over the current central difference method outside of contact applications.

PR #28175, opened 2024-07-18
Adding direct central difference method to solid mechanics

TheGreatCid:
This is a follow up to #27991.

## Reason

This PR addresses the first half of https://github.com/idaholab/moose/issues/28008

## Design 

The "direct" solution update, is calculated as the following:

$a_n=M^{-1}(F^{ext}-F^{int})$
$v_{n+1/2}=v_{n-1/2}+\frac{\Delta t_(old) + \Delta t}{2} a_n$
$d_{n+1}=d_n + \Delta t v_{n+1/2}$

A new class is added to implement the above formulation: `CentralDifferenceDirect`
`CentralDifferenceDirect` performs the solution update while also retrieving the mass matrix from `MassMatrix` which needs to be present in any input file that uses the new time integration method. 

Additionally, a new boundary condition base, `DirectDirichletBCBase` was added to enforce Dirichlet BC's using this new time integrator. It computes the residual force needed to enforce the BC during the solution update.


modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
```diff
@@ -0,0 +1,56 @@
...
+
+  void computeADTimeDerivatives(ADReal & ad_u_dot,
+                                const dof_id_type & dof,
+                                ADReal & ad_u_dotdot) const override
+  {
+    ad_u_dotdot.value();
+    ad_u_dot.value();
+    std::cout << dof << std::endl;
+    return mooseError("NOT SUPPORTED");
+  }
```
hugary1995, line 46: What's going on here?

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
```diff
@@ -0,0 +1,56 @@
...
+
+protected:
+  /**
+   * Helper function that actually does the math for computing the time derivative
+   */
```
hugary1995, line 51:
```suggestion
```

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
```diff
@@ -0,0 +1,194 @@
...
+#include "CentralDifferenceDirect.h"
+#include "MooseTypes.h"
+#include "NonlinearSystem.h"
+#include "FEProblem.h"
+// libMesh includes
+#include "libmesh/nonlinear_solver.h"
+#include "libmesh/sparse_matrix.h"
+#include "DirichletBCBase.h"
+
+registerMooseObject("MooseApp", CentralDifferenceDirect);
```
hugary1995, line 22:
```suggestion
registerMooseObject("SolidMechanicsApp", CentralDifferenceDirect);
```

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
```diff
@@ -0,0 +1,194 @@
...
+  // "Invert" the diagonal mass matrix
+  _mass_matrix_diag.reciprocal();
+
+  // Calculate acceleration
+  auto & accel = *_sys.solutionUDotDot();
+  accel.pointwise_mult(_mass_matrix_diag, _explicit_residual);
+
+  // Scaling the acceleration
+  auto accel_scaled = accel.clone();
+  accel_scaled->scale((_dt + _dt_old) / 2);
```
hugary1995, line 171: This is different from the math you showed in the PR description.

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
```diff
@@ -0,0 +1,194 @@
...
+
+  // Adding old vel to new vel
+  auto & vel = *_sys.solutionUDot();
+  const auto & old_vel = _sys.solutionUDotOld();
+  vel = *old_vel;
+  vel += *accel_scaled;
+
+  // Scale velocity by dt
+  auto vel_scaled = vel.clone();
+  vel_scaled->scale(_dt);
```
hugary1995, line 181: This is however consistent with your math...

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
```diff
@@ -0,0 +1,194 @@
...
+  vel.close();
+  accel.close();
+
+  // Check for convergence by seeing if there is a nan or inf
+  auto sum = _solution_update.sum();
+  converged = std::isfinite(sum);
+
+  // The linear iteration count remains zero
+  _n_linear_iterations = 0;
+  return converged;
```
hugary1995, line 193: This is fine with me for now. In the long term, we need a better convergence check for explicit dynamics, or rather a _divergence check_.

Imagine we use what you have here as a convergence check: When the solve didn't not "converge", the framework handles solve failure, restores the solution, and travel back in time to reattempt the solve with a reduced time step. For implicit, that will work -- as we are moving closer to the basin of a traction. But for explicit, that doesn't help, in general.

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
```diff
@@ -0,0 +1,194 @@
...
+CentralDifferenceDirect::computeTimeDerivatives()
+{
+  /*
+  Because this is called in NonLinearSystemBase
+  this should not actually compute the time derivatives.
+  Calculating time derivatives here will cause issues for the
+  solution update.
+  */
+  return;
+}
```
hugary1995, line 61: Where can I check to see if the second time derivative is implemented correctly according to your math?

hugary1995 (changes requested):

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
hugary1995: nvmd, found it below.

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
```diff
@@ -0,0 +1,194 @@
...
+  _current_time = _fe_problem.time();
+
+  // Set time to the time at which to evaluate the residual
+  _fe_problem.time() = _fe_problem.timeOld();
+  _nonlinear_implicit_system->update();
+
+  // Compute the residual
+  _explicit_residual.zero();
+  _fe_problem.computeResidual(
+      *_nonlinear_implicit_system->current_local_solution, _explicit_residual, _nl.number());
```
hugary1995, line 83: could you help me understand why this _explicit_residual is Fext-Fint in the equation?

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: I agree, for now I was simply copying the current way of doing things found in `ExplicitTimeIntegrator.C`

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: Good catch, I need to update the documentation. This is the correct version

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
TheGreatCid: This is a copy of what was found in `CentralDifference.C` to overcome the override requirement for `ComputeADTimeDerivatives`

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: It will make more sense once I get some test input files pushed in. I've stalled on this PR while I am at WCCM. 
This is the case because Fint will be the residual returned from your stress divergence only. Fext will return any relevant nodal BCs

TheGreatCid: There will be no inertia kernel in the input file

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
hugary1995: But do you need these three lines?
```
    ad_u_dotdot.value();
    ad_u_dot.value();
    std::cout << dof << std::endl;
```

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
hugary1995: Okay, then this is different from the original paper you referenced. I think this will require some documentation (i.e. why (dt+dt_old)/2 instead of dt).

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
hugary1995: > Fint will be the residual returned from your stress divergence only. Fext will return any relevant nodal BCs

That makes sense.

> There will be no inertia kernel in the input file

How do you tag the mass matrix then?

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: I'm guessing the paper I referenced assumed constant dt. The way I've implemented does not

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
TheGreatCid: Those are too suppress compiler messages. It's a bit jank but I assumed it was acceptable since it was already implemented.

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: I've added a test to the PR. This shows how the mass matrix is tagged

TheGreatCid:
I've added perhaps the more questionable aspect of this PR, which is classes for Dirichlet BC enforcement. I've introduced a new base class which accounts for the central difference time integration to return the proper force required to enforce the Dir BC. I've added a constant BC and function BC on top of this base class.

TheGreatCid:
Not sure if this is an issue, but this implementation only outputs midpoint velocities. So the velocity fields will be $v_{n+1/2}$.

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
lindsayad: >This is a copy of what was found in CentralDifference.h to overcome the override requirement for ComputeADTimeDerivatives

Not quite sure what you mean here. There are no errors in the AD routine in `CentralDifference` in upstream MOOSE

TheGreatCid: Hmm. you're right. I'm not sure what I mean either. I can't remember where I got this from. 
What is the correct way to handle this case?

modules/solid_mechanics/include/bcs/DirectDirichletBCBase.h
```diff
@@ -0,0 +1,62 @@
...
+   */
+  virtual Real computeQpValue() = 0;
+
+  /// Whether or not the value is to be preset
+  const bool _preset;
+
+  /// Nonlinear implicit system, if applicable; otherwise, nullptr
+  // NonlinearImplicitSystem * _nonlinear_implicit_system;
+  NonlinearImplicitSystem * _nonlinear_implicit_system;
+
```
lindsayad, line 60:
```suggestion
```

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
```diff
@@ -0,0 +1,49 @@
...
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ExplicitTimeIntegrator.h"
+#include "MooseError.h"
+#include "libmesh/sparse_matrix.h"
+
+class MassMatrixHelper;
```
lindsayad, line 16:
```suggestion
```

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
```diff
@@ -0,0 +1,49 @@
...
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ExplicitTimeIntegrator.h"
+#include "MooseError.h"
+#include "libmesh/sparse_matrix.h"
```
lindsayad, line 14: can forward declare

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
```diff
@@ -0,0 +1,66 @@
...
+    _preset(false),
+    _nonlinear_implicit_system(dynamic_cast<NonlinearImplicitSystem *>(&_sys.system())),
+    _mass_matrix()
+{
+}
+
+void
+DirectDirichletBCBase::initialSetup()
+{
+  _mass_matrix = &_nonlinear_implicit_system->get_system_matrix();
```
lindsayad, line 33: this confuses me. In your test you have an extra tagged matrix called "mass" which is what the `MassMatrix` kernel adds into. What is in the system matrix? It seems like it's not the mass. What I would expect to see is the `_mass_matrix` corresponding to the tagged mass matrix, not the system matrix. Else the `_mass_matrix` data member should have some different name

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
```diff
@@ -0,0 +1,66 @@
...
+  InputParameters params = NodalBC::validParams();
+  params.addParam<TagName>("mass_matrix_tag", "mass", "The tag for the mass matrix");
+
+  return params;
+}
+
+DirectDirichletBCBase::DirectDirichletBCBase(const InputParameters & parameters)
+  : NodalBC(parameters),
+    _preset(false),
+    _nonlinear_implicit_system(dynamic_cast<NonlinearImplicitSystem *>(&_sys.system())),
```
lindsayad, line 25:
```suggestion
```

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
lindsayad: If you still need to access the system matrix in this class, I would do it with `_sys.system().get_matrix("System Matrix")`. In that case you will get a map find error if there is no system matrix which could happen if the system is not an `ImplicitSystem`. Using `_sys.system().get_matrix("System Matrix")` also has the conceptual benefit of working with both `NonlinearImplicitSystem` and `LinearImplicitSystem`.

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
```diff
@@ -0,0 +1,66 @@
...
+Real
+DirectDirichletBCBase::computeQpResidual()
+{
+  // Get dof for current var
+  auto dofnum =
+      _current_node->dof_number(_sys.number(), variable().number(), _sys.dofMap().sys_number());
+
+  // Get lumped component of mass matrix
+  Real masslump = 0;
+  for (int j = 0; j < int(_mass_matrix->n()); j++)
```
lindsayad, line 57:
```suggestion
  for (const auto j : make_range(_mass_matrix->n()))
```

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
```diff
@@ -0,0 +1,66 @@
...
+    const dof_id_type & dof_idx = _var.nodalDofIndex();
+    current_solution.set(dof_idx, computeQpValue());
+  }
+}
+
+Real
+DirectDirichletBCBase::computeQpResidual()
+{
+  // Get dof for current var
+  auto dofnum =
```
lindsayad, line 52:
```suggestion
  const auto dofnum =
```

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
```diff
@@ -0,0 +1,161 @@
...
+  _n_linear_iterations = 0;
+
+  _current_time = _fe_problem.time();
+
+  auto & mass_matrix = _nonlinear_implicit_system->get_system_matrix();
+
+  // Compute the mass matrix
+  if (!_constant_mass || (_constant_mass && _t_step == 1))
+    _fe_problem.computeJacobianTag(
+        *_nonlinear_implicit_system->current_local_solution, mass_matrix, mass_tag);
```
lindsayad, line 86: I see. It looks like you are using the mass matrix tag ID, but you are not using the extra matrix that is added for that tag ID. So right now you will have two matrices in your system, one of which is not used. Maybe you should just add the `MassMatrix` contributions directly into the "system" matrix tag instead of creating the extra matrix with `extra_tag_matrices`

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,135 @@
...
+        function = forcing_fn
+    []
+[]
+
+[Executioner]
+    type = Transient
+
+    [TimeIntegrator]
+        type = CentralDifferenceDirect
+        mass_matrix_tag = 'mass'
```
lindsayad, line 110:
```suggestion
        mass_matrix_tag = 'system'
```

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,135 @@
...
+[]
+
+[Kernels]
+    [DynamicSolidMechanics]
+        displacements = 'disp_x disp_y'
+    []
+    [massmatrix]
+        type = MassMatrix
+        density = 1
+        matrix_tags = 'mass'
```
lindsayad, line 52:
```suggestion
        matrix_tags = 'system'
```

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,135 @@
...
+        variable = disp_x
+    []
+    [massmatrix_y]
+        type = MassMatrix
+        density = 1
+        matrix_tags = 'mass'
```
lindsayad, line 58:
```suggestion
        matrix_tags = 'system'
```

modules/solid_mechanics/test/tests/dynamics/time_integration/tests
```diff
@@ -72,4 +72,13 @@
     design = "Dynamics.md"
     issues = "#18687"
   []
+
+  [direct_central_difference]
+    type = 'Exodiff'
+    input = 'direct_central_difference.i'
+    exodiff = 'direct_central_difference_out.e'
+    abs_zero = 1e-09
+    requirement = "The solid mechanics app shall support a generalized form of central difference time integration"
```
lindsayad, line 81: I think you should be more descriptive than "generalized", communicating what this new time integrator does that the framework integrator does not

lindsayad (changes requested):

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: Hmm. Perhaps I am confused as to how these systems works. My understanding is that the mass matrix is computed using `computeJacobianTag`. The `MassMatrix` tag is used in `computeJacobianTag` to properly calculate the mass matrix from `MassMatrix.C`. 
 It is then stored in `_nonlinear_implicit_system` as that is where `mass_matrix` is referencing. This leaves open the option to use a constant mass.

TheGreatCid: Does the mass matrix tag ID actually generate a matrix, even without the call to `computeJacobianTag`?

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
TheGreatCid: The logic here is that `CentralDifferenceDirect::performExplicitSolve` compute the mass matrix and stores it in `_nonlinear_implicit_system->get_system_matrix()` before the residual is calculated. `DirectDirichletBCBase.C` is retrieving that matrix. 
I don't believe the system matrix is ever written by anything else. This is the same logic currently used in `ActuallyExplicitEuler`.

modules/solid_mechanics/test/tests/dynamics/time_integration/tests
TheGreatCid: Fair point, Ill change that

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
TheGreatCid: I don't understand the reason, could you explain?

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: Oh I see what you are saying. Ignore the previous comments. That make sense.

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: [9ece25d](https://github.com/idaholab/moose/pull/28175/commits/9ece25dbf1c22cb308158e04798e1903c05d45dc)

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: [9ece25d](https://github.com/idaholab/moose/pull/28175/commits/9ece25dbf1c22cb308158e04798e1903c05d45dc)

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: [9ece25d](https://github.com/idaholab/moose/pull/28175/commits/9ece25dbf1c22cb308158e04798e1903c05d45dc)

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
TheGreatCid: I believe this is resolved here, by tagging the system matrix instead of generating a new tag.
[9ece25d](https://github.com/idaholab/moose/pull/28175/commits/9ece25dbf1c22cb308158e04798e1903c05d45dc)

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
hugary1995: https://stackoverflow.com/questions/9906402/should-one-use-forward-declarations-instead-of-includes-wherever-possible

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
TheGreatCid: [62bfb7d](https://github.com/idaholab/moose/pull/28175/commits/62bfb7d35c75281000aea0b1f35f3dd6acffa259)

modules/solid_mechanics/include/bcs/DirectDirichletBCBase.h
TheGreatCid: [62bfb7d](https://github.com/idaholab/moose/pull/28175/commits/62bfb7d35c75281000aea0b1f35f3dd6acffa259)

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
TheGreatCid: I wasn't aware of this aspect of C++, thanks

TheGreatCid: [d8b86f1](https://github.com/idaholab/moose/pull/28175/commits/d8b86f16fa90d2992962003d08f61c78bf953921)

modules/solid_mechanics/test/tests/dynamics/time_integration/tests
TheGreatCid: [d9b75bb](https://github.com/idaholab/moose/pull/28175/commits/d9b75bb14970b3f38865683f49e6bac2fd62edc1)

modules/solid_mechanics/include/bcs/DirectDirichletBCBase.h
```diff
@@ -0,0 +1,62 @@
...
+public:
+  static InputParameters validParams();
+
+  DirectDirichletBCBase(const InputParameters & parameters);
+
+  virtual void initialSetup() override;
+
+  void computeValue(NumericVector<Number> & current_solution);
+
+  bool preset() const { return _preset; }
```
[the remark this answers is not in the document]
lindsayad: It would if your class was derived from `DirichletBCBase` but your class is derived directly from `NodalBC`. Another way to think about it is that since this function is not an overridden virtual it is impossible for it to be called from the framework

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
lindsayad: If you don't want to support AD, you can just make this function a one-line `mooseError`. And then you can delete the function parameters so that you don't get unused variable warnings

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "ExplicitTimeIntegrator.h"
+
+// Forward declarations
+namespace libMesh
+{
+template <typename T>
+class SparseMatrix;
+}
+class MooseError;
```
lindsayad, line 20: Since you actually call the function here in the header you need the definition of `mooseError` available so you actually need to `#include "MooseError.h"`. However, you're "getting lucky" because it's already indirectly included via `#include "ExplicitTimeIntegrator.h"`. I would remove this line or alternatively, move your `computeADTimeDerivatives` implementation to the .C file

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,135 @@
...
+#
+# Testing that the first and second time derivatives
+# are calculated correctly using the Central Difference Direct
+# method
+###########################################################
+
+# [Problem]
+#     extra_tag_matrices = 'mass'
+# []
+
```
lindsayad, line 14:
```suggestion
```

modules/solid_mechanics/test/tests/dynamics/time_integration/tests
```diff
@@ -72,4 +72,16 @@
     design = "Dynamics.md"
     issues = "#18687"
   []
+
+  [direct_central_difference]
+    type = 'Exodiff'
+    input = 'direct_central_difference.i'
+    exodiff = 'direct_central_difference_out.e'
+    abs_zero = 1e-09
+    requirement = "The solid mechanics app shall support a form of central difference time integration
```
lindsayad, line 81: this is a great requirement, thanks!

lindsayad (commented):
this is looking quite good to me

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
lindsayad: agreed!

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: [ab8b9fe](https://github.com/idaholab/moose/pull/28175/commits/ab8b9fe85692cb185dda7f6264efbf1c75a0e597)

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
TheGreatCid: [ab8b9fe](https://github.com/idaholab/moose/pull/28175/commits/ab8b9fe85692cb185dda7f6264efbf1c75a0e597)

I just added the include statement and removed the forward declaration

modules/solid_mechanics/include/bcs/DirectDirichletBCBase.h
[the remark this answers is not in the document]
TheGreatCid: [ab8b9fe](https://github.com/idaholab/moose/pull/28175/commits/ab8b9fe85692cb185dda7f6264efbf1c75a0e597)

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
TheGreatCid: [ab8b9fe](https://github.com/idaholab/moose/pull/28175/commits/ab8b9fe85692cb185dda7f6264efbf1c75a0e597)

TheGreatCid:
Should I manually squash all the additional commits, or is the repo set to automatically do that?

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
```diff
@@ -38,13 +37,9 @@ class CentralDifferenceDirect : public ExplicitTimeIntegrator
...
 
-  void computeADTimeDerivatives(ADReal & ad_u_dot,
-                                const dof_id_type & dof,
-                                ADReal & ad_u_dotdot) const override
+  void computeADTimeDerivatives(ADReal &, const dof_id_type &, ADReal &) const override
   {
-    ad_u_dotdot.value();
-    ad_u_dot.value();
-    return mooseError("On %d - NOT SUPPORTED", dof);
+    return mooseError("NOT SUPPORTED");
```
lindsayad, line 42: no need to `return` here

lindsayad:
down to 2 conversations!

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
hugary1995: Why did you choose (dt+dt_old)/2? That doesn't seem to follow the constant acceleration assumption.

hugary1995: Check my math, assuming constant acceleration and linear velocity
```
v_{n+1/2} = v_{n-1/2} + dt_old / 2 * a_{n-1} + dt / 2 * a_n
```
Your implementation assumes
```
(dt_old+dt) / 2 * a_n = dt_old / 2 * a_{n-1} + dt / 2 * a_n
```
which is questionable.

hugary1995:
I know one of the 2 remaining conversations is mine. I'll have to insist on resolving that one before merging.

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: The first statement should be written as 
`v_{n+1/2} = v_{n-1/2} + dt_old / 2 * a_n + dt / 2 * a_n`

Really this is just defining an effective time step to jump between mid points when the time step is not constant. 
`dt_eff = (dt_{n} + dt_{n+1})/2`
`v_{n+1/2} = v_{n-1/2} +  dt_eff  * a_n`

[Abaqus uses the same method. ](https://classes.engineering.wustl.edu/2009/spring/mase5513/abaqus/docs/v6.6/books/gsx/default.htm?startat=ch03s02.html)

If `t_{n-1} = 1`, `t_{n}=2` and `t_{n+1} = 2.5`, we want `t_{n-1/2}+dt_eff = 2.25 = t_{n+1/2}`
In this case `dt_eff=0.75`, so this would return the correct value for`t_{n+1/2}`

TheGreatCid:
I renamed `CentralDifferenceDirect` to `DirectCentralDifference` for consistency.

modules/solid_mechanics/include/timeintegrators/CentralDifferenceDirect.h
TheGreatCid: [736c3f4](https://github.com/idaholab/moose/pull/28175/commits/736c3f45bea7fbcebac66ba88f30a5be7e0b0c79)

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
hugary1995: Okay, I see. So this assumes constant acceleration between midpoints, not between time step begin and end. Your last comment can be a good documentation. Make sure you also reference the Abaqus manual. I'll resolve this conversation once I see the relevant documentation added.

lindsayad:
> Should I manually squash all the additional commits, or is the repo set to automatically do that?

the repo does not automatically squash

TheGreatCid:
> > Should I manually squash all the additional commits, or is the repo set to automatically do that?
> 
> the repo does not automatically squash

Would you prefer a consolidation of the commits before merging?

GiudGiud:
It would be good to squash the "Apply suggestions from code review" with the commits they belong to.
And more importantly, it would be helpful to have messages that are more descriptive than "apply suggestions" or "address reviewer comments"

modules/solid_mechanics/src/timeintegrators/CentralDifferenceDirect.C
TheGreatCid: [f4073ac](https://github.com/idaholab/moose/pull/28175/commits/f4073ace5c7337485b751e904d6d55d78ed7b604)

I've updated the documentation here

TheGreatCid:
Well it looks like I made some mistakes with my rebasing... I'll need to clean this up.

TheGreatCid:
Ok. Back to normal. I'll be more careful rebasing this time.

TheGreatCid:
Alright, I've correctly consolidated my commits

modules/solid_mechanics/include/bcs/DirectDirichletBCBase.h
```diff
@@ -0,0 +1,49 @@
...
+#include "NodalBC.h"
+
+// libMesh forward declarations
+namespace libMesh
+{
+template <typename T>
+class SparseMatrix;
+template <typename T>
+class NumericVector;
+class NonlinearImplicitSystem;
```
lindsayad, line 21:
```suggestion
```

modules/solid_mechanics/include/bcs/DirectDirichletBCBase.h
```diff
@@ -0,0 +1,49 @@
...
+{
+public:
+  static InputParameters validParams();
+
+  DirectDirichletBCBase(const InputParameters & parameters);
+
+  virtual void initialSetup() override;
+
+  bool preset() const { return _preset; }
+
```
lindsayad, line 36:
```suggestion
```

modules/solid_mechanics/include/bcs/DirectDirichletBCBase.h
```diff
@@ -0,0 +1,49 @@
...
+  virtual Real computeQpResidual() override;
+
+  /**
+   * Compute the value of the DirichletBC at the current quadrature point
+   */
+  virtual Real computeQpValue() = 0;
+
+  /// Whether or not the value is to be preset
+  const bool _preset;
+
```
lindsayad, line 47:
```suggestion
```

modules/solid_mechanics/include/timeintegrators/DirectCentralDifference.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "ExplicitTimeIntegrator.h"
+
+// Forward declarations
+namespace libMesh
+{
+template <typename T>
+class SparseMatrix;
+}
+class MooseError;
```
lindsayad, line 20:
```suggestion
```

modules/solid_mechanics/include/timeintegrators/DirectCentralDifference.h
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Implements a truly explicit (no nonlinear solve) first-order, forward Euler
+ * time integration scheme.
```
lindsayad, line 24: Seems like not quite the right doxygen

modules/solid_mechanics/include/timeintegrators/DirectCentralDifference.h
```diff
@@ -0,0 +1,53 @@
...
+  virtual void computeTimeDerivatives() override;
+
+  virtual void solve() override;
+  virtual void postResidual(NumericVector<Number> & residual) override;
+
+  virtual bool performExplicitSolve(SparseMatrix<Number> & mass_matrix) override;
+
+  // // Applying pre-set nodal BCs
+  // void setPresetBCs();
+
```
lindsayad, line 43:
```suggestion
```

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
lindsayad, line 14:
```suggestion
```

lindsayad (changes requested):
seems like some changes might have gotten lost in the rebase

TheGreatCid:
> seems like some changes might have gotten lost in the rebase

Thank you for catching this.

modules/solid_mechanics/include/timeintegrators/DirectCentralDifference.h
TheGreatCid: Fixed this in the latest push

lindsayad (approved):

modules/solid_mechanics/doc/content/source/bcs/DirectDirichletBC.md
```diff
@@ -0,0 +1,21 @@
...
+
+This applies a Dirichlet BC meant to be used in conjunction with [DirectCentralDifference](source/timeintegrators/DirectCentralDifference.md).
+
+BC's are applied by calculating the residual force needed,$\mathbf{F}_g$, to enforce BC's during a central difference solution update.
+
+At each boundary node:
+
+\begin{equation}
+    F_g = \frac{u_g-u_n}{\Delta t^2} - \frac{v_{n-\frac{1}{2}}}{\Delta t}
+\end{equation}
```
hugary1995, line 15: Explain the meaning of the variables.

modules/solid_mechanics/doc/content/source/bcs/DirectFunctionDirichletBC.md
```diff
@@ -0,0 +1,21 @@
...
+
+This applies a function Dirichlet BC meant to be used in conjunction with [DirectCentralDifference](source/timeintegrators/DirectCentralDifference.md).
+
+BC's are applied by calculating the residual force needed to enforce BC's during a central difference solution update.
+
+At each boundary node:
+
+\begin{equation}
+    F_g = \frac{u_g-u_n}{\Delta t^2} - \frac{v_{n-\frac{1}{2}}}{\Delta t}
+\end{equation}
```
hugary1995, line 15: Explain the meaning of the variables. Point out u_g is the function value.

modules/solid_mechanics/doc/content/source/bcs/DirectDirichletBC.md
hugary1995: Provide an equation for Delta t for changing step size.

modules/solid_mechanics/doc/content/source/bcs/DirectFunctionDirichletBC.md
hugary1995: Provide an equation for Delta t for charging step size.

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
```diff
@@ -0,0 +1,53 @@
...
+  // Get lumped component of mass matrix
+  Real masslump = 0;
+  for (const auto j : make_range(_mass_matrix->n()))
+    masslump += (*_mass_matrix)(dofnum, j);
+
+  // Compute residual to enforce BC
+  // This is the force required to enforce the BC in a central difference scheme
+  Real avg_dt = (_dt + _dt_old) / 2;
+  Real resid =
+      (computeQpValue() - _u[_qp]) / (avg_dt * _dt) - (*_sys.solutionUDotOld())(dofnum) / avg_dt;
```
hugary1995, line 49: I'd like to understand the derivation, i.e. why the denominator is (avg_dt * _dt)

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,131 @@
...
+    [strain_block]
+        type = ComputeFiniteStrain # ComputeIncrementalSmallStrain
+        displacements = 'disp_x disp_y'
+        implicit = false
+    []
+    [stress_block]
+        type = ComputeFiniteStrainElasticStress
+    []
+    [density_one]
```
hugary1995, line 74: Is the density material property used?

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,131 @@
...
+    []
+    [massmatrix]
+        type = MassMatrix
+        density = 1
+        matrix_tags = 'system'
+        variable = disp_x
+    []
+    [massmatrix_y]
+        type = MassMatrix
+        density = 1
```
hugary1995, line 53: Please add one more test case with density value that's not one and compare the result against analytical solutions.

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,131 @@
...
+    type = Transient
+
+    [TimeIntegrator]
+        type = DirectCentralDifference
+        mass_matrix_tag = 'system'
+    []
+
+    start_time = 0.0
+    num_steps = 6
+    dt = 0.1
```
hugary1995, line 111: Please one more test case with varying time step sizes (given CFL satisfied)

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,131 @@
...
+        y = '0.0 0.0 0.0025 0.01 0.0175 0.02 0.02'
+    []
+[]
+
+[Kernels]
+    [DynamicSolidMechanics]
+        displacements = 'disp_x disp_y'
+    []
+    [massmatrix]
+        type = MassMatrix
```
hugary1995, line 46: Does the mass matrix object accept density defined as a material property? If not I suggest adding one.

hugary1995 (changes requested):

hugary1995:
Some additional suggestions/questions as I forgot to review your DirichletBC changes...

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: It does not. It inherits from from `Reaction.h`, and renames the rate param, to be density. It's a bit odd. So most likely the best way to do this would be to add an additional optional parameter that takes in density as a material property. But that would still be a bit weird.

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: It feeds into `WaveSpeed` but that is also not used. I'll remove both and regold.

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
hugary1995: I would be fine with only keeping the material property parameter. In the input file, you can still assign plain numeric literals to that parameter.

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
TheGreatCid: CD Solution update:
`a_n = M^-1 F` 
`v_(n+1/2) = v_(n-1/2) + avg_dt * a_n`
`d_(n+1) = d_n + dt * v_(n+1/2)`
Here u_(n+1) can be thought of as the boundary value that needs to be enforced, So u_(n+1) = u_g
Then combine the solution update into one equation,

`u_g = u_n + _dt * (v_(n-1/2) + avg_dt M^-1 * F)`
`u_g = u_n + _dt * v_(n-1/2) + _dt * avg_dt * M^-1 * F`
Then solve for F

`F = M*(u_g - u_n  - v_(n-1/2) * _dt)/(_dt * avg_dt)`
 `F = M*[(u_g-u_n) / (_dt*avg_dt) - v_(n-1/2)/avg_dt]`

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: In `MassMatrix` the density parameter is defined like this `  params.renameParam("rate", "density", "Optional density for scaling the computed mass.");`. 
Where the original parameter `rate` is a `Real`. 
I'm not sure if I can change this input to a material property without also modifying `Reaction.h`.

hugary1995: or just create a new object called something like `MatMassMatrix`...

TheGreatCid: That would be the more straight forward option

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: [23fce96](https://github.com/idaholab/moose/pull/28175/commits/23fce96798e42135935011643212a8d05882c9c3)

The test case with varied time steps uses a non-one density

TheGreatCid: I can add a separate test if you think that would be better

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: [92cea51](https://github.com/idaholab/moose/pull/28175/commits/92cea51147cc62906a06ae97d9587e1f9fee24fa)

Added that test here

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: [23fce96](https://github.com/idaholab/moose/pull/28175/commits/23fce96798e42135935011643212a8d05882c9c3)

I took a crack at it here

hugary1995: It looks good to me. My only complaint is that `MatMassMatrix` and `MassMatrix` have many duplicate code. Could you perhaps factor out a common base class called `MassMatrixBase` and let both derive from it?

hugary1995: Tagging @lindsayad as it technically belongs to the framework.

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
hugary1995: This works! I assume you also ran the test locally (with a much longer time and more time steps) to make sure the solution matches the analytical solution. Could you do me a favor to attach the figure here (for record keeping)? Thanks!

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
hugary1995: Thanks! The derivation makes sense to me. Then shouldn't `(computeQpValue() - _u[_qp])` be `(computeQpValue() - _u_old[_qp])` or am I missing something?

hugary1995: And more importantly, shouldn't this be whole thing be multiplied by the mass matrix?

TheGreatCid: The mass matrix is multiplied into the residual in the line after the highlighted section here. 

As for your first point, I think you are correct as u_n should equal the old solution at this point in the solve. Though using u_n = _u[_qp] still produces the right answer. There may an assignment of _u[_qp] that I am unaware of that is making the current formulation work by happenstance.

Using `_sys.solutionOld()(dofnum)` returns the same result

hugary1995: I can understand why _u and _u_old hold the same value at the time of computing the BC residual. I recommend you to request _u_old (and also _u_dot_old) in the constructor. That way we don't rely on undefined (or undocumented) behavior.

hugary1995:
This also reminds me that we should throw an error if user attempts to use `Predictor` together with central differencing. If you don't already know how to query the warehouse then nevermind, just create an issue so that we don't forget about this in the future.

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: The results from a longer simulation are a bit puzzling. I ran simulations on a single element with `dt=0.01*sin(16*t)+0.015` to oscillate the time steps. This is the same input file as the test added to this PR with the forcing function/dt changed.

When I set `u=t` using a `DirichletFunctionBC` the results look fine, the velocities and the displacement match up.  I plotted the points of the simulations to show the changing time step.

linearvel

When I set`u=t*t`, then the results are odd. The solution appears to be correct, however there are oscillations in the mid point velocity that I cannot make sense of. 

quadvel
Here is a zoom in of the phenomenon. 
quadvel_zoom

The odd thing is that I still get the correct solution despite the mid point velocity appearing incorrect. 

I'm looking into this, but let me know if you have any ideas as to the cause.

hugary1995: It's difficult for me to visually identify the errors here. Can you perhaps integrate the velocity error in time and show that? I mean something like `\int_t 0.5*(v-v_analytical)^2 dt`. I think some integration error is expected with adaptive time stepping, as long as the error converges uniformly with time step size, it's all good.

TheGreatCid: Figure_1
I plotted the average time step(the y-offset of the sin curve for dt) vs. the time integral of the difference between analytical and solution. It does look it is converging ok.

hugary1995: okay, all good then.

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
TheGreatCid: I know in `_fe_problem` you can set `u_dot_old`  but I don't believe there is a call to to requisition the old solution.

hugary1995: Use `coupledDotOld`

TheGreatCid:
> This also reminds me that we should throw an error if user attempts to use `Predictor` together with central differencing. If you don't already know how to query the warehouse then nevermind, just create an issue so that we don't forget about this in the future.

If would just be this right?

```
  if (_nl.getPredictor())
    mooseError("Predictor not supported for Direct Central Difference!");
```

hugary1995:
Oh if it's that easy just add it I guess. Also add a test to make sure the error can be triggered.

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
TheGreatCid: Are you saying to use `coupledDotOld(variable name)` instead of `_sys.solutionOld()(dofnum)` to call the old velocity?
That would involve an extra step as I would need to retrieve the correct variable name

modules/solid_mechanics/doc/content/source/bcs/DirectDirichletBC.md
TheGreatCid: [a2a74d6](https://github.com/idaholab/moose/pull/28175/commits/a2a74d6e380e6d7fbbd23093bb3390cfa0741a73)

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
hugary1995: right... then just use `_var.uDotOld()`, `_var` is a reference to the underlying `MooseVariable`. Similarly use `_var.slnOld()` to get `_u_old`.

TheGreatCid: Why is that better than this?
```
  Real resid = (computeQpValue() - _var.uDotOld()[dofnum]) / (avg_dt * _dt) -
               (_var.slnOld()[dofnum]) / avg_dt;
```

TheGreatCid: Nevermind, that segfaults....

TheGreatCid: Ah but the other method seg faults as well. The `dofnum` parameter is defined based on the total solution vector. I need a separate param for the local index

TheGreatCid: Oh I am overthinking... I can just the _qp index.

TheGreatCid:
Hmm. The `_nl.getPredictor();` does not seem to work actually. 
It returns a nullptr even if there is a predictor.

modules/solid_mechanics/src/bcs/DirectDirichletBCBase.C
TheGreatCid: [8b672f9](https://github.com/idaholab/moose/pull/28175/commits/8b672f9a8c599fc0b8c3018bbd220daabc3c9884)

hugary1995:
I followed your discussion in #27991. If I understand it correctly, the reason for moving all this into the solid mechanics module is that the `InertialForce` implementation is a bit hacky, which I agree. Now that you've switched to the `MassMatrix` approach which is indeed much cleaner and is receiving first-class framework support, why don't we move the `DirectCentralDifference` stuff into the framework? I don't see anything being very specific to solid mechanics at this point. Or put it another way, this should work for a fluid mechanics problem as well, for example.

hugary1995:
...or even more general than that -- a reaction diffusion problem.

hugary1995:
Other than that, we are down to one final unresolved conversation. I think so far this has been very good work! Consider doing an internship at ANL next summer.

TheGreatCid:
That's a good point. Now that we are using the MassMatrix kernel, this should be generalized now. I can draft up some tests with other physics, though I probably won't be able to do that until Monday. 

> Other than that, we are down to one final unresolved conversation. I think so far this has been very good work! Consider doing an internship at ANL next summer.

If I can work on my dissertation while I'm there...

TheGreatCid:
Gary and I discussed off GitHub to leave this time integrator in solid mechanics instead of moving it to framework.

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
GiudGiud: weigh one code duplication against another. how much does it realistically inherit from Reaction.h?

lindsayad:
Don't let him poach you @TheGreatCid or else I will have to request changes again 😆

TheGreatCid:
Well I do agree that there should be a mass matrix base class...

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: [f32f1e7](https://github.com/idaholab/moose/pull/28175/commits/f32f1e751f63f51fb0763df97dd8fae602363d41)

I took a crack at it here

framework/include/kernels/MassMatrixBase.h
```diff
@@ -0,0 +1,27 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "Kernel.h"
+
+/**
+ *  Base class for mass matrix calculation
+ */
+
```
lindsayad, line 17:
```suggestion
```

framework/include/kernels/MatMassMatrix.h
```diff
@@ -0,0 +1,30 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "MassMatrixBase.h"
+
+/**
+ * Calculates a mass matrix using a material property for density.
+ */
+
```
lindsayad, line 17:
```suggestion
```

framework/include/kernels/MatMassMatrix.h
```diff
@@ -0,0 +1,30 @@
...
+{
+public:
+  static InputParameters validParams();
+
+  MatMassMatrix(const InputParameters & parameters);
+
+protected:
+  virtual Real computeQpJacobian() override;
+
+  // The density of the material
```
lindsayad, line 28:
```suggestion
  /// The density of the material
```

modules/solid_mechanics/test/tests/dynamics/time_integration/tests
```diff
@@ -72,4 +72,29 @@
...
+
+  [direct_central_difference_varied_dt]
+    type = 'Exodiff'
+    input = 'direct_central_difference_varied_dt.i'
+    exodiff = 'direct_central_difference_varied_dt_out.e'
+    abs_zero = 1e-09
+    requirement = "The solid mechanics app shall support a form of central difference time integration
+                 that uses a direct calculation of the acceleration from the residual forces to 
+                 perform a solution update. The direct central difference method should properly use 
+                 a time step averaging method to handle non-constant time steps"
```
lindsayad, line 96:
```suggestion
                 a time step averaging method to handle non-constant time steps."
```

lindsayad (changes requested):

lindsayad (dismissed):

lindsayad:
I'm unsubscribing to reduce my notifications. Mention me if I'm needed

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference_varied_dt.i
```diff
@@ -0,0 +1,136 @@
...
+        displacements = 'disp_x disp_y'
+    []
+    [massmatrix]
+        type = MatMassMatrix
+        density = density
+        matrix_tags = 'system'
+        variable = disp_x
+    []
+    [massmatrix_y]
+        type = MatMassMatrix
```
hugary1995, line 59:
```suggestion
        type = MassMatrix
```

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
```diff
@@ -0,0 +1,126 @@
...
+        displacements = 'disp_x disp_y'
+    []
+    [massmatrix]
+        type = MatMassMatrix
+        density = density
+        matrix_tags = 'system'
+        variable = disp_x
+    []
+    [massmatrix_y]
+        type = MatMassMatrix
```
hugary1995, line 52:
```suggestion
        type = MassMatrix
```

hugary1995 (changes requested):

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference.i
TheGreatCid: Fixed in the latest force push

modules/solid_mechanics/test/tests/dynamics/time_integration/direct_central_difference_varied_dt.i
TheGreatCid: Fixed in the latest force push

hugary1995 (approved):

framework/src/kernels/MassMatrix.C
```diff
@@ -27,8 +30,22 @@ MassMatrix::validParams()
...
 {
   if (!isParamValid("matrix_tags") && !isParamValid("extra_matrix_tags"))
     mooseError("One of 'matrix_tags' or 'extra_matrix_tags' must be provided");
 }
+
+Real
+MassMatrix::computeQpResidual()
+{
+  mooseError("Residual should not be calculated for the MassMatrix kernel");
+  return 0;
```
dschwen, line 44:
```suggestion
```

dschwen (approved):
Merging is currently blocked, so if you have a chance and remove that unnecessary return 0 that'd be great. I'll merge as soon as possible.

framework/src/kernels/MassMatrix.C
TheGreatCid: [890051f](https://github.com/idaholab/moose/pull/28175/commits/890051f10c04ed475c25d51463f6a01c7b2abb42)

PR #28202, opened 2024-07-23
fixing inertialForce use_displaced_mesh

### LINKED ISSUE #28174: InertialForce should not set use_displaced_mesh to true by default
### Motivation

In `InertialForce` we have params.set<bool>("use_displaced_mesh") = true;.
This technically correct if we calculate the density correctly. 
If we set it to false then it would always be correct with just using a constant density.
$\int_{\Omega_0} \rho_0 a d\Omega_0 =  \int_{\Omega} \rho  a d\Omega  \neq  \int_{\Omega} \rho_0  a d\Omega$ if the volume is change during deformation. Many of the tests using `InertialForce` just use a constant generic material for the density so they are doing it incorrectly by default. 

### Design

Get rid of params.set<bool>("use_displaced_mesh") = true;

### Impact

Physically correct. As far as I can see, I don't see any test with InertialForce that calculates the density correctly.

ddm42:
- use_displaced_mesh defaults to false to match common use of GenericConstantMaterial for density. If use_displaced_mesh=True is required, use Density object from "Misc" module, which updates density based on deformation.
- fixed specifying density MaterialPropertyName

closes #28174


maxnezdyur:
@bwspenc Can you look at the discussion on the issue? Danny and I think the fix prevents a mistake that many users would do (ahem one I fell into multiple times), but we would have to regold multiple files. If we are thinking about the physics correctly and this fix is needed, then the two ways to go about this PR is to either regold or add `use_displaced_mesh=true` to all the diffing input files but that may lead users into adding that to their input files and again doing the problem wrong again.

hugary1995:
This is the right thing to do. I suggest you first check how many tests require regolding. If there's a lot, we'll have to take a different approach of making the transition.

maxnezdyur:
It looks like we have about 13 tests that diff, a mix of csv and exodiffs.

hugary1995:
And that includes internal, public, and controlled apps, right? That's pretty manageable.

hugary1995:
Oh if the action handles it correctly, then...

maxnezdyur:
So the action doesn't allow dynamics with the new kernels, and if a user were to try to set them up themselves they may have into this trap.

ddm42:
@bwspenc I know you said this could be expanded into a bigger PR that addresses similar issues with the dynamics system. I will stop updating this here until someone looks at it.

dschwen:
@maxnezdyur is taking this over

PR #28218, opened 2024-07-25
Add two function for coupling variables by name in Coupleable

### LINKED ISSUE #28214: A new function coupledValueByName in Coupleable interface
### Motivation

In Griffin, we have some objects having parameters in type of `std::map<std::string, VariableName>`, which map string names to variable names. Those variables need to be the dependent variables of the objects. Currently, Griffin has to call some low-lever functions to get the moose variable from the variable name and call `addMooseVariableDependency` to add the dependency. So I propose to add a new function `coupledValueByName`, which takes a variable name and return the reference for accessing the variable values on quadrature points.

### Design

This function will be in `Coupleable`. Because currently we only need the variable values on qps, we can just add one function. If in the future, more quantities on qps such as gradient are needed, we can add more.

### Impact

A new way to introduce couplings in a MOOSE object.

YaqiWang:
Closes #28214.


YaqiWang:
I am thinking another design, like add a new function in `InputParameters` that can mark certain parameters to check variable names. Then we can extend `InputParameters::getVecMooseType` to support more data types. The new function can check the parameter type. With this, `coupledValueByName` can skip those dependency codes. @lindsayad do you like the current code in this PR or you want to pursue the one I suggested.

YaqiWang:
@lindsayad This has been updated based on our discussions.

framework/doc/content/source/interfaces/Coupleable.md
```diff
@@ -190,6 +190,15 @@ process. These functions can be found here:
 
 !listing framework/include/interfaces/Coupleable.h start=coupled-dof-values-begin end=coupled-dof-values-end include-start=false
 
+## Coupling of variables through their names
+
+Typically variables are coupled in through calling "params.addCoupledVar" or "params.addRequiredCoupledVar" of the valid parameters of an object.
+There are cases where the variable names are provided through parameters in other types with "params.addParam" or "params.addRequiredParam".
+For these cases, this interface provides two functions `coupledValueByName` and `coupledArrayValueByName` that take a variable name directly.
+This suffix "ByName" is only available for the two functions currently because they are only needed by MOOSE and MOOSE applications.
```
lindsayad, line 198:
```suggestion
This suffix `ByName` is only available for the two functions currently because they are only needed by MOOSE and MOOSE applications.
```

lindsayad (dismissed):

lindsayad (approved):

Issue #28227, opened 2024-07-25
Enhancements to AverageSectionValueSampler


bwspenc:
### Motivation

The AverageSectionValueSampler, originally added for issue #26165, needs a few enhancements to meet user needs:

- [x] Currently, the positions at which average values are computed must be defined by the user. Automatic definition of those positions at node locations would greatly improve usability.
- [x] Output of the positions corresponding to the computed values would also improve usability.
- [ ] It should have an option to properly account for symmetry planes not aligned with the Cartesian coordinates

### Design

To automatically compute the positions, find the axial positions of all nodes in the component during initialization and upon mesh updates.
Outputing posiitons simply involves adding another output vector.
To properly account for vector values computed with off-axis symmetry planes, the computed value should be zeroed out in the direction normal to the symmetry plane, and then converted to the Cartesian coordinates.

### Impact

This will not change the computation procedure, but will add new features.

PR #28228, opened 2024-07-25
Add features to AverageSectionValue

bwspenc:
ref #28227 

1. Deprecate 'lengths' parameter in favor of more descriptive 'positions'
2. If 'positions' are not defined, automatically define positions based on locations of nodes in the component.
3. Generate an error (which can be overridden) if the number of nodes at each position varies. This is to avoid errors that might be encountered due to tolerances in locating nodes at a position.
4. Add an output vector with the axial positions.


modules/solid_mechanics/src/vectorpostprocessors/AverageSectionValueSampler.C
```diff
@@ -143,30 +198,96 @@ AverageSectionValueSampler::execute()
...
+      }
+    }
   }
 
   for (const auto j : make_range(_variables.size()))
-    *(_output_vector[j]) = output_vector_partial[j];
+    *_output_vector[j] = output_vector_partial[j];
+  const auto pos_idx = _variables.size();
+  for (const auto li : index_range(_positions))
+    (*_output_vector[pos_idx])[li] = _positions[li];
```
jasondhales, line 211:
```suggestion
  (*_output_vector[pos_idx]) = _positions;
```

modules/solid_mechanics/src/vectorpostprocessors/AverageSectionValueSampler.C
```diff
@@ -143,30 +198,96 @@ AverageSectionValueSampler::execute()
...
+  return axial_position;
+}
+
+void
+AverageSectionValueSampler::automaticallyLocatePositions()
+{
+  _positions.clear();
+
+  // Data structure used to collect the locations with nodes on each processor,
+  // and gather into in parallel.
```
jasondhales, line 239:
```suggestion
  // and gather in parallel.
```

jasondhales (approved):

PR #28241, opened 2024-07-27
improvement of some error messages in MooseVariableBase

### LINKED ISSUE #28240: improvement of some error messages in MooseVariableBase
### Motivation

Some of the error messages in MooseVariableBase.C are telling the user what is wrong, but do not give the variable name.

### Design

It would be nice, if the error message also contains the variable name.

### Impact

Improved error messages. Easing debugging for the user.

jmeier:
Some of the error messages in MooseVariableBase.C are telling the user what is wrong, but do not give the variable name. It would be nice, if the error message also contains the variable name.
closes #28240

### DIFF
--- a/framework/src/variables/MooseVariableBase.C
+++ b/framework/src/variables/MooseVariableBase.C
@@ -128,8 +128,11 @@ MooseVariableBase::MooseVariableBase(const InputParameters & parameters)
   }
   else
   {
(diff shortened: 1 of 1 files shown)


lindsayad (commented):
If you're hitting this from your input file then we should change these to errors instead of asserts

lindsayad (approved):

lindsayad:
thank you!

PR #28244, opened 2024-07-28
Libmesh submodule update

roystgnr:
- Bounding box testing for .msh $Entities in GmshIO::read(), to detect
  file inconsistencies
- Added Netgen as an optional submodule.  With Netgen enabled, libMesh
  now includes a `NetGenMeshInterface` class which can be used to
  generate tetrahedral meshes within triangulated boundaries or within
  volume mesh domains.
- Added `libmesh/fuzzy_equals.h` methods: `absolute_fuzzy_equals()`
  and `relative_fuzzy_equals()`, for checking equality to within a
  specified absolute or relative tolerance.  Added `l1_norm()` and
  `l1_norm_diff()` methods as necessary to classes (`TypeVector`,
  `NumericVector`, `SparseMatrix`) and subclasses which might be
  compared.
- Fixes for `make distcheck` testing of release candidate tarballs
- Updates to EIM error indicator normalization in Reduced Basis code

### DIFF
--- a/modules/doc/content/newsletter/2024/2024_07.md
+++ b/modules/doc/content/newsletter/2024/2024_07.md
@@ -58,6 +58,24 @@ for a complete description of all MOOSE changes.
 - `System::is_initialized()` is now `const`
 - Added `LibmeshPetscCall` macro that can wrap any PETSc function call without need for `LIBMESH_CHKERR(ierr)`
 
+### `2024.07.27` Update
+
+- Bounding box testing for .msh $Entities in GmshIO::read(), to detect
+  file inconsistencies
+- Added Netgen as an optional submodule.  With Netgen enabled, libMesh
+  now includes a `NetGenMeshInterface` class which can be used to
+  generate tetrahedral meshes within triangulated boundaries or within
+  volume mesh domains.
+- Added `libmesh/fuzzy_equals.h` methods: `absolute_fuzzy_equals()`
+  and `relative_fuzzy_equals()`, for checking equality to within a
+  specified absolute or relative tolerance.  Added `l1_norm()` and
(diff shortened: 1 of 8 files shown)


modules/doc/content/newsletter/2024/2024_07.md
```diff
@@ -58,6 +58,24 @@ for a complete description of all MOOSE changes.
 - `System::is_initialized()` is now `const`
 - Added `LibmeshPetscCall` macro that can wrap any PETSc function call without need for `LIBMESH_CHKERR(ierr)`
 
+### `2024.07.27` Update
+
+- Bounding box testing for .msh $Entities in GmshIO::read(), to detect
```
lindsayad, line 63: If for some reason you have to push again, I might suggest moving the `LibmeshPetscCall` item down to this new update

lindsayad (commented):
👍

modules/doc/content/newsletter/2024/2024_07.md
roystgnr: Yeah; I'm still not sure how I got the news item in the previous update but not the actual commit.  I usually just `git log --merges hash..` and expand+reorder what I read there.  I wonder if I somehow managed to `git checkout origin/master` right before the `LibmeshPetscCall` merge and `git log --merges hash..origin/master` right after.

lindsayad: It's because I suggested that you add it to the last newsletter entry. So it was my flub

lindsayad (approved):

milljm (approved):
for my part, Conda looks good to me.

PR #28255, opened 2024-07-29
Do not loop over boundary nodes when there is no NodalBC

### LINKED ISSUE #28254: Do not loop over boundary nodes when there is no NodalBC
### Motivation

`NonlinearSystemBase::computeNodalBCs` loops over boundary nodes whenever there are boundary nodes (including reinit), regardless of whether there is any `NodalBC` in the system. This incurs some overhead and can be avoided easily.

### Design

`NonlinearSystemBase::computeNodalBCs` will return early when the size of `_nodal_bcs` is zero, and the loop will not be executed when `nbc_warehouse->size()` is zero.

### Impact

Reduced overhead.

NamjaeChoi:
Closes #28254.

### DIFF
--- a/framework/src/systems/NonlinearSystemBase.C
+++ b/framework/src/systems/NonlinearSystemBase.C
@@ -2043,6 +2043,20 @@ NonlinearSystemBase::computeNodalBCs(const std::set<TagID> & tags)
   if (_has_save_in)
     _fe_problem.getAuxiliarySystem().solution().close();
 
+  // Select nodal kernels
+  MooseObjectWarehouse<NodalBCBase> * nbc_warehouse;
+
+  if (tags.size() == _fe_problem.numVectorTags(Moose::VECTOR_TAG_RESIDUAL) || !tags.size())
+    nbc_warehouse = &_nodal_bcs;
+  else if (tags.size() == 1)
+    nbc_warehouse = &(_nodal_bcs.getVectorTagObjectWarehouse(*(tags.begin()), 0));
+  else
+    nbc_warehouse = &(_nodal_bcs.getVectorTagsObjectWarehouse(tags, 0));
+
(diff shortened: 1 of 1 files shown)


framework/src/systems/NonlinearSystemBase.C
```diff
@@ -2061,24 +2064,25 @@ NonlinearSystemBase::computeNodalBCs(const std::set<TagID> & tags)
       else
         nbc_warehouse = &(_nodal_bcs.getVectorTagsObjectWarehouse(tags, 0));
 
-      for (const auto & bnode : bnd_nodes)
-      {
-        BoundaryID boundary_id = bnode->_bnd_id;
-        Node * node = bnode->_node;
-
-        if (node->processor_id() == processor_id())
+      if (nbc_warehouse->size())
```
lindsayad, line 2067: I think the early check should be enough. What I would do is move the creation of `nbc_warehouse` pointer ahead of your early return check, and so then do 
```c++
if (!nbc_warehouse->size())
  return;
```
And then you don't need the later `if` checks that you've added

lindsayad: It would reduce the diff too as a nice byproduct

lindsayad (changes requested):

framework/src/systems/NonlinearSystemBase.C
NamjaeChoi: Sounds good.

NamjaeChoi: Do you think I can also move `reinitNodeFace` inside the following `if`?

NamjaeChoi:
The controlled app test failure is unrelated.

lindsayad (approved):

PR #28285, opened 2024-08-02
Update MooseServer formatting to not remove include directives

brandonlangley:
refs #22766

### DIFF
--- a/framework/src/base/MooseServer.C
+++ b/framework/src/base/MooseServer.C
@@ -1090,10 +1090,14 @@ MooseServer::gatherDocumentFormattingTextEdits(wasp::DataArray & formattingTextE
                                                int tab_size,
                                                bool /* insert_spaces */)
 {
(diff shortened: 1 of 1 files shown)


brandonlangley:
@dschwen -

This addresses an [issue](https://github.com/idaholab/moose-language-support/issues/24) that was reported on the MOOSE language support VS Code extension GitHub repo.

The formatter needed to give the input path to the WASP parse method so `!include` can find relative paths.

It was not an issue with the `!include` in the formatting test since the included file is in the working directory.

But this showed up in VS Code when `!include` file paths were not relative to the VS Code working directory.

I have tested this with both VS Code as well as Workbench to verify it fixes the problem reported by the user.

Could you review this three-line update and merge it if you approve?

brandonlangley:
@grunerjmeier - FYI, here is the fix for the issue of `!include` directives being removed by VS Code document formatting.

dschwen (approved):

PR #28317, opened 2024-08-06
Set verify_holes to false in RGMB FlexiblePatternGenerator mesh subgenerator calls

shikhar413:
## Impact

refs #28316

### DIFF
--- a/modules/reactor/src/meshgenerators/AssemblyMeshGenerator.C
+++ b/modules/reactor/src/meshgenerators/AssemblyMeshGenerator.C
@@ -531,6 +531,7 @@ AssemblyMeshGenerator::generateFlexibleAssemblyBoundaries()
     params.set<boundary_id_type>("external_boundary_id") = _assembly_boundary_id;
     params.set<BoundaryName>("external_boundary_name") = _assembly_boundary_name;
     params.set<SubdomainName>("background_subdomain_name") = block_to_delete + "_TRI";
+    params.set<bool>("verify_holes") = false;
     params.set<unsigned short>("background_subdomain_id") = RGMB::ASSEMBLY_BLOCK_ID_TRI_FLEXIBLE;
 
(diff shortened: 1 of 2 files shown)


shikhar413:
@milljm this PR should address the test failure mentioned in https://github.com/idaholab/moose/pull/28231

milljm:
Keeping an eye on it: https://civet.inl.gov/job/2370876/

roystgnr:
I assume the problem here is downstream of our failure to fix https://github.com/libMesh/libmesh/pull/3497 ?

But I'd love to be reassured that that's more than an assumption, that the `point (x,y,z)=( 2.04665, 1.11022e-16,        0) on hole boundary but outside outer boundary` has been checked and *definitely* isn't outside the outer boundary.

It's probably a good idea to turn off this check either way, since it is going to give us false positives until we can fix it, but if we let a bad test case slip through then we'll find ourselves stuck getting *true* positives *after* we fix it.

shikhar413:
> I assume the problem here is downstream of our failure to fix https://github.com/libMesh/libmesh/pull/3497 ?

@roystgnr yes these changes are needed to get the Conda Intel Mac tests pass, which was triggered when adding the  feature in this PR - https://github.com/idaholab/moose/pull/28231. Basically the idea here is to delete the outermost layer created by the RGMB mesh generators and triangulate this layer instead, which will facilitate stitching of assembly mesh structures without having any hanging nodes. Since we're just triangulating a layer that was previously deleted through the mesh subgenerators, we can be sure that the hole (layers not deleted in assembly) is contained within the outer boundary (outermost assembly boundary). These checks happen through the mesh subgenerators that are called before this deletion and triangulation step.

shikhar413:
@milljm looks like the devel tests that were previously failing now pass

GiudGiud (approved):
the holes are placed programmatically, and the flexible pattern sub-generator already checks that they dont overlap iirc

this should be safe to do for now though we can re-enable it if we find that users have a way to make the hole placement invalid

PR #28338, opened 2024-08-08
Add `Correctors` syntax

### LINKED ISSUE #28339: Syntax for UserObjects that would be the equivalent of auxkernels but for nonlinear variables
### Motivation

for when the nonlinear solution isnt quite where the user wants (not normalized but doesnt want to use a constraint in the NL system for example)

### Design

Adding new syntax called Correctors to moose/framework

### Impact

Allow better organization of Moose objects by capabilities

kyleeswanson:
## Reason
Adding syntax for Correctors 

## Impact
Allows better organization of different UserObject capabilities

Closes  #28339


dschwen:
This is neat. I was about to complain that we already have a user object that renormalizes solutions, but it looks like you're moving it to a new system dedicated to solution "postprocessing".

framework/doc/content/source/correctors/PointwiseRenormalizeVector.md
```diff
@@ -0,0 +1,21 @@
+# PointwiseRenormalizeVector
+
+!syntax description /Correctors/PointwiseRenormalizeVector
+
+This user object can pointwise renormalize the solution for a set of variables, taking each variable as the component of a vector and scaling the variables to obtain the user specified L2-norm.
```
dschwen, line 5: Not a user object anymore

GiudGiud:
I ll clean this up.
For future reference objects in modules are not to be moved in the framework casually

modules/navier_stokes/src/correctors/NSPressurePin.C
```diff
@@ -48,6 +48,7 @@ NSPressurePin::validParams()
       "pressure_average", "A postprocessor that computes the average of the pressure variable");
 
   params.addClassDescription("Pins the pressure after a solve");
+  params.registerBase("Corrector");
```
hugary1995, line 51: Just curious, is this now necessary for every corrector class?

hugary1995 (changes requested):
LGTM

modules/navier_stokes/src/correctors/NSPressurePin.C
GiudGiud: until we make a base class and decide what goes in it, yes

framework/include/actions/AddCorrectorAction.h
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Action for creating Corrector, a UserObject
+ * designed for nonlinear variable solutions without constraints.
+ */
+class AddCorrectorAction : public MooseObjectAction
+{
+public:
+  /**
+   * Class constructor
```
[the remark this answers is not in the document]
GiudGiud: ugh this was copy-pasted. I ll find the source and remove it there too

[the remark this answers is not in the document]
GiudGiud: done

framework/doc/content/source/correctors/PointwiseRenormalizeVector.md
hugary1995: I have a related comment: I understand that right now correctors are basically userobjects in the backend. But should we still advertise that fact? Not just in this documentation markdown, I also saw similar emphasis in docstrings elsewhere. For example, the postprocessors also derive from userobject, but we don't explicitly emphasize that fact anywhere. Likewise, I feel we probably don't need to mention correctors being userobjects from times to times. What do you think?

framework/include/actions/AddCorrectorAction.h
```diff
@@ -0,0 +1,27 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+// MOOSE includes
+#include "MooseObjectAction.h"
+
+/**
+ * Action for creating Corrector, a UserObject
+ * designed for nonlinear variable solutions without constraints.
```
GiudGiud, line 17:
```suggestion
 * designed for modifying nonlinear variable solutions, for example outside of the nonlinear solve
```

framework/doc/content/source/correctors/PointwiseRenormalizeVector.md
GiudGiud: I think it's fine. If users can't understand inheritance between Correctors and User Objects they won't like MOOSE much anyway. 
And developers can always use the reminder since the execution scheduling of Correctors is tied with the user objects, same as the postprocessors. So they do need to know

hugary1995: Yeah, I think we don't require users to have OOP background, so maybe we won't expect them to understand the specific word "inheritance", but I agree we can communicate to users that "correctors are user objects behind the scenes".  So I would suggest explaining this bit in `Correctors/index.md`.

hugary1995:
This looks good to me otherwise. I'll approve when all tests pass.

framework/doc/content/syntax/Correctors/index.md
```diff
@@ -0,0 +1,14 @@
+# Correctors
+
+The `Correctors` system is designed to modify the values of nonlinear variable solutions.
+This can be as part of a predictor-corrector time integration scheme for example.
+
```
GiudGiud, line 5:
```suggestion

!alert note
Correctors are [UserObjects](UserObjects/index.md) behind the scene. They simply have a dedicated role and syntax.
```

GiudGiud:
@hugary1995

hugary1995 (approved):
For future reference: Having to call `registerBase` in every corrector is suboptimal. @GiudGiud said it will no longer be needed once a common base class for all Correctors is added.

GiudGiud:
I've been thinking about it and the inheritance tree of UOs would force us to create multiple base classes. One under generalUO, and maybe one under ElementUO

So maybe instead of inserting base classes in between UOs and Correctors, we should have an interface/helper like parent class that changes the base and adds whichever other attributes.

PR #28346, opened 2024-08-11
Update MooseServer formatting to not edit input that is malformed

brandonlangley:
refs #22766

### DIFF
--- a/framework/src/base/MooseServer.C
+++ b/framework/src/base/MooseServer.C
@@ -1097,7 +1097,10 @@ MooseServer::gatherDocumentFormattingTextEdits(wasp::DataArray & formattingTextE
   // input check expanded any brace expressions in cached tree so reprocess
   std::stringstream input_errors, input_stream(document_text);
   wasp::DefaultHITInterpreter interpreter(input_errors);
-  interpreter.parseStream(input_stream, parse_file_path);
+
+  // return without adding any formatting text edits if input parsing fails
+  if (!interpreter.parseStream(input_stream, parse_file_path))
+    return true;
 
   // return without adding any formatting text edits if parser root is null
   if (interpreter.root().is_null())
--- a/framework/src/base/MooseServer.C
+++ b/framework/src/base/MooseServer.C
@@ -1154,7 +1154,7 @@ MooseServer::formatDocument(wasp::HITNodeView parent, std::size_t & prev_line, s
     std::string blank = child.line() > prev_line + 1 ? "\n" : "";
 
     // format include directive with indentation and collapse extra spacing
-    if (wasp::is_nested_file(child))
+    if (child.type() == wasp::FILE)
       format_string += blank + newline_indent + MooseUtils::trim(collapse_spaces(child.data()));
 
     // format normal comment with indentation and inline comment with space
--- a/unit/src/MooseServerTest.C
+++ b/unit/src/MooseServerTest.C
@@ -1563,6 +1563,7 @@ xy_data = " 0 0.0
           100 22.30
          1000 42.210"
 [../]
+  !include    missing_file.i
 [./multi_line_indent_decrease_01]
             type = PiecewiseConstant
             xy_data = "100 1.0
@@ -1658,7 +1659,7 @@ expression = '0.1 - 2.0 * 0.2 * x^1 + 3.0 * 0.3 * x^2 - 4.0 * 0.4 * x^3 + 5.0 * 
   // expected textedits with zero-based lines and columns
 
   std::string textedits_expect = R"INPUT(
-textedit_position: [2.0]-[58.11]
+textedit_position: [2.0]-[59.11]
 textedit_new_text:
 num_dim = 2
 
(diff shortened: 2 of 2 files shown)


brandonlangley:
@dschwen -

My other update fixed the formatter removal of `!include file.i` when the file exists but not when the file is missing.

So `!include file.i` still gets removed by the formatter if `file.i` does not exist, but this one-line change fixes that.

I also added this `!include` with a nonexistent file scenario to the server formatting unit test to verify that it is written.

Sorry about a second PR for this when they total only about four lines of changes, but could you review it and merge?

@grunerjmeier - Here is the `!include` formatter fix for the case that you found when the included file does not exist.

grunerjmeier:
Dear @brandonlangley,

I took the liberty of testing the code modification. Now "!include" is preserved for existing and non-existing files.

But I found another problem (not sure if this is scope of this PR):

This
```
!include 'bla[c].i'
```
gets changed to
```
!include 'bla
[c]
  .i = .i
```

Jörg

brandonlangley:
@grunerjmeier -

> But I found another problem (not sure if this is scope of this PR):
> 
> This
> 
> ```
> !include 'bla[c].i'
> ```
> 
> gets changed to
> 
> ```
> !include 'bla
> [c]
>   .i = .i
> ```

Yes the `!include` file paths are not allowed to contain a newline character, `#` character, or `[` character.

So that is bad HIT syntax, and clicking View > Problems will show you all of your input validation problems.

I recommend leaving that Problems panel open while editing to have a list of all input diagnostics available.

You really should only format your document after fixing any errors that have to do with actual HIT syntax.

So you could disable formatOnSave and only format when you want using right click > Format Document.

I considered disabling formatting if there is any syntax error so you cannot format at all if it has things like:

 -  `!include 'bla[c].i'`
 -  `!include missing_file.i`

Would this be a better option and just not allow users to format input when it is not in proper HIT syntax?

I think good general advice is that you may see undefined formatting if you try on syntactically bad input.

grunerjmeier:
Dear @brandonlangley,

> Yes the !include file paths are not allowed to contain a newline character, # character, or [ character.

Thanks for pointing that out. I did not knew. This should be documented [here](https://mooseframework.inl.gov/application_usage/input_syntax.html).


> I recommend leaving that Problems panel open while editing to have a list of all input diagnostics available.
> You really should only format your document after fixing any errors that have to do with actual HIT syntax.
> So you could disable formatOnSave and only format when you want using right click > Format Document.
> I considered disabling formatting if there is any syntax error so you cannot format at all if it has things like: `!include 'bla[c].i'`; `!include missing_file.i`
> Would this be a better option and just not allow users to format input when it is not in proper HIT syntax?

Ok. I'll have an eye to the Problems panel in the future. Unfortunately, to switch off auto format collides with the recommendation given [here](https://mooseframework.inl.gov/help/development/VSCode.html).

So on your question: Yes, I'd prefer to leave the auto-format on. In addition "Format Document" should not format documents having errors (or only do a partial formatting of the part without errors) and showing an error message.

brandonlangley:
@dschwen -

If you look at the earlier comments, would it be best if I just disabled formatting when there are HIT syntax errors like:

 - `param value` (with no `=`)
 - `param =` (with no `value`)
 - `[]` block closing missing
 - etc.

So in order to do document formatting, the user must first make it at least structurally sound according to HIT syntax.

Note that it would still allow formatting of inputs with diagnostics from the app build, but just not structural HIT errors.

This is how the VS Code extension acted before the language server, it did not format an input that failed HIT parsing.

Else if it still lets structurally bad input be formatted, maybe `formatOnSave` should be removed from the advice [here](https://mooseframework.inl.gov/help/development/VSCode.html#0941bde6-dafc-4d16-b9d8-e1f8c534b6b0)?

I looked and it would just be a one line update in the format method to return early if the WASP interpreter parse fails.

And this would return the VS Code extension to the formatting behavior it had before integrating the language server.

dschwen:
In the old extension a warning would pop up that the input is malformed and formatting would just not happen. So you can leave formatOnSave on, but just keep you unformatted input.

brandonlangley:
@dschwen -

> In the old extension a warning would pop up that the input is malformed and formatting would just not happen. So you can leave formatOnSave on, but just keep you unformatted input.

Thanks, I updated the formatting here to not format input that is malformed (i.e. that fails the WASP syntax parse).

So the user will just need to make their input at least structurally sound according to HIT syntax before formatting.

And they can see issues that should be fixed from the syntax error messages reported in the Problems panel, e.g.,

> syntax error, unexpected invalid token, expecting = or := [Ln 4, Col 19]

> syntax error, unexpected end of file, expecting block terminator [Ln 8, Col 1]

This also prevents `formatOnSave` from formatting inputs that are malformed and making possible undefined edits.

I updated this PR title accordingly and the original case for bad include paths no longer gets formatted incorrectly.

FYI @grunerjmeier

dschwen (approved):

PR #28353, opened 2024-08-12
Loosen partitioning tolerance

lindsayad:
Refs 32 bit failure from #27691


GiudGiud (dismissed):
see my PR :D

GiudGiud (approved):
my only concern is that these parMETIS tests now use ptscotch. hence my other PR. The two PRs are compatible though

lindsayad:
I agree. I saw this as a quick fix independent of the partitioner used

lindsayad:
@GiudGiud's associated PR: https://github.com/idaholab/moose/pull/28355

PR #28362, opened 2024-08-13
Allow `MooseControl` to inherit the current Python env

### LINKED ISSUE #28366: Allow `MooseControl` object to provide current environment to the MOOSE app subrocess
### Motivation

When executing MOOSE apps with webserver controls on HPC systems, the environment variables related to the queue scheduler/MPI implementation may be required for the `subprocess` module to successfully start another process with `mpiexec`.

### Design

An addition of the option to provide the current Python shell environment via `os.environ` to the `env` keyword arg of `subprocess.Popen`

### Impact

This should be a pure addition to the API.

pshriwise:
## Reason

In HPC environments, some environment variables are required when spawning calls using `mpiexec` through the Python `subprocess` module (e.g.  variables that are part of the PMIx standard).

Resolves #28366 

## Design

This adds an attribute to the `MooseControl` class to indicate whether or not the current Python environment should be passed on to the `subprocess` call when `MooseControl.initialize` is called.

## Impact

This is a pure addition to the interfaces, though the default I've chosen does pass the current environment by default.

### DIFF
--- a/python/MooseControl/MooseControl.py
+++ b/python/MooseControl/MooseControl.py
@@ -56,6 +56,7 @@ class MooseControl:
             moose_command (list[str]): The command to use to start the moose process
             moose_port (int): The webserver port to connect to
             moose_control_name (str): The name of the input control object
+            inherit_environment (bool): Whether or not the MOOSE command will inherit the current shell environment
         """
         # Setup a basic logger
         logging.basicConfig(level=logging.INFO,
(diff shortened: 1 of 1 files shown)


loganharbour (approved):
👍🏼

loganharbour:
@pshriwise with sqa we need the contents of your PR message in an issue, with that issue referenced in the commits

loganharbour:
Failure is unrelated.

PR #26073, opened 2023-11-15
Initial work on contact for explicit finite element dynamics

recuero:
Closes #25666.

Closes #26243.

PR #28373, opened 2024-08-14
Modifying the implementation of explicit contact using updated central difference method

TheGreatCid:
This PR addresses the second half of #28008. Now that #28175 has been merged the updated formulation of [Heinstein 2000](https://www.sciencedirect.com/science/article/abs/pii/S0045782599003424) can be implemented. 

This is a node-face contact method for explicit dynamics. 
With this formulation the acceleration is calculated as follows,

$a_n=M^{-1}(F^{ext}-F^{int} + \lambda^n G^T_n)$,

where $\lambda$ is the contact force and $G_n$ is the contact surface Jacobian. 

The majority of the system for this contact model was implemented by @recuero #26073 however at the time there was no central difference time integrator that calculated acceleration directly from residual forces like that used by Heinstein. As such, some workarounds had to be made to implement this contact method. Naturally, this lead to some issues with the contact model. These (I hope) have been cleaned up with the updated model. 

The updated model now computes the contact force, $\lambda_n$ which is then fed into the acceleration calculation to enforce the contact constraint, rather than performing an override of the solution variable. 

Tests have been performed to compare against implicit mortar methods to validate. 
The tests are an extended version of the drop tests described in `moose/modules/contact/test/tests/explicit_dynamics/settlement.i`
The following shows the energies in both the implicit and explicit simulations. Note that the stiffness damping is manually calibrated to match the two simulations as (to my knowledge) there is not a straightforward way to make sure implicit and explicit simulations receive the same amount of damping. 

Another  test was done under more extreme deformations. Here a slug is shoved into a stiff anvil. An impact velocity is created using a body force as I prefer to use capabilities currently available in MOOSE and #28083 was not merged at the time of running. Using a body force leads to some odd behavior but all that is relevant here is the comparison between explicit and implicit simulations. 

The following compares the qualitative behavior of both cases,

https://github.com/user-attachments/assets/29de4c78-ee7c-4f95-a890-311b6cc0caa3

The following is a comparison of the max and average values plastic strain in the slug. The comparison somewhat falters at later time steps, however I believe explicit matches implicit to a reasonable degree. 

And here is a video of 5-body contact. The middle block is an elastic-plastic material, while the rest are elastic with varying stiffness. The bottom plate is pushing the system up in the y-direction. The topmost plate is fixed in all direction on its top surface. 

https://github.com/user-attachments/assets/ba707ffb-c43f-4b5c-90b1-e4b908189a41


TheGreatCid:
Looks like I am getting exodiffs when running parallel. I'll start looking into that, but any suggestions on direction would be much appreciated.

recuero:
Looks good to me. The agreement between implicit and explicit is great. You could even add the implicit test to the repo so that it runs for a few representative time steps. Regardless, I think the comparison shows the velocity-based contact is working as intended for verification problems.

One additional analysis that can be done in the future to assess the energy behavior of this kinematic approach is to compare it with the penalty approach. For increasingly "optimal" values of the penalty approach, both methods should converge to (sort of) the same results. Since the penalty method is available in explicit, that would allow to look into more dynamic problems and also make sure penalty works well in this setting (which I don't believe I paid much attention to). 

Happy to see this explicit dynamics work moving forward.

TheGreatCid:
@maxnezdyur @dschwen 

I'm going to look at this again next week. I was hesitant to merge because of inadequacies in modeling contact between two objects of significantly different masses. 
Most likely, this is an issue with the model itself rather than a coding issue.

TheGreatCid:
Also @maxnezdyur I know you were planning on looking at this. Have you managed to make any fixes?

TheGreatCid:
Right now, the contact surfaces are sticking to each other. I'm confused because I thought I fixed this previously. I'll need to figure out if I accidentally reverted a change.

TheGreatCid:
> Right now, the contact surfaces are sticking to each other. I'm confused because I thought I fixed this previously. I'll need to figure out if I accidentally reverted a change.

I was mistaken, I had look at the data wrong. It should be fine in this regard. I did not properly regold previously.

maxnezdyur:
@TheGreatCid 
Is there a version of `settlement.i` for implicit? So I have a reference to recreate the energy graphs?

TheGreatCid:
Here ya go
[settlement_implicit.zip](https://github.com/user-attachments/files/18377936/settlement_implicit.zip)

Also, from how the tests are failing, it's clear that I need to update this PR to match some of the more recent changes on "next"

maxnezdyur:
The PR is failing due to inconsistencies between Clang and parallel processing. I suspect the issue lies within the contact module rather than the changes in this PR. The failure occurs during exodiff comparisons at time steps 70, 80, and 90. The previous test interval was 50, which may have hidden the problem by not saving these time steps, while the PR changed the interval to 10.

TheGreatCid:
It is failing in parallel specifically, which is odd.
I can look into your interval theory. I just wont be able to get to it today.

modules/contact/test/tests/explicit_dynamics/highvel.i
```diff
@@ -206,66 +206,67 @@
         stiffness_damping_coefficient = 1.0e-3
         generate_output = 'stress_zz strain_zz'
     []
-    [inertia_x]
-        type = InertialForce
+    [Mass_x]
+        type = MassMatrix
         variable = disp_x
+        density = density
+        matrix_tags = 'system'
```
maxnezdyur, line 213: I found the issue with the parallel inconsistency.
We should not be using the system matrix as the "mass" matrix.
Add 
```
[Problem]
  extra_tag_matrices = 'mass'
[]
```
to the top of the file then change 
`matrix_tags = "system"` to `matrix_tags = "mass"`.

modules/contact/test/tests/explicit_dynamics/highvel.i
```diff
@@ -206,66 +206,67 @@
...
     []
-    [inertia_y]
-        type = InertialForce
+    [Mass_y]
+        type = MassMatrix
         variable = disp_y
+        density = density
+        matrix_tags = 'system'
```
maxnezdyur, line 219: repeat mass matrix here

modules/contact/test/tests/explicit_dynamics/highvel.i
```diff
@@ -206,66 +206,67 @@
...
     []
-    [inertia_z]
-        type = InertialForce
+    [Mass_z]
+        type = MassMatrix
         variable = disp_z
+        density = density
+        matrix_tags = 'system'
```
maxnezdyur, line 225: repeat mass matrix here

modules/contact/test/tests/explicit_dynamics/highvel.i
```diff
@@ -332,13 +333,13 @@
...
-    end_time = 0.02
+    end_time = 0.03
     dt = 2e-4
     timestep_tolerance = 1e-6
 
     [TimeIntegrator]
-        type = CentralDifference
-        solve_type = lumped
+        type = DirectCentralDifference
+        mass_matrix_tag = system
```
maxnezdyur, line 342: repeat mass matrix here

maxnezdyur (changes requested):
You can get rid of the extra debug information that I asked you to add. Look at `high_vel.i` we can't use the system matrix as the mass matrix. What happens is that all the other kernels will also contribute to this matrix. So we can create another matrix that will be our mass matrix. The other input files need the fix I showed in `high_vel.i`

modules/contact/test/tests/explicit_dynamics/highvel.i
TheGreatCid: Hm. We use the same method in the tests for DirectCentralDifference and it does not lead to errors in parallel.
Though I can still give this a shot

TheGreatCid:
@recuero @dschwen 
I would really appreciate a code review, thank you!

modules/contact/test/tests/explicit_dynamics/tests
```diff
@@ -21,39 +21,16 @@
...
-    cli_args = 'Outputs/file_base=test_balance_short_out Executioner/end_time=-0.009'            
+    cli_args = 'Outputs/file_base=test_balance_short_out Executioner/end_time=0.001'            
     requirement = 'The system shall be able to solve a simple few-element normal contact problem '
                   'using explicit dynamics solving uncoupled, local equations of momentum balance in debug mode.'
   []
-  [test_balance_overwrite_short]
-    type = 'CSVDiff'
-    input = 'test_balance.i'
-    csvdiff = 'test_balance_overwrite_short_out.csv'
-    cli_args = 'ExplicitDynamicsContact/my_contact/overwrite_current_solution=true Outputs/file_base=test_balance_overwrite_short_out Executioner/end_time=-0.009'
```
recuero, line 50: If you are removing these tests, remove gold and input files as well.

recuero (commented):
Looks good to me. I left some comments.

The problems I had as verification tests were relatively simple. It may make sense to use more complex ones, with other geometries and conditions. Also, if things are now working as expected, I would remove the comment and warning below:

https://github.com/idaholab/moose/blob/0f8738066bbb1ef8480ed138ab4613739815ccb9/modules/contact/src/actions/ExplicitDynamicsContactAction.C#L82

TheGreatCid:
Thank you for the review @recuero!

What other geometries would you be looking for, for additional tests?

dschwen:
Let's get this merged and have @maxnezdyur fix the issues Antonio pointed out. He's working on this anyways.

modules/contact/src/constraints/ExplicitDynamicsContactConstraint.C
```diff
@@ -302,19 +312,20 @@ ExplicitDynamicsContactConstraint::solveImpactEquations(const Node & node,
...
 
   auto & u_dot = *_sys.solutionUDot();
-  auto & u_old = _sys.solutionOld();
-  auto & u_old_old = _sys.solutionOlder();
-  // Mass proxy for secondary node.
-  const Real mass_proxy = density_secondary * wave_speed_secondary * _dt * nodal_area;
+
+  // Get lumped mass value
+  auto & diag = _sys.getVector("mass_matrix_diag");
+  Real mass_q = diag(dof_x);
```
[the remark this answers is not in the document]
dschwen: @maxnezdyur will address this in a follow on PR

modules/contact/test/tests/explicit_dynamics/highvel.i
```diff
@@ -206,66 +210,67 @@
         stiffness_damping_coefficient = 1.0e-3
         generate_output = 'stress_zz strain_zz'
     []
```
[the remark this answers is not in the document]
dschwen: @maxnezdyur will address this in a later follow on PR

modules/contact/test/tests/explicit_dynamics/tests
dschwen: @maxnezdyur will address this in a follow on PR

dschwen (approved):
Thanks David, I vote we get this in as is. Max is feverishly working on a follow on PR already.

TheGreatCid:
> Thanks David, I vote we get this in as is. Mex is feverishly working on a follow on PR already.

Ah good to hear! I'm sure he will iron out any issues.

recuero (approved):

PR #28405, opened 2024-08-16
better unique id approach for RevolveGenerator

### LINKED ISSUE #28401: Unique ID Issue in RevolveGenerator
### Bug Description

When we set a `unique_id()` before adding an object to a mesh, the mesh uses that id to update it's cache of the max `unique_id()` (on that processor, if it's distributed), so that `parallel_max_unique_id()` can be O(1) instead of O(N).

When we set a `unique_id()` *after* adding an object to a mesh, the mesh has no idea anything happened.  `parallel_max_unique_id()` is wrong.  Any code that uses it (including XYZDelaunay, and ironically RevolveGenerator) afterward is wrong.

There's a comment saying that RevolveGenerator was "Inspired by AdvancedExtruderGenerator", which used to have the same bug.  The simplest fix is just a `set_next_unique_id` call that takes into account the generator's formula for setting new unique ids.

_Originally posted by @roystgnr in https://github.com/idaholab/moose/issues/28298#issuecomment-2292446675_
          

### Steps to Reproduce

See Bug Description

### Impact

Followup MGs may get wrong starting unique id and then the new unique id assignment would be messed.

### [Optional] Diagnostics

_No response_

miaoyinb:
closes #28401

### DIFF
--- a/framework/src/meshgenerators/AdvancedExtruderGenerator.C
+++ b/framework/src/meshgenerators/AdvancedExtruderGenerator.C
@@ -1024,12 +1024,14 @@ AdvancedExtruderGenerator::generate()
     }
   }
 
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
   // Update the value of next_unique_id based on newly created nodes and elements
   // Note: Number of element layers is one less than number of node layers
   unsigned int total_new_node_layers = total_num_layers * order;
   unsigned int new_unique_ids = orig_unique_ids + (total_new_node_layers - 1) * orig_elem +
                                 total_new_node_layers * orig_nodes;
   mesh->set_next_unique_id(new_unique_ids);
+#endif
 
   // Copy all the subdomain/sideset/nodeset name maps to the extruded mesh
   if (!input_subdomain_map.empty())
--- a/modules/reactor/src/meshgenerators/RevolveGenerator.C
+++ b/modules/reactor/src/meshgenerators/RevolveGenerator.C
@@ -511,8 +511,8 @@ RevolveGenerator::generate()
           const unique_id_type uid =
               (current_node_layer == 0)
                   ? node->unique_id()
-                  : orig_unique_ids + (current_node_layer - 1) * (orig_nodes + orig_elem * 2) +
-                        node->id();
+                  : (orig_unique_ids + (current_node_layer - 1) * (orig_nodes + orig_elem * 2) +
+                     node->id());
           new_node->set_unique_id(uid);
 #endif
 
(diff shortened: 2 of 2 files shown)


miaoyinb:
@roystgnr FYI

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -1024,12 +1024,14 @@ AdvancedExtruderGenerator::generate()
     }
   }
 
+#ifdef LIBMESH_ENABLE_UNIQUE_ID
```
roystgnr, line 1027: I approve as a matter of principle, but I don't think MOOSE has compiled without unique_id support for years.

roystgnr (dismissed):

miaoyinb:
Okay, I will double check. 

Do you happen to have an input file that leads to duplicate ids?

roystgnr:
There's one in the `Modules debug` recipe failure:

```
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: Working Directory: /data/civet0/build/moose/modules/reactor/test/tests/meshgenerators/revolve_generator
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: Running command: /data/civet0/build/moose/modules/combined/combined-dbg -i revolve_2d.i --mesh-only revolve_2d_quad8_2.e Mesh/gmg/elem_type=QUAD8 Mesh/rg/axis_direction="-1.0 1.0 0.0" --error --error-override --timing Outputs/perf_graph=true --libtorch-device cpu
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: 
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: 
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: *** ERROR ***
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: 
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: Stack frames: 15
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: 0: libMesh::print_trace(std::ostream&)
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: 1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*, std::ostream&)
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: 2: libMesh::MeshTools::libmesh_assert_valid_unique_ids(libMesh::MeshBase const&)
reactor/test:meshgenerators/revolve_generator.test/simple_quad8_2: 3: libMesh::MeshBase::prepare_for_use()
```

The failing line is an assertion that an iterator over nodes doesn't see a unique_id that it's already seen on a previous node or element.

roystgnr (approved):

PR #28414, opened 2024-08-17
Forward parameter errors to `MooseBase` when possible 

### LINKED ISSUE #28413: Forward to `MooseBase::callMooseError()` in `InputParameters` when possible
### Motivation

The `mooseError()` calls in `InputParameters` could have additional error context (the object that is using the parameters).

### Design

Set the `MooseBase` object in `InputParameters` when parameters are associated with an object, and forward all errors for those parameters to the `MooseBase` object.

### Impact

More error context.

loganharbour:
Closes #28413

This ties all errors in `InputParameters` to the moose object that uses the parameters, if possible.

### DIFF
--- a/framework/include/base/MooseApp.h
+++ b/framework/include/base/MooseApp.h
@@ -1493,14 +1493,14 @@ template <typename T>
 const T &
 MooseApp::getParam(const std::string & name)
 {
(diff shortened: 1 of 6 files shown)


loganharbour:
Apologies for all of the spam... I think I'm good now.

PR #28415, opened 2024-08-17
Emit an error if retrieving a MooseEnum value without it being set

### LINKED ISSUE #28412: Throw error when getting a `MooseEnum` parameter without a value set
### Motivation

If you set a`MooseEnum` as a parameter and do not specify a default and then try to `getParam<MooseEnum>(...)`, you'll get seg fault instead of an error.

### Design

Make this an error instead, checking if a value is set when getting the parameter.

### Impact

Improved user experience.

loganharbour:
Closes #28412

@grmnptr came across issues with a MooseEnum being retrieved from parameters without being required. The resulting value was invalid when not set, causing indexing with a phony value. This forces us to check for validity when using a `MooseEnum` value before actually getting it. In Peter's case, it will result in an error instead of a crash.

### DIFF
--- a/framework/include/base/MooseApp.h
+++ b/framework/include/base/MooseApp.h
@@ -1493,14 +1493,14 @@ template <typename T>
 const T &
 MooseApp::getParam(const std::string & name)
 {
-  return InputParameters::getParamHelper(name, _pars, static_cast<T *>(0));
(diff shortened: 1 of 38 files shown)


loganharbour:
Requires https://github.com/idaholab/moose/pull/28414

PR #28417, opened 2024-08-17
Make test valgrind heavy

### LINKED ISSUE #28416: Make `outputs/format.yaml_dump_test` valgrind heavy
### Motivation

The test `outputs/format.yaml_dump_test` times out quite a bit with valgrind because it skirts the 600s limit.

### Design

Make `outputs/format.yaml_dump_test` valgrind heavy.

### Impact

Fewer CIVET failures.

loganharbour:
Closes #28416

### DIFF
--- a/test/tests/outputs/format/tests
+++ b/test/tests/outputs/format/tests
@@ -32,6 +32,7 @@
     cli_args = '--yaml'
     # suppress error checking, the word 'ERROR' shows up in the yaml dump
     errors = 'zzzzzzzzzz'
+    valgrind = HEAVY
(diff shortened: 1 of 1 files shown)


grmnptr (approved):
Do we expect this test to be slow by default? Or was there a change which slowed this down?

loganharbour:
It's been slow for a while. We don't really care too much about optimizing this dump because it is a pretty string/map/parse intensive thing that isn't used in a production run

grmnptr:
Sounds good to me!

PR #28420, opened 2024-08-18
Allow eight-byte integers in input files

### LINKED ISSUE #28419: Parser does not parse large integers in type of dof_id_type
### Motivation

Looks like currently the parser does not parse integers high than 2^32-1 while users may expect a 8-bite integer in their valid parameters.

### Design

It seems we can slightly change `parser.h`, `parser.cc` and `Builder.C` to enable this. I will push up a quick PR for this. I do not know this part of the code well though.

### Impact

`dof_id_type` does not work as intended in a valid parameter.

YaqiWang:
Closes #28419.


framework/contrib/hit/include/hit/parse.h
```diff
@@ -212,6 +212,7 @@ class Node
   virtual std::vector<double> vecFloatVal();
   virtual std::vector<bool> vecBoolVal();
   virtual std::vector<int> vecIntVal();
+  virtual std::vector<int64_t> vecInt64Val();
```
lindsayad, line 215: How about
```suggestion
  virtual std::vector<dof_id_type> vecDofIdTypeVal();
```

roystgnr: I prefer vecInt64Val here, no need to make it libMesh-aware.

Though we should cast_int before trying to put the result in a `dof_id_type`

framework/contrib/hit/src/hit/parse.cc
```diff
@@ -823,6 +828,21 @@ Field::vecIntVal()
...
+  for (const auto & s : vecStrVal())
+  {
+    std::istringstream iss(s);
+    int64_t conversion;
+    if ((iss >> conversion).fail() || !iss.eof())
+      throw Error("cannot convert field '" + fullpath() + "' value '" + s + "' to integer");
+    vec.push_back(conversion);
+  }
+  return vec;
+}
```
lindsayad, line 844: I would make a helper template method that you can instantiate with `int64_t` and `int` to remove the code duplication

test/tests/parser/parse_int64_t/tests
```diff
@@ -0,0 +1,12 @@
+[Tests]
+  [parse_int64_t]
+    design = 'Parser.md'
+    issues = '#28419'
+    type = 'Exodiff'
+    input = 'parse_int64_t.i'
+    cli_args = '--mesh-only'
+    exodiff = 'parse_int64_t_in.e'
+    recover = false
+    requirement = 'The system shall support reading in eight-bites integers from an input file.'
```
lindsayad, line 10:
```suggestion
    requirement = 'The system shall support reading in eight-byte integers from an input file.'
```

test/tests/parser/parse_int64_t/parse_int64_t.i
```diff
@@ -0,0 +1,16 @@
...
+    nx = 3
+    ny = 3
+    xmin = 0
+    xmax = 3
+    ymin = 0
+    ymax = 3
+    extra_element_integers = test_id
+    extra_element_integer_ids = '0 -1 2
+                                 1  3 450359962
+                                -2  3 18446744073709551614'
```
lindsayad, line 14: Could you make this test use the vector of vectors? E.g. add multiple `extra_element_integers`

framework/src/meshgenerators/GeneratedMeshGenerator.C
```diff
@@ -82,6 +82,8 @@ GeneratedMeshGenerator::validParams()
 
   params.addParam<std::vector<ExtraElementIDName>>("extra_element_integers",
                                                    "Names of extra element integers");
+  params.addParam<std::vector<std::vector<dof_id_type>>>("extra_element_integer_ids",
+                                                         "IDs of extra element integers");
```
lindsayad, line 86: So this is a capability you want in this generator?

lindsayad (changes requested):

framework/src/meshgenerators/GeneratedMeshGenerator.C
YaqiWang: No but this is the easiest way of testing the parser. Also this can be a nice capability for users to test things.

test/tests/parser/parse_int64_t/parse_int64_t.i
YaqiWang: Yes.

framework/contrib/hit/include/hit/parse.h
YaqiWang: I think supporting the underling raw type is preferred. This is in the contrib folder thus I suppose should be independent on libMesh defined types.

framework/src/meshgenerators/GeneratedMeshGenerator.C
lindsayad: Could you instead create a test generator object, a la something similar to the test object I created for testing remote element deletion, `AlwaysDeleteRemotesGeneratedMeshGenerator`?

lindsayad: We should not add capabilities to production objects that won't or shouldn't be used in production

lindsayad: If you think this parameter can/should be used in production, that's a good argument for keeping it. Making your life a little more convenient because you don't need to make a test object is not a good argument

Issue #28424, opened 2024-08-19
Variational Smoother 


oanaoana:
### Motivation

A variational smoother is a superior tool for assuring mesh quality. The current libMesh implementation should be tested, updated and extended to support the entire range of elements and mesh features used in MOOSE.

### Design

Provide MOOSE access to the libMesh implementation in [mesh__smoother__vsmoother.h](https://mooseframework.inl.gov/docs/doxygen/libmesh/mesh__smoother__vsmoother_8h_source.html). Add associated tests to check the implementation.

### Impact

Allow users to improve mesh quality by applying a smoother to current meshes. 
Increase computational efficiency by performing simulations on higher quality meshes.

PR #28425, opened 2024-08-19
Mesh smoother

oanaoana:
This PR is associated issue  #28424


roystgnr:
It's hard to argue that we shouldn't have this available when it's working on some meshes; I just wish I knew which meshes (3D? a particular element type or combination thereof?) it *didn't* work on.

In the meantime is there any way in MOOSE to put a warning on it, something like the `libmesh_experimental();` call?

GiudGiud:
It does not seem to respect volumes by default. Below, the gap grew quite a bit after 5 iterations.

I tried it on Kyle's really bad tet mesh and it hits an assert. In opt, it increases overlap between elements, and negative Jacobians. Not really a surprise given the starting mesh, it's really bad.

On my ATF mesh which could use some smoothing, the 2D did not work, Paraview was crashing on all the 2D elements for some reason, only showing the shell
In 3D (so after the extrusion) it did smooth, with no created issues. The two meshes look the same

Screenshot 2024-09-13 at 3 28 27 PM

PR #28399, opened 2024-08-16
Remove flags. Default Python packages moose-mpi

milljm:
Remove DEBUG and other offending flags from CXXFLAGS/CPPFLAGS.

When installing moose-mpi, supply a default set of python packages necessary to build MOOSE.

Continue to constrain to a Python range. Default Python is latest supported version (3.11 at this commit)

Closes #28283

PR #28437, opened 2024-08-20
set a necessary option to allow exec

### LINKED ISSUE #28427: Fix strange exec/shell behavior in moose-dev container
### Motivation

We can't exec properly:

```bash
apptainer exec oras://image 'ls /'
```

### Design

For now, provide the fix in top-level moose-dev container. A more permanent solution should be placed in a lower container, such as mpi.def.

### Impact

`exec` subcommands will function once more.

milljm:
I am still not sure why this is needed, but it is.

Closes #28427

### DIFF
--- a/apptainer/moose-dev.def
+++ b/apptainer/moose-dev.def
@@ -136,6 +136,14 @@ From: {{ APPTAINER_FROM }}
     # Install node.js
     dnf install -y nodejs npm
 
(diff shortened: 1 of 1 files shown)


milljm:
Do not merge... #28399 will incorporate the necessary change.

milljm:
Closed via #28813

PR #28473, opened 2024-08-26
Modifing base vectors of shell element and added test example

### LINKED ISSUE #28471: New contravariant base vectors for shell elements
### Motivation

The MITC4 shell element produces incorrect stresses when used with unstructured meshes.

### Design

The contravariant base vectors were modified in ADComputeIncrementalShellStrain.C 

### Impact

The stress values computed from both structured and unstructured meshes become essentially the same and show a similar pattern. The modification has no impact on the results of tests with structured meshes.

Kavan-Khaledi:
closes #28471
This is the first PR related to the shell elements, as discussed in the meeting with @bwspenc and @dschwen 

## Reason
The MITC4 produces incorrect stresses when used with unstructured meshes

## Design

- New contravariant base vectors were defined in ComputeIncrementalShellStrain.C
- One test was added
- The explanations were added to the doc file

## Impact
- The stress calculation for unstructured meshes is improved
- No impact on tests with structured meshes
- With the new formulation for the contravariant base vectors, the tapered.i test fails because it produces slightly different stress values compared to the gold results. This discrepancy is expected due to the trapezoidal shape of the element in this example, leading to different stress values with the modified formulation compared to the previous one.


dschwen:
This PR makes `test:shell/static.tapered` diff, but I suspect that's an intended side effect....

jmeier:
> This PR makes `test:shell/static.tapered` diff, but I suspect that's an intended side effect....

This is because this model now (with the proposed changes) produces different results. We suggest to re-gold this test. Please find our findings below:

@Kavan-Khaledi wrote in the initial message of this PR on this test: 
> This discrepancy is expected due to the trapezoidal shape of the element in this example, leading to different stress values with the modified formulation compared to the previous one.

For example, see disp_y at t=1 (left new implementation; right "old" gold):
image

Please consider disp_y for the node at (0, 1, 0); the node at the left-top:
- old gold: -0.0309852
- new implementation: -0.0637957

The difference is factor two.

To investigate this, I did the same model with a different 3D-FE-Programm "Plaxis 3D" with a different mesh and TRI6:
image

This gives disp_y = -0.05971 what supports the "new" results.

Therefore, we suggest to re-gold this test with the results of the "new" behavior.

jmeier:
There is also a problem with `test:shell/static.pinched_cylinder_symm_unstructured`:

At least for the following checks, this test fails:
- Modules debug: FAILED (TIMEOUT)

The same test succeeds for other checks, e.g.:
- Modules
- Modules parallel
- LibTorch

It is unclear to us, why this check fails. I'm unable to dig deeper into this problem, as running the test in debug mode yields for me:
```
Floating point exception signaled (invalid floating point operation)!
libMesh terminating:
```

I assume this is the same error as issued in https://github.com/idaholab/moose/issues/28350

How do we deal with this failed test?

dschwen:
@Kavan-Khaledi I missed that last point in your PR description. Yes, we should definitely regold that test. I can do that.

dschwen:
I pushed some small edits and will clean-up the commit history once you're done on this PR (otherwise you'll get some git conflicts).

jmeier:
Dear @dschwen and Moose Team,

With this PR one of our tests (`test:shell/static.pinched_cylinder_symm_unstructured`) fails with an "invalid floating point operation". According to our findings listed below, this could be an unrelated error in Moose that is triggered by our test. @Kavan-Khaledi and I have had no luck finding the underlying reason for this error. 

We need your support here. How should we proceed?

 
**The error:**
When running the test, the following error is shown *only* if run in dbg-mode:
```
Floating point exception signaled (invalid floating point operation)!
libMesh terminating
```

<details><summary>The output of gdb</summary>
<p>

```
This is the output of gdb:
Time Step 1, time = 1, dt = 1
 
    Computing Residual.
Thread 1 "solid_mechanics" received signal SIGFPE, Arithmetic exception.
fedisableexcept (excepts=5) at ../sysdeps/x86_64/fpu/fedisblxcpt.c:30
30      ../sysdeps/x86_64/fpu/fedisblxcpt.c: No such file or directory.
(gdb) bt
#0  fedisableexcept (excepts=5) at ../sysdeps/x86_64/fpu/fedisblxcpt.c:30
#1  0x00007ffff01e0ddb in libMesh::enableFPE(bool) () from /home/mjg/mambaforge3/envs/moose/libmesh/lib/libmesh_dbg.so.0
#2  0x00007ffff57c5aee in FloatingPointExceptionGuard::~FloatingPointExceptionGuard (this=0x7fffffff8bf0, __in_chrg=<optimized out>)
    at /home/mjg/projects/moose-kavan/framework/build/header_symlinks/FloatingPointExceptionGuard.h:39
#3  0x00007ffff57dab72 in NonlinearSystemBase::computeResidualTags (this=0x555555c51e50, tags=...) at /home/mjg/projects/moose-kavan/framework/src/systems/NonlinearSystemBase.C:882
#4  0x00007ffff61f558a in FEProblemBase::computeResidualTags (this=0x555555c302e0, tags=...) at /home/mjg/projects/moose-kavan/framework/src/problems/FEProblemBase.C:6823
#5  0x00007ffff61f27f5 in FEProblemBase::computeResidualInternal (this=0x555555c302e0, soln=..., residual=..., tags=...) at /home/mjg/projects/moose-kavan/framework/src/problems/FEProblemBase.C:6663
#6  0x00007ffff61f0b70 in FEProblemBase::computeResidual (this=0x555555c302e0, soln=..., residual=..., nl_sys_num=0) at /home/mjg/projects/moose-kavan/framework/src/problems/FEProblemBase.C:6467
#7  0x00007ffff61f06a3 in FEProblemBase::computeResidualSys (this=0x555555c302e0, sys=..., soln=..., residual=...) at /home/mjg/projects/moose-kavan/framework/src/problems/FEProblemBase.C:6438
#8  0x00007ffff57c459f in ComputeResidualFunctor::residual (this=0x555555c53730, soln=..., residual=..., sys=...) at /home/mjg/projects/moose-kavan/framework/src/systems/ComputeResidualFunctor.C:27
#9  0x00007ffff10dbedb in libmesh_petsc_snes_residual () from /home/mjg/mambaforge3/envs/moose/libmesh/lib/libmesh_dbg.so.0
#10 0x00007fffe9e27281 in SNESComputeFunction () from /home/mjg/mambaforge3/envs/moose/petsc/lib/libpetsc.so.3.21
#11 0x00007fffe9e01369 in SNESSolve_NEWTONLS () from /home/mjg/mambaforge3/envs/moose/petsc/lib/libpetsc.so.3.21
#12 0x00007fffe9e2ef45 in SNESSolve () from /home/mjg/mambaforge3/envs/moose/petsc/lib/libpetsc.so.3.21
#13 0x00007ffff10e3727 in libMesh::PetscNonlinearSolver<double>::solve(libMesh::SparseMatrix<double>&, libMesh::NumericVector<double>&, libMesh::NumericVector<double>&, double, unsigned int) ()
   from /home/mjg/mambaforge3/envs/moose/libmesh/lib/libmesh_dbg.so.0
#14 0x00007ffff1180dbe in libMesh::NonlinearImplicitSystem::solve() () from /home/mjg/mambaforge3/envs/moose/libmesh/lib/libmesh_dbg.so.0
#15 0x00007ffff578c48a in TimeIntegrator::solve (this=0x555555b700d0) at /home/mjg/projects/moose-kavan/framework/src/timeintegrators/TimeIntegrator.C:54
#16 0x00007ffff57d052a in NonlinearSystem::solve (this=0x555555c51e50) at /home/mjg/projects/moose-kavan/framework/src/systems/NonlinearSystem.C:189
#17 0x00007ffff61e69bf in FEProblemBase::solve (this=0x555555c302e0, nl_sys_num=0) at /home/mjg/projects/moose-kavan/framework/src/problems/FEProblemBase.C:6029
#18 0x00007ffff59eafec in FEProblemSolve::solve (this=0x555555cd24b8) at /home/mjg/projects/moose-kavan/framework/src/executioners/FEProblemSolve.C:291
--Type <RET> for more, q to quit, c to continue without paging--c
#19 0x00007ffff59f0508 in FixedPointSolve::solveStep (this=0x555555cd2770, begin_norm=@0x555555ed0fa0: 0, end_norm=@0x555558a92a50: 0, transformed_dofs=...) at /home/mjg/projects/moose-kavan/framework/src/executioners/FixedPointSolve.C:439
#20 0x00007ffff59ef39d in FixedPointSolve::solve (this=0x555555cd2770) at /home/mjg/projects/moose-kavan/framework/src/executioners/FixedPointSolve.C:291
#21 0x00007ffff4de7204 in TimeStepper::step (this=0x555555bc6600) at /home/mjg/projects/moose-kavan/framework/src/timesteppers/TimeStepper.C:168
#22 0x00007ffff59ff700 in Transient::takeStep (this=0x555555cd2100, input_dt=-1) at /home/mjg/projects/moose-kavan/framework/src/executioners/Transient.C:416
#23 0x00007ffff59fe184 in Transient::execute (this=0x555555cd2100) at /home/mjg/projects/moose-kavan/framework/src/executioners/Transient.C:291
#24 0x00007ffff640a674 in MooseApp::executeExecutioner (this=0x5555557c59f0) at /home/mjg/projects/moose-kavan/framework/src/base/MooseApp.C:1172
#25 0x00007ffff6411a3d in MooseApp::run (this=0x5555557c59f0) at /home/mjg/projects/moose-kavan/framework/src/base/MooseApp.C:1554
#26 0x0000555555557e7e in Moose::main<SolidMechanicsTestApp> (argc=3, argv=0x7fffffffbe58) at /home/mjg/projects/moose-kavan/framework/build/header_symlinks/MooseMain.h:47
#27 0x00005555555576e6 in main (argc=3, argv=0x7fffffffbe58) at /home/mjg/projects/moose-kavan/modules/solid_mechanics/src/main.C:17
 
```

</p>
</details> 

 
**Our findings**

The test completes successfully in opt-mode and the results look reasonable.

In dbg-mode the results of @Kavan-Khaledi and me are as follows:

1)	New code (modified base vectors) with structured mesh &rarr; no floating exception in DBG

2)	Old code (without modified base vectors) with structured mesh &rarr; no floating exception in DBG

3)	New code with unstructured mesh &rarr; floating exception in DBG

4)	Old code with unstructured mesh &rarr; floating exception in DBG

This leads us to the assumption that the error is not caused by our adjustments to the code.

There are so more (perhaps obvious) findings:

5) The error looks very similar to the one in #28350. We hope that if the issue over there is solved, "our" error will also vanish? Is it foreseeable whether and when this issue can be resolved so that we can check it?

6) If one would comment all lines `FloatingPointExceptionGuard fpe_guard(_app);` in "FEProblemBase.C" (this line is there 2x) and in "NonlinearSystemBase.C" (the line is there 3x) the "dbg" build works. Obviously, we will not include this code change to this PR.

7) Without the hacky changes in 6), using the argument `--no-trap-fpe` will also suppress the error even in dbg-mode

dschwen:
I sprinkled some calls to `fetestexcept(FE_ALL_EXCEPT)` and found that an `FE_INEXACT` exception is thrown [here](https://github.com/idaholab/moose/pull/28473/files#diff-ff235476d288d7fdfc5388b25b78a8065396b6f99f5412bf3e9cbc619733dbffL288) (`ADRealVectorValue normal = x.cross(y);`). However we only trap `FE_DIVBYZERO | FE_INVALID`. So I'm suspecting that this unhandled flag gets trapped when `fedisableexcept` is called :-/

jmeier:
Dear @dschwen and @bwspenc,

It looks like the flag `FE_INVALID` of `fetestexcept` gets set here:
https://github.com/idaholab/moose/blob/7c55923541c02ca9abfdef6b90b73e77eab77467/framework/src/outputs/Exodus.C#L316-L322

After that, Moose does not check and libmesh finds it in the next step and complains. It is not clear to me what inside `write_timestep_discontinuous`/`write_timestep` happens so I presume the flag is caused in there somewhere.

BTW: By following the idea of @dschwen this had been found by sprinkeling the lambda function at the end of this comment all over Moose.

**How do we proceed?** 
- Separate issue here in Moose?
- Issue over in libMesh?
- Ignorantly just add `feclearexcept(FE_ALL_EXCEPT)` in `Exodus.C`?
- Should Moose check `fetestexcept` before and after all the output has been done (e.g. in `FEProblemBase::outputStep`) to avoid this kind of error in the future? 
- Something else?

<details><summary>Lambda function to check fetestexcept</summary>
<p>

```
auto show_fe_exceptions = [](std::string prefix)
    {
        int code = fetestexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW | FE_UNDERFLOW);
        if (code != 0)
        {
          std::cout << "fetestexcept " << prefix << ": ";
          if(fetestexcept(FE_DIVBYZERO))     std::cout << " FE_DIVBYZERO";
          if(fetestexcept(FE_INEXACT))       std::cout << " FE_INEXACT";
          if(fetestexcept(FE_INVALID))       std::cout << " FE_INVALID";
          if(fetestexcept(FE_OVERFLOW))      std::cout << " FE_OVERFLOW";
          if(fetestexcept(FE_UNDERFLOW))     std::cout << " FE_UNDERFLOW";
          std::cout << "\n" << std::flush;
          feclearexcept(FE_ALL_EXCEPT);
        };
        return code;
    };

  show_fe_exceptions("Exodus::outputNodalVariables: initial");
```

</p>
</details>

dschwen:
Ok, I'll build this PR with Roy's changes applied. Last night i arrived at the conclusion that nowhere in this object an exception is raised.

dschwen:
No more FPE with Roy's PR. However the new pinched cylinder test takes 20 times longer than the old one with the structured mesh, despite having only 5 times more DOFs. Why is that?

Kavan-Khaledi:
> No more FPE with Roy's PR. However the new pinched cylinder test takes 20 times longer than the old one with the structured mesh, despite having only 5 times more DOFs. Why is that?


Great news, Daniel! Thanks.
Just to clarify, the old test had only 100 structured elements (with 605 primary DOFs), while the new test has 1,583 unstructured elements (with 8,320 primary DOFs). I used a refined mesh to better visualize the stress compatibility resulting from the modified base vectors in the post-processing. I also added two auxiliary variables for $\sigma_{11}$ in the input file, which introduced 3,166 additional auxiliary DOFs. So, the calculation time you observed makes sense to me, as the DOFs in the new model are nearly 19 times greater than in the old one. Kavan

bwspenc:
Our general guideline for the standard regression tests is that they should run in optimized mode in less than about 2 seconds. I assume with the finer mesh that unstructued case would take longer than 2 seconds. Would it be possible to make a coarsely meshed version of it that is similar in size to the structured mesh? If it's important to have a finer version of it, we could add that but mark it as 'heavy', but I'm not really sure that's even necessary.

bwspenc:
@dschwen Which PR is Roy's? Is it merged into next now?

Kavan-Khaledi:
> Our general guideline for the standard regression tests is that they should run in optimized mode in less than about 2 seconds. I assume with the finer mesh that unstructued case would take longer than 2 seconds. Would it be possible to make a coarsely meshed version of it that is similar in size to the structured mesh? If it's important to have a finer version of it, we could add that but mark it as 'heavy', but I'm not really sure that's even necessary.



Yes, i replaced the test with a coarse meshed version.

dschwen:
I'm on this right now, checking if we can run the unstructured pinched cylinder test in dbg with the coarsend mesh.

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -353,6 +353,7 @@ ADComputeIncrementalShellStrain::computeGMatrix()
         gmn(0, 2) += (*_dxyz_dxi[j])[i](component) * (*_dxyz_dzeta[j])[i](component);
         gmn(1, 2) += (*_dxyz_deta[j])[i](component) * (*_dxyz_dzeta[j])[i](component);
 
+        // why are we dropping derivatives here?!
```
bwspenc, line 356: There are a bunch of these that don't actually have derivatives, and don't need to be AD properties. That's separate from this, though.

bwspenc (approved):

bwspenc:
I tested the pinched cylinder problem with the regular mesh with these changes, and as I suspected, the solution is unchanged (since these changes only affect the stress output). It still behaves differently when pinched in the x and y directions, so that's still an outstanding issue. I think we should merge this now, though.

PR #28490, opened 2024-08-29
Revert to previous behavior in deepdiff

### LINKED ISSUE #28489: Account for `threshold_to_diff_deeper` in `DeepDiff`
### Motivation

`DeepDiff` introduces the `threshold_to_diff_deeper` parameter, which it describes as:

> threshold_to_diff_deeper=float is a number between 0 and 1. When comparing dictionaries that have a small intersection of keys, we will report the dictionary as a new_value instead of reporting individual keys changed. If you set it to zero, you get the same results as DeepDiff 7.0.1 and earlier, which means this feature is disabled. The new default is 0.33 which means if less that one third of keys between dictionaries intersect, report it as a new object.

We want to report individual keys instead of the entire diff.

### Design

Set `threshold_to_diff_deeper=0` for the `DeepDiff.deepdiff` calls in the test harness.

### Impact

Reverts to previous behavior.

loganharbour:
Closes #28489

### DIFF
--- a/python/TestHarness/testers/SchemaDiff.py
+++ b/python/TestHarness/testers/SchemaDiff.py
@@ -30,6 +30,8 @@ class SchemaDiff(RunApp):
             self.specs['required_python_packages'] = 'deepdiff>=6.1.0'
         elif 'deepdiff' not in self.specs['required_python_packages']:
             self.specs['required_python_packages'] += ' deepdiff>=6.1.0'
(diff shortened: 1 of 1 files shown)


loganharbour:
@GiudGiud @milljm this resolves the deepdiff issue.

GiudGiud (approved):

PR #28514, opened 2024-08-30
Using field variables instead of scalar variables for THM junctions, part 1

### LINKED ISSUE #25593: Consider alternatives to scalar variables for THM junction variables
## Reason
Scalar variables as libMesh considers them are global variables (well they can be sudomain restricted) that should introduce a dense row and a dense column in the system matrix. Take for instance setting a mean zero pressure in a cavity problem to remove the nullspace: the scalar Lagrange multiplier variable residual is a function of every pressure dof in the cavity and every pressure residual is a function of the Lagrange multiplier dof. libMesh constructs its sparsity pattern anticipating a coupling structure like this, which can be unnecessarily expensive if the coupling structure is *not* that way.

## Design
Consider an alternative like a Lagrange variable restricted to a `NodeElem` associated with the junction

## Impact
Potentially much faster sparsity build times. See https://github.com/idaholab/moose/discussions/25555 for the impetus of this issue

joshuahansel:
Closes #25593

This work is broken into parts to prevent breaking RELAP-7:

1. This PR: does inconsequential prep work and adds a virtual that RELAP-7 will need to override to prevent breakage.
2. RELAP-7 PR: override the virtual
3. THM PR: actually implement the node element option for 1-phase components (scalar option will be deprecated)
4. RELAP-7 PR: implement the node element option for 2-phase components and fix all deprecated test failures.

This first part does some preliminary work:

- Moved some mesh-related methods from `GeometricalComponent` to `Component`. The former is interpreted as components having a mesh > 0D. Now, 0D components (e.g., junctions) may need to do some mesh-related things since we will add node elements.
- Added some getter methods for some junction connection data.
- Cleanup in  `ADVolumeJunction1PhaseBC` to facilitate future changes.
- Removed `ShaftConnectedTestComponent`, which performs some redundant Jacobian testing.
- Added a virtual method `getJunctionVariables()`, which is used for nothing yet, but will be overridden in RELAP-7 for part 2 and then used in THM in part 3.


lindsayad:
I see you don’t want CIVET testing but this isn’t marked as draft. Do you want human review then?

joshuahansel:
It's not quite ready for human review because I'm working through a distributed mesh bug. I marked as WIP while I was transferring WIP between machines.

joshuahansel:
@lindsayad This is ready for review now.

joshuahansel:
@licharlot Can you review this when convenient? There's nothing juicy in this one, so I'm letting @lindsayad off the hook for now.

licharlot (approved):

PR #28532, opened 2024-09-03
Revert "Cache the results for the civet_results tests"

loganharbour:
Reverts idaholab/moose#23296

We can revert now to getting actual results because CIVET doesn't timeout.


loganharbour:
Failures are unrelated. This only touches python tests.

dschwen (approved):

PR #28556, opened 2024-09-06
Throw a paramError when an incorrect strain base name is provided

### LINKED ISSUE #25673: PorousFlow mass/heat postprocessors can silently ignore strain
## Bug Description
As noted in #24668 there could be an issue obtaining strain in these postprocessors:

```
This logic seems dangerous

if i mispelled base_name, it wont retrieve a total strain and will change the result.

why would not always retrieve the total strain if a base name is passed?
```

The logic is
```
_base_name(isParamValid("base_name") ? getParam<std::string>("base_name") + "_" : ""), _has_total_strain(hasMaterialProperty<RankTwoTensor>(_base_name + "total_strain")),
_total_strain(_has_total_strain
                      ? &getMaterialProperty<RankTwoTensor>(_base_name + "total_strain")
                      : nullptr),
```

So, if the user has a strain calculator, but provides an incorrect `base_name` (maybe a typo in the input file), `_has_total_strain` would evaluate to false, and `_total_strain` would just be a null pointer, and no strain would be included in the calculations.

It would probably be better to throw a param error in this case that `base_name` doesn't exist.

## Steps to Reproduce
See above

## Impact
May silently produce an incorrect value

cpgr:
Instead of silently failing, alert the user that the strain base_name does not exist.

Fixes #25673

### DIFF
--- a/modules/porous_flow/src/postprocessors/PorousFlowFluidMass.C
+++ b/modules/porous_flow/src/postprocessors/PorousFlowFluidMass.C
@@ -143,6 +143,10 @@ PorousFlowFluidMassTempl<is_ad>::PorousFlowFluidMassTempl(const InputParameters 
   if (_phase_index.empty())
     for (unsigned int i = 0; i < num_phases; ++i)
       _phase_index.push_back(i);
+
+  // Error if a strain base_name is provided but doesn't exist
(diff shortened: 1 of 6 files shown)


GiudGiud (approved):

cpgr:
Thanks @GiudGiud

PR #28576, opened 2024-09-08
Suppress PETSc configure warnings

### LINKED ISSUE #28575: PETSc configure warning
### Motivation

If users follow the [HPC clusters](https://mooseframework.inl.gov/getting_started/installation/hpc_install_moose.html) installation instructions, they will be presented with the following (in my opinion) concerning messages while configuring PETSc. In fact, I have received several questions regarding these warning messages.

```
=============================================================================================
                         Configuring PETSc to compile on your system
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Found environment variable: CC=mpicc. Ignoring it! Use "./configure CC=$CC" if you really
  want to use this value
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Found environment variable: CXX=mpicxx. Ignoring it! Use "./configure CXX=$CXX" if you
  really want to use this value
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Found environment variable: FC=mpif90. Ignoring it! Use "./configure FC=$FC" if you
  really want to use this value
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Found environment variable: F77=mpif77. Ignoring it! Use "./configure F77=$F77" if you
  really want to use this value
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Found environment variable: F90=mpif90. Ignoring it! Use "./configure F90=$F90" if you
  really want to use this value
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Using default C optimization flags "-g -O". You might consider manually setting optimal
  optimization flags for your system with COPTFLAGS="optimization flags" see
  config/examples/arch-*-opt.py for examples
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Using default Cxx optimization flags "-g -O". You might consider manually setting optimal
  optimization flags for your system with CXXOPTFLAGS="optimization flags" see
  config/examples/arch-*-opt.py for examples
=============================================================================================
=============================================================================================
                                     ***** WARNING *****
  Using default FC optimization flags "-g -O". You might consider manually setting optimal
  optimization flags for your system with FOPTFLAGS="optimization flags" see
  config/examples/arch-*-opt.py for examples
=============================================================================================
```

### Design

The first few are related to environment variables for compilers. If we always instruct users to set
```
export CC=mpicc CXX=mpicxx FC=mpif90 F90=mpif90 F77=mpif77
```
those warnings can be "fixed" by passing these values via the PETSc configure script.

The last three are related to compiler flags. Again, since we are using the default, it doesn't hurt to explicitly specifying the flags via the configure script.

### Impact

Suppress the warnings so users are less concerned.

hugary1995:
- Pass compilers via PETSc's configure script
- Pass compiler flags via PETSc's configure script

This suppresses some potentially concerning warning messages (see issue).

close #28575

### DIFF
--- a/modules/doc/content/getting_started/installation/hpc_install_moose.md
+++ b/modules/doc/content/getting_started/installation/hpc_install_moose.md
@@ -27,20 +27,17 @@ This *usually* involves `module load` commands. Please note again, that Intel co
 supported.
 !style-end!
 
-Sometimes after loading a proper MPI environment, it is still necessary to set some variables.
-Check to see if the following variables are set:
-
+After loading a proper MPI environment, check to see if the following variables are set:
 ```bash
 echo $CC $CXX $FC $F90 $F77
 ```
 
-If nothing returns, or what does return does *not* include MPI naming conventions (`$CC` is `gcc`
-and not `mpicc` like we need), you need to set them manually each and every time you load said
-environment:
-
+If nothing returns, the following default values are used:
 ```bash
-export CC=mpicc CXX=mpicxx FC=mpif90 F90=mpif90 F77=mpif77
+CC=mpicc CXX=mpicxx FC=mpif90 F90=mpif90 F77=mpif77
 ```
+If the returned variable values do *not* include MPI naming conventions (`$CC` is `gcc`
+and not `mpicc` like we need), you need to set the values manually.
 
 !include installation/start_up_profile.md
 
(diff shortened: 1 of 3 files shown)


hugary1995:
Sigh, I guess there's not way to reasonably suppress these warnings because we need to rely on PETSc to detect MPI-wrapped compilers on HPCs. I guess I'll live with these warnings for now.

PR #28594, opened 2024-09-10
Swap coeff by function in VariableTimeIntegrationAux

nmnobre:
## Reason
A function is significantly more flexible than a constant coefficient, allowing us to, for instance, a) integrate over an arbitrary time period smaller than the entire simulation time and to b) weight that integral freely.

There wasn't technically any test for the functionality I'm improving here, so I've added one.


nmnobre:
I no longer think this a good idea.

PR #28601, opened 2024-09-11
Don't do this once

loganharbour:
refs #27097


joshuahansel (approved):

loganharbour:
The applications that this effects have been properly tested against this change:

- relap-7: https://civet.inl.gov/event/196752/
- sockeye: https://civet.inl.gov/event/196753/

cticenhour (approved):

PR #28618, opened 2024-09-13
Update language server in moose-dev

### LINKED ISSUE #28617: Update code language server to v1.2.0
### Motivation

The language server plugin was updated in https://github.com/idaholab/moose-language-support/releases/tag/v1.2.0 to honor the `MOOSE_LANGUAGE_SERVER` environment variable for the executable to use for language support.

### Design

Update the language server plugin in `moose-dev.def`.

### Impact

Will simplify the process for users using the language server plugin within code server instances for applications, where they don't need to choose an executable at all.

loganharbour:
Closes https://github.com/idaholab/moose/issues/28617

### DIFF
--- a/apptainer/moose-dev.def
+++ b/apptainer/moose-dev.def
@@ -85,7 +85,7 @@ From: {{ APPTAINER_FROM }}
     # by the "setup-code-server-extensions" script (in PATH)
     mkdir /opt/code-server/extensions
     # Download extensions for code-server, currently just moose language support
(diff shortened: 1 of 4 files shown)


loganharbour:
@dschwen this is all it takes to update the language server.

@milljm This is a moose-dev update, but it's not a priority. I'm using this PR to test Daniel's new plugin. We can get this in after your big changes.

Issue #23118, opened 2023-01-09
Add solid properties interface for specific internal energy


joshuahansel:
## Reason
For heat conduction applications, it is useful to compute a solid's specific internal energy for checking or requiring energy conservation.

## Design
This would add a new interface `e(T)` to the `ThermalSolidProperties` base class, along with a default implementation:
```
e(T) = integral_0^T cp(T') dT'
```
This will require a new interface for the integral of `cp(T)`, which by default will give a not-implemented error. Note that for the function formulation, https://github.com/idaholab/moose/issues/23116 will be needed.

Derived classes may want to override `e(T)` because they use a different reference point, instead, doing something like
```
e(T) = integral_{T_ref}^T cp(T') dT'
```

## Impact
This will add a new interface.

PR #28620, opened 2024-09-13
Added specific internal energy to ThermalSolidProperties

joshuahansel:
This completes Task 1 in #23118.


licharlot (approved):
The code here looks good. Are you planning on adding `e` in `ThermalSolidPropertiesMaterial` (and the one with constant density) and in `ThermalSolidPropertiesPostprocessor` in another PR?

joshuahansel:
I will add those object changes to this PR (and address coverage).

joshuahansel:
I decided to modify only the functor material, not the materials or PP. I'm tentatively planning to deprecate those in favor of the functor material, but I haven't committed, since the functor material may be more inconvenient.

licharlot (approved):

PR #28640, opened 2024-09-16
Capture return code in `MooseApp`

### LINKED ISSUE #28639: Enable the setting of the exit code from moose
### Motivation

The `--run` option for running application tests is currently flawed. It only returns exit code 1 on failure, and does not pass the actual return code from the test harness.

### Design

Add `MooseEpp::exitCode()` and use it when returning from main. 

Capture the exit code when using `--run` and exit with that return code instead of using `mooseError()`.

### Impact

Will enable using `--run` with intermittent failures, because we can pass the true exit code from the test harness. Will also let others exit non-zero from moose without a mooseError.

loganharbour:
Closes #28639

### DIFF
--- a/examples/ex01_inputfile/src/main.C
+++ b/examples/ex01_inputfile/src/main.C
@@ -20,7 +20,5 @@
 int
 main(int argc, char * argv[])
 {
-  Moose::main<ExampleApp>(argc, argv);
-
-  return 0;
+  return Moose::main<ExampleApp>(argc, argv);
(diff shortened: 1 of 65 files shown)


loganharbour:
@dschwen this is ready.

Will need help from folks with app patches once it's in.

permcody (approved):

loganharbour:
Failures are unrelated and fixed in current master.

PR #28644, opened 2024-09-17
Update stale.yml

mosfet80:
## Reason

updated the runtime to Node.js 20.
Node.js 16 is deprecated


loganharbour:
Unsure of intentions. Closing.

PR #28645, opened 2024-09-17
Update hdf5 into meta.yaml

mosfet80:
## Reason

Improved support for Intel oneAPI.
Added support for AOCC and classic Flang w/ the Autotools. 
Thread-safety + static library disabled on Windows w/ CMake.
Running H5make_libsettings is no longer required for cross-compiling. 
Running H5detect is no longer required for cross-compiling. 
Removed "-commons" linking option on Darwin, as COMMON and EQUIVALENCE are no longer used in the Fortran source. 

## Design

Added a simple cache to the read-only S3 (ros3) VFD. 
Added new API function H5Pget_actual_selection_io_mode(). 
Added optimized support for the parallel compression feature when using the multi-dataset I/O API routines collectively. Changed H5Pset_evict_on_close so that it can be called with a parallel build of HDF5. Fortran async APIs H5A, H5D, H5ES, H5G, H5F, H5L and H5O were added. 
## Impact

Disabled running of MPI Atomicity tests for OpenMPI major versions < 5.

### DIFF
--- a/conda/mpi/meta.yaml
+++ b/conda/mpi/meta.yaml
@@ -9,7 +9,7 @@
 {% set version = "2024.09.12" %}
 
 # HDF5 Version
-{% set hdf5_version = "1.14.2" %}
+{% set hdf5_version = "1.14.3" %}
 
 package:
   name: moose-mpi


loganharbour:
Unsure of intentions. Closing.

milljm:
We found an issue with bumping HDF5 to 1.14.3. https://github.com/idaholab/moose/issues/28350. There is a current PR in place to limit this to 1.14.2. (https://github.com/idaholab/moose/pull/28399 which was reverted, but not for reasons of HDF5, and we will continue to limit HDF5 to 1.14.2 until we can properly solve the issues found with 1.14.3)

Issue #28283, opened 2024-08-01
Strip out DEBUG flags in Conda activation routines


milljm:
### Motivation

The following being set:
```pre
CPP/CXXFLAGS=-DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -ftree-vectorize
```
is preventing some of us from building libMesh in debug mode using Conda compilers.

Like we do with:
```pre
TEMP_CXXFLAGS=${CXXFLAGS//-std=c++[0-9][0-9]}
# The  "-fdebug-prefix-map" flags then need to be removed, as they are specific
# to the conda build process. Finally, `-std=c++17` can be appended to the end.
ACTIVATION_CXXFLAGS=${TEMP_CXXFLAGS%%-fdebug-prefix-map*}-std=c++17
```
we should add routines to remove the offending flags.

### Design

Add necessary regex to remove offending activation flags

### Impact

Users should be able to build libMesh in debug mode.

PR #28587, opened 2024-09-09
Add profiling to moose-dev container

grmnptr:
Closes #28583

PR #28656, opened 2024-09-18
Update moose-mpi & PETSc: default Python packages, pinnings, apptainer exec fix, optimize compiler flags

### LINKED ISSUE #28643: MPI Change: Add influential file. Correct a small permissions issue in mpi.def container
### Motivation

There was a new file added to apptainer/mpi.def that was mistakenly not added to the influential files.

There is a sourced file that has incorrect permissions. While they work at present, they should be world-readable.

Edit:
Using this PR to include changes from a reverted PR #28283 (with fixes). Otherwise, this would be too many rapid changes in such a low lever Conda package.

### Design

Track the `01-fix-exec.sh` influential file.
Fix permissions on said file during %post.

### Impact

The versioner.yaml file will correctly track all related Apptainer files.
The source file will have correct permissions.

milljm:
Remove DEBUG and other offending flags from CXXFLAGS/CPPFLAGS.

When installing moose-mpi, supply a default set of python packages necessary to build MOOSE.

Continue to constrain to a Python range. Default Python is latest supported version (3.11 at this commit)

Pin HDF5 to 1.14.2 due to discovered issue: https://github.com/idaholab/moose/issues/28350

As requested, incorporating `set` options we require when launching containers with exec subcommand.

Bump SDK for Mac Intel (PETSc contrib requirement).

Transition to files section for exec fix

Pin several packages, as we are tired of experiencing issues with only small bumps to build versions.

Refs #28427 #28283

Closes #28643


apptainer/moose-dev.def
```diff
@@ -152,9 +164,42 @@ From: {{ APPTAINER_FROM }}
...
 
+{% if PROFILING is defined %}
+    # Install gperftools
+    GPERFTOOLS_VERSION=gperftools-2.15
+    cd ${ROOT_BUILD_DIR}
+    git clone -b ${GPERFTOOLS_VERSION} --single-branch https://github.com/gperftools/gperftools.git
+    GPERFTOOLS_SRC=${ROOT_BUILD_DIR}/gperftools
+    cd ${GPERFTOOLS_SRC}
+    ./autogen.sh
+    ./configure --prefix={{ GPERF_DIR }} --enable-frame-pointers
```
lindsayad, line 178: Are people hoping to do a lot of heap profiling? That would be the reason I'd recommend going to the frame-pointer route. Otherwise using libunwind gives more accurate stack traces, particularly in my experience if fortran libraries like mumps are involved

lindsayad:
I'm confused. Is this meant to replace https://github.com/idaholab/moose/pull/28587 ?

loganharbour:
> I'm confused. Is this meant to replace #28587 ?

It combines it into this one. No need to update mpi and up, and then petsc and up in two PRs.

loganharbour (approved):

apptainer/moose-dev.def
milljm: No one seems to be commenting on this. And I am only adding what other have been asking for.

loganharbour: We talked about this. This is fine as is, and we can change it around later if needed.

PR #28658, opened 2024-09-18
Using field variables instead of scalar variables for THM junctions, part 3

joshuahansel:
Step 3 of https://github.com/idaholab/moose/pull/28514.

Refs #25593 

This adds a new option "use_scalar_variables" to all volume-junction-based components, which is true by default but gives a deprecation message. The deprecation message points to a URL with relevant info on transitioning.

There were several re-golds, but none of them were from actually different results. They were all from:

New names for junction variables
New elements in the mesh (NodeElems)


joshuahansel:
@loganharbour I added a function `MooseUtils::mooseDocsURL(path)` here for referring to a doc page in an error/warning message. Opinions on this? Maybe there should be a virtual method in the app, so that other apps can do a similar thing? With non-open apps, there are two URLs, one for binary and one for source. Maybe that can be some build preprocessor to flip those?

joshuahansel:
@lindsayad This is ready. The juicy bits are in `VolumeJunction1Phase` and `ADVolumeJunctionBaseUserObject`.

joshuahansel:
@licharlot Can you review this when you get a chance? @lindsayad is on paternity leave now, so he no longer gets the great privilege.

licharlot (approved):

PR #28668, opened 2024-09-19
Restrict the volume force baseline check to a subdomain.

### LINKED ISSUE #28667: Volume force correction in  `INSFVRhieChowInterpolator` is not restricted to given blocks.
### Bug Description

Volume force correction in  `INSFVRhieChowInterpolator` is not restricted to given blocks. 
Results in access error when we have a simulation which is only part fluid.

### Steps to Reproduce

Define a fluid block and solid block and add force corrections to the Rhie-Chow interpolator.

### Impact

Renders force correction impossible with complex reactor cases.

### [Optional] Diagnostics

_No response_

grmnptr:
Closes #28667


grmnptr:
This should be an easy fix and people doing MSR need this bad, considering they have their reflector region without any fluid flow.

GiudGiud (approved):

grmnptr:
Failure is not related.

Issue #25275, opened 2023-08-22
Add fin heat transfer enhancement capability


joshuahansel:
## Reason
Some thermal systems features fins to enhance heat transfer.

## Design
This would add two functor materials:

- `FinEnhancementFactorFunctorMaterial`
- `FinEfficiencyFunctorMaterial`

The former would be for the factor that is used in a convective boundary condition, and the latter computes (one option for) the fin efficiency used by the former.

A later step is needed for the boundary conditions of interest to make sure that they can accept a functor material property to multiply the term.

## Impact
New classes. No change to existing inputs.

PR #28669, opened 2024-09-19
Added scale parameter to HeatTransferFromHeatStructure1Phase

joshuahansel:
This adds a parameter `scale` to `HeatTransferFromHeatStructure1Phase`, which can be used to enhance heat transfer due to fins.

Refs #25275


modules/thermal_hydraulics/test/tests/components/heat_transfer_from_heat_structure_1phase/fin_enhancement.i
```diff
@@ -0,0 +1,240 @@
...
+p_outlet = 1e5
+T_initial_hs = 800
+mfr_inlet = ${fparse rho_inlet * vel_inlet * area}
+
+htc = 100
+
+# Suppose that there are 10 rectangular, 1-mm-thick fins of height 1 cm over the length
+# of the cooled section.
+n_fin = 10
+h_fin = 0.01
```
licharlot, line 27: The heat structure is coupled with the flow channel on the inner side, so it'd be more realistic if the fin hight was smaller than the channel diameter.

licharlot (commented):
Should you also add a test for the `HSBoundaryAmbientConvection` component?

joshuahansel:
> Should you also add a test for the `HSBoundaryAmbientConvection` component?

There already is one.

licharlot (approved):

PR #28675, opened 2024-09-20
Fix behavior of four-noded shell for curved geometries

### LINKED ISSUE #28674: Shell elements give inaccurate results on curved geometries
### Bug Description

The implementation of the MITC4 shell element in MOOSE produces results that are inaccurate when modeling curved geometries. 

### Steps to Reproduce

This can be illustrated with the pinched_cylinder and pinched_cylinder_symm_unstructured tests in the solid_mechanics module. Those tests model a section of a cylinder with symmetry planes that is pinched in the x-direction. If the BCs are modified to pinch in the y-direction instead, they should produce the same displacement, but in a different direction. However, the results differ. 

### Impact

This prevents the use of this element for anything but flat geometries.

### [Optional] Diagnostics

_No response_

bwspenc:
closes #28674 

With these changes the pinched cylinder problem now gives equivalent results when loading is applied in the x or y directions.


modules/solid_mechanics/test/tests/shell/static/pinched_cylinder_symm_unstructured.i
```diff
@@ -1,17 +1,5 @@
 # Test for displacement of pinched cylinder (similar to pinch_cyl_symm.i)
-# The Test is run with an unstructured mesh
```
bwspenc, line 2: Note that I'm removing this results table because the results differ somewhat from the prior results. I don't have the three meshes used for convergence, so these are likely not correct. This could be added back in once the follow-on commit with the cylinder coordinate system gets added.

jmeier:
Dear @bwspenc,

Please allow me to ask two questions regarding this PR:

1. This PR changes the behaviour in `modules/solid_mechanics`. So it leaves the (deprecated) `modules/tensor_mechanics` unchanged. Should we also change `modules/tensor_mechanics` to keep the behaviour consistent even if it will be removed in the future?
2. The same normalization can be found in `ADInertialForceShell.C`. Again for `solid_mechanics` and `tensor_mechanics`. Is this code also affected by the same problem and shoul be changed?
    https://github.com/idaholab/moose/blob/5570e4a9f30ba29240404425793a219d21fe57dd/modules/solid_mechanics/src/kernels/ADInertialForceShell.C#L241-L245

Jörg

bwspenc:
> 1. This PR changes the behaviour in `modules/solid_mechanics`. So it leaves the (deprecated) `modules/tensor_mechanics` unchanged. Should we also change `modules/tensor_mechanics` to keep the behaviour consistent even if it will be removed in the future?

We actually moved all the code over from `modules/tensor_mechanics` into `modules/solid_mechanics`, so there isn't anything there to update.

> 2. The same normalization can be found in `ADInertialForceShell.C`. Again for `solid_mechanics` and `tensor_mechanics`. Is this code also affected by the same problem and shoul be changed?
>    https://github.com/idaholab/moose/blob/5570e4a9f30ba29240404425793a219d21fe57dd/modules/solid_mechanics/src/kernels/ADInertialForceShell.C#L241-L245

Thanks for pointing that out. I'll probably just address that on my next iteration of this change set.

modules/solid_mechanics/test/tests/shell/static/pinched_cylinder_symm.i
```diff
@@ -11,137 +11,112 @@
...
+#                      |  (MOOSE implementation) | (Reported by Dvorkin)
 #----------------------|-------------------------|-------------------------
-#     10 x 10          |          0.806          |        0.83
-#     20 x 20          |          1.06           |        0.96
-#     40 x 40          |          0.95           |         -
-#     80 x 160         |          0.96           |         -
+#     10 x 10          |          0.82           |        0.83
+#     20 x 20          |          0.95           |        0.96
+#     40 x 40          |          0.99           |         -
+#     80 x 160         |          1.01           |         -
```
dschwen, line 28:
```suggestion
#     80 x 80          |          1.01           |         -
```

dschwen (approved):

Issue #27979, opened 2024-06-24
Improve error message for PolygonConcentricCircleMeshGenerator


lewisgross1296:
### Motivation

I spent an embarrassing amount of time trying to figure out why my a block used in two mesh generation scripts
```
  [fuel_compact]
    type = PolygonConcentricCircleMeshGenerator
    num_sides = 6
    polygon_size = ${fparse pin_lattice_pitch/2}
    ring_radii = '${fparse 0.6*fuel_compact_radius} ${fparse 0.8*fuel_compact_radius} ${fuel_compact_radius}'
    ring_intervals = '2 2 2'
    num_sectors_per_side = '6 6 6 6 6 6'
    ring_block_ids = '2 2 2'
    ring_block_names = 'fuel fuel fuel'
    background_block_ids = '1'
    background_block_names = 'graphite'
    background_intervals = 5
  []
```
was failing in one and suceeding in another via this error message
```

*** ERROR ***
solid_mesh.i:17: (Mesh/fuel_compact/ring_block_ids):
    This parameter must have the appropriate size if it is provided. The size should be the same as the size
    of 'ring_intervals' if the innermost ring interval (including boundary layers) is unity; otherwise the 
    size should  be greater than the size of 'ring_intervals' by one. If 'quad_center_elements' is true, 
    it is optional to only provide this parameter with the same size as 'ring_intervals'
```

Turns out it was because I forgot to copy over
```
[GlobalParams]
  quad_center_elements = true
[]
```

### Design

Looking at `modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C`, the error messages are pretty verbose and could probably be improved with some more case checking and printing out the specific case which caused the error, instead of printing all the possible issues. 

E.g. here, it could have figured out that `quad_center_elements` was not set and print something like 
```
Expecting a ring_block_ids with length of 4, since ring_intervals has a length of 3.
Either add one more block id or set quad_center_elements = true
```

### Impact

Users will spend less time debugging their mesh generation script

lewisgross1296:
I see what you are saying. The last sentence does read better as you wrote it.

I guess my complaint is more that having this
```
If 'quad_center_elements' is true, it is optional to only provide this parameter with the same size as 'ring_intervals'
 ```
in the error message puts more work on the user to parse an already long and technical message, when instead we can just check the value of `quad_center_elements`.

[...]

PR #28685, opened 2024-09-23
Improve error message for PolygonConcentricCircleMeshGenerator #27979

kyriv-lab:
This PR addresses issue: #27979

### DIFF
--- a/modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
+++ b/modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
@@ -342,25 +342,55 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
     if (!_ring_block_ids.empty() &&
         _ring_block_ids.size() !=
             (_ring_intervals.size() + (unsigned int)(num_innermost_ring_layers != 1)))
-      paramError("ring_block_ids",
-                 "This parameter must have the appropriate size if it is provided. The size should "
-                 "be the same as the size of 'ring_intervals' if the innermost ring interval "
-                 "(including boundary layers) is unity; otherwise the size should be greater than "
-                 "the size of 'ring_intervals' by one. If 'quad_center_elements' is true, it is "
-                 "optional to only provide this parameter with the same size as 'ring_intervals'");
+    {
+      _console << "quad_center_elements is : " << (_quad_center_elements ? "true" : "false")
+               << std::endl;
+      _console << "ring_block_ids size is : " << _ring_block_ids.size() << std::endl;
+      _console << "ring_intervals size is : " << _ring_intervals.size() << std::endl;
+      _console << "first element of ring_intervals is : " << num_innermost_ring_layers << std::endl;
+      if (!_quad_center_elements)
+        paramError("ring_block_ids",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the first element of 'ring_intervals' is more than one, "
+                   "the size of 'ring_block_ids' must be "
+                   "equal to the size of 'ring_intervals' + 1. Else, equal to the size of "
+                   "'ring_intervals'");
+      else
+        paramError("ring_block_ids",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the first element of 'ring_intervals' is more than one, "
+                   "the size of 'ring_block_ids' must be "
+                   "equal to the size of 'ring_intervals' or 'ring_intervals' + 1. Else, equal to "
+                   "the size of ring_intervals' ",
+                   _quad_center_elements);
+    }
     if (!_ring_block_names.empty() && _quad_center_elements && num_innermost_ring_layers > 1 &&
         _ring_block_names.size() == _ring_intervals.size())
       _ring_block_names.insert(_ring_block_names.begin(), _ring_block_names.front());
     if (!_ring_block_names.empty() &&
         _ring_block_names.size() !=
             (_ring_intervals.size() + (unsigned int)(num_innermost_ring_layers != 1)))
-      paramError(
-          "ring_block_names",
-          "This parameter must have the appropriate size if it is set. The size should be the "
-          "same as the size of 'ring_intervals' if the innermost ring interval (including "
-          "boundary layers) is unity; otherwise the size should be greater than the size of "
-          "'ring_intervals' by one. If 'quad_center_elements' is true, it is optional to only "
-          "provide this parameter with the same size as 'ring_intervals'");
+    {
+      _console << "quad_center_elements is : " << (_quad_center_elements ? "true" : "false")
+               << std::endl;
+      _console << "ring_block_names size is : " << _ring_block_names.size() << std::endl;
+      _console << "ring_intervals size is : " << _ring_intervals.size() << std::endl;
+      _console << "first element of ring_intervals is : " << num_innermost_ring_layers << std::endl;
+      if (!_quad_center_elements)
+        paramError("ring_block_names",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the first element of 'ring_intervals' is more than one, "
+                   "the size of 'ring_block_names' must be "
+                   "equal to the size of 'ring_intervals' + 1. Else, equal to the size of "
+                   "'ring_intervals'");
+      else
+        paramError("ring_block_names",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the first element of 'ring_intervals' is more than one, "
+                   "the size of 'ring_block_names' must be "
+                   "equal to the size of 'ring_intervals' or 'ring_intervals' + 1. Else, equal to "
+                   "the size of 'ring_intervals'");
+    }
     for (unsigned int i = 0; i < _ring_radii.size(); i++)
     {
       const Real layer_width = _ring_radii[i] - (i == 0 ? 0.0 : _ring_radii[i - 1]);
--- a/modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
+++ b/modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
@@ -336,60 +336,85 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
     const unsigned int num_innermost_ring_layers =
         _ring_inner_boundary_layer_params.intervals.front() + _ring_intervals.front() +
         _ring_outer_boundary_layer_params.intervals.front();
+    // If conditions are met, duplicate the first element of _ring_block_ids at the start.
     if (!_ring_block_ids.empty() && _quad_center_elements && num_innermost_ring_layers > 1 &&
         _ring_block_ids.size() == _ring_intervals.size())
       _ring_block_ids.insert(_ring_block_ids.begin(), _ring_block_ids.front());
+    // check if the number of ring block ids is appropiate
     if (!_ring_block_ids.empty() &&
         _ring_block_ids.size() !=
             (_ring_intervals.size() + (unsigned int)(num_innermost_ring_layers != 1)))
     {
-      _console << "quad_center_elements is : " << (_quad_center_elements ? "true" : "false")
-               << std::endl;
-      _console << "ring_block_ids size is : " << _ring_block_ids.size() << std::endl;
-      _console << "ring_intervals size is : " << _ring_intervals.size() << std::endl;
-      _console << "first element of ring_intervals is : " << num_innermost_ring_layers << std::endl;
+      // Create an ostringstream for the debug information
+      std::ostringstream debug_info;
+      debug_info << "quad_center_elements is : " << (_quad_center_elements ? "true" : "false")
+                 << std::endl;
+      debug_info << "ring_block_ids size is : " << _ring_block_ids.size() << std::endl;
+      debug_info << "ring_intervals size is : " << _ring_intervals.size() << std::endl;
+      debug_info << "first element of ring_intervals is : " << num_innermost_ring_layers
+                 << std::endl;
+
+      // error message
       if (!_quad_center_elements)
+      {
         paramError("ring_block_ids",
                    "This parameter must have the appropriate size if it is provided: "
                    "If the first element of 'ring_intervals' is more than one, "
                    "the size of 'ring_block_ids' must be "
                    "equal to the size of 'ring_intervals' + 1. Else, equal to the size of "
-                   "'ring_intervals'");
+                   "'ring_intervals'\n",
+                   debug_info.str());
+      }
       else
+      {
         paramError("ring_block_ids",
                    "This parameter must have the appropriate size if it is provided: "
                    "If the first element of 'ring_intervals' is more than one, "
                    "the size of 'ring_block_ids' must be "
                    "equal to the size of 'ring_intervals' or 'ring_intervals' + 1. Else, equal to "
-                   "the size of ring_intervals' ",
-                   _quad_center_elements);
+                   "the size of 'ring_intervals'\n",
+                   debug_info.str());
+      }
     }
+    // If conditions are met, duplicate the first element of _ring_block_names at the start.
     if (!_ring_block_names.empty() && _quad_center_elements && num_innermost_ring_layers > 1 &&
         _ring_block_names.size() == _ring_intervals.size())
       _ring_block_names.insert(_ring_block_names.begin(), _ring_block_names.front());
+    // check if the number of ring block names is appropiate
     if (!_ring_block_names.empty() &&
         _ring_block_names.size() !=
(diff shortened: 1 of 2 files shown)


kyriv1980:
@lewisgross1296 , @lindsayad  please review at your convenience.

lewisgross1296:
I still think it would be better to check `quad_center_elements` and print two different messages based on whether it is true or false. Once we know that, we can print an error message that includes the allowed size of the parameters based on user input and prints what they are currently are producing. IMO this is the quickest way to help users understand their mistake.

The proposed changes to the error messages made them longer, which to me makes them harder to understand and defeats the purpose of my original issue.

miaoyinb:
I agree with @lewisgross1296. The message is kind of too long now.

lewisgross1296:
This is starting to look better. What do people think about reporting the value of `quad_center_elements`, as well as the values of size(ring_block_ids) and size(ring_intervals) in the error message? While maybe not strictly necessary, I could see it being useful to the user.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,40 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
         _ring_block_ids.size() !=
             (_ring_intervals.size() + (unsigned int)(num_innermost_ring_layers != 1)))
-      paramError("ring_block_ids",
-                 "This parameter must have the appropriate size if it is provided. The size should "
-                 "be the same as the size of 'ring_intervals' if the innermost ring interval "
-                 "(including boundary layers) is unity; otherwise the size should be greater than "
-                 "the size of 'ring_intervals' by one. If 'quad_center_elements' is true, it is "
-                 "optional to only provide this parameter with the same size as 'ring_intervals'");
+    {
+      if (!_quad_center_elements)
```
miaoyinb, line 346: Please mention the value of "quad_center_elements" in the error messages.

Also, please add the corresponding tests for code coverage. Thx.

lewisgross1296 (changes requested):
This is more what I was imagining. Some tests would be nice (and to make sure the output looks good). 

A low hanging fruit for tests would be the original code blocks I posted in #27979, or a simplified version of those.

Probably want a test for both cases of `quad_center_elements` (true + wrong size and false + wrong size).

Thanks for being open to feedback!

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: I have added some print statements of the input parameters before the error message. I have also added two tests.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,55 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
         _ring_block_ids.size() !=
             (_ring_intervals.size() + (unsigned int)(num_innermost_ring_layers != 1)))
-      paramError("ring_block_ids",
-                 "This parameter must have the appropriate size if it is provided. The size should "
-                 "be the same as the size of 'ring_intervals' if the innermost ring interval "
-                 "(including boundary layers) is unity; otherwise the size should be greater than "
-                 "the size of 'ring_intervals' by one. If 'quad_center_elements' is true, it is "
-                 "optional to only provide this parameter with the same size as 'ring_intervals'");
+    {
+      _console << "quad_center_elements is : " << (_quad_center_elements ? "true" : "false")
```
miaoyinb, line 346: @GiudGiud Is the the preferred way of doing this? or we want this to be done by constructing a string type message first and then using `paramError` or `mooseError` to print it out?

modules/reactor/test/tests/meshgenerators/polygon_concentric_circle_mesh_generator/tests
```diff
@@ -495,7 +495,27 @@
...
+    []
+    [err_ring_block_ids_quad]
+      type = 'RunException'
+      input = 'poly_2d.i'
+      cli_args = 'Mesh/gen/ring_radii="2 3" Mesh/gen/ring_intervals="2 2"
+                  Mesh/gen/ring_block_ids="20"
+                  Mesh/gen/quad_center_elements="true"
+                  --mesh-only "poly_2d_err.e"'
+      expect_err = 'This parameter must have the appropriate size if it is provided.'
+      detail = "ring_block_ids size is not consistent with number of rings in the mesh."
```
miaoyinb, line 509: Make the detail different from each other to avert the CIVET error.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
miaoyinb: This printing only happens when there is an error. The `if` after this is used to print different messages based on another input parameter's value.

GiudGiud: it could be fine either way then

modules/reactor/test/tests/meshgenerators/polygon_concentric_circle_mesh_generator/tests
kyriv1980: I edited the detail. Thank you.

miaoyinb (approved):

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,55 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
     if (!_ring_block_ids.empty() &&
```
GiudGiud, line 342: add a comment for each of the two statements above

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,76 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
+
+      // error message
+      if (!_quad_center_elements)
+      {
+        paramError("ring_block_ids",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the first element of 'ring_intervals' is more than one, "
+                   "the size of 'ring_block_ids' must be "
+                   "equal to the size of 'ring_intervals' + 1. Else, equal to the size of "
+                   "'ring_intervals'",
```
GiudGiud, line 363:
```suggestion
                   "'ring_intervals'\n",
```

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,76 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
+      debug_info << "ring_intervals size is : " << _ring_intervals.size() << std::endl;
+      debug_info << "first element of ring_intervals is : " << num_innermost_ring_layers
+                 << std::endl;
+
+      // error message
+      if (!_quad_center_elements)
+      {
+        paramError("ring_block_ids",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the first element of 'ring_intervals' is more than one, "
```
GiudGiud, line 360: it's not just that, it s also the size of the boundary layers.
if either:
- there is more than one interval in the innermost ring
- there is a boundary layer (inner or outer) in the innermost ring
and
- we are not using quad center elements
Then there must be an extra block id (for the non-quad element)

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,76 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
     if (!_ring_block_ids.empty() &&
         _ring_block_ids.size() !=
             (_ring_intervals.size() + (unsigned int)(num_innermost_ring_layers != 1)))
```
GiudGiud, line 344: I dont understand why we need the extra block id if _quad_center_elements is true ?

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -343,26 +343,36 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
       if (!_quad_center_elements)
+      {
         paramError("ring_block_ids",
                    "This parameter must have the appropriate size if it is provided: "
                    "If the first element of 'ring_intervals' is more than one, "
                    "the size of 'ring_block_ids' must be "
                    "equal to the size of 'ring_intervals' + 1. Else, equal to the size of "
-                   "'ring_intervals'");
+                   "'ring_intervals'",
+                   debug_info.str());
```
roystgnr, line 364: Do we need a newline in between here to make sure the output doesn't look like `'ring_intervals'quad_center_elements`?

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,76 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
-          "same as the size of 'ring_intervals' if the innermost ring interval (including "
-          "boundary layers) is unity; otherwise the size should be greater than the size of "
-          "'ring_intervals' by one. If 'quad_center_elements' is true, it is optional to only "
-          "provide this parameter with the same size as 'ring_intervals'");
+    {
+      // Create an ostringstream for the debug information
+      std::ostringstream debug_info;
+      debug_info << "quad_center_elements is : " << (_quad_center_elements ? "true" : "false")
+                 << std::endl;
+      debug_info << "ring_block_names size is : " << _ring_block_ids.size() << std::endl;
```
GiudGiud, line 388:
```suggestion
      debug_info << "ring_block_names size is : " << _ring_block_names.size() << std::endl;
```

GiudGiud (changes requested):

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: If quad elements is true, the user can choose to provide it or not. But if he doesn't the one gets duplicated.

GiudGiud: ok seems fine but it s an implementation detail. focus on making the error message clear, then you can mention that as a note at the end.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: I think that wont hurt at all. I will add it at the beginning of the debug string.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -342,25 +342,55 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
-                 "be the same as the size of 'ring_intervals' if the innermost ring interval "
-                 "(including boundary layers) is unity; otherwise the size should be greater than "
-                 "the size of 'ring_intervals' by one. If 'quad_center_elements' is true, it is "
-                 "optional to only provide this parameter with the same size as 'ring_intervals'");
+    {
+      _console << "quad_center_elements is : " << (_quad_center_elements ? "true" : "false")
+               << std::endl;
+      _console << "ring_block_ids size is : " << _ring_block_ids.size() << std::endl;
+      _console << "ring_intervals size is : " << _ring_intervals.size() << std::endl;
+      _console << "first element of ring_intervals is : " << num_innermost_ring_layers << std::endl;
```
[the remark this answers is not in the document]
kyriv1980: how would you rephrase it?
innermost ring layers?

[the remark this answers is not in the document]
GiudGiud: works

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: will do.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: added the new line

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: fixed that.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: The boundary layer and the ring intervals together make the num_innermost_ring_layers. I can rephrase it like that

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -336,31 +336,86 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
+      debug_info << "number of innermost ring layers is : " << num_innermost_ring_layers
+                 << std::endl;
+
+      // error message
+      if (!_quad_center_elements)
+      {
+        paramError("ring_block_names",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the number of the innermost ring layers is more than one, "
+                   "and we have non-quad elements, the size of 'ring_block_names' must be "
```
GiudGiud, line 403:
```suggestion
                   "If the number of the innermost ring layers (" + std::to_string(num_innermost_ring_layers)) + " :first ring interval and inner/outer boundaries of the first ring) is more than one, "
                   "and since the central elements are not QUADs, the size of 'ring_block_names' must be "
```

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -336,31 +336,86 @@ PolygonConcentricCircleMeshGeneratorBase::PolygonConcentricCircleMeshGeneratorBa
...
+                   "equal to the size of 'ring_intervals' + 1. Else, equal to the size of "
+                   "'ring_intervals'\n",
+                   debug_info.str());
+      }
+      else
+      {
+        paramError("ring_block_names",
+                   "This parameter must have the appropriate size if it is provided: "
+                   "If the number of the innermost ring layers is more than one, "
+                   "and we have quad elements, the size of 'ring_block_names' must be "
```
GiudGiud, line 413: customize this message more, we know we have all quads in this case

GiudGiud (changes requested):

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: i will edit.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
kyriv1980: we know that we either have or not have quad central elements. If we mention that we don't have quad central elements in one error message we should also mention that we have them in the other.

GiudGiud: this "we have quad elements" is inside the 'if' condition, so we are not asserting that we have quads here

GiudGiud: I think this whole thing would be a lot clearer if you had a message structured like this:
- there are this many inner most ring layers
- we are / are not using quad elements for the center elements
Therefore the number of block ids should equal to the number ... ( <that number> )

GiudGiud (approved):
we could code-golf this a little more to reduce the number of lines but let's move on

thanks for the reviews everyone

PR #28692, opened 2024-09-24
Update MooseServer formatting to not use wrong document text

brandonlangley:
refs #22766


brandonlangley:
@dschwen @loganharbour -

This fixes the issue where document formatting was using the text from the last document edited.

The MooseServer was updated to keep a map from document paths to their current updated text.

The document formatting now uses that map to find the correct text based on the document path.

You can observe this issue in VS Code before the fix then notice it works as expected after the fix.

Could I ask one of you to review and merge this when you get a chance?

dschwen (approved):

PR #28742, opened 2024-10-01
Calculating local quantities for Q4 shell elements

### LINKED ISSUE #28741: Calculation of local quantities for shell elements
### Motivation

A uniform and compatible local coordinate system is needed when using shells for curved geometries. This allows for the definition of local quantities such as stresses, forces, and moments.

### Design

- A convention was used to either automatically calculate local coordinates or allow users to define them manually.
- Aux variables were added to visualize the local vectors
- The local stresses, forces and bending moments were calculated and stored using aux variables.
- Tests were added, and the doc file was edited

### Impact

- With the new local coordinate system, only one test, **large_beam_bending**, fails. However, the difference between the new and old results for this test is minimal and negligible, so we regold it.
- Other features, such as the calculation of local quantities, will not affect existing capabilities.

Kavan-Khaledi:
@bwspenc @dschwen 
## Reason

To calculate the local quantities such as local stresses, forces and bending moments for shell elements

## Design

- A convention was used to either automatically calculate local coordinates or allow users to define them manually.

- Aux variables were added to visualize the local vectors

- The local stresses, forces and bending moments were calculated and stored using aux variables.

- Tests were added, and the doc file was edited

## Impact

- With the new local coordinate system, only one test, **large_beam_bending**, fails. However, the difference between the new and old results for this test is minimal and negligible, so we regold it.

- Other features, such as the calculation of local quantities, will not affect existing capabilities.

Closes #28741


bwspenc:
My apologies -- I just saw this! I get so many github notifications that I'm not very good at staying on top of them. I'll start looking at this.

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -388,12 +425,26 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
       (*_transformation_matrix)[i] = J;
 
-      // calculate ge
-      ADRealVectorValue e3 = (*_dxyz_dzeta[j])[i] / (*_dxyz_dzeta[j])[i].norm();
-      ADRealVectorValue e1 = (*_dxyz_deta[j])[i].cross(e3);
-      e1 /= e1.norm();
-      ADRealVectorValue e2 = e3.cross(e1);
-      e2 /= e2.norm();
+      // compute element's local coordinate
+      computeLocalCoordinate();
```
bwspenc, line 429:
```suggestion
      computeLocalCoordinates();
```

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -410,17 +461,65 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
+{
+  // default option:the convention used to calculate the local nodal coordinates is also adopted for
+  // calculating the element's local coordinates
+  // the first local vector can be also defined by the user
+
+  // All nodes have the same normal vector. Therefore, here, we use only the normal vecor of the
+  // first element as "the element's normal vector"
+  _e3 = _node_normal[0];
+
+  if (_is_v1_defined)
```
bwspenc, line 488: With the suggested changes to have `_ref_first_local_dir` always exist and default to the y-direction, you don't need this `if` statment anymore, and you don't need to get the parameter (which is expensive, so you don't want to do it frequently like you're doing it currently).

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -410,17 +461,65 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
+  {
+    _e1 = getParam<RealVectorValue>("first_local_vector_ref");
+
+    if (MooseUtils::absoluteFuzzyEqual(_e1.norm(), 0.0, 1e-3))
```
bwspenc, line 492: This check can be moved to the class's constructor so you just need to do it once up front.

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -410,17 +461,65 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
+      mooseError(
+          "The norm of the defined first local axis is less than the acceptable telerance (1e-3)");
+
+    _e1 /= _e1.norm();
+
+    // vector v1 and the normal are considered parallel if the angle between them is less
+    // than 0.05 degree
+    if (MooseUtils::absoluteFuzzyEqual(std::abs(_e1 * _e3), 1.0, 0.05 * libMesh::pi / 180.0))
```
bwspenc, line 500: I'd just use `libMesh::TOLERANCE` as your small number. That way you don't need to do this math every time you come in here.

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -410,17 +461,65 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
     }
   }
+  else
+  {
+    _e1 = _x2.cross(_e3);
+    // If x2 is parallel to the element's normal, set e1 to x3 (less than 0.05 degree is considered
+    // parallel)
+    if (MooseUtils::absoluteFuzzyEqual(std::abs(_x2 * _e3), 1.0, 0.05 * libMesh::pi / 180.0))
+      _e1 = _x3;
+  }
```
bwspenc, line 518:
```suggestion
```

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -434,17 +533,11 @@ void
...
-
-    // If x2 is parallel to node normal, set V1 to x3
-    if (MooseUtils::absoluteFuzzyEqual(_v1[k].norm(), 0.0, 1e-6))
-      _v1[k] = _x3;
-
-    _v1[k] /= _v1[k].norm();
-
-    _v2[k] = _node_normal[k].cross(_v1[k]);
+    _v1[k] = _e1;
+    _v2[k] = _e2;
```
bwspenc, line 540: If you get rid of _v1 and _v2 you don't need to do this.

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -260,6 +280,18 @@ ADComputeIncrementalShellStrain::computeProperties()
                                       _unrotated_total_strain *
                                       (*_contravariant_transformation_matrix[j])[i].transpose();
     }
+    // the element's local coordinate is saved here for the post processing and visualization
```
bwspenc, line 283: I'm not sure this is a great idea to do this every time this gets called. A few alternatives to think about:
- Do this optionally, only if the user asks for it
- Make these direction vectors stateful properties, so you're just computing them once at the beginning like we're doing for the B matrix, etc. This would require a little memory but would cut down on compute cost.
- Have an AuxKernel that computes this

Maybe we should test how much overhead is involved in this. If it's negligible, then just keep it as-is.

modules/solid_mechanics/src/materials/ADComputeShellStress.C
```diff
@@ -38,12 +38,16 @@ ADComputeShellStress::ADComputeShellStress(const InputParameters & parameters)
...
+  _t_weights = t_qrule->get_weights();
   _elasticity_tensor.resize(_t_points.size());
   _stress.resize(_t_points.size());
   _stress_old.resize(_t_points.size());
   _strain_increment.resize(_t_points.size());
+  _local_transformation_matrix.resize(_t_points.size());
   _covariant_transformation_matrix.resize(_t_points.size());
   _global_stress.resize(_t_points.size());
+  _local_stress.resize(_t_points.size());
+  _t_shell = &getMaterialProperty<Real>("shell_thickness");
```
bwspenc, line 50: As I noted for the strain calculator class, I think it's better to directly couple in the shell thickness as a variable here rather than computing it as a material property in the other class and retrieving it here.

modules/solid_mechanics/src/materials/ADComputeShellStress.C
```diff
@@ -54,11 +58,24 @@ ADComputeShellStress::ADComputeShellStress(const InputParameters & parameters)
...
+        &getMaterialProperty<RankTwoTensor>("local_transformation_t_points_" + std::to_string(t));
     _covariant_transformation_matrix[t] = &getMaterialProperty<RankTwoTensor>(
         "covariant_transformation_t_points_" + std::to_string(t));
     _global_stress[t] =
         &declareProperty<RankTwoTensor>("global_stress_t_points_" + std::to_string(t));
+    _local_stress[t] =
+        &declareProperty<RankTwoTensor>("local_stress_t_points_" + std::to_string(t));
   }
+
+  _shell_force_1 = &declareProperty<Real>("shell_force_1");
```
bwspenc, line 71: I'm not sure I like the idea of declaring and computing these properties all the time in this class since they're just used for output. What would you think about making an elemental AuxKernel that computes the average resultants? Since you have to add an AuxKernel to output the averages of these material properties, it wouldn't be any more work than you're already doing in the input file.

bwspenc (changes requested):

modules/solid_mechanics/include/materials/ADComputeIncrementalShellStrain.h
```diff
@@ -186,4 +199,7 @@ class ADComputeIncrementalShellStrain : public Material
   ADRealVectorValue _g2_b;
   ADRealVectorValue _g2_d;
   RankTwoTensor _unrotated_total_strain;
+  ADRealVectorValue _e1;
+  ADRealVectorValue _e2;
+  ADRealVectorValue _e3;
```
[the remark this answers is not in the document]
Kavan-Khaledi: I wanted to make this change, but ADComputeFiniteShellStrain.C also contains numerous _V1 and _V2 vectors with different indices of p and k, which is confusing. I think it would be better to leave this for later.

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -41,6 +41,9 @@ ADComputeIncrementalShellStrain::validParams()
                                        "Quadrature order in out of plane direction");
   params.addParam<bool>(
       "large_strain", false, "Set to true to turn on finite strain calculations.");
+  params.addParam<RealVectorValue>(
+      "first_local_vector_ref",
+      "Optional vector defining the reference direction for the first local vector e1");
```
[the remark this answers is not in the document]
Kavan-Khaledi: I updated the description as you suggested, but I don't want to default the variable to RealVectorValue(0, 1, 0) because the local coordinate computation is handled differently when the user defines the "reference_first_local_direction". To clarify this, I added more details to the .md file. I hope, my explaination is clear!

modules/solid_mechanics/doc/content/modules/solid_mechanics/ShellElements.md
```diff
@@ -63,13 +63,13 @@ g^2=\dfrac{g_3 \times g_1}{g_1 \cdot (g_2 \times g_3)}
 \begin{equation}
 g^3=\dfrac{g_1 \times g_2}{g_1 \cdot (g_2 \times g_3)}
 \end{equation}
-These contravariant base vectors satisfy the orthogonality condition: $g^i \cdot g_j =\delta_{ij}$, where $\delta_{ij}$ is Kronecker delta. The $\hat{e}_i$ vectors are the local cartesian coordinate system computed using the covariant base vectors, i.e.,
+These contravariant base vectors satisfy the orthogonality condition: $g^i \cdot g_j =\delta_{ij}$, where $\delta_{ij}$ is Kronecker delta. The $\hat{e}_i$ vectors are the local cartesian coordinate system computed using the following rule:
```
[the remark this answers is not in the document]
Kavan-Khaledi: Well, the Y-axis (0, 1, 0 ) cannot really be the default value for `reference_first_local_direction` parameter. I think, we need an if condition anyway? The if condition is necessary because the first local vector is calculated differently when the user defines the reference direction vector. By default, the first local vector is determined by the **cross product** of the Y-axis with the element's normal vector. However, if the user provides a reference vector, the given vector is directly projected using a **dot product instead of a cross product**.  Please see my explanation in the .md file

modules/solid_mechanics/doc/content/modules/solid_mechanics/ShellElements.md
```diff
@@ -63,13 +63,13 @@ g^2=\dfrac{g_3 \times g_1}{g_1 \cdot (g_2 \times g_3)}
...
 
 \begin{equation}
-\hat{e}_3 = \frac{g_3}{|g_3|}; \hat{e}_1 = \frac{g_2 \times \hat{e}^3}{|g_2 \times \hat{e}^3|}; \hat{e}_2 = \hat{e}_3 \times \hat{e}_1
+\hat{e}_3 = V_n; \hat{e}_1 = \frac{X_2 \times \hat{e}^3}{|X_2 \times \hat{e}^3|}; \hat{e}_2 = \hat{e}_3 \times \hat{e}_1
```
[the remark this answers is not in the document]
Kavan-Khaledi: please see the updated doc file (.md file)

modules/solid_mechanics/doc/content/modules/solid_mechanics/ShellElements.md
```diff
@@ -63,13 +63,13 @@ g^2=\dfrac{g_3 \times g_1}{g_1 \cdot (g_2 \times g_3)}
...
 \end{equation}
 
-The elasticity tensor at each quadrature point (qp) is computed in [ADComputeIsotropicElasticityTensorShell](/ADComputeIsotropicElasticityTensorShell.md).
+In this equation, $V_n$ represents the element's normal vector, and $X_2$ is the unit vector in the global y-direction, given by $X_2={0, 1, 0}$. Users also have the option to define the local coordinate system by introducing the parameter "first_local_vector_ref" in the input file. When this option is selected, the first local vector $e_1$ of each shell element is calculated by projecting the user_defined vector onto the element, ensuring it lies "in plane" and is perpendicular to the element's normal. The second local vector $e_2$ is then determined using right-hand-rule, via the cross product of $e_3$ and $e_1$. The elasticity tensor at each quadrature point (qp) is computed in [ADComputeIsotropicElasticityTensorShell](/ADComputeIsotropicElasticityTensorShell.md).
```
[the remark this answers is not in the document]
Kavan-Khaledi: I have updated the .md file and added more detailed explanations about how the default method and the user-defined approach for computing the element's local vectors work. here it is:

The $\hat{e}_i$ vectors are the local cartesian coordinate system computed using the following rule:


$\hat{e}_3 = V_n; \hat{e}_1 = \frac{X_2 \times \hat{e}_3}{|X_2 \times \hat{e}_3|}; \hat{e}_2 = \hat{e}_3 \times \hat{e}_1$


In this equation, $V_n$ represents the element's normal vector and $X_2$ is the unit vector in the global y-direction, given by $X_2={0, 1, 0}$. Users also have the option to specify the direction of the first local vector by introducing the parameter `reference_first_local_direction` in the input file. When this option is selected, the first local vector $e_1$ of each shell element is calculated by projecting the user-defined vector $\hat{e}_{1r}$ onto the element, ensuring it lies "in-plane" and is perpendicular to the element's normal. The second local vector $\hat{e}_2$ is then determined via the cross product of $\hat{e}_3$ and $\hat{e}_1$. In this case, the computation of the shell's local coordinate is as follows:

\begin{equation}
\hat{e}_3 = V_n; \hat{e}_1 = \dfrac{\hat{e}_{1r}-(\hat{e}_{1r} \dot \hat{e}_3)~\hat{e}_3}{|\hat{e}_{1r}-(\hat{e}_{1r} \dot \hat{e}_3)~\hat{e}_3|} ; \hat{e}_2 = \hat{e}_3 \times \hat{e}_1
\end{equation}

modules/solid_mechanics/include/materials/ADComputeIncrementalShellStrain.h
```diff
@@ -172,11 +179,17 @@ class ADComputeIncrementalShellStrain : public Material
...
+  MaterialProperty<RealVectorValue> * _second_local_vector;
+  MaterialProperty<RealVectorValue> * _normal_local_vector;
+  MaterialProperty<Real> * _shell_thickness;
+
   /// simulation variables
   ADRealVectorValue _x2;
   ADRealVectorValue _x3;
   const NumericVector<Number> * const & _sol;
   const NumericVector<Number> & _sol_old;
+  const bool _is_v1_defined;
```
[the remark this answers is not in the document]
Kavan-Khaledi: I think, for **curved geometries** we need to get that direction from the parameters for each material point during initialialization?? am i wrong? this was my assumption for the curved geometries

modules/solid_mechanics/include/materials/ADComputeIncrementalShellStrain.h
```diff
@@ -172,11 +179,17 @@ class ADComputeIncrementalShellStrain : public Material
   /// Rotation matrix material property
   ADMaterialProperty<RankTwoTensor> * _transformation_matrix;
 
+  MaterialProperty<RealVectorValue> * _first_local_vector;
+  MaterialProperty<RealVectorValue> * _second_local_vector;
+  MaterialProperty<RealVectorValue> * _normal_local_vector;
+  MaterialProperty<Real> * _shell_thickness;
```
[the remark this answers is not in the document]
Kavan-Khaledi: Done

modules/solid_mechanics/include/materials/ADComputeIncrementalShellStrain.h
```diff
@@ -172,11 +179,17 @@ class ADComputeIncrementalShellStrain : public Material
   /// Rotation matrix material property
   ADMaterialProperty<RankTwoTensor> * _transformation_matrix;
 
+  MaterialProperty<RealVectorValue> * _first_local_vector;
+  MaterialProperty<RealVectorValue> * _second_local_vector;
+  MaterialProperty<RealVectorValue> * _normal_local_vector;
```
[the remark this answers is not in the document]
Kavan-Khaledi: Done

modules/solid_mechanics/include/materials/ADComputeShellStress.h
```diff
@@ -44,15 +44,34 @@ class ADComputeShellStress : public Material
   /// Material property for elasticity tensor
   std::vector<const ADMaterialProperty<RankFourTensor> *> _elasticity_tensor;
 
+  /// shell thickness
+  const MaterialProperty<Real> * _t_shell;
```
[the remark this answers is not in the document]
Kavan-Khaledi: Done

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -410,17 +461,65 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
+}
+
+void
+ADComputeIncrementalShellStrain::computeLocalCoordinate()
+{
+  // default option:the convention used to calculate the local nodal coordinates is also adopted for
+  // calculating the element's local coordinates
+  // the first local vector can be also defined by the user
+
+  // All nodes have the same normal vector. Therefore, here, we use only the normal vecor of the
```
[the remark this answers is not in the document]
Kavan-Khaledi: Done

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -260,6 +280,18 @@ ADComputeIncrementalShellStrain::computeProperties()
...
+    for (unsigned int ii = 0; ii < 3; ++ii)
+    {
+      (*_first_local_vector)[i](ii) = MetaPhysicL::raw_value(_e1(ii));
+      (*_second_local_vector)[i](ii) = MetaPhysicL::raw_value(_e2(ii));
+      (*_normal_local_vector)[i](ii) = MetaPhysicL::raw_value(_e3(ii));
+    }
+
+    // save the shell thickness here; to be used later in ComputeShellStress for calculating the
+    // local force and moments
+    (*_shell_thickness)[i] = _thickness[i];
```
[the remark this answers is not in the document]
Kavan-Khaledi: Done

modules/solid_mechanics/include/materials/ADComputeIncrementalShellStrain.h
```diff
@@ -47,6 +47,9 @@ class ADComputeIncrementalShellStrain : public Material
   /// Computes the transformation matrix from natural coordinates to local cartesian coordinates for elasticity tensor transformation
   virtual void computeGMatrix();
 
+  /// Computes the element's local coordinate
```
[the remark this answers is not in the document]
Kavan-Khaledi: done

modules/solid_mechanics/include/materials/ADComputeIncrementalShellStrain.h
```diff
@@ -47,6 +47,9 @@ class ADComputeIncrementalShellStrain : public Material
...
+  virtual void computeLocalCoordinate();
```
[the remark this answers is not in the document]
Kavan-Khaledi: done

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -260,6 +280,18 @@ ADComputeIncrementalShellStrain::computeProperties()
                                       _unrotated_total_strain *
                                       (*_contravariant_transformation_matrix[j])[i].transpose();
     }
+    // the element's local coordinate is saved here for the post processing and visualization
+    computeLocalCoordinate();
```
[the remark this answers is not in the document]
Kavan-Khaledi: done

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
Kavan-Khaledi: done

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
Kavan-Khaledi: done

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
Kavan-Khaledi: Agreed, I added an AuxKernel (ShellLocalCoordinatesAux) that saves the vector components in Aux variables.

modules/solid_mechanics/src/materials/ADComputeShellStress.C
Kavan-Khaledi: done

modules/solid_mechanics/src/materials/ADComputeShellStress.C
Kavan-Khaledi: I added an AuxKernel (ShellResultantsAux) that stores the resultants in Aux variables.

modules/solid_mechanics/include/auxkernels/ShellResultantsAux.h
```diff
@@ -0,0 +1,51 @@
...
+#include "AuxKernel.h"
+#include "RankTwoTensor.h"
+#include "MooseEnum.h"
+
+namespace libMesh
+{
+class QGauss;
+}
+
+// Forward declarations
```
dschwen, line 21:
```suggestion
// Forward declarations
namespace libMesh
{
class QGauss;
}
```

modules/solid_mechanics/include/auxkernels/ShellLocalCoordinatesAux.h
```diff
@@ -0,0 +1,37 @@
...
+
+// Forward declarations
+
+class ShellLocalCoordinatesAux : public AuxKernel
+{
+public:
+  static InputParameters validParams();
+
+  ShellLocalCoordinatesAux(const InputParameters & parameters);
+  virtual ~ShellLocalCoordinatesAux() {}
```
dschwen, line 24:
```suggestion
```

modules/solid_mechanics/src/auxkernels/ShellLocalCoordinatesAux.C
```diff
@@ -0,0 +1,79 @@
...
+  params.addClassDescription(
+      "This AuxKernel stores a specific component of a shell element's local coordinate "
+      "vector in an auxiliary variable.");
+  params.addParam<std::string>("base_name", "Mechanical property base name");
+  MooseEnum property("first_local_vector second_local_vector normal_local_vector");
+  params.addRequiredParam<MooseEnum>(
+      "property",
+      property,
+      "The local axis to output, calculated on the shell element. Valid options are: "
+      "first_local_vector, second_local_vector and normal_local_vector");
```
dschwen, line 29: Listing valid options here is redundant as the MooseEnum already provides this information.

modules/solid_mechanics/src/auxkernels/ShellLocalCoordinatesAux.C
```diff
@@ -0,0 +1,79 @@
...
+  params.addRequiredParam<unsigned int>(
+      "component", "The vector component of the local coordinate vector: 0, 1 or 2");
+
+  return params;
+}
+
+ShellLocalCoordinatesAux::ShellLocalCoordinatesAux(const InputParameters & parameters)
+  : AuxKernel(parameters),
+    _base_name(isParamValid("base_name") ? getParam<std::string>("base_name") + "_" : ""),
+    _property(getParam<MooseEnum>("property")),
```
dschwen, line 39:
```suggestion
    _property(getParam<MooseEnum>("property").getEnum<PropertyType>()),
```

modules/solid_mechanics/include/auxkernels/ShellLocalCoordinatesAux.h
```diff
@@ -0,0 +1,37 @@
...
+
+protected:
+  virtual Real computeValue() override;
+
+  /// Base name of the material system used to calculate the elastic energy
+  const std::string _base_name;
+
+  MooseEnum _property;
```
dschwen, line 32:
```suggestion
  enum class PropertyType {
    first_local_vector, 
    second_local_vector,
    normal_local_vector
  } _property;
```

modules/solid_mechanics/src/auxkernels/ShellLocalCoordinatesAux.C
```diff
@@ -0,0 +1,79 @@
...
+  {
+    case 0: // Corresponds to the first local axis
+      output_value = (*_local_coordinates)[_qp](0, _component);
+      break;
+
+    case 1: // Corresponds to the second local axis
+      output_value = (*_local_coordinates)[_qp](1, _component);
+      break;
+
+    case 2: // Corresponds to the normal vector to the shell element
```
dschwen, line 67:
```suggestion
    case PropertyType::first_local_vector:
      output_value = (*_local_coordinates)[_qp](0, _component);
      break;

    case PropertyType:: second_local_vector:
      output_value = (*_local_coordinates)[_qp](1, _component);
      break;

    case PropertyType:: normal_local_vector:
```

modules/solid_mechanics/src/auxkernels/ShellLocalCoordinatesAux.C
```diff
@@ -0,0 +1,79 @@
...
+      output_value = (*_local_coordinates)[_qp](2, _component);
+      break;
+
+      // Add other cases as needed
+
+    default:
+      mooseError("Invalid definition of the shell's local axis: valid options are : "
+                 "first_local_vector, second_local_vector, normal_local_vector");
```
dschwen, line 75: IMO it's fine to just put Internal Error here. This is a developer issue and the backtrace will be enough information for someone to fix this. Again, we want to avoid redundant specification of the available options.

modules/solid_mechanics/src/auxkernels/ShellResultantsAux.C
```diff
@@ -0,0 +1,150 @@
...
+  _t_qrule = std::make_unique<QGauss>(
+      1, Utility::string_to_enum<Order>(getParam<std::string>("through_thickness_order")));
+  _t_points = _t_qrule->get_points();
+  _t_weights = _t_qrule->get_weights();
+  _local_stress_t_points.resize(_t_points.size());
+  for (unsigned int t = 0; t < _t_points.size(); ++t)
+  {
+    _local_stress_t_points[t] = &getMaterialProperty<RankTwoTensor>(
+        _base_name + "local_stress_t_points_" + std::to_string(t));
+  }
```
dschwen, line 59:
```suggestion
  for (const auto t: index_range(_t_points))
    _local_stress_t_points[t] = &getMaterialProperty<RankTwoTensor>(
        _base_name + "local_stress_t_points_" + std::to_string(t));
```

we prefer using these ranged based loops

modules/solid_mechanics/src/auxkernels/ShellResultantsAux.C
```diff
@@ -0,0 +1,150 @@
...
+}
+
+Real
+ShellResultantsAux::computeValue()
+{
+  Real _shell_resultant = 0.0;
+
+  switch (_resultant)
```
dschwen, line 67: for the `_resultant` we could do the same enum class pattern. I'd be happy to make those menial changes.

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -388,12 +406,26 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
+      ADRankTwoTensor local_rotation_mat;
+      local_rotation_mat(0, 0) = _e1(0);
+      local_rotation_mat(0, 1) = _e1(1);
+      local_rotation_mat(0, 2) = _e1(2);
+      local_rotation_mat(1, 0) = _e2(0);
+      local_rotation_mat(1, 1) = _e2(1);
+      local_rotation_mat(1, 2) = _e2(2);
+      local_rotation_mat(2, 0) = _e3(0);
+      local_rotation_mat(2, 1) = _e3(1);
+      local_rotation_mat(2, 2) = _e3(2);
```
dschwen, line 423: here we have a "from rows" named constructor to make this a one liner

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
```diff
@@ -388,12 +406,26 @@ ADComputeIncrementalShellStrain::computeGMatrix()
...
+
+      for (unsigned int ii = 0; ii < 3; ++ii)
+        for (unsigned int jj = 0; jj < 3; ++jj)
```
dschwen, line 426:
```suggestion
      for (const auto ii: make_range(Moose::dim))
        for (const auto jj: make_range(Moose::dim))
```

dschwen (commented):
Hope those picky comments are not a turn-off. Most of them you could just bulk accept. The remainder I'm happy to help with!

modules/solid_mechanics/src/auxkernels/ShellLocalCoordinatesAux.C
Kavan-Khaledi: you are right, we don't need this error message at all.

modules/solid_mechanics/src/auxkernels/ShellResultantsAux.C
Kavan-Khaledi: Done:) thank you

modules/solid_mechanics/src/materials/ADComputeIncrementalShellStrain.C
Kavan-Khaledi: done;)

modules/solid_mechanics/include/materials/ADComputeIncrementalShellStrain.h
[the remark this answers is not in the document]
bwspenc: OK, that makes sense. That's something that's probably worth considering in the future.

bwspenc (commented):
I've looked through this, and I'm happy with it, except that I think we need to resolve an important point. I would really like it if the behavior could be consistent between when you specify a reference direction, and if you don't specify a reference direction. Unfortunately, it is not--if you specify the reference direction it takes the projection of your direction onto the shell, while if you don't, it takes a cross product. I attempted to unify these two by adding another commit on top of yours in my local branch, but I'm struggling because it breaks a bunch of tests. That's probably because they have rotational BCs that depend on the current default scheme for determining orientations. I'm still thinking about this.

bwspenc:
I have been thinking about this some more, and I am leaning toward resolving the issue I raised above by requiring the user to specify `reference_first_local_direction`, and get rid of the old default behavior. That would require anyone using shell elements to reevaluate how the local directions are defined, and make any needed changes to the rotational BCs in their model.

This would make it so current models fail, but I don't think these are being used very widely (because of this issue). It's better to have a model fail to run than to have it silently behave differently than it did before.

It wouldn't be too hard to make this change. It would require modifying the BCs on some of the rotational DOFs in a few of the regression tests.

Kavan-Khaledi:
> I have been thinking about this some more, and I am leaning toward resolving the issue I raised above by requiring the user to specify `reference_first_local_direction`, and get rid of the old default behavior. That would require anyone using shell elements to reevaluate how the local directions are defined, and make any needed changes to the rotational BCs in their model.
> 
> This would make it so current models fail, but I don't think these are being used very widely (because of this issue). It's better to have a model fail to run than to have it silently behave differently than it did before.
> 
> It wouldn't be too hard to make this change. It would require modifying the BCs on some of the rotational DOFs in a few of the regression tests.

Dear Ben, I remember we had long discussions with Jörg about this topic in the past. But our main assumption was that we should leave the default calculations in Moose unchanged to minimize the impact on the existing tests. I think it makes sense to implement a unified approach, even if this requires changing the BC definition in some tests. I will go through it after the New Year vacations. Thank you and I wish you all happy holidays and a good start into 2025. Kavan

modules/solid_mechanics/test/tests/shell/static/beam_bending_moment_AD_2.i
```diff
@@ -163,117 +163,117 @@
...
 [Postprocessors]
-  [./disp_z_tip]
+  [disp_z_tip]
     type = PointValue
     point = '1.0 0.0 10.0'
     variable = disp_y
-  [../]
-  [./rot_y_tip]
+  []
+  [rot_y_tip]
```
bwspenc, line 232: Minor detail, but should this block be renamed to `rot_x_tip`?

modules/solid_mechanics/doc/content/modules/solid_mechanics/ShellElements.md
```diff
@@ -63,19 +63,13 @@ g^2=\dfrac{g_3 \times g_1}{g_1 \cdot (g_2 \times g_3)}
 \begin{equation}
 g^3=\dfrac{g_1 \times g_2}{g_1 \cdot (g_2 \times g_3)}
 \end{equation}
-These contravariant base vectors satisfy the orthogonality condition: $g^i \cdot g_j =\delta_{ij}$, where $\delta_{ij}$ is Kronecker delta. The vectors $\hat{e}_i$ define the local Cartesian coordinate system on the shell element. By default, these local coordinates are computed using the following rule
+These contravariant base vectors satisfy the orthogonality condition: $g^i \cdot g_j =\delta_{ij}$, where $\delta_{ij}$ is Kronecker delta. The $\hat{e}_i$ vectors are the local cartesian coordinate system defined based on a user-defined reference first local direction vector $\hat{e}_{1r}$, which is defined by the `reference_first_local_direction` parameter, which defaults to the $X$ direction (i.e. $\hat{e}_{1r}=(1, 0, 0)$). The projection of $\hat{e}_{1r}$ onto the local element's plane provides the first local direction, so that the $\hat{e}$ vectors are computed as:
```
bwspenc, line 66: Cartesian should be capitalized.

bwspenc (commented):
Thanks for addressing my concern and sorting out those tests! I'm happy with this now. I made comments on a couple of really minor things I noticed. If we have to go through another iteration, please update these, but let's just fix them in the next round of changes if the tests pass. I invalidated a couple of the tests because they failed for unrelated reasons.

bwspenc (approved):

PR #21665, opened 2022-07-22
Create the moose-seacas package

cticenhour:
## Reason
[SEACAS](https://github.com/sandialabs/seacas) was historically a part of our [old MOOSE package](https://github.com/idaholab/package_builder/blob/master/build_from_source/template/seacas-tools). While it generally saw light use, some members of the application teams have requested that it be restored.

## Design
A basic SEACAS installation, using most of the defaults (minus GNU parallel ~~and X11 support~~), should be available as part of a conda package named `moose-seacas`. It should ideally be interoperable and able to sit side-by-side with `moose-mpich/petsc/libmesh`. While SEACAS does build it's own copy of a serial HDF5, containing it within `$CONDA_PREFIX/seacas` and adding it to the `PATH` in such a way as to prioritize the HDF5 installation provided by `moose-mpich` should be sufficient for most user cases.

## Impact
This will add a new package for users who need Exodus manipulation options.

PR #28747, opened 2024-10-01
Update seacas to 2024-08-15

cticenhour:
Remove patch for Intel Macs (not required since 2024-04-03 release)

refs #21665


milljm (approved):
👍

PR #28752, opened 2024-10-02
AnisotropicRankTwoTensor material

abhi1587:
It allows adding an anisotropic material property dependent on a field variable.

## Reason

The current application defines thermal conductivity tensors. It can also be used to define anisotropic material properties in solid mechanics as well.

## Design

A rank two tensor is created in which functions define the principal elements. The new function can also take basis values.

## Impact

It is a new addition.


dschwen:
We already have https://github.com/idaholab/moose/blob/next/framework/src/materials/GenericConstantRankTwoTensor.C which seems to fulfill the requirements.

dschwen:
We also have that in 

https://github.com/idaholab/moose/blob/next/modules/phase_field/src/materials/CompositeMobilityTensor.C

dschwen (commented):
Use https://github.com/idaholab/moose/blob/next/framework/include/materials/CompositeTensorBase.h

dschwen (changes requested):
Base this on https://github.com/idaholab/moose/blob/next/framework/include/materials/CompositeTensorBase.h

abhi1587:
Hi Daniel, I looked into your suggestion. Using the compositetensor class will not allow me to achieve the intended purpose. We want an object where each element of the property tensor can be controlled individually as a function of a field variable to get the desired anisotropy along the three axes. Additionally, the basis vector definition allows us to define the anisotropy in the cartesian for each axes of the property object.

hugary1995:
@abhi1587 I think we only intend to use **AD** for computing the property. @dschwen is suggesting you to utilize `CompositeTensorBase` to correctly compute the derivatives using chain rule relying on the `DerivativePropertyInterface` **without AD**. It's probably too challenging for you to generalize `CompositeTensorBase` to handle the case of AD.

I think there are other derivatives that cannot be computed even with the help of `DerivativePropertyInterface` unless we switch our entire app to using `DerivativePropertyInterface`. So, for the time being, let's (1) implement the object only for the AD case (2) rename it to `ADAnisotropicRankTwoTensor` (3) move the object to our app (4) close this PR.

PR #28772, opened 2024-10-03
Add support for non-archive library dependencies in install

### LINKED ISSUE #28771: Enable installs of just `.so` libraries with make install
### Motivation

The make install target currently only works on a list of library archives, and then determines the associated library (`.so` or `.dylib`) based on said library. This won't work when we only need to install `.so` or `.dylib` libraries.

### Design

Separate out logic of library vs library archive installs so that additional libraries that don't have archives can be supported.

### Impact

Expands install capability.

loganharbour:
Closes #28771


permcody (approved):
Looks good to me and since I watched you work on this, I know what it all does for now. Let's see what CIVET says!

PR #28795, opened 2024-10-07
Fix check for suffix when suffix doesn't exist

loganharbour:
refs #27091


GiudGiud (approved):

loganharbour:
Failure is unrelated.

PR #28804, opened 2024-10-08
Add ability to compute relative sensitivities directly in STM.

### LINKED ISSUE #28799: Add option to create relative sensitivities from DirectPerturbationSampler
### Motivation

Absolute sensitivities can be very confusing if the different parameters have different orders of magnitude.

### Design

Add an option to compute relative sensitivities.

### Impact

Better way to interpret the sensitivity coefficients.

grmnptr:
Closes #28799


loganharbour (approved):
Looks good.

The extra +1 index could be unclear in the future if you keep adding more data, but it's fine for now.

PR #28808, opened 2024-10-08
Testing new civet recipes

milljm:
Test if we can detect a failure to update packages after modifying an influential file being tracked by versioner.py.

Ref: #000

DO NOT MERGE


milljm:
This PR has served its purpose. Closing.

PR #28810, opened 2024-10-08
Fix bounds and initial conditions input parameter

### LINKED ISSUE #28748: Remove redundant input variable for setting up initial conditions and bounds in optimization problems
### Motivation

``ParameterMeshOptimization`` will read in input parameters ``initial_condition``, ``lower_bound``, and ``upper_bound`` and not do anything with the data it gets from these parameters.  This is because ``ParameterMeshOptimization`` derives off ``GeneralOptimization`` and ``GeneralOptimization`` uses these as input parameters.  However, these parameters are not used to set data in the constructor because they sometimes require data from a sub-app for their creation.  So these parameters are used to set data in function called ``setICsandBounds``.  ``ParameterMeshOptimization`` overrides ``setICsandBounds`` and sets values using from its own input parameters named ``constant_group_initial_condition``, ``constant_group_lower_bounds``, ``constant_group_upper_bounds``.    An error when ``initial_condition``, ``lower_bound``, and ``upper_bound`` are used in ``ParameterMeshOptimization`` would be good.  A better solution would be to use to get rid of ``constant_group_initial_condition``, ``constant_group_lower_bounds``, ``constant_group_upper_bounds`` from ``ParameterMeshOptimization`` and instead use the parameters set in ``GeneralOptimization`` to initialize data.

### Design

 A better solution would be to use to get rid of ``constant_group_initial_condition``, ``constant_group_lower_bounds``, ``constant_group_upper_bounds`` from ``ParameterMeshOptimization`` and instead use the parameters set in ``GeneralOptimization`` to initialize data.

### Impact

This will help users avoid a difficult to find input file error.

lynnmunday:
closes #28748

### DIFF
--- a/modules/optimization/doc/content/source/optimizationreporters/ParameterMeshOptimization.md
+++ b/modules/optimization/doc/content/source/optimizationreporters/ParameterMeshOptimization.md
@@ -11,19 +11,19 @@ This optimization reporter performs the same type of optimization as [Optimizati
 - finite-element order ([!param](/OptimizationReporter/ParameterMeshOptimization/parameter_orders)) where a single value input is applied to all parameters,
 - initial condition which can be set by the following inputs:
 
-  - from the input file using ([!param](/OptimizationReporter/ParameterMeshOptimization/constant_group_initial_condition)) where one value is given for each group of parameters
+  - from the input file using ([!param](/OptimizationReporter/OptimizationReporter/initial_condition)) where all values or one value per group can be specified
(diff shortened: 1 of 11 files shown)


lynnmunday:
@maxnezdyur will you review?

maxnezdyur (approved):
The PR looks good. I don't see much out of the ordinary in the refactor.

GiudGiud (approved):

PR #28837, opened 2024-10-11
Workaround gcc 13 -Wdangling-reference warning

### LINKED ISSUE #28835: -Wdangling-reference warning with newer GCC
### Bug Description

When building MOOSE with gcc 13, instantiations of `MaterialAuxBaseTempl` give me warnings such as:

```
/home/stogrh/git/moose-fresh/framework/build/header_symlinks/MaterialAuxBase.h: In instantiation of 'MaterialAuxBaseTempl<T, is_ad, is_functor, RT>::MaterialAuxBaseTempl(const InputParameters&) [with T = libMesh::DenseMatrix<double>; bool is_ad = false; bool is_functor = false; RT = double]':
/home/stogrh/git/moose-fresh/framework/src/auxkernels/MaterialRealDenseMatrixAux.C:28:42:   required from here
/home/stogrh/git/moose-fresh/framework/build/header_symlinks/MaterialAuxBase.h:47:24: warning: possibly dangling reference to a temporary [-Wdangling-reference]
   47 |   const PropertyType & _prop;
      |                        ^~~~~
/home/stogrh/git/moose-fresh/framework/build/header_symlinks/MaterialAuxBase.h:102:6: note: the temporary was destroyed at the end of the full expression '<lambda closure object>MaterialAuxBaseTempl<libMesh::DenseMatrix<double>, false, false, double>::MaterialAuxBaseTempl(const InputParameters&)::<lambda()>{((MaterialAuxBaseTempl<libMesh::DenseMatrix<double>, false, false, double>*)this)}.MaterialAuxBaseTempl<libMesh::DenseMatrix<double>, false, false, double>::MaterialAuxBaseTempl(const InputParameters&)::<lambda()>()'
   97 |     _prop([this]() -> const auto & {
      |           ~~~~~~~~~~~~~~~~~~~~~~~~~~
   98 |       if constexpr (is_functor)
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~
   99 |         return this->template getFunctor<Moose::GenericType<T, is_ad>>("functor");
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  100 |       else
      |       ~~~~
  101 |         return this->template getGenericMaterialProperty<T, is_ad>("property");
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  102 |     }()),
      |     ~^~
```

### Steps to Reproduce

Compile MOOSE with `g++ (Ubuntu 13.2.0-23ubuntu4) 13.2.0`, the default compiler on Ubuntu 24.04.1 LTS

### Impact

If there *isn't* a dangling reference bug here, this is just an annoyance, a false positive warning that would at worst encourage developers to ignore more serious warnings in new code.

I don't believe there is a dangling reference bug here - the code looks good, and recent GCC versions have had false positives emitted for that warning.  The false positives I've tested do seem to be fixed for my GCC build, but there are some very similar cases in bug reports.

If there is a dangling reference bug here, of course, that's the worst kind of bug; undefined behavior with "nasal demons" in the best case and silent corruption of data in the worst (which is also the typical) case.

I've got a workaround that I'll put up as a PR for consideration, but with *that* the impact is that we'd be replacing 5 lines of tight C++17-style code with 20 lines of C++98-style template metafunctions.

### [Optional] Diagnostics

_No response_

roystgnr:
Closes #28835

## Reason
Building MOOSE with gcc 13.2.0 gives me dangling reference warnings.

I don't see how this could be an actual dangling reference, and gcc has bug reports specifically related to dangling reference false positives from lambda functions, but the test case in bug report https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111607 doesn't actually trigger my gcc version.

## Design
Removing the nice beautiful modern-style `if constexpr` lambda function in a constructor initializer list and replacing it with an ugly 20th-century-style template metafunction silences the warning for me.

## Impact
This PR fixes a nasty false-positive compiler warning if I'm right, or fixes a serious bug if I'm wrong.  Either way it hurts the code aesthetics so I'd like to hope there's a better solution.

`git blame` seems to point to @dschwen for the code I'm removing here, so I'm pinging him in case he sees another solution or any non-aesthetic issues with my solution.

### DIFF
--- a/framework/include/auxkernels/MaterialAuxBase.h
+++ b/framework/include/auxkernels/MaterialAuxBase.h
@@ -13,6 +13,10 @@
 #include "AuxKernel.h"
 #include "Assembly.h"
 
+// Forward declarations
+template <typename T, bool is_ad, bool is_functor, typename RT>
+struct PropSetter;
+
 /**
  * A base class for the various Material related AuxKernal objects.
  * \p RT is short for return type
@@ -61,6 +65,9 @@ private:
 
   /// ID of the subdomain currently being iterated over
   const SubdomainID & _current_subdomain_id;
+
+  /// Hack to avoid "dangling reference" warnings from gcc 13
+  friend struct PropSetter<T, is_ad, is_functor, RT>;
 };
 
 template <typename T, bool is_ad, bool is_functor, typename RT>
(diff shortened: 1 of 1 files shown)


framework/include/auxkernels/MaterialAuxBase.h
```diff
@@ -90,16 +97,31 @@ MaterialAuxBaseTempl<T, is_ad, is_functor, RT>::validParams()
...
+  {
+    return mabt.template getGenericMaterialProperty<T, is_ad>("property");
+  }
+};
+
 template <typename T, bool is_ad, bool is_functor, typename RT>
 MaterialAuxBaseTempl<T, is_ad, is_functor, RT>::MaterialAuxBaseTempl(
     const InputParameters & parameters)
   : AuxKernelTempl<RT>(parameters),
-    _prop([this]() -> const auto & {
```
loganharbour, line 97: 🤮

dschwen:
My vote would be to locally silence this bogus warning.  The new code adds too much maintenance burden.

roystgnr:
> My vote would be to locally silence this bogus warning. The new code adds too much maintenance burden.

That's fair.  I'll put up a draft PR doing that.

framework/include/auxkernels/MaterialAuxBase.h
dschwen: This is beautiful!

loganharbour:
Should be resolved instead by https://github.com/idaholab/moose/pull/30856

Issue #28863, opened 2024-10-16
Use actual normals for 1D elements in THM


joshuahansel:
### Motivation

LibMesh is changing edge elements to use the actual 3D space normals, rather than always (1,0,0) or (-1,0,0): https://github.com/libMesh/libmesh/pull/3971. THM assumes the old calculation and needs to transition.

### Design

There are places where the old assumption is used as `_normals[_qp](0)`. To get the equivalent for the first task, try the trick `_current_side * 2 - 1.0` to get the equivalent -1 or +1.

### Impact

None; internal changes only.

PR #28864, opened 2024-10-16
Patched THM DG kernel for upcoming normal vector changes

joshuahansel:
Refs #28863

Motivating libMesh PR: https://github.com/libMesh/libmesh/pull/3971

### DIFF
--- a/modules/thermal_hydraulics/src/dgkernels/ADNumericalFlux3EqnDGKernel.C
+++ b/modules/thermal_hydraulics/src/dgkernels/ADNumericalFlux3EqnDGKernel.C
@@ -61,10 +61,12 @@ ADNumericalFlux3EqnDGKernel::computeQpResidual(Moose::DGResidualType type)
   std::vector<ADReal> U1 = {_rhoA1[_qp], _rhouA1[_qp], _rhoEA1[_qp], _A_elem[_qp]};
   std::vector<ADReal> U2 = {_rhoA2[_qp], _rhouA2[_qp], _rhoEA2[_qp], _A_neig[_qp]};
 
+  const Real nLR_dot_d = _current_side * 2 - 1.0;
+
   const std::vector<ADReal> & flux_elem =
-      _numerical_flux.getFlux(_current_side, _current_elem->id(), true, U1, U2, _normals[_qp](0));
+      _numerical_flux.getFlux(_current_side, _current_elem->id(), true, U1, U2, nLR_dot_d);
   const std::vector<ADReal> & flux_neig =
-      _numerical_flux.getFlux(_current_side, _current_elem->id(), false, U1, U2, _normals[_qp](0));
+      _numerical_flux.getFlux(_current_side, _current_elem->id(), false, U1, U2, nLR_dot_d);
 
   ADReal re = 0.0;
   switch (type)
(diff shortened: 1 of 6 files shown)


modules/thermal_hydraulics/src/dgkernels/ADNumericalFlux3EqnDGKernel.C
```diff
@@ -61,10 +61,14 @@ ADNumericalFlux3EqnDGKernel::computeQpResidual(Moose::DGResidualType type)
   std::vector<ADReal> U1 = {_rhoA1[_qp], _rhouA1[_qp], _rhoEA1[_qp], _A_elem[_qp]};
   std::vector<ADReal> U2 = {_rhoA2[_qp], _rhouA2[_qp], _rhoEA2[_qp], _A_neig[_qp]};
 
+  // Temporary hack to allow libMesh fix. After the fix, this should be changed to:
+  // const Real nLR_dot_d = _normals[_qp] * _direction[_qp];
```
roystgnr, line 65: What's `_direction` in this context?  Sorry for asking with such ignorance of the context here, I just want to make we don't end up with a 

If we just want a unit direction at each side, +1 or -1, then there's no point futzing with the normals at all.  We're simulating pipes here, right?  If a pipe bends at an angle between two elements then that shouldn't affect fluxes, just friction loss.

roystgnr:
This is resolving all the exodiffs for me, but I'm still seeing something worrying from `test:components/heat_transfer_from_heat_structure_1phase.phy.energy_heatstructure_ss_1phase` - libMesh throws an error from a `TypeVector<double>::unit()` call in `ADNumericalFlux3EqnBase::getFlux()` (as called from `ADFlowBoundaryFlux1Phase::computeQpIntegral()`, claiming that it's being asked to normalize a zero-length vector.  Looks like we've still got a raw use of `_normals[_qp]` in the `getFlux()` call there to fix?

joshuahansel:
I'm building locally with your libmesh to make sure that I get all of the normals fixed.

modules/thermal_hydraulics/src/dgkernels/ADNumericalFlux3EqnDGKernel.C
roystgnr: Hmm... it looks like it'll be equivalent for curved pipes approximated by straight Edge elements, but if you've actually got a curved Edge3 won't you end up getting dot product absolute values under 1.0?

(Unrelated: `dir.unit()` is a terser `dir / dir.norm()` that will also scream about division by 0 in `dbg` mode)

joshuahansel: Yeah you're right, it'd be wrong in that case. I don't know that we'll ever use curved Edge3, but I guess I should go ahead and use the safer calculation - so do you just recommend keeping `side * 2 - 1.0`?

roystgnr: `side*2-1.0` should work for everything *except* if you have a chain of edges in your mesh and one is flipped relative to the others.  Hopefully that hasn't happened before, because it would have broken the previous code too; the whole reason we're here is that I added a "test normals on flipped elements" unit test to libMesh and it caught us flipping normals on a flipped Edge.  But I'm not sure how to keep that problem from happening in the future, since here "flipped" depends not on anything geometric but solely on orientation relative to neighbors.

Actually, maybe we should just check that (in dbg+devel)?  On side N, check neighbor N, and if it's not null and not `remote_elem` then assert that it has a neighbor 1-N who is either us or one of our ancestors.  If it does then we're good, if it doesn't then that mesh wasn't THM compatible.

roystgnr (approved):

PR #28901, opened 2024-10-21
RunCommand job_slot control

### LINKED ISSUE #28900: Allow RunCommand to alert using more than one core/slot/job
### Motivation

When using RunCommand tester type, `max_parallel || min_parallel` is not an allowed parameter. For those using this tester type to run a parallel job of some kind, we need a way to consume _x_ job slots in the TestHarness Scheduler.

### Design

Add a param that will inform the TestHarness this job requires multiple cores.

### Impact

TestHarness will not error when one tries to set `min_parallel` (`max_parallel`, I see no obvious use here, as the TestHarness is designed to run `command = ` unmolested.

milljm:
Seems simple enough. We already have a `getSlots` method, should be able to use that.

Closes #28900

### DIFF
--- a/python/TestHarness/testers/RunCommand.py
+++ b/python/TestHarness/testers/RunCommand.py
@@ -16,11 +16,18 @@ class RunCommand(Tester):
         params = Tester.validParams()
         params.addRequiredParam('command',      "The command line to execute for this test.")
         params.addParam('test_name',          "The name of the test - populated automatically")
+        params.addParam('job_slots',    1, "Inform the TestHarness how many cores this job is"
+                                           " expected to consume")
         return params
 
     def __init__(self, name, params):
         Tester.__init__(self, name, params)
         self.command = params['command']
+
+    def getSlots(self, options) ->int:
+        """ return param job slots, default to 1 if 0 is provided """
+        # insure zero can't be used
+        return max(1, int(self.specs['job_slots']))
 
     def getCommand(self, options):
         # Create the command line string to run
--- a/python/TestHarness/tests/test_Allocations.py
+++ b/python/TestHarness/tests/test_Allocations.py
@@ -101,3 +101,24 @@ class TestHarnessTester(TestHarnessTestCase):
         self.assertNotIn('MAX_CPUS', output)
         self.assertNotIn('OVERSIZED', output)
         self.assertRegex(output, 'tests/test_harness.allocation_test.*OK')
+
+    def testRunCommand(self):
+        """
+        RunCommand [insufficient slots] SKIP / OK check
+        """
+        output = self.runTests('--no-color', '-i', 'allocation_runcommand', '-j 2').decode('utf-8')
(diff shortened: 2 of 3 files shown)


loganharbour:
Who needs this capability?

milljm:
> Who needs this capability?

@vanwdani 

https://github.com/idaholab/moose/issues/28900#issuecomment-2427314325

loganharbour:
I don't want to support this unless we cannot find an alternative. 

Manual commands without the test harness having context of what's being ran is very dangerous. Won't work with something like HPC execution. The harness should manage MPI execution, never a tester. We've shot ourselves in the foot supporting this.

milljm:
Agreed, but, RunCommand is there to stay I think. If I decided to run a test, which ran `conda index`, and I wanted to consume 4 cores doing so, shouldn't the TestHarness support that?

milljm:
I was curious where we might already be in violation, and I think so?

https://github.com/idaholab/moose/blob/f5a01db01841e690890a2a245f55ea13fefb199f/python/MooseDocs/test/tests

vanwdani:
The easy way to remove some of the cases where this is needed it for the TestHarness to be able to be dependent on tests which occur outside of the working directory.  Is this a possibility?

milljm:
> The easy way to remove some of the cases where this is needed it for the TestHarness to be able to be dependent on tests which occur outside of the working directory. Is this a possibility?

You mean?:
```pre
[test]
  pre_req = ../path/to/different/test
```
No, this is not allowed... The 'tester' is a DAG object, oblivious to other tester DAG objects...

vanwdani:
> Member

Right now this is basically a bash script that I'm running to see when jobs outside of the directory have run, then it allows a python plotting script to run.  We can make this just a Bison type of command, but depending on tests from outside a directory seems like something that could be useful for others.

milljm:
hmmm... well, I think it would be super cool for an entire DAG object to depend on another. That would be a neat feature I think.

vanwdani:
https://airflow.apache.org/docs/apache-airflow/1.10.3/_api/airflow/sensors/external_task_sensor/index.html ?

milljm:
> https://airflow.apache.org/docs/apache-airflow/1.10.3/_api/airflow/sensors/external_task_sensor/index.html ?

Oh I know it's possible, but probably will require a new param. Unless I get hacky somehow.

vanwdani:
Can at least Bison use the unique test ids for the daq ids?

milljm:
The problem, is that the Scheduler is shuffling around DAG objects with no recognition of whats inside them. We will have to create some key=value pair, and iterate over that

```python
testers = [('tests/simple_diffusion', <DAGobjectXabc123>), ]
```
where as now, its basically:
```python
testers = [<DAGobjectXabc123>, ]
```

milljm:
Closing in favor of https://github.com/idaholab/moose/issues/28947

PR #28923, opened 2024-10-23
Fix use of TAO solution vector

### LINKED ISSUE #28922: Fix how the optimization module receives parameters from TAO
### Bug Description

@lindsayad noticed that OptimizeSolve gives TAO PETSC ownership of a solution vector pointer that is then also being modified by OptimizeSolve.  So both the TAO optimization algorithms and OptimizeSolve are changing values in this vectors.  This behavior is wrong and causes the TAO optimization algorithms to be incorrectly applied.  This is also not allowed in a PETSC 3.22 update that uses vecloc on solution vectors.  

### Steps to Reproduce

This test now converges in one fewer iteration after fixing this: 
 modules/optimization/executioners/constrained/shape_optimization.equality_constraint

You will also note that the gradient is reduced at a much faster rate.

### Impact

This was causing linesearch to fail and resulted in more tests using a trust-region solver or a unit line-search.  Fixing htis will enable us to use line-search algorithms again.  Modifying the tao parameter vector probably has other unintended consequences I am unaware of.

### [Optional] Diagnostics

_No response_

lynnmunday:
closes #28922


lynnmunday:
@lindsayad Will you review this?  I documented the changes to the tests in the commit messages.

modules/optimization/src/executioners/OptimizeSolve.C
```diff
@@ -71,7 +71,6 @@ OptimizeSolve::solve()
 
   // Initialize solution and matrix
   _obj_function->setInitialCondition(*_parameters);
-  _obj_function->setInitialCondition(*_parameters);
```
lindsayad: nice catch, thanks!

lindsayad (approved):

PR #28989, opened 2024-11-01
Store tests by relative instead of absolute paths in the previous results file

### LINKED ISSUE #28988: Simplify previous test results output
### Motivation

The current previous test results output uses mostly absolute paths for defining each test. This is excessive and makes it difficult to move test data.

### Design

Store the test root in the previous results and then store the tests using keys that do not include the absolute path to the test directory.

### Impact

Will make it easier to move test results when reading from the previous results.

loganharbour:
Closes #28988

### DIFF
--- a/python/TestHarness/schedulers/Job.py
+++ b/python/TestHarness/schedulers/Job.py
@@ -764,9 +764,7 @@ class Job(OutputInterface):
         joint_status = self.getJointStatus()
 
         # Base job data
-        job_data = {'name'                 : self.getTestNameShort(),
-                    'long_name'            : self.getTestName(),
-                    'timing'               : self.timer.totalTimes(),
+        job_data = {'timing'               : self.timer.totalTimes(),
                     'status'               : joint_status.status,
                     'status_message'       : joint_status.message,
                     'fail'                 : self.isFail(),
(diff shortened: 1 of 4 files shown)


loganharbour:
This is ready. It makes it easier to copy content from HPC test results into an external directory (for the bison assessment, for example).

milljm:
I think you need to update the TestHarness hash. Unless we're ignoring this for the time being.

python/TestHarness/schedulers/Job.py
```diff
@@ -764,9 +764,7 @@ def storeResults(self, scheduler):
         joint_status = self.getJointStatus()
 
         # Base job data
-        job_data = {'name'                 : self.getTestNameShort(),
-                    'long_name'            : self.getTestName(),
```
milljm, line 768: I trust we are not using these anymore?

milljm (approved):

python/TestHarness/schedulers/Job.py
loganharbour: Those were actually added in my last PR: they're unused now because they're contained in the keys for the main map

loganharbour:
> I think you need to update the TestHarness hash. Unless we're ignoring this for the time being.

It's already updated, hence why https://civet.inl.gov/job/2526480/ ran

PR #28997, opened 2024-11-04
Use the app name from the binary when running installed tests

### LINKED ISSUE #28996: Use app name from running binary when running installed tests
### Motivation

The installed `moose_test_runner` uses the `app_name` from the `testroot`. We shouldn't be parsing this app name because the app name is already known. It's the run that we ran:

```
app-opt --run ...
```

with (`app` in this case).

### Design

Use the binary name to produce `app_name` for the `TestHarness` when running installed tests.

### Impact

Greater flexibility when running installed app tests.

loganharbour:
Closes #28996

### DIFF
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -1716,6 +1716,9 @@ MooseApp::runInputs()
           " --run <dir>\" again.");
     }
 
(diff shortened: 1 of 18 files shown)


permcody (approved):

milljm (approved):
looks good to me!

PR #29027, opened 2024-11-07
Ensure postprocessors are computed during linear solve with PJFNKMO

### LINKED ISSUE #28993: Need to make Problem/bx_norm work with PJFNKMO eigenvalue calculation
### Motivation

When adding `Problem/bx_norm` parameter with `Executioner/solve_type=PJFNKMO` with `Eigenvalue` executioner, calculations will produce wrong results.

### Design

Users would expect `Problem/bx_norm` works regardless the solve_type. This turns out to be passing the unnecessary normalization to SLEPc in case of PJFNKMO. A test will be added after the code fix.

### Impact

A combination of input parameters that possibly has not be used by users so far.

lindsayad:
Closes #28993

### DIFF
--- a/framework/src/utils/SlepcSupport.C
+++ b/framework/src/utils/SlepcSupport.C
@@ -951,6 +951,13 @@ mooseSlepcEigenFormFunctionB(SNES snes, Vec x, Vec r, void * ctx)
 
     ierr = MatMult(B, x, r);
     CHKERRQ(ierr);
+
+    if (eigen_problem->bxNormProvided())
+    {
+      // User has provided a postprocessor. We need it updated
+      updateCurrentLocalSolution(eigen_nl.sys(), x);
+      eigen_problem->execute(EXEC_LINEAR);
+    }
   }
   else
     moosePetscSNESFormFunction(snes, x, r, ctx, eigen_nl.eigenVectorTag());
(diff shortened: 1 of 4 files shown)


test/tests/problems/eigen_problem/jfnk_mo/tests
```diff
@@ -57,6 +57,7 @@
     type = 'CSVDiff'
     input = 'ne_coupled_mo.i'
     csvdiff = 'ne_coupled_eigenvalues_0001.csv'
+    cli_args = 'Problem/bx_norm=uint'
```
YaqiWang, line 60: I feel this deserve a separate test with its own requirement.

framework/src/utils/SlepcSupport.C
```diff
@@ -951,6 +951,13 @@ mooseSlepcEigenFormFunctionB(SNES snes, Vec x, Vec r, void * ctx)
 
     ierr = MatMult(B, x, r);
     CHKERRQ(ierr);
+
+    if (eigen_problem->bxNormProvided())
+    {
+      // User has provided a postprocessor. We need it updated
+      updateCurrentLocalSolution(eigen_nl.sys(), x);
+      eigen_problem->execute(EXEC_LINEAR);
```
YaqiWang, line 959: I realized that we do not have a check to make sure `execute_on` of the bx_norm postprocessor includes linear. Care to add it in `EigenProblem`?

YaqiWang (changes requested):
The code change looks good. I checked with a Griffin test and it does work as expected! Just need to address my two comments and this should be good to go.

lindsayad:
Nice suggestions. Hopefully should be addressed

GiudGiud (approved):

YaqiWang:
My comments are addressed. I cannot resolve the comments by myself.

lindsayad:
thanks for raising the issue and the review @YaqiWang !

PR #29028, opened 2024-11-07
Restricted convergence test to 4 processes

joshuahansel:
This fixes the sweep testing failures caused by https://github.com/idaholab/moose/pull/28848.

### DIFF
--- a/test/tests/convergence/default_nonlinear_convergence/default_nonlinear_convergence.i
+++ b/test/tests/convergence/default_nonlinear_convergence/default_nonlinear_convergence.i
@@ -35,9 +35,8 @@
 [Convergence]
   [res_conv]
     type = DefaultNonlinearConvergence
-    # With default tolerances, the solve takes 2 iterations, but with this
-    # value, the solve takes only 1:
-    nl_abs_tol = 1e-5
+    # With this, convergence should occur immediately (zero iterations)
+    nl_abs_tol = 1e15
   []
 []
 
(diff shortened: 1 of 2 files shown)


joshuahansel:
Enabled parallel sweep testing on this PR

lindsayad (changes requested):
I should have caught this on review. Is there any way to test this other than iteration counting? The fact that this fails in parallel means it's likely not future proof against slight changes in PETSc implementation, etc.

joshuahansel:
I think that iteration counting is the right way to test a convergence object because this is ultimately the impact a convergence has. I don't see that not testing at high process counts will be a detriment. At low counts, I think it should be stable enough.

lindsayad (approved):

loganharbour:
Looks like this is now failing SQA, not sure how it didn't fail before

```

ERRORS:
/Convergence/AugmentedLagrangianContactFEProblemConvergence is missing a markdown file.
  The page should be located at modules/contact/doc/content/source/convergence/AugmentedLagrangianContactConvergence.md
  A stub page can be created using using the './moosedocs.py generate' command
Control this message using: 'log_missing_markdown'

/Convergence/AugmentedLagrangianContactReferenceConvergence is missing a markdown file.
  The page should be located at modules/contact/doc/content/source/convergence/AugmentedLagrangianContactConvergence.md
  A stub page can be created using using the './moosedocs.py generate' command
Control this message using: 'log_missing_markdown' 
```

https://civet.inl.gov/job/2540128/

joshuahansel:
Ok I'll fix now. And then try to find why this didn't show up before.

joshuahansel:
Fix: https://github.com/idaholab/moose/pull/29046

PR #29049, opened 2024-11-11
Added verbose output option for Convergence objects

### LINKED ISSUE #29048: Add verbose output for Convergence objects
### Motivation

This will allow printing out helpful information when using Convergence objects.

### Design

This would take the verbose flag from the executioner for Convergence objects and use that flag in a conditional to check information on convergence/divergence.

### Impact

New console output if `verbose` set to `true` in Executioner.

joshuahansel:
Closes #29048


GiudGiud (approved):
once testing is fixed

PR #29068, opened 2024-11-13
Added diagnostic check for watertight nodesets

### LINKED ISSUE #28822: Add diagnostic check for water-tight nodesets
### Motivation

Adding more features to the mesh diagnostic system will help users debug issues with their meshes.  

Linked to #25278 and #28816 

### Design

Should be added to MeshDiagnostics Generator. There is a diagnostic feature in progress that would detect when a mesh has external sides without a sideset. This would be similar, but instead look for missing nodesets instead of sidesets.

### Impact

Makes it easier to find issues with user meshes.

DanielYankura:
-Loops through all elements and sides
-If side is external it checks if all its nodes are assigned to a nodeset

closes #28822

## Reason

Adds new diagnostic tool for users

## Design

If feature it turned on, it loops through each element, then the element's sides. If the sides are external it checks each node to ensure it's been assigned to a nodeset. If not, the node id is printed for reference.

## Impact

Adds watertight nodeset check to Mesh Diagnostics

### DIFF
--- a/framework/src/meshgenerators/MeshDiagnosticsGenerator.C
+++ b/framework/src/meshgenerators/MeshDiagnosticsGenerator.C
@@ -46,6 +46,10 @@ MeshDiagnosticsGenerator::validParams()
       chk_option,
       "whether to check for external sides that are not assigned to any sidesets");
   params.addParam<MooseEnum>(
+      "check_for_watertight_nodesets",
+      chk_option,
+      "whether to check for external nodes that are not assigned to any nodeset");
+  params.addParam<MooseEnum>(
       "examine_element_volumes", chk_option, "whether to examine volume of the elements");
   params.addParam<Real>("minimum_element_volumes", 1e-16, "minimum size for element volume");
   params.addParam<Real>("maximum_element_volumes", 1e16, "Maximum size for element volume");
@@ -80,6 +84,7 @@ MeshDiagnosticsGenerator::MeshDiagnosticsGenerator(const InputParameters & param
     _input(getMesh("input")),
     _check_sidesets_orientation(getParam<MooseEnum>("examine_sidesets_orientation")),
     _check_watertight_sidesets(getParam<MooseEnum>("check_for_watertight_sidesets")),
+    _check_watertight_nodesets(getParam<MooseEnum>("check_for_watertight_nodesets")),
     _check_element_volumes(getParam<MooseEnum>("examine_element_volumes")),
     _min_volume(getParam<Real>("minimum_element_volumes")),
     _max_volume(getParam<Real>("maximum_element_volumes")),
@@ -103,10 +108,10 @@ MeshDiagnosticsGenerator::MeshDiagnosticsGenerator(const InputParameters & param
     paramError("examine_non_conformality",
                "You must set this parameter to true to trigger mesh conformality check");
   if (_check_sidesets_orientation == "NO_CHECK" && _check_watertight_sidesets == "NO_CHECK" &&
-      _check_element_volumes == "NO_CHECK" && _check_element_types == "NO_CHECK" &&
-      _check_element_overlap == "NO_CHECK" && _check_non_planar_sides == "NO_CHECK" &&
-      _check_non_conformal_mesh == "NO_CHECK" && _check_adaptivity_non_conformality == "NO_CHECK" &&
-      _check_local_jacobian == "NO_CHECK")
+      _check_watertight_nodesets == "NO_CHECK" && _check_element_volumes == "NO_CHECK" && 
+      _check_element_types == "NO_CHECK" && _check_element_overlap == "NO_CHECK" && 
+      _check_non_planar_sides == "NO_CHECK" && _check_non_conformal_mesh == "NO_CHECK" 
+      && _check_adaptivity_non_conformality == "NO_CHECK" && _check_local_jacobian == "NO_CHECK")
     mooseError("You need to turn on at least one diagnostic. Did you misspell a parameter?");
 }
 
@@ -128,6 +133,9 @@ MeshDiagnosticsGenerator::generate()
 
   if (_check_watertight_sidesets != "NO_CHECK")
     checkWaterTightSidesets(mesh);
+
+  if (_check_watertight_nodesets != "NO_CHECK")
+    checkWatertightNodesets(mesh);
 
   if (_check_element_volumes != "NO_CHECK")
     checkElementVolumes(mesh);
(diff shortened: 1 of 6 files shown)


framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -45,6 +45,10 @@ MeshDiagnosticsGenerator::validParams()
       "check_for_watertight_sidesets",
       chk_option,
       "whether to check for external sides that are not assigned to any sidesets");
+  params.addParam<MooseEnum>(
+      "check_for_watertight_nodesets",
+      chk_option,
+      "whether to check for external nodes that are not assigned to any nodeset");
   params.addParam<MooseEnum>(
```
GiudGiud, line 52: let s add a parameter for the nodesets that are supposed to englobe the entire mesh. when this parameter is in use, the nodes are only checked against those "blessed" nodesets

same for sidsets
actually now that I see that this is not an option for sidesets, let s keep it for future work

DanielYankura: Sounds good. I'll open a new issue for adding that parameter to the sideset and nodeset checks.

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -335,6 +343,65 @@ MeshDiagnosticsGenerator::checkWaterTightSidesets(const std::unique_ptr<MeshBase
...
+  3) Loop through all sides of that element
+  4) If side is external loop through its nodes
+  5) If node is not associated with any nodeset add to list
+  6) Print out node id
+  */
+  if (mesh->mesh_dimension() < 2)
+    mooseError("The nodeset check only works for 2D and 3D meshes");
+  auto & boundary_info = mesh->get_boundary_info();
+  boundary_info.build_side_list();
+  const auto nodeset_map = boundary_info.get_nodeset_map();
```
GiudGiud, line 362: where is the _boundary_node_id being initialized in libmesh? is the previous call doing it?

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -335,6 +343,65 @@ MeshDiagnosticsGenerator::checkWaterTightSidesets(const std::unique_ptr<MeshBase
...
+  unsigned int num_nodes_without_nodeset = 0;
+  std::set<Node> checked_nodes;
+
+  for (const auto elem : mesh->active_element_ptr_range())
+  {
+    for (auto i : elem->side_index_range())
```
GiudGiud, line 368:
```suggestion
    for (const auto i : elem->side_index_range())
```

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -335,6 +343,65 @@ MeshDiagnosticsGenerator::checkWaterTightSidesets(const std::unique_ptr<MeshBase
...
+    {
+      // Check if side is external
+      if (elem->neighbor_ptr(i) == nullptr)
+      {
+        // Side is externa, now check nodes
```
GiudGiud, line 373:
```suggestion
        // Side is external, now check nodes
```

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -335,6 +343,65 @@ MeshDiagnosticsGenerator::checkWaterTightSidesets(const std::unique_ptr<MeshBase
...
+          if (nodeset_map.count(node) == 0)
+          {
+            // This node does not have a nodeset!!!
+            num_nodes_without_nodeset++;
+            checked_nodes.insert(*node);
+            std::string message;
+            if (num_nodes_without_nodeset < _num_outputs)
+            {
+              message = "Node " + std::to_string(node->id()) +
+                        " is external, but has not been assigned to a nodeset";
```
GiudGiud, line 391:
```suggestion
                        " is on an external boundary of the mesh, but has not been assigned to a nodeset";
```

test/tests/meshgenerators/mesh_diagnostics_generator/tests
```diff
@@ -45,6 +45,22 @@
       expect_out = 'Number of external element faces that have not been assigned to a sideset: 4'
       detail = '3D meshed cube missing a sideset on one side'
     []
+    [2D_watertight_nodesets]
+      type = 'RunApp'
+      input = '2D_missing_nodeset.i'
+      cli_args = '--mesh-only'
+      mesh_mode = 'replicated'
+      expect_out = 'Number of external nodes that have not been assigned to a nodeset: 1'
+      detail = '2D meshed square missing a nodeset on one side'
```
GiudGiud, line 54:
```suggestion
      detail = '2D meshed square with one external node that does not belong to a nodeset'
```

test/tests/meshgenerators/mesh_diagnostics_generator/tests
```diff
@@ -45,6 +45,22 @@
...
+    []
+    [3D_watertight_nodesets]
+      type = 'RunApp'
+      input = '3D_missing_nodeset.i'
+      cli_args = '--mesh-only'
+      mesh_mode = 'replicated'
+      expect_out = 'Number of external nodes that have not been assigned to a nodeset: 1'
+      detail = '3D meshed cube missing a nodeset on one side'
```
GiudGiud, line 62: same

GiudGiud (commented):
looks good

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
DanielYankura: So I looked into it. The node ids get added to _boundary_node_id after execute_mesh_generators is called in ActionWarehouse.C. From what I can tell the mesh is always initialized before this test could ever be called.

GiudGiud: >  after execute_mesh_generators

wouldnt that be too late then?

DanielYankura: Sorry, it's the execute_mesh_generators task that initializes _boundary_node_id. After initialization mesh diagnostics runs. So that task runs both I believe, but it should initialize first.

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -335,6 +343,66 @@ MeshDiagnosticsGenerator::checkWaterTightSidesets(const std::unique_ptr<MeshBase
...
+        {
+          const auto node = node_list[j];
+          if (checked_nodes.count(*node))
+            continue;
+          // if node is not part of nodeset map add it to list of bad nodes
+          if (nodeset_map.count(node) == 0)
+          {
+            // This node does not have a nodeset!!!
+            num_nodes_without_nodeset++;
+            checked_nodes.insert(*node);
```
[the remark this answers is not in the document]
DanielYankura: Thanks! I was trying to figure out why it was failing that test. I think I have the same issue in #28701 so I'll fix it in that one as well.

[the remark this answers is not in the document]
GiudGiud: I didnt remember that one was still open. Do you need anything else from me for that? Are we waiting on Roy's review?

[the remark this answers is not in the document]
DanielYankura: I think I'm just waiting on Roy's review. Though if this one get's merged first I might have to make some changes to the other before it can be merged.

GiudGiud (approved):

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -335,6 +343,66 @@ MeshDiagnosticsGenerator::checkWaterTightSidesets(const std::unique_ptr<MeshBase
...
+  2) Loop through all elements
+  3) Loop through all sides of that element
+  4) If side is external loop through its nodes
+  5) If node is not associated with any nodeset add to list
+  6) Print out node id
+  */
+  if (mesh->mesh_dimension() < 2)
+    mooseError("The nodeset check only works for 2D and 3D meshes");
+  auto & boundary_info = mesh->get_boundary_info();
+  boundary_info.build_side_list();
```
roystgnr, line 361: What's going on here?  This is the `build_side_list()` variant that returns a vector of tuples, but if nothing uses the return value then it's just an expensive no-op.

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -335,6 +343,66 @@ MeshDiagnosticsGenerator::checkWaterTightSidesets(const std::unique_ptr<MeshBase
...
+        // Side is external, now check nodes
+        auto side = elem->side_ptr(i);
+        const auto & node_list = side->get_nodes();
+        for (unsigned int j = 0; j < side->n_nodes(); j++)
+        {
+          const auto node = node_list[j];
+          if (checked_nodes_id.count(node->id()))
+            continue;
+          // if node is not part of nodeset map add it to list of bad nodes
+          if (nodeset_map.count(node) == 0)
```
roystgnr, line 382: Could we just test `boundary_info.n_boundary_ids(node)` here?  The raw `get_nodeset_map()` accessor was an encapsulation mistake I should never have approved, and I still vaguely dream of deprecating it someday.

roystgnr (changes requested):
In principle, I oppose the very concept of checking for watertight nodesets, and would be happy to rant at length about why.  (nobody else would be happy, so I won't)

In practice, I recognize that some MOOSE workflows will be using nodesets as pseudo-sidesets until the end of time; this is a great way for users of those workflows to check for meshing problems.  I just have one minor problem and one minor request here.

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
DanielYankura: I think I copied that over from a similar function I made previously and then forgot to remove it. I'l fix that real quick.

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
DanielYankura: Just want to check, would that return true if the node is assigned to a sideset but not a nodeset?

roystgnr (approved):

Issue #27553, opened 2024-05-06
Improve modular NEML2 interface


dschwen:
### Motivation

A use case for nodal NEML2 model execution has popped up, and the execution during the `INITIAL` execution state needs to be enabled. Investigate micro optimizations for tensor conversion and contiguous storage for fast memcopy transfers back to MOOSE. Add some more error checking.

### Design

Template `ExecuteNEML2Model` and add a nodal version. Figure out why the `INITIAL` stage is not properly executing. Check the compatibility of block restrictions of the involved user objects.

### Impact

Added capability

PR #28597, opened 2024-09-11
NEML2Action overhaul

hugary1995:
- Factor out a common action for NEML2 for block-restriction
- Remove legacy objects
- Unify gatherers for variable and parameter
- Migrate all input files to use NEML2Action
- Remove legacy tests

close #27553

PR #29071, opened 2024-11-13
Use `from_blob` based transfer to NEML2

dschwen:
Refs #27553

### DIFF
--- a/modules/solid_mechanics/src/neml2/userobjects/MOOSEMaterialPropertyToNEML2.C
+++ b/modules/solid_mechanics/src/neml2/userobjects/MOOSEMaterialPropertyToNEML2.C
@@ -55,7 +55,8 @@ MOOSEMaterialPropertyToNEML2<T, state>::validParams()
 template <typename T, unsigned int state>
 MOOSEMaterialPropertyToNEML2<T, state>::MOOSEMaterialPropertyToNEML2(const InputParameters & params)
   : MOOSEToNEML2(params),
-    _mat_prop(getGenericMaterialProperty<T, false>("moose_material_property", state))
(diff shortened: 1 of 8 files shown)


modules/solid_mechanics/src/neml2/userobjects/MOOSEMaterialPropertyToNEML2.C
```diff
@@ -74,10 +75,18 @@ MOOSEMaterialPropertyToNEML2<T, state>::MOOSEMaterialPropertyToNEML2(const Input
 }
 
 template <typename T, unsigned int state>
-torch::Tensor
-MOOSEMaterialPropertyToNEML2<T, state>::convertQpMOOSEData() const
+void
+MOOSEMaterialPropertyToNEML2<T, state>::insertIntoInput(neml2::LabeledVector & input) const
 {
-  return NEML2Utils::toNEML2<T>(_mat_prop[_qp]);
+  input.base_index_put_(_neml2_variable, neml2::Tensor(torch::from_blob(const_cast<double *>(_buffer.data()), {_buffer.size() / _item_size,_item_size}), 1));
```
dschwen, line 81: This needs to permit an arbitrary number of tensor dimensions.

hugary1995:
`from_blob` is definitely the way to go.

In addition to the current conflicts, it'd be ideal to wait for https://github.com/applied-material-modeling/neml2/pull/177, as it will save some tensor reshaping operations on the moose side.

dschwen:
Rebasing this is not feasible anymore and it is probably superseeded by #28597.

dschwen:
I'll wait for the NEML2 2.0 update and might revisit `from_blob`. It turned out to have a modest impact on speed. However, the relative speedup might be more significant given the expected NEML2 performance increases related to 2.0.

PR #29104, opened 2024-11-19
Use new conda system, allow user overrides, update apptainer base images

milljm:
Fix apptainer's user override issue...

Need to source user overrides before sourcing Conda environment.


milljm:
Before this is merged, please alert me. As there are Civet Recipes that need to go in before that happens.

apptainer/moose-dev.def
```diff
@@ -46,7 +43,7 @@ From: {{ APPTAINER_FROM }}
 
 %environment
     export PATH={{ MOOSE_MINIFORGE }}/bin:/opt/code-server/bin:$PATH
-    source activate {{ MOOSE_MINIFORGE }}/envs/{{ MOOSE_CONDA_ENV_NAME }}
+    source activate {{ MOOSE_MINIFORGE }}
```
milljm, line 46: @loganharbour This is my proposed fix. e.g. Still need to activate the base environment.

loganharbour (approved):

PR #29120, opened 2024-11-22
Add a flag to allow derived action to suppress a unnecessary warning

### LINKED ISSUE #29119: Add a flag to let MaterialOutputAction not output unsupported property names
### Motivation

`MaterialOutputAction` provides a convenient way of visualize material properties. However, it does not and cannot support all types of material properties. In such a case, it outputs the list of unsupported properties on screen. In Griffin, we have an action derived from `MaterialOutputAction` that uses the machinery in it to output more material properties ONLY unsupported by it. The derived action is not meant to replace `MaterialOutputAction` and not supposed to output all properties. It should not issue the warning on unsupported properties. This is possibly the fix for a failing BludCRAB test preventing its Griffin submodule update.

### Design

If we can add a flag in `MaterialOutputAction` and guard the warning with the flag, the derived class can reset that flag to false to suppress the warning output.

### Impact

Should not have impact on any calculations.

YaqiWang:
Closes #29119.

### DIFF
--- a/framework/src/actions/MaterialOutputAction.C
+++ b/framework/src/actions/MaterialOutputAction.C
@@ -31,6 +31,10 @@ InputParameters
 MaterialOutputAction::validParams()
 {
   InputParameters params = Action::validParams();
+  /// A flag to tell this action whether or not to print the unsupported properties
+  /// Note: A derived class can set this to false, override materialOutput and output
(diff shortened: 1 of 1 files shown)


GiudGiud (approved):

YaqiWang:
The failing test does not seem related. Should we invalidate it?

PR #29143, opened 2024-11-26
Register data per application and allow arbitrary registration

### LINKED ISSUE #29141: Enable an application to retrieve its data path
### Motivation

We need to be able to obtain the specific installed and in-tree data path for Griffin content.

### Design

Store the data paths per-application and enable them to be queried as such.

### Impact

New capability.

loganharbour:
Closes #29141

Summary:

- Enables the association of data file directories with a name, so that you can (1) query the directories by name, (2) resolve ambiguities when data is found in multiple places, (3) output this information in a readable way
- Removes the need to call `DataFileInterface::getDataFileName("param")` and instead sets the param value internally so that `getParam<DataFileName>("param")` gets you the absolute path to whatever we searched

You can directly query an application's registered data path with:

```
Registry::getDataPath("griffin");
```

In the event that you have ambiguous data, you'll get an error like the following (where you can provide `name:path/to/file` to be explicit)

```
*** ERROR ***
CLI_ARGS:1: (UserObjects/data_file/data_file) Multiple files were found when searching for the data file 'file':

  moose: /path/to/data/file
  other: /path/to/data/file

You can resolve this ambiguity by appending a prefix with the desired data name, for example:

  moose:file
```

In the event that you're searching from a parameter (you have a base to search from) and nothing is found, you get a useful printout of everything searched, like:

```
*** ERROR ***
CLI_ARGS:1: (UserObjects/data_file/data_file) Unable to find the data file 'file' anywhere.

Paths searched:

  moose data: /path/to/data
  working directory: /path/to/input/base
```

The command line options `--show-data-params` and `--show-data-paths` were added so that you can see this info in the header (because we no longer call `mooseInfo()` every time we resolve params because it would be too verbose):

```
Data File Paths:
  moose: /data/harblh/projects/griffin/moose/framework/data

Data File Parameters:
  UserObjects/data_file/data_file = /data/harblh/projects/griffin/moose/framework/data/README.md
```

If the search path starts with `./` it will not search data directories, and you'll get information that this is the case:

```
CLI_ARGS:1: (UserObjects/data_file/data_file) Unable to find the data file './file' anywhere.

Paths searched:

  working directory: /path/to/input/base

Data path(s) were not searched because search path begins with './'
```

and the same for when the search path resolves outside of `.`:

```
CLI_ARGS:1: (UserObjects/data_file/data_file) Unable to find the data file '../file' anywhere.

Paths searched:

  working directory: /path/to/input/base

Data path(s) were not searched because search path resolves behind '.'
```

You can interact directly with the registry to get files (like needed for the NEML case) with:

```
Moose::DataFileUtils::getPathExplicit("bison", "path/to/data");
```


dschwen (dismissed):

loganharbour:
@dschwen this is ready for review.

dschwen (approved):

PR #29144, opened 2024-11-26
Split lower-d mesh blocks based on side types

### LINKED ISSUE #29142: MooseMesh::buildLowerDMesh needs to add separate blocks for lower-d elements in different types
### Motivation

This is an issue often encountered in 3D meshes. Currently, the function adds two lower-d blocks, one for all interior sides and another for all boundary sides. In 3D, the sides could be in different element types, like either triangle, quad. This will cause an issue when writing the mesh into an Exodus mesh, where a mesh block must have the same element type.

### Design

Even though calculations does not care lower-d element types, it could be a good idea to separate the mesh block into blocks based on the element types. We will need to add getters in `MooseMesh` to get a vector of interior side blocks and boundary side blocks, so that applications do not use `Moose::INTERNAL_SIDE_LOWERD_ID` or `INTERNAL_SIDE_LOWERD_SUBDOMAIN` directly. We need to add a test that contains elements like prism which have sides in different types.

### Impact

Make calculations with lower-d mesh blocks able to perform Exodus outputs.

YaqiWang:
Closes #29142. Without this fix, the new test `3d-prism.i` will error out in `--mesh-only` mode.


YaqiWang:
This will need a Griffin patch. Tag @NamjaeChoi

GiudGiud (commented):
Can we use contains instead of count and empty() instead of size() ?
Using a fixed subdomain Id was likely an optimization so we should try to stay optimized

framework/include/loops/ThreadedElementLoopBase.h
```diff
@@ -255,8 +255,8 @@ ThreadedElementLoopBase<RangeType>::operator()(const RangeType & range, bool byp
 
         onElement(elem);
 
-        if (elem->subdomain_id() == Moose::INTERNAL_SIDE_LOWERD_ID ||
-            elem->subdomain_id() == Moose::BOUNDARY_SIDE_LOWERD_ID)
+        if (_mesh.interiorLowerDBlocks().count(elem->subdomain_id()) > 0 ||
+            _mesh.boundaryLowerDBlocks().count(elem->subdomain_id()) > 0)
```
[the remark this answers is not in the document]
YaqiWang: Yes, looks like it is C++20 https://en.cppreference.com/w/cpp/container/set/contains. Do you want me make this change?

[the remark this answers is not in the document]
GiudGiud: you cant if it s C++20

framework/src/bcs/LowerDIntegratedBC.C
```diff
@@ -34,15 +34,15 @@ LowerDIntegratedBC::LowerDIntegratedBC(const InputParameters & parameters)
...
-  if (!lower_domains.count(Moose::BOUNDARY_SIDE_LOWERD_ID) && lower_domains.size() != 1)
-    paramError(
-        "lowerd_variable",
-        "Must be only defined on the subdomain BOUNDARY_SIDE_LOWERD_SUBDOMAIN subdomain that is "
-        "added by Mesh/build_all_side_lowerd_mesh=true");
-
-  if (_var.activeSubdomains().count(Moose::BOUNDARY_SIDE_LOWERD_ID))
-    paramError("variable",
-               "Must not be defined on the subdomain INTERNAL_SIDE_LOWERD_SUBDOMAIN subdomain");
+  for (const auto & id : _mesh.boundaryLowerDBlocks())
```
GiudGiud, line 37: this isnt quite the error we had before

we are basically forbidding the block restriction (on lower D blocks) of the variable. it must be defined on all lower D blocks

this should be fine for now

YaqiWang: Right. It is the current behavior. We can make it fancier in the future easily if someone requests.

GiudGiud: it is only the current behavior over our current set of tests. If I block restrict this variable to one of 2 lower D blocks , I will hit the error when I should not

framework/src/mesh/MooseMesh.C
```diff
@@ -632,19 +635,69 @@ MooseMesh::buildLowerDMesh()
...
       deleteable_elems.insert(elem);
     else if (elem->n_sides() > max_n_sides)
       max_n_sides = elem->n_sides();
 
   for (auto & elem : deleteable_elems)
     mesh.delete_elem(elem);
+  for (const auto & id : _lower_d_interior_blocks)
+    _mesh_subdomains.erase(id);
+  for (const auto & id : _lower_d_boundary_blocks)
+    _mesh_subdomains.erase(id);
```
GiudGiud, line 649: why do that?

framework/src/mesh/MooseMesh.C
```diff
@@ -632,19 +635,69 @@ MooseMesh::buildLowerDMesh()
...
+  // we assume this id is not used by the mesh
+  auto id = libMesh::Elem::invalid_subdomain_id - 2;
+  for (const auto & tpid : interior_side_types)
+  {
+    const auto type = ElemType(tpid);
+    mesh.subdomain_name(id) = "INTERNAL_SIDE_LOWERD_SUBDOMAIN_" + Utility::enum_to_string(type);
+    interior_block_ids[type] = id;
+    _lower_d_interior_blocks.insert(id);
+    if (_mesh_subdomains.count(id) > 0)
+      mooseError("Trying to add a mesh block with id ", id, " that has existed in the mesh");
```
GiudGiud, line 685: you wont hit that if you deleted it earlier though?

test/tests/kernels/hfem/3d-prism.i
```diff
@@ -0,0 +1,119 @@
+[Mesh]
+  [square]
+    type = GeneratedMeshGenerator
+    nx = 3
+    ny = 3
+    nz = 3
```
GiudGiud, line 6:
```suggestion
    nx = 2
    ny = 2
    nz = 2
```
we only really need an interior face, no need to overdo it

framework/src/mesh/MooseMesh.C
YaqiWang: No. Adding this line is to make the code more defensive.

framework/src/bcs/LowerDIntegratedBC.C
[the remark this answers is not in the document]
YaqiWang: The logic has to be this way. The variable on lower-d mesh actually can be defined both on the boundary and interior. In such a case, `lower_domains` will contain a block id that not in `_mesh.boundaryLowerDBlocks()`. The check here is to make sure the lower-d variable is valid on all boundary lower-d mesh blocks.

framework/src/mesh/MooseMesh.C
YaqiWang: This is for reentering this function safely. First call does not need this, but if we recalling this function without this, we will hit the error checks. This part of code is essentially performing clearance of what has been done by this function.

framework/src/bcs/LowerDIntegratedBC.C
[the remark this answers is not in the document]
GiudGiud: Ok let’s check with the two sets then

there is no reason to force the variable to be defined on all lowerD blocks

the intent of this check is to prevent the variable from being defined on a higher D block. No need to change that.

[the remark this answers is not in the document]
YaqiWang: >let’s check with the two sets then

What do you mean? Checking this lower-d variable is defined on all boundary lower-d blocks is too much but that is consistent with the current code, because the lower-d block is covering the entire boundary sides.

More rigorous check should be making sure the boundary sides in `boundary` parameter of this BC are covered by the variable on the corresponding lower-d block, but we currently do not have a clean way to do that check.

Theoretically speaking, this variable can be defined on a higher-d block at the same time.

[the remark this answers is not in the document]
YaqiWang: This is pretty mind boggling...

The old logic is
```
  if (!lower_domains.count(Moose::BOUNDARY_SIDE_LOWERD_ID) && lower_domains.size() != 1)
    paramError(
        "lowerd_variable",
        "Must be only defined on the subdomain BOUNDARY_SIDE_LOWERD_SUBDOMAIN subdomain that is "
        "added by Mesh/build_all_side_lowerd_mesh=true");
```
which means when the lower-d variable only has ONE block, it will never hit this error, no matter what that block is. So this check is actually not right. What we want is to make sure this lower-d variable is defined on ALL boundary lower-d blocks. And the variable is not necessarily only on the boundary lower-d blocks. Let us say the variable is indeed on multiple blocks that makes this check meaningful, this logic is indeed to make sure the variable is defined on all boundary lower-d blocks.

The current logic
```
  const auto & lower_domains = _lowerd_var.activeSubdomains();
  for (const auto & id : _mesh.boundaryLowerDBlocks())
    if (lower_domains.count(id) == 0)
      paramError("lowerd_variable",
                 "Must be defined on the boundary lower-dimensional subdomain '" +
                     _mesh.getSubdomainName(id) +
                     "' that is added by Mesh/build_all_side_lowerd_mesh=true");
```
makes sure the variable is defined on the boundary lower-d block, and the code is consistent with the error message.

framework/src/mesh/MooseMesh.C
GiudGiud: If I define lower D blocks in my mesh generation, this will delete all of them

framework/src/bcs/LowerDIntegratedBC.C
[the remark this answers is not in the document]
YaqiWang: Not need all, but for now, this is required. When the variable is on multiple blocks, this check is consistent with the old check. This is a BC, not on all blocks. But currently we require the lower-d variable is defined on all boundary blocks.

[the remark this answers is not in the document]
GiudGiud: The current check does not impose for the variable to be defined on all lower D blocks. It imposes that the variable is at least defined on `Moose::BOUNDARY_SIDE_LOWERD_ID` though in a weird way

> This is a BC, not on all blocks.

you made the same mistake in the kernels
```
  for (const auto & id : _mesh.interiorLowerDBlocks())
    if (lower_domains.count(id) == 0)
      paramError("lowerd_variable",
                 "Must be defined on the interior lower-dimensional subdomain '" +
                     _mesh.getSubdomainName(id) +
                     "'that is added by Mesh/build_all_side_lowerd_mesh=true");
```

framework/src/mesh/MooseMesh.C
YaqiWang: No because `_lower_d_interior_blocks` only contains blocks in this name `INTERNAL_SIDE_LOWERD_SUBDOMAIN_xxx`.

framework/src/bcs/LowerDIntegratedBC.C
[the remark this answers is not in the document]
YaqiWang: >it imposes that the variable is at least defined on Moose::BOUNDARY_SIDE_LOWERD_ID

Yes, that will guarantee this BC will work properly.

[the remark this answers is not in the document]
YaqiWang: The same for the kernels, if the lower-d variable is defined on all lower-d interior blocks, the kernel will work properly, otherwise, it may not because there could be an interior side where the variable does not have definition.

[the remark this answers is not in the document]
YaqiWang: Added an issue for overly restrict check here: https://github.com/idaholab/moose/issues/29151

framework/src/mesh/MooseMesh.C
YaqiWang: There could be a change that a crazy user created a block with this id. Then this error will be useful to tell the users to give a new id to his block.

YaqiWang:
@GiudGiud More comments?

framework/src/dgkernels/DGLowerDKernel.C
```diff
@@ -41,15 +41,21 @@ DGLowerDKernel::DGLowerDKernel(const InputParameters & parameters)
...
-  if (!lower_domains.count(Moose::INTERNAL_SIDE_LOWERD_ID) && lower_domains.size() != 1)
-    paramError(
-        "lowerd_variable",
-        "Must be only defined on the subdomain INTERNAL_SIDE_LOWERD_SUBDOMAIN subdomain that is "
-        "added by Mesh/build_all_side_lowerd_mesh=true");
-
-  if (_var.activeSubdomains().count(Moose::INTERNAL_SIDE_LOWERD_ID))
-    paramError("variable",
-               "Must not be defined on the subdomain INTERNAL_SIDE_LOWERD_SUBDOMAIN subdomain");
+  for (const auto & id : _mesh.interiorLowerDBlocks())
```
GiudGiud, line 44: if there is more code like these two checks that is shared between HFEM Kernels you should consider a base class

framework/src/interfaces/BlockRestrictable.C
```diff
@@ -338,8 +338,9 @@ BlockRestrictable::checkVariable(const MooseVariableFieldBase & variable) const
 {
   // a variable defined on all internal sides does not need this check because
   // it can be coupled with other variables in DG kernels
-  if (variable.activeSubdomains().count(Moose::INTERNAL_SIDE_LOWERD_ID) > 0)
-    return;
+  for (const auto & id : variable.activeSubdomains())
+    if (_blk_mesh->interiorLowerDBlocks().count(id) > 0)
```
GiudGiud, line 342: this isnt quite the same check now. it's on ANY lowerD blocks when it used to be on ALL (just 1)

GiudGiud: so lets update the comment

framework/src/mesh/MooseMesh.C
```diff
@@ -632,19 +635,69 @@ MooseMesh::buildLowerDMesh()
...
+        boundary_side_types.insert(side_elem->type());
+    }
+  mesh.comm().set_union(interior_side_types);
+  mesh.comm().set_union(boundary_side_types);
+
+  // assign block ids for different side types
+  std::map<ElemType, SubdomainID> interior_block_ids;
+  std::map<ElemType, SubdomainID> boundary_block_ids;
+  // we assume this id is not used by the mesh
+  auto id = libMesh::Elem::invalid_subdomain_id - 2;
```
GiudGiud, line 677: this is risky. what id did we used to use?

YaqiWang: The same, otherwise we would see exodiffs.

framework/src/interfaces/BlockRestrictable.C
YaqiWang: Good catch, I changed the code here to make the comment valid.

framework/src/dgkernels/DGLowerDKernel.C
YaqiWang: `ArrayDGLowerDKernel` and `DGLowerDKernel` are the two base classes. Same for the two integrated BCs. `HFEMDirichletBC` needs one more check on a new coupled lower-d variable `u_hat`.

GiudGiud (approved):
We are trading one issue (same block name different element types) for another here (broken block restriction of lower D variables for the kernels checked in the framework) but they will both be resolved in time

PR #29146, opened 2024-11-26
Allowing Dirichlet BCs to take state - needed for limiters

### LINKED ISSUE #29145: Allowing FVDrichletBCs to take state
### Motivation

Currently Dirichlet BCs in the FV framework are hardcoded to use the current state. This is inconvenient if we want Dirichlet BCs to take a previous state of a functor.

### Design

Modify the `getDirichletBoundaryFaceValue` function in `MooseVariableFV` to use the state provided to this function and extend the `boundaryValue` getter in Dirichlet BCs to take the state provided to this function. Also, remove the assert that ensures that state.state is 0

### Impact

More flexibility when using FV Dirichlet BCs or developing new derived classes from FVDirichletBCs

tanoret:
Closes #29145

### DIFF
--- a/framework/include/fvbcs/FVDirichletBC.h
+++ b/framework/include/fvbcs/FVDirichletBC.h
@@ -21,7 +21,7 @@ public:
 
   static InputParameters validParams();
 
-  ADReal boundaryValue(const FaceInfo & fi) const override;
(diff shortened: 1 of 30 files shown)


grmnptr (changes requested):
Nice, just one minor thing when using it with functions.

grmnptr (approved):
Looks good to me!

PR #29167, opened 2024-11-28
Add a missing communication for interior and boundary lower-d blocks

### LINKED ISSUE #29166: Small problem with distributed mesh with HFEM can hit an error
### Motivation

We recently added a test input `moose/test/tests/kernels/hfem/3d-prism.i` that has a 2*2*2 3D mesh and we hit an error when running it with 4 processors with distributed mesh option:
```
kernels/hfem.mixed_side_types_part1: *** ERROR ***
kernels/hfem.mixed_side_types_part1: Each subdomain must contain at least one Kernel.
kernels/hfem.mixed_side_types_part1: The following block(s) lack an active kernel: INTERNAL_SIDE_LOWERD_SUBDOMAIN_QUAD4  (ids: 65532 )
```
While typically we do not spread the mesh so thin in real calculations, this can be an issue in some extreme cases.

### Design

`NonlinearSystemBase::checkKernelCoverage` will need more checks.

### Impact

Better sanity check to avoid unnecessary errors in HFEM.

YaqiWang:
Closes #29166. This is follow-up fix for PR #29144.

### DIFF
--- a/framework/src/mesh/MooseMesh.C
+++ b/framework/src/mesh/MooseMesh.C
@@ -1419,6 +1419,8 @@ MooseMesh::cacheInfo()
       _block_node_list[node.id()].insert(elem->subdomain_id());
     }
   }
+  _communicator.set_union(_lower_d_interior_blocks);
+  _communicator.set_union(_lower_d_boundary_blocks);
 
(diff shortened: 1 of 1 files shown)


loganharbour (approved):
Failures are unrelated.

PR #29172, opened 2024-11-30
Copy out-of-directory symlinks during install

### LINKED ISSUE #29171: Fix installed symlinks out of test folder
### Motivation

We don't currently copy test symlinks during install that resolve outside of the test directory that is being copied.

Example: https://github.com/idaholab/moose/blob/cedfd055b01b8ccebc50e5d10a89847caa8e24ae/test/tests/meshdivisions/depletion_id_in.e points outside of the `tests` directory. The installed symlink is broken:

```
$ apptainer exec oras://mooseharbor.hpc.inl.gov/moose/moose-openmpi-x86_64:latest realpath /opt/moose/share/combined/framework/tests/meshdivisions/depletion_id_in.e
INFO:    Using cached SIF image
realpath: /opt/moose/share/combined/framework/tests/meshdivisions/depletion_id_in.e: No such file or directory
```

### Design

Resolve the symlinks and copy additional data (removing the symlinks) as needed.

### Impact

Added capability.

loganharbour:
Closes #29171

Relies on rsync to do the copy, with the `--copy-unsafe-links` which will copy full files instead of links for links that no longer resolve. Could do it more manually, but this is cleaner. The "unsafe" links are checked. We could do this without rsync, but it's a bit more complex. And seeing that we require it in the installed environment, I'm fine with requiring it in the install environment.

Resolves the issue of symlinks that point out of one installed test directory into another.

### DIFF
--- a/conda/moose/run_test.sh
+++ b/conda/moose/run_test.sh
@@ -4,7 +4,7 @@ set -e
 if [[ $(uname) == 'Darwin' ]] && [[ $(uname -m) == 'x86_64' ]]; then
     exit 0
 fi
-combined-opt --copy-inputs framework
-cd combined/framework
-combined-opt --run -j 4 --re=kernels/simple_diffusion
+moose_test-opt --copy-inputs tests
+cd moose_test/tests
+moose_test-opt --run -j 4 --re=kernels/simple_diffusion
 exit 0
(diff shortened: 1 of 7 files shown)


permcody (approved):

loganharbour:
Failure is unrelated.

PR #29181, opened 2024-12-03
Try just installing moose-tools in base

loganharbour:


milljm:
Yeah I like that idea! No need to have other environments in a specific container like these.

PR #29363, opened 2024-12-05
Let UserObject initialSetup honor execution group

### LINKED ISSUE #29362: Let UserObject initialSetup honor execution group
### Motivation

The order of execution of UserObjects can be controlled by setting the execution group. However, it only applies to the execution and `initialSetup` does not respect it. It will be beneficial if `initialSetup` respect the execution group as well.

### Design

Perform `initialSetup` of UserObjects following the execution group just like `execute`.

### Impact

Better dependency handling of UserObjects.

NamjaeChoi:
Closes #29362.


NamjaeChoi:
I'm not sure what the failing tests are about. There is no test added, but there were two tests that were relying on the order of `initialSetup` between UserObjects that broke after this change, so they can serve as the test.

lindsayad (approved):

lindsayad:
failures are unrelated

PR #29366, opened 2024-12-05
Handle created but uninitialized NumericVector in dataLoad

### LINKED ISSUE #29365: dataLoad specialization for NumericVector needs to handle cases where vector was created but not initialized
### Bug Description

A `NumericVector` can be created early but initialized later, but if `dataLoad` is called for such a `NumericVector` between its creation and initialization, it will error.

### Steps to Reproduce

Define a restartable `NumericVector`, create it in a constructor and initialize it in `initialSetup`.

### Impact

A confusing error.

### [Optional] Diagnostics

_No response_

NamjaeChoi:
Closes #29365.

### DIFF
--- a/framework/src/restart/DataIO.C
+++ b/framework/src/restart/DataIO.C
@@ -705,12 +705,12 @@ dataLoad(std::istream & stream, std::unique_ptr<libMesh::NumericVector<Number>> 
   // Construct the vector given the type, only if we need to. v could be non-null here
   // if we're advancing back and loading a backup
   if (!v)
-  {
     v = NumericVector<Number>::build(comm, solver_package);
-    v->init(size, local_size);
-  }
   else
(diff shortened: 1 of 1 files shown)


lindsayad (changes requested):
I don't fully understand this yet. Can you add a framework test?

NamjaeChoi:
I think I might just close this PR. Now I'm thinking that this is unimportant and easy to workaround.

My initial intention was, let's say I have a restartable `NumericVector` declared as restrable, which should be in the form of `std::unique_ptr<NumericVector<Real>>` in order to leverage the default `dataLoad` and `dataStore`. One can allocate this vector early (e.g. in a constructor)
```
_vector = std::make_unique<PetscVector<Real>>(comm());
```
but choose to initialize later (e.g. in `initialSetup`)
```
_vector->init(_n);
```
`dataLoad` occurs between those two and triggers an assertion.

PR #29382, opened 2024-12-09
Enhanced UEL support

dschwen:
Add support for arbitrary user elements, defined in Abaqus input files. This work consists of

- A new mesh class that manages a libMesh mesh with node elements and separately manages arbitrary element connectivity
- Actions to create the Variables required by the Abaqus input file with the correct block restrictions to assign variables to subsets of element nodes as defined in the input
- A new user object to interface with the UEL plugin 
- A relationship manager that utilizes the custom element connectivity from the mesh object to enable proper ghosting and coupling


dschwen:
Those failing tests should all be skipped by the capabilties parameter at top level.

framework/src/base/Assembly.C
```diff
@@ -169,15 +169,19 @@ Assembly::Assembly(SystemBase & sys, THREAD_ID tid)
...
-                : FEGenericBase<Real>::build(_mesh_dimension - 1, FEType(FIRST, LAGRANGE));
-  // This FE object should not take part in p-refinement
-  _fe_msm->add_p_level_in_reinit(false);
-  _JxW_msm = &_fe_msm->get_JxW();
-  // Prerequest xyz so that it is computed for _fe_msm so that it can be used for calculating
-  // _coord_msm
-  _fe_msm->get_xyz();
+  if (_mesh_dimension > 0)
+  {
+    // no mortar in 0D meshes
```
lindsayad, line 176: I would swap the placement of these two comments

framework/src/bcs/DirichletBCBase.C
```diff
@@ -28,7 +28,7 @@ DirichletBCBase::computeValue(NumericVector<Number> & current_solution)
 {
   mooseAssert(_preset, "BC is not preset");
 
-  if (_var.isNodalDefined())
+  if (_var.isNodalDefined() && shouldApply())
```
lindsayad, line 31: I would prefer we put the `shouldApply()` in `NonlinearSystemBase.C` to be consistent with the `nbc->shouldApply()` check we do before `nbc->computeResidual()`. We could (probably should) then `mooseAssert(shouldApply())` here. 

Alternatively you could move the `nbc->shouldApply()` out of `NonlinearSystemBase` and put it in the nodal bc class itself. I just want consistency

lindsayad: The latter seems better encapsulated

framework/src/executioners/SteadyBase.C
```diff
@@ -30,7 +30,8 @@ SteadyBase::SteadyBase(const InputParameters & parameters)
     _system_time(getParam<Real>("time")),
     _time_step(_problem.timeStep()),
     _time([this]() -> Real & { return this->_problem.time() = this->_system_time; }()),
-    _output_iteration_number(0)
+    _output_iteration_number(0),
+    _last_solve_converged(true)
```
lindsayad, line 34: this is already populated in the header isn't it?

lindsayad (commented):
Just reviewing the framework pieces

PR #29516, opened 2024-12-10
Add PIMPLE transient solve for linearFV

### LINKED ISSUE #29150: Add PIMPLE solver to the NS module
### Motivation

We need a segregated transient fluid solver in the NS module. We could add PIMPLE (PISO+SIMPLE).

### Design

Add PIMPLE, potentially abstract some functions in SIMPLE.

### Impact

Would allow transient multiphysics solves with segregated fluid dynamics solvers.

grmnptr:
Closes #29150


modules/navier_stokes/src/executioners/PIMPLESolve.C
```diff
@@ -0,0 +1,456 @@
...
+  // different momentum components
+  std::vector<std::pair<unsigned int, Real>> its_normalized_residuals;
+
+  LinearImplicitSystem & momentum_system_0 =
+      libMesh::cast_ref<LinearImplicitSystem &>(_momentum_systems[0]->system());
+
+  PetscLinearSolver<Real> & momentum_solver =
+      libMesh::cast_ref<PetscLinearSolver<Real> &>(*momentum_system_0.get_linear_solver());
+
+  momentum_solver.reuse_preconditioner(true);
```
GiudGiud, line 93: where is the step to "refresh" it on the next time step?

modules/navier_stokes/src/executioners/PIMPLESolve.C
```diff
@@ -0,0 +1,456 @@
...
+    LinearImplicitSystem & momentum_system =
+        libMesh::cast_ref<LinearImplicitSystem &>(_momentum_systems[system_i]->system());
+
+    NumericVector<Number> & solution = *(momentum_system.solution);
+    NumericVector<Number> & rhs = *(momentum_system.rhs);
+    SparseMatrix<Number> & mmat = *(momentum_system.matrix);
+
+    auto diff_diagonal = solution.zero_clone();
+
+    // We plug zero in this to get the system matrix and the right hand side of the linear problem
```
GiudGiud, line 112: which step plugs the zeros?

modules/navier_stokes/src/executioners/PIMPLESolve.C
```diff
@@ -0,0 +1,456 @@
...
+    // Very important, for deciding the convergence, we need the unpreconditioned
+    // norms in the linear solve
+    LIBMESH_CHKERR(KSPSetNormType(momentum_solver.ksp(), KSP_NORM_UNPRECONDITIONED));
+    // Solve this component. We don't update the ghosted solution yet, that will come at the end
+    // of the corrector step. Also setting the linear tolerances and maximum iteration counts.
+    _momentum_linear_control.real_valued_data["abs_tol"] = _momentum_l_abs_tol * norm_factor;
+    momentum_solver.set_solver_configuration(_momentum_linear_control);
+
+    _console << " solution after solve " << std::endl;
+    solution.print();
```
GiudGiud, line 132:
```suggestion
    if (_print_fields)
    {
      _console << " solution before solve " << std::endl;
      solution.print();
     }
```

GiudGiud (commented):
looks good to me.
all the SIMPLE tests use this new code indirectly so imo 90% of it is already proven. Let's just get some PIMPLE-SIMPLE comparisons for some steady state solutions?

modules/navier_stokes/src/executioners/PIMPLESolve.C
grmnptr: Great catch! I will add the preconditioner refresh after every simple iteration. At some point I'll add an adaptive preconditioner refresh as well (so when residuals are low we don't refresh anymore).

grmnptr:
@tanoret I think this is ready for a technical review.

modules/thermal_hydraulics/src/base/Simulation.C
```diff
@@ -633,7 +633,7 @@ Simulation::sortAddedComponentVariables() const
 void
 Simulation::addVariables()
 {
-  Transient * trex = dynamic_cast<Transient *>(getApp().getExecutioner());
+  TransientBase * trex = dynamic_cast<TransientBase *>(getApp().getExecutioner());
```
tanoret, line 636: Maybe same here for the dynamic_cast. Feel free to disregard

modules/optimization/src/userobjects/DensityUpdateTwoConstraints.C
```diff
@@ -100,7 +100,7 @@ DensityUpdateTwoConstraints::DensityUpdateTwoConstraints(const InputParameters &
                  "This parameter needs to be provided when including thermal sensitivity.");
   }
 
-  auto transient = dynamic_cast<Transient *>(_app.getExecutioner());
+  auto transient = dynamic_cast<TransientBase *>(_app.getExecutioner());
```
tanoret, line 103: Same here for the dynamic_cast

modules/navier_stokes/examples/flow-over-circle/flow_over_circle_linearfv.i
```diff
@@ -0,0 +1,228 @@
+[Problem]
```
tanoret, line 1: Very nice verification!!!

tanoret: Would you please compute the Strouhal number and add a validation test against the analytical solution? See for example: https://www.grc.nasa.gov/www/wind/valid/lamcyl/Study1_files/Study1.html

tanoret (commented):
Nice work! A couple of general suggestions:
- Please add a test case for the Strouhal number so that we can actually validate the transients capabilities
- I would suggest more separation between PIMPLE and SIMPLE. In that way it would be easier to test and implement new schemes in the future

Otherwise, it is great!

modules/navier_stokes/examples/flow-over-circle/flow_over_circle_linearfv.i
grmnptr: This exercise was based on a benchmark example in the literature. When I first added this example for the other solvers I split the input files into multiple files to facilitate mesh and parameter reuse. In `header.i` I have this:

```
# -----------------------------------------------------------------------------
# Flow around a cylinder (2D) benchmark validation case
# This example showcases a flow around a cylinder which results in vortex
# shedding. The problem specification has been taken from the following paper:
#
# @incollection{schafer1996benchmark,
#   title={Benchmark computations of laminar flow around a cylinder},
#   author={Sch{\"a}fer, Michael and Turek, Stefan and Durst, Franz and Krause, Egon and Rannacher, Rolf},
#   booktitle={Flow simulation with high-performance computers II},
#   pages={547--566},
#   year={1996},
#   publisher={Springer}
# }
# The Reyndols number is Re=100.
# The expected Strouhal number (St) is in the [0.2950, 0.3050] range, with
# refinement=8 and sufficiently fine time steps, we expect to get St=0.2941 with the model below.
# Run it using the following command:
# ./navier_stokes-opt -i header.i mesh.i flow_over_circle.i executioner_postprocessor.i
# -----------------------------------------------------------------------------
```

grmnptr: Unfortunately, adding a postprocessor for the Strouhal number could be tricky in this case. A good approach would be through a Fourier transform. The number I am reporting was based on a manual calculation based on paraview data. I would not like to add that postprocessor in this PR.

framework/src/actions/ComposeTimeStepperAction.C
```diff
@@ -59,9 +59,9 @@ ComposeTimeStepperAction::act()
       if (isParamValid(param_name_value.first))
         new_params.applyParameter(_pars, param_name_value.first);
 
-    Transient * transient = dynamic_cast<Transient *>(_app.getExecutioner());
+    TransientBase * transient = dynamic_cast<TransientBase *>(_app.getExecutioner());
```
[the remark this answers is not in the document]
grmnptr: This comment covers most of the cast-related comments, I will address other comments where this does not apply. We store the executioner in an `Executioner *` pointer. This goes towards either steadybase/transientbase. So the cast in this case is from abstract to derived. If I call this with a steadybase and I use other casts, like static_cast for example, we get an undefined behavior. Dynamic cast will do a runtime check if the cast is complete and if the cast is not complete, dynamic cast returns a `nullptr`. This nullpointer is used with a mooseError/Asssert in most of the cases after these casts (like the line below this which was not modified). So the user will get an error/assert something like this:
 `"The transient executioner does not exist"` or similar in other files.
 
In this specific action, this cast is only used to check during runtime if we have a transient executioner or not. We do this because this adds a time stepper which would make no sense for a steady executioner. So storing it in a member is not necessary.

framework/src/multiapps/FullSolveMultiApp.C
```diff
@@ -72,7 +72,7 @@ FullSolveMultiApp::initialSetup()
 
       if (_ignore_diverge)
       {
-        Transient * tex = dynamic_cast<Transient *>(ex);
+        TransientBase * tex = dynamic_cast<TransientBase *>(ex);
```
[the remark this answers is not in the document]
grmnptr: This is also a case where we are using the pointer from the dynamic cast (would be nullptr if Executioner* cant be cast to the derived transient) to see if the simulation is a transient simulation or not. There is an error message after this cast.

framework/src/multiapps/TransientMultiApp.C
```diff
@@ -659,7 +659,7 @@ void
 TransientMultiApp::setupApp(unsigned int i, Real /*time*/) // FIXME: Should we be passing time?
 {
   auto & app = _apps[i];
-  Transient * ex = dynamic_cast<Transient *>(app->getExecutioner());
+  TransientBase * ex = dynamic_cast<TransientBase *>(app->getExecutioner());
```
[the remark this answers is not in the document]
grmnptr: Same here. We have an error that follows this:
```
  TransientBase * ex = dynamic_cast<TransientBase *>(app->getExecutioner());
  if (!ex)
    mooseError("MultiApp ", name(), " is not using a Transient Executioner!");
```

framework/src/outputs/ProgressOutput.C
```diff
@@ -29,7 +29,7 @@ ProgressOutput::validParams()
 
 ProgressOutput::ProgressOutput(const InputParameters & parameters)
   : Output(parameters),
-    _transient_executioner(dynamic_cast<Transient *>(_app.getExecutioner())),
+    _transient_executioner(dynamic_cast<TransientBase *>(_app.getExecutioner())),
```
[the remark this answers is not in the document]
grmnptr: We use dynamic cast to get the nullpointer which is checked at other parts of this object.

framework/src/postprocessors/NumFailedTimeSteps.C
```diff
@@ -25,7 +25,7 @@ NumFailedTimeSteps::NumFailedTimeSteps(const InputParameters & parameters)
 {
   if (_subproblem.isTransient())
   {
-    _timestepper = dynamic_cast<Transient *>(_app.getExecutioner())->getTimeStepper();
+    _timestepper = dynamic_cast<TransientBase *>(_app.getExecutioner())->getTimeStepper();
```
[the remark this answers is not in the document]
grmnptr: Considering that we already have a check right before this, we could potentially use other casts. I'll check.

framework/src/predictors/Predictor.C
```diff
@@ -59,7 +59,7 @@ Predictor::Predictor(const InputParameters & parameters)
     _skip_times(getParam<std::vector<Real>>("skip_times")),
     _skip_times_old(getParam<std::vector<Real>>("skip_times_old")),
     _skip_after_failed_timetep(getParam<bool>("skip_after_failed_timestep")),
-    _timestep_tolerance(dynamic_cast<Transient *>(_app.getExecutioner())->timestepTol())
+    _timestep_tolerance(dynamic_cast<TransientBase *>(_app.getExecutioner())->timestepTol())
```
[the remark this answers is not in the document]
grmnptr: This could potentially be a different cast, but it is on the safer side. This will never fail without a `mooseError` because the `SetupPredictorAction` will error when this object is added if the executioner is not Transient.

framework/src/postprocessors/RelativeSolutionDifferenceNorm.C
```diff
@@ -24,7 +24,7 @@ RelativeSolutionDifferenceNorm::validParams()
 }
 
 RelativeSolutionDifferenceNorm::RelativeSolutionDifferenceNorm(const InputParameters & params)
-  : GeneralPostprocessor(params), _trex(dynamic_cast<Transient *>(_app.getExecutioner()))
+  : GeneralPostprocessor(params), _trex(dynamic_cast<TransientBase *>(_app.getExecutioner()))
```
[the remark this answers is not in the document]
grmnptr: We use the potential nullpointer to throw and error:
```
if (!_trex)
    mooseError("RelativeSolutionDifferenceNorm postprocessor is only for transient calculations");
```

modules/navier_stokes/include/executioners/SIMPLESolve.h
```diff
@@ -10,67 +10,25 @@
...
 
 /**
- * SIMPLE-based solution object with linear FV system assembly.
+ * Implementation of Semi-Implicit Method for Pressure-Linked Equations (SIMPLE)
+ * with MOOSE-based applications. It is a slightly restricted version of
+ * the PIMPLE algorithm: (i) no PISO iterations are allowed, (ii) no time
+ * kernels are allowed.
  */
-class SIMPLESolve : public SIMPLESolveBase
+class SIMPLESolve : public PIMPLESolve
```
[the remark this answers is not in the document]
grmnptr: We will remove SIMPLESolveBase once we can remove the nonlinear assembly approach. I can add a few lines to this header to justify this inheritance. To me SIMPLE is just the same as PIMPLE without the PISO loop. So the difference is just one parameter (number of piso corrections). Both will be able to benefit from the SIMPLE-C option so in that sense this inheritance is still good.

modules/navier_stokes/examples/flow-over-circle/flow_over_circle_linearfv.i
tanoret: Thanks for the comments. 

How would you validate non-regression for the transient implantation then? Someone can mess up the time integration scheme by preserving the drag and lift coefficient in the few steps that we compare. The actual validation would be the St number. It is important that this is correct

I suggest the following:
1. Reduce the mesh density and run a couple cycles (not a couple time steps)
2. Export the csv for all time steps
3. Use a Python test to determine the oscillation frequency

modules/navier_stokes/include/executioners/SIMPLESolve.h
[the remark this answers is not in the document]
tanoret: Agree, but this kind of hardcoding don't allow us to experiment with different SIMPLE projection methods without having to change the backend of SIMPLE. The pressure lumping, face interpolation, etc. are not necessarily the same for SIMPLE and PIMPLE. I would suggest to follow the OF way and separating PIMPLE and SIMPLE in two different methods. Thanks and sorry if this requires extra work

tanoret:
Thanks again for this work. My comments are mostly to make sure that the code is robust and we can keep improving the code in the future without much refactoring. Thanks again!

modules/navier_stokes/include/executioners/SIMPLESolve.h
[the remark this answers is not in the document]
GiudGiud: If a method needs to be implemented differently, you can make it virtual in the base class then override it in the SiMPLE derived class. 

Copy pasting code will not make things more tinkerable enough to be worth it imo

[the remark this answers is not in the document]
tanoret: Mmm.. maybe, but then things get a bit entangled an not really clean. PIMPLE and SIMPLE are ultimately two different methods and as we move to more engineering ready applications they both have different schemes to implement. The best way to keep them clean would be to generate a common base class for both and then deriving SIMPLE and PIMPLE from the base class, e.g.

BaseClass
  |--------> SIMPLE 
  |--------> PIMPLE
  ...
  |--------> All the other methods that we will try in the future
  
  We can host all the common methods in the base class if need be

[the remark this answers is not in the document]
grmnptr: I will change this to something that is closer to what you want, but the two solvers are 99% the same if we don't consider moving reference frames/meshes. Even in OF, the face interpolation methods and all the details are consistent between the two and if we add an option to one we should definitely add it to the other one too. I don't see a lot of need for customization at the moment, but assuming that we might touch moving meshes/reference frames I will do this change.

modules/navier_stokes/examples/flow-over-circle/flow_over_circle_linearfv.i
grmnptr: We will add these types of tests to the flow test bed. To be able to get a reasonable St number, we will need to run this with reasonable refinement and time step size (Implicit Euler is pretty diffusive). Without these the St will be 20-ish % off and I would rather not show that.

modules/optimization/src/userobjects/DensityUpdateTwoConstraints.C
grmnptr: The cast is used for error checking later on to make sure we have a transient executioner driving this.

modules/thermal_hydraulics/src/base/Simulation.C
grmnptr: The cast is used for differentiating between steady and transient behavior later on this.

modules/navier_stokes/examples/flow-over-circle/flow_over_circle_linearfv.i
grmnptr: Done!

modules/navier_stokes/include/executioners/SIMPLESolve.h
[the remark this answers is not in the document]
grmnptr: Done!

grmnptr:
I think this ready for a second round!

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/linear-segregated/steady-transient-compare/compare.cmp
```diff
@@ -0,0 +1,3 @@
+TIME STEPS relative 1.0 floor 0.0
+
+ELEMENT VARIABLES relative 5e-5 floor 1e-6
```
GiudGiud, line 3: note that you dont need a .cmp file if you are not assigning tolerances on a per-variable basis. you can set rel_err / abs_zero in the tests config

GiudGiud (approved):
This is a great capability, and I did not see any issue with the code.  
new inheritance structure will be good post nonlinear SIMPLE deprecation. Looks odd now but no real issue.

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/linear-segregated/steady-transient-compare/compare.cmp
grmnptr: I did this just for the time step tolerance.

GiudGiud: oh nice

tanoret (approved):
Thank you for implementing some of the requested changes and addressing the rest. Good to go for me

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -0,0 +1,459 @@
...
+    NS::FV::relaxMatrix(mmat, _momentum_equation_relaxation, *diff_diagonal);
+    NS::FV::relaxRightHandSide(rhs, solution, *diff_diagonal);
+
+    // The normalization factor depends on the right hand side so we need to recompute it for this
+    // component
+    Real norm_factor = NS::FV::computeNormalizationFactor(solution, mmat, rhs);
+
+    // Very important, for deciding the convergence, we need the unpreconditioned
+    // norms in the linear solve
+    LIBMESH_CHKERR(KSPSetNormType(momentum_solver.ksp(), KSP_NORM_UNPRECONDITIONED));
```
roystgnr, line 115: This (and the other `LIBMESH_CHKERR` calls below) just broke libMesh testing, thanks to me foolishly assuming that just because MOOSE tested clean against https://github.com/libMesh/libmesh/pull/4037 a month ago it would still be clean when I noticed the overlooked PR and hit Merge today.

Can we get them changed to LibmeshPetscCall?

grmnptr: Yes, I will submit the PR in 15 mins.

PR #28929, opened 2024-10-24
Switch to using LibmeshPetscCall wrapper macros

nmnobre:
Hi @lindsayad,

## Reason and Design
Same as that of libMesh/libmesh#3973.
Shortens the code and allows exception throwing.

~I'm missing `LIBMESH_CHKERR`, I'll do that in another PR, I suspect this will be enough for a few CI problems already.~

Cheers,
-N

PR #29563, opened 2024-12-17
Amendment to #28929

nmnobre:
There were some new code additions to the Navier-Stokes module in between while I was working on #28929 and its merger, which meant I missed a very small number of incidences.


GiudGiud (approved):
I think I saw Alex catch a few of those in his PR but it should not conflict

nmnobre:
Fix for Ba Hun here: WilkAndy/ba_hun#63.
Fix for Raccoon here: hugary1995/raccoon#189.
I don't know why Blue Crab is failing.

PR #29566, opened 2024-12-17
Change MooseDocs CIVET mergeresults command defaults and add extension documentation

### LINKED ISSUE #29565: MooseDocs CIVET extension mergeresults command always shows most up-to-date hashes
### Bug Description

When using the `!civet mergeresults` command in MooseDocs, it currently shows links associated with the CI for the most up-to-date git commit SHAs for the repository being referenced in the extension. This is great when websites are built for production (e.g., https://mooseframework.inl.gov), but awful if a local copy of the website is being built using an older copy of the code (say, when you want to generate documentation for an older release). This causes confusion because the links to the CI references no longer connect to the code or documentation being rendered.

### Steps to Reproduce

Use `!civet mergeresults` in MooseDocs with the CIVET Extension properly enabled when using an older version of MOOSE or MOOSE-based applications.

### Impact

This is an annoyance, and possible violation of SQA compliance if this is being used in audited documentation.

### [Optional] Diagnostics

_No response_

cticenhour:
Closes #29565 

@gambka I discovered this while testing with your BISON branch. This should make the output of `!civet mergeresults` consistent for the release versions of BISON docs (or any older version of the docs built locally) by default.

### DIFF
--- a/python/doc/content/python/MooseDocs/extensions/civet.md
+++ b/python/doc/content/python/MooseDocs/extensions/civet.md
@@ -1,0 +1,108 @@
+# CIVET Extension
+
+The [!ac](CIVET) extension adds the ability for MooseDocs to download results from a [CIVET] client
+(e.g., [https://civet.inl.gov](https://civet.inl.gov)) and present them as reports, links, badges,
+and other information in a MooseDocs page. This extension is most-often used with the [extensions/sqa.md].
+
+## Extension Configuration
+
+[config-civet-ext] lists the available configuration settings for the CIVET extension. These
+configuration items should be included in the MooseDocs configuration file (e.g., config.yml).
+
+!devel settings id=config-civet-ext
+                caption=Configuration options for the CIVET extension.
+                module=MooseDocs.extensions.civet
+                object=CivetExtension
+
+## Extension Commands
+
+The following sections provide information regarding each of the commands available to the extension.
+
+### Results Inline Linking
+
+The `results` inline command allows for linking to testing results for the version of the code repository
+used to build the documentation. It generates a single link displaying the git commit SHA which links
+to the associated CIVET results for that SHA.
+
+!devel! example id=civet-results-example
+                caption=Example of the CIVET extension `results` command.
+
+[!civet!results]
+
+!devel-end!
+
+!devel settings id=civet-results
+                caption=Options for the CIVET extension `results` command.
+                module=MooseDocs.extensions.civet
+                object=CivetResultsCommand
+
+### Merge Results Linking
+
+The `civet mergeresults` command provides a set of links to CIVET results pages for the version of
+the code repository used to build the documentation (by default). This is similar to the [#results-inline-linking]
+command, except multiple links are provided. Each link is labeled using the git SHA associated with
+the merge event that prompted testing. This means that MOOSE, for example, might have multiple links
+associated with `next`, `devel`, and `master` testing events associated with a single `master` branch
+merge constituting the current version of the code and documentation. If the `use_current_hash`
+settings option is set to `False`, then the most up-to-date version of the these links is retrieved
+from the online remote repository.
+
+!devel! example id=civet-merge-results-example
+                caption=Example of the CIVET extension `mergeresults` command.
+
+!civet mergeresults
+!devel-end!
+
+!alert! tip title=Inline usage
+Note that this command can also be used inline, using the `[!civet!mergeresults]` syntax.
+!alert-end!
+
+!devel settings id=civet-merge-results
+                caption=Options for the CIVET extension `mergeresults` command.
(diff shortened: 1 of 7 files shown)


loganharbour (commented):
Unit tests an option?

cticenhour:
@loganharbour Yeah, I can add one.

python/doc/content/python/MooseDocs/extensions/civet.md
```diff
@@ -0,0 +1,108 @@
...
+The [!ac](CIVET) extension adds the ability for MooseDocs to download results from a [CIVET] client
+(e.g., [https://civet.inl.gov](https://civet.inl.gov)) and present them as reports, links, badges,
+and other information in a MooseDocs page. This extension is most-often used with the [extensions/sqa.md].
+
+## Extension Configuration
+
+[config-civet-ext] lists the available configuration settings for the CIVET extension. These
+configuration items should be included in the MooseDocs configuration file (e.g., config.yml).
+
+!devel settings id=config-civet-ext
```
loganharbour, line 12: Image

This looks a little goofy (red box). Maybe we should have a global cache folder for all of moosedocs that is a command line option or something. Isn't really important... just odd

python/doc/content/python/MooseDocs/specification.md
```diff
@@ -55,6 +55,7 @@ tables. [user-ext] provides a list of extensions that are useful for those writi
 | [extensions/media.md] | Extension for including images and movies. |
 | [extensions/appsyntax.md] | Enables the use of MOOSE application syntax within markdown files. |
 | [extensions/sqa.md] | Provides tools for writing software quality documentation using templates. |
+| [CIVET Extension](extensions/civet.md optional=True) | Provides linking and results connections to [CIVET](https://github.com/idaholab/civet) testing results. |
```
loganharbour: Why does this need to be optional exactly? Does it require that the civet extension also be used in the docs that build it?

loganharbour (commented):
Minor. Looks good to me otherwise

python/doc/content/python/MooseDocs/extensions/civet.md
cticenhour: It is odd, indeed. I hadn't noticed it. I need to remind myself of how that cache is set, and we might align with another default instead for more consistency.

python/doc/content/python/MooseDocs/specification.md
cticenhour: Yes, it's turned off for most applications, since most of them don't utilize the CIVET extension (or need to do so). The usage examples I give in civet.md set this off.

An alternative might be to allow for the devel extension syntax to have a sensible default if the referenced extension isn't enabled. Then we'd have full doco, but the interactive page elements could be made static if demanded by the configuration. What do you think?

That alternative might risk being big enough to be a separate PR though.

python/doc/content/python/MooseDocs/extensions/civet.md
cticenhour: As an update on this one, I looked yesterday and saw that `$HOME/.local/share` is the common root location default for all of MooseDocs cache. I am exploring a new flag to set this location at the time of building, but at the end of the day, I think this display will look a bit goofy because it's pulling from the local build environment information to set it. We should probably consider this for another PR. In the meantime, I will address this immediate concern by adjusting the description to make clear we're referring to `$HOME/.local/share/civet/jobs` as the default.

cticenhour: With the most recent changes, I was able to make a minimal change to get this to present the environment variable version of this default, so this looks much better:

image

python/doc/content/python/MooseDocs/specification.md
cticenhour: This has been addressed in the more recent changes as well. A new `requires_extension` parameter was introduced for the `devel` extension commands that were breaking, which then gracefully issues a note that the extension is missing when it is not available. This still allows the rest of the documentation to be rendered for reference.

cticenhour:
All patches for failing apps have been staged:

- MALAMUTE: idaholab/malamute#223
- Mastodon: idaholab/mastodon#478
- BlueCRAB: https://github.inl.gov/ncrc/blue_crab/pull/373
- Dire Wolf: https://github.inl.gov/ncrc/dire_wolf/pull/326   ---> MERGED
- Sockeye: https://github.inl.gov/ncrc/sockeye/pull/952   ---> MERGED
- Griffin: https://github.inl.gov/ncrc/griffin/pull/3366   ---> MERGED

loganharbour (approved):

cticenhour:
All submodule updates pushed, and some approvals recieved. Will start pinging for more.

Issue #29573, opened 2024-12-18
Deprecated MooseDocs options and commands still exist after multiple years


cticenhour:
### Bug Description

Multiple MooseDocs features that were deprecated and then replaced still have remnants hanging around the code base (the `--dump` and `-generate` flags for `./moosedocs.py check` are two examples). These items are confusing and should be removed.

### Steps to Reproduce

Search for "deprecated" in the MooseDocs code base. 

### Impact

An annoyance, as there might be multiple ways to perform certain activities in MooseDocs, but only one we want to support.

PR #29581, opened 2024-12-19
Remove HTML style comments from MooseDocs

cticenhour:
These have been deprecated for multiple years. It appears as though patches will have to occur across multiple applications as well as MOOSE itself.

Refs #29573


joshuahansel:
Is the new way documented somewhere? I know I've looked for it before and didn't find it, so I just looked around in source files and found the old way.

cticenhour:
Good question. I plan to doublecheck the documentation and make sure the markdown-style method is clear as the last step before opening for review! I don't think it's very clear right now, if I'm remembering correctly.

joshuahansel (approved):

loganharbour (approved):

cticenhour:
Thanks for the reviews!

PR #29597, opened 2024-12-20
Fix the Positioning Bug of CombinerGenerator

### LINKED ISSUE #29596: The First Position in CombinerGenerator Not Applied
### Bug Description

`CombinerGenerator` provides an option to provide multiple input meshes with corresponding `positions` for translation. However, the first input mesh is not translated to the first position. Instead, it is placed at the origin.

Thank Soon Kyu Lee of ANL for reporting this bug.

### Steps to Reproduce

The example provided in the MG's [documentation](https://mooseframework.inl.gov/source/meshgenerators/CombinerGenerator.html) has this issue too. See Case 2 and the corresponding figure.

### Impact

Cannot position the first input mesh.

### [Optional] Diagnostics

_No response_

miaoyinb:
closes #29596


GiudGiud (approved):
good catch! thanks for the fix

PR #29589, opened 2024-12-20
Fix find_moose_executable utility

nmnobre:
Gotta love Python. 🐍

PR #29624, opened 2024-12-30
Regen hit.cpp w/ cython 3.0.11, remove conda/pyhit and fix find_moose_executable util

nmnobre:
So we can build against Python 3.13.


loganharbour (changes requested):
Not ready for this until conda infrastructure changes are complete

@nmnobre this will also require a bump to conda packages, as the `moose-pyhit` conda package relies on this.

nmnobre:
> Not ready for this until conda infrastructure changes are complete
> 
> @nmnobre this will also require a bump to conda packages, as the `moose-pyhit` conda package relies on this.

I'm a bit confused... I bumped the version in `conda/pyhit/meta.yaml` but as far as I can tell, that points to https://github.com/milljm/moose-pyhit.git anyway, which seems severely outdated? It both points to an old moose commit (which will of course not have the changes here to `hit.cpp` - and it's missing some previous changes too) but also doesn't seem to be up to date with `python/pyhit/pyhit.py` in this repo?

But, surely the apptainer/docker failures are unrelated to conda?

Cheers!

nmnobre:
@loganharbour All green. Thanks for the help and apologies for messing up #29589. Can I get this in before #29104? 😇

milljm:
Changing Python versions requires a change to the entire stack. You'll need to start at the beginning with `moose-mpi`:

https://github.com/idaholab/moose/blob/1e17680adfcc2117439231c38b5bb9cb3a102636/conda/mpi/meta.yaml#L100

From experience, I can tell you I had issues getting our TestHarness to work with Python 3.12. So 3.13 may be quite the project!

nmnobre:
> Edit: My mistake, I thought we were attempting to bump Python, not make something else compatible with Python 3.13...

Yes!
It's just I've got 3.13 on my machine and had to keep 3.12 around just because of MOOSE, in particular `hit.cpp`.
I tried regenerating with the latest Cython and it just worked, so that's all I need. :)

conda/pyhit/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version = "2024.05.08" %}
+{% set version = "2025.01.07" %}
```
milljm, line 1: FYI, Changing anything in conda/pyhit is no longer necessary, as PyHIT is now being built as an output package in conda/wasp. Changes in this recipe will actually be ignored by Civet.

Ideally we should remove conda/pyhit. I would support and help in its removal if you would like to make that a part of this PR?

```bash
git rm -rf conda/pyhit
```

milljm: We can also ignore my request, and I will remove it in another PR another day :)

nmnobre: No, this makes sense and explains my confusion above with conda/pyhit and https://github.com/milljm/moose-pyhit.

scripts/tests/versioner_hashes.yaml
```diff
@@ -397,3 +397,6 @@ cece3f711bd9c235af31c9d5a4c54e292400ecc1: #29074
   libmesh: b8b3d69
   wasp: 2fbcef5
   moose-dev: 5e9be02
+92ed1c1fe9080d2340bb7b564558ff38b8559a62: #29624
+  wasp: 6827a12
+  moose-dev: 12cd548
```
milljm, line 402: We need to list all libraries here. There is an argument you can supply to `./versioner.py` simplify the process:

- Commit all your changes, then
- run versioner as so: `./versioner.py --summary`

Append those results to this file, and commit those changes as the last separate commit to your PR. Keeping it separate allows for easy cherry-picking when making additional edits to your PR.

> [!TIP]
> There is also another feature of versioner.py that you may find useful:
> `./versioner.py --verify 1e17680a`
> _where `1e17680a` is an upstream hash of some kind_
> It will supply a summary of changes, and hints of invalid entries or changes required and not detected.

milljm (changes requested):
Need to address versioner.yaml file

conda/pyhit/meta.yaml
milljm: I can imagine! Plus having it be my own personal repository... gross.

My apologies not removing this when I moved PyHIT into the WASP Conda package.

milljm (approved):

scripts/tests/versioner_hashes.yaml
```diff
@@ -397,3 +397,10 @@ cece3f711bd9c235af31c9d5a4c54e292400ecc1: #29074
   libmesh: b8b3d69
   wasp: 2fbcef5
   moose-dev: 5e9be02
+52f0c08c60a3c03326972087a160fe0a0c6039f2: #29624
+  tools: 5abb5f2
+  mpi: 82bbd28
+  petsc: e78a9a4
+  libmesh: b8b3d69
+  wasp: e97f1e3
+  moose-dev: 5a18091
```
milljm, line 406: ❤️

milljm:
Fenix documentation test failures are unrelated. Merging...

loganharbour (approved):

PR #29635, opened 2025-01-02
Add mfem build to moose-dev container

loganharbour:


scripts/update_and_rebuild_mfem.sh
```diff
@@ -0,0 +1,125 @@
...
+fi
+
+CONDUIT_BUILD_DIR="${CONDUIT_SRC_DIR}/build"
+LIBMESH_DIR=${LIBMESH_DIR:-$(realpath -m "${SCRIPT_DIR}/../libmesh/installed")}
+if [ -z "$PETSC_DIR" ]; then
+  PETSC_DIR=$(realpath -m "${SCRIPT_DIR}/../petsc/arch-moose")
+fi
+HDF5_DIR=${HDF5_DIR:-$PETSC_DIR}
+HYPRE_DIR=${HYPRE_DIR:-$PETSC_DIR}
+METIS_DIR=${METIS_DIR:-$PETSC_DIR}
```
lindsayad, line 70: For integration in MOOSE I don't think we should make these settable by the user. We don't want different versions of HDF5 or hypre or metis being used than what are used by PETSc

scripts/update_and_rebuild_mfem.sh
```diff
@@ -0,0 +1,125 @@
...
+    -DSuperLUDist_DIR="$HYPRE_DIR" \
+    -DBUILD_SHARED_LIBS=ON \
+    -DHDF5_DIR="$HDF5_DIR" \
+    -DBLAS_DIR="$HYPRE_DIR" \
+    -DMFEM_ENABLE_EXAMPLES=yes \
+    -DMFEM_ENABLE_MINIAPPS=yes \
+    -DBLAS_LIBRARIES="$HYPRE_DIR/lib/libfblas.a" \
+    -DLAPACK_LIBRARIES="$HYPRE_DIR/lib/libflapack.a" \
+    -DBLAS_INCLUDE_DIRS="$HYPRE_DIR/include" \
+    -DLAPACK_INCLUDE_DIRS="$HYPRE_DIR/include" \
```
lindsayad, line 113: If not making HYPRE/METIS/HDF5_DIR configurable, we can just set everything with the PETSc variables

loganharbour: Removed

scripts/update_and_rebuild_mfem.sh
```diff
@@ -0,0 +1,121 @@
...
+make install
+
+# Build and install mfem
+mkdir -p "$MFEM_BUILD_DIR"
+cd "$MFEM_BUILD_DIR"
+cmake .. \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=YES \
+    -DMFEM_USE_OPENMP=NO \
+    -DMFEM_THREAD_SAFE=NO \
+    -DHYPRE_DIR="$PETSC_DIR" \
```
lindsayad, line 91: This won't work if someone is using the classic definitions of `PETSC_DIR` and `PETSC_ARCH` because the installation directory will then be `$PETSC_DIR/$PETSC_ARCH`. It might be useful to create a `PETSC_INSTALL_DIR` that is `$PETSC_DIR` when `PETSC_ARCH` is not set and `$PETSC_DIR/$PETSC_ARCH` when `PETSC_ARCH` is set. Then you can sub that for hypre, metis, parmetis, hdf5, blas, etc. The `-DPETSC_DIR=$PETSC_DIR` and `-DPETSC_ARCH=$PETSC_ARCH` would remain unchanged.

Also it seems like things are working fine with `-DPETSC_ARCH=$PETSC_ARCH` when `PETSC_ARCH` is empty, but it would feel a little cleaner to only add that argument to `cmake` if it is non-empty

PR #26878, opened 2024-02-23
Command line rewrite

loganharbour:
Closes #29058

- Limits the use of command line arguments with optional values
- Adds global and non-global cli arguments, global being those that propagate to multiapps
- Simplifies how cli arguments are passed to multiapps
- Adds input file(s) and cli args/hit cli args to the header

Requires a patch in griffin: https://github.inl.gov/ncrc/griffin

PR #29650, opened 2025-01-06
Fix (sub)subapp command line args

nmnobre:
Hi Logan (@loganharbour),

Fixes a couple of bugs introduced in 4f72723 of #26878:

1) In `initSubAppCommandLine()`, the use of `std::next()` meant we were losing an argument with each spawned subapp. This is okay for the first subapp because it only skipped the executable invocation (which I think was the reason it was there? - but note this is unnecessary because for that entry, `global == false`) but then meant that if the subapp created a further subapp, the latter would lose an argument.
2) Similarly, when printing in `outputFrameworkInformation`, the use of `std::next()` meant we weren't printing the entire argument list. We now only skip with `std::next()` on the main app, so we don't print the executable invocation line.

Cheers,
-Nuno

### DIFF
--- a/framework/src/outputs/ConsoleUtils.C
+++ b/framework/src/outputs/ConsoleUtils.C
@@ -57,7 +57,8 @@ outputFrameworkInformation(const MooseApp & app)
   oss << "\n";
 
   const auto & cl = std::as_const(*app.commandLine());
-  const auto cl_range = as_range(std::next(cl.getEntries().begin()), cl.getEntries().end());
+  const auto cl_range =
+      as_range(std::next(cl.getEntries().begin(), app.multiAppLevel() == 0), cl.getEntries().end());
 
   std::stringstream args_oss;
   for (const auto & entry : cl_range)
--- a/framework/src/outputs/ConsoleUtils.C
+++ b/framework/src/outputs/ConsoleUtils.C
@@ -57,6 +57,7 @@ outputFrameworkInformation(const MooseApp & app)
   oss << "\n";
 
   const auto & cl = std::as_const(*app.commandLine());
+  // We skip the 0th argument of the main app, i.e., the name used to invoke the program
   const auto cl_range =
       as_range(std::next(cl.getEntries().begin(), app.multiAppLevel() == 0), cl.getEntries().end());
 
--- a/framework/src/parser/CommandLine.C
+++ b/framework/src/parser/CommandLine.C
@@ -211,6 +211,8 @@ CommandLine::initSubAppCommandLine(const std::string & multiapp_name,
     subapp_args.push_back(MooseUtils::removeExtraWhitespace(arg));
 
   // Pull out all of the arguments that are relevant to this multiapp from the parent
+  // Note that the 0th argument of the main app, i.e., the name used to invoke the program,
+  // is neither a global entry nor a subapp entry and, as such, won't be passed on
   for (auto & entry : as_range(getEntries().begin(), getEntries().end()))
     if (entry.global || (entry.subapp_name && (*entry.subapp_name == multiapp_name ||
                                                *entry.subapp_name == subapp_name)))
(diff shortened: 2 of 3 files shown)


loganharbour:
@nmnobre I actually have some pretty rigorous unit tests for everything in `CommandLine` in `CommandLineTest.C`. 

Would you mind creating a test there so you can show the case you're trying to fix with this?

nmnobre:
> Would you mind creating a test there so you can show the case you're trying to fix with this?

Done. :)
We don't test the output utility anywhere do we?

loganharbour:
> > Would you mind creating a test there so you can show the case you're trying to fix with this?
> 
> Done. :) We don't test the output utility anywhere do we?

Nope, I don't think we do. But `getEntries()` is tested pretty well which I think is the significant piece.

Should we just remove the executable path within `getEntries()` all together? That might be simpler and then we don't any special logic?

tophmatthews:
At the risk of mixing problems, I think the first command line gets drop for just one subapp (instead of sub-subapp problems you're trying to fix here.) For example, I should have two command line arguments (and indeed two are actually being passed), but the header only shows one.
```
sub000: Input File(s):
sub000:   /Users/cmatthews/projects/centipede/problems/brouwer/sub.i
sub000: 
sub000: Command Line Input Argument(s): 
***THERE SHOULD BE ANOTHER cli_arg HERE, BUT ISN'T***
sub000:   po2_input=2.1149999999999998e-09
sub000: 
sub000: Parallelism:
sub000:   Num Processors:          1
sub000:   Num Threads:             1
```
Does this fix this problem or is it a separate issue?

nmnobre:
> Should we just remove the executable path within getEntries() all together?

Probably, but maybe you had a good reason to do this?

https://github.com/idaholab/moose/blob/1e17680adfcc2117439231c38b5bb9cb3a102636/framework/src/parser/CommandLine.C#L143-L149

If you change that, I *think* we'd only need to revert this as well(?):

https://github.com/idaholab/moose/blob/1e17680adfcc2117439231c38b5bb9cb3a102636/framework/src/parser/CommandLine.C#L368-L376

nmnobre:
> Does this fix this problem or is it a separate issue?

This is the output bug I mention in bullet point (2) above and this PR fixes it. :)
As you noted the subapp actually takes in the argument, so it's a "soft" bug in that case.
For subsubapps we are _actually_ losing arguments though, that's bullet point (1).

loganharbour (approved):
LGTM. I'm fine with it like this. Unit test picks up what we were missing.

@nmnobre if you have a strong desire to make `getEntries` a little more straight forward, I won't object either!

nmnobre:
> @nmnobre if you have a strong desire to make `getEntries` a little more straight forward, I won't object either!

No, because I actually think what you did here:

https://github.com/idaholab/moose/blob/1e17680adfcc2117439231c38b5bb9cb3a102636/framework/src/parser/CommandLine.C#L143-L149

Is indeed the right thing to do. The main app's 0th argument, i.e. the name used to invoke the program, is a command line entry; so conceptually, I think it belongs with the other entries. But we perhaps can add a couple of comments to make this clearer in the code. Hang on.

loganharbour (approved):
Ah yeah, so the next isn't needed the second time because it's never hit. Works for me!

I'm honestly glad that you were able to follow `CommandLine`. Trying to get the command line behavior to be more well defined was a pain and produced some not so beautiful logic :D

nmnobre:
> Ah yeah, so the next isn't needed the second time because it's never hit. Works for me!

Yep, `std::next` isn't needed beyond the main app because we drop the program's name from the first subapp level anyway.

> I'm honestly glad that you were able to follow `CommandLine`. Trying to get the command line behavior to be more well defined was a pain and produced some not so beautiful logic :D

gdb helped I'll be honest, but it looks neater.

Issue #29662, opened 2025-01-08
use `_console` for app header


tophmatthews:
### Motivation

When running multiapps, a nice color indentation appears for outputs from the sub apps that looks like:
```
Time Step 0, time = 0

Time Step 1, time = 0.25, dt = 0.25
sub0: 
sub0: Time Step 1, time = 0.25, dt = 0.25
sub0_sub0: 
sub0_sub0: Time Step 1, time = 0.25, dt = 0.25
```

Some subapps have headers defined in `SomeApp::header() const`. The current implementation of the header call in `MooseApp::setupOptions()` gets close by coloring the header and adding `:`, but misses out on the subapp:

```
// Print the header, this is as early as possible
  auto hdr = header();
  if (hdr.length() != 0)
  {
    if (multiAppLevel() > 0)
      MooseUtils::indentMessage(_name, hdr);
    Moose::out << hdr << std::endl;
  }
```

Sometimes there are things in that header that are worth seeing for each sub app instance, so having `sub0:` output before the header would be nice.

Even other times, it would be nice to suppress it all together.

### Design

Change the header output to simply
```
// Print the header, this is as early as possible
  _console << header() << std::endl;
```

Second, use command line options to provide a flag to suppress output altogether.

### Impact

More consistant multiapps.

pbehne:
> The current implementation of the header call in MooseApp::setupOptions() gets close by coloring the header and adding :, but misses out on the subapp:

> Sometimes there are things in that header that are worth seeing for each sub app instance, so having sub0: output before the header would be nice.

I don't fully understand these statements. Could you please elaborate?

tophmatthews:
Sure. It's basically this:
```
sub009: Centipede version:  git commit da4a829 on 2024-12-17
sub009: Units:              nanometer, second, kelvin, eV
sub009: Command line:       some_value=2.9276796160740592
sub009: 
sub009: The following total 20 aux variables:
sub009:   C_Oi
```

vs this:
```
: Centipede version:  git commit da4a829 on 2024-12-17
: Units:              nanometer, second, kelvin, eV
: Command line:       some_value=2.9276796160740592
: 
: The following total 20 aux variables:
```

tophmatthews:
Since `some_value` might be different depending on which app, I would like to be able to correlate header to subapp number.

PR #29663, opened 2025-01-08
Header output

tophmatthews:
Changes header output to `_console` stream. In order to test this, I needed an app to actually have a header out output. I added one to the `MooseTestApp`, hopefully that is okay.

ref #29662

### DIFF
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -730,13 +730,7 @@ MooseApp::setupOptions()
   TIME_SECTION("setupOptions", 5, "Setting Up Options");
 
   // Print the header, this is as early as possible
-  auto hdr = header();
-  if (hdr.length() != 0)
-  {
-    if (multiAppLevel() > 0)
-      MooseUtils::indentMessage(_name, hdr);
-    Moose::out << hdr << std::endl;
-  }
+  _console << header() << std::endl;
 
   if (getParam<bool>("error_unused"))
     setCheckUnusedFlag(true);
--- a/test/include/base/MooseTestApp.h
+++ b/test/include/base/MooseTestApp.h
@@ -25,4 +25,6 @@ public:
 
   static void registerAll(Factory & f, ActionFactory & af, Syntax & s, bool use_test_objs = false);
   static void registerApps();
+
+  virtual std::string header() const override;
 };
--- a/test/src/base/MooseTestApp.C
+++ b/test/src/base/MooseTestApp.C
@@ -141,6 +141,12 @@ MooseTestApp::registerApps()
   registerApp(MooseTestApp);
 }
 
+std::string
+MooseTestApp::header() const
+{
+  return std::string("MOOSE Test App\n");
+}
+
 extern "C" void
 MooseTestApp__registerAll(Factory & f, ActionFactory & af, Syntax & s)
 {
--- a/test/tests/multiapps/multilevel/dt_from_parent_parent.i
+++ b/test/tests/multiapps/multilevel/dt_from_parent_parent.i
@@ -6,30 +6,30 @@
 []
 
 [Variables]
-  [./u]
-  [../]
+  [u]
+  []
 []
 
 [Kernels]
-  [./diff]
+  [diff]
     type = Diffusion
     variable = u
-  [../]
+  []
 []
 
 [BCs]
-  [./left]
+  [left]
     type = DirichletBC
     variable = u
     boundary = left
     value = 0
-  [../]
-  [./right]
+  []
+  [right]
     type = DirichletBC
     variable = u
     boundary = right
     value = 1
-  [../]
+  []
 []
 
 [Executioner]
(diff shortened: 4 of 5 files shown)


pbehne:
In the screenshot, the console on the left contains test results before the proposed changes and the console on the right contains test results after the proposed changes. Did you intend to lose the color in `sub0:`?
image

tophmatthews:
Yeah, I had to drop the color via the input file in order to make the regexp work in the test. The difference addition here is the "Moose Test App" header and it's prefixes.

framework/src/base/MooseApp.C
```diff
@@ -730,13 +730,7 @@ MooseApp::setupOptions()
   TIME_SECTION("setupOptions", 5, "Setting Up Options");
 
   // Print the header, this is as early as possible
-  auto hdr = header();
-  if (hdr.length() != 0)
-  {
-    if (multiAppLevel() > 0)
-      MooseUtils::indentMessage(_name, hdr);
```
pbehne, line 737: I was confused as to why this line wasn't incorporating the app name into the message. So I investigated and it turns out that `_name` = "". So I investigated more and found a bug. Could you please pull #29671 and see if the headers behave as you want?

tophmatthews: Yup! That took care of that problem, thanks! Is there a reason for all the complications with `indentMessage`? Seems like `_console` takes care of it altogether.

tophmatthews: I guess the difference is slight, for example this is your newest fix:
image

tophmatthews: And using `_console` tracks that extra new line:
```
MOOSE Test App

sub0: MOOSE Test App
sub0: 
sub0_sub0: MOOSE Test App
sub0_sub0: 
```

tophmatthews: This is because
```
std::string
MooseTestApp::header() const
{
  return std::string("MOOSE Test App\n");
}
```

tophmatthews: I think it comes down to if you want the header in `_console` or `Moose::out`?

pbehne: > Yup! That took care of that problem, thanks! Is there a reason for all the complications with `indentMessage`? Seems like `_console` takes care of it altogether.

The issue was that `indentMessage` was taking `_name` as a parameter for the  message prefix. `_name` was not behaving as expecting due to it being redeclared, but not explicitly initialized in `MooseApp`. `console`, on the other hand, uses the method `name()`, which returns the expected `MooseBase::_name` attribute. The solution was to remove the declaration of `MooseApp::_name` and let it used the inherited `_name` attribute from `MooseBase`.

tophmatthews: Yes, I get that, and I get that the fix in #29671 is good no matter what. I suppose my I'm getting at is `Moose::out` the correct output stream for headers, or is `_console` the right way. Certainly `_console` is much cleaner in the code, and captures _all_ of the header, including whitespace.

pbehne: > I think it comes down to if you want the header in `_console` or `Moose::out`?

After speaking with @permcody, it seems our official stance is to prefer `_console` over `Moose::out`. I would be fine with your changes, just re-add the check `if (hdr.length() != 0)` so we don't print empty lines.

Also, I do think that polluting all console outputs with a bunch of "MOOSE Test App" is not worth having the test for these changes, so I would also get rid of the virtual `header` method in `MooseTestApp` and drop the test.

As a final caveat, I would also consider how setting `Outputs/console=false` would affect the output as opposed to using `Moose::out` and if we are willing to accept that difference (if it exists).

So in conclusion, if you want to change `Moose::out` to `_console`, I'm fine with that. If you don't want to bother and are happy with my existing fix, I'm also fine with that. Just let me know.

tophmatthews: > seems our official stance is to prefer

Yeah, that's what I seemed to remember...

>  not worth having the test

I suppose so...but my next PR is going to suppress headers all together via a command line option #29669. It seems somewhat anti-moose to not have tests...but up to you all.

> I would also consider how setting Outputs/console=false would affect the output as opposed to using Moose::out

`Outputs/console=false` with `_console` and `Moose::out` suppresses output from the parent app, but not the header. I don't see a difference. Do you anticipate one?

tophmatthews:
I'm just going to close this and use `_console` in #29669 since I'm getting rid of tests.

framework/src/base/MooseApp.C
pbehne: > Do you anticipate one?

I didn't know what to anticipate, it was just something to check.

pbehne: > I suppose so...but my next PR is going to suppress headers all together via a command line option https://github.com/idaholab/moose/pull/29669

That one will definitely need test(s). If another reviewer tells you that you need tests related to something like this, please tag me, as I just got review authority and am still learning.

PR #29669, opened 2025-01-09
Add cmd line arg to suppress header output

### LINKED ISSUE #29662: use `_console` for app header

tophmatthews:
Combined #29663 into this. Changed the header output stream to `_console`, and added command line arg to suppress header output. Added command line arg to `MooseTestApp` to optionally print a header to allow testing.

Closes #29662

### DIFF
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -247,8 +247,10 @@ MooseApp::validParams()
       "Continue the calculation. Without <file base>, the most recent recovery file will be used");
   params.setGlobalCommandLineParam("recover");
 
-  params.addCommandLineParam<bool>(
-      "suppress_header", "--suppress-header", false, "Flag to print the App header");
+  params.addCommandLineParam<bool>("suppress_header",
+                                   "--suppress-header",
+                                   false,
+                                   "Disables the output of the application header.");
   params.setGlobalCommandLineParam("suppress_header");
 
   params.addCommandLineParam<bool>(
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -248,6 +248,10 @@ MooseApp::validParams()
   params.setGlobalCommandLineParam("recover");
 
   params.addCommandLineParam<bool>(
+      "suppress_header", "--suppress-header", false, "Flag to print the App header");
+  params.setGlobalCommandLineParam("suppress_header");
+
+  params.addCommandLineParam<bool>(
       "test_checkpoint_half_transient",
       "--test-checkpoint-half-transient",
       "Run half of a transient with checkpoints enabled; used by the TestHarness");
@@ -730,13 +734,8 @@ MooseApp::setupOptions()
   TIME_SECTION("setupOptions", 5, "Setting Up Options");
 
   // Print the header, this is as early as possible
-  auto hdr = header();
-  if (hdr.length() != 0)
-  {
-    if (multiAppLevel() > 0)
-      MooseUtils::indentMessage(_name, hdr);
-    Moose::out << hdr << std::endl;
-  }
+  if (header().length() && !getParam<bool>("suppress_header"))
+    _console << header() << std::endl;
 
   if (getParam<bool>("error_unused"))
     setCheckUnusedFlag(true);
--- a/test/src/base/MooseTestApp.C
+++ b/test/src/base/MooseTestApp.C
@@ -52,6 +52,9 @@ MooseTestApp::validParams()
       "--test-check-legacy-params",
       "Check for legacy parameter construction with CheckLegacyParamsAction; for testing");
 
+  params.addCommandLineParam<std::string>(
+      "append_header", "--append-header <header>", "", "String to print at top of console output");
+
   params.set<bool>("automatic_automatic_scaling") = false;
   params.set<bool>("use_legacy_material_output") = false;
   params.set<bool>("use_legacy_initial_residual_evaluation_behavior") = false;
@@ -139,6 +142,12 @@ void
 MooseTestApp::registerApps()
 {
   registerApp(MooseTestApp);
+}
+
+std::string
+MooseTestApp::header() const
+{
+  return getParam<std::string>("append_header");
 }
 
 extern "C" void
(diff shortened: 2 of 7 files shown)


framework/src/base/MooseApp.C
```diff
@@ -247,6 +247,10 @@ MooseApp::validParams()
       "Continue the calculation. Without <file base>, the most recent recovery file will be used");
   params.setGlobalCommandLineParam("recover");
 
+  params.addCommandLineParam<bool>(
+      "suppress_header", "--suppress-header", "Flag to print the App header");
+  params.setGlobalCommandLineParam("suppress_header");
```
tophmatthews, line 252: Ugh, this comment got stuck in "review", sorry it didn't go out...anyways, heres the command line and old comment...

@loganharbour This PR builds on #29663 , so it's a bit convoluted with that update, but here is the cmd line I'm trying to add. I would like to be able to optionally suppress the header being printed. I tried to use `CommonOutputAction`, similar to #29665, but it's not built at this point (I think?).

tophmatthews:
Can someone pass judgement on including a test on this? The simplest is a `header()` in `MooseTestApp` like https://github.com/idaholab/moose/pull/29669/commits/9e54f637e5a2ccc0d6c3a3abd1db5c75539f976d

loganharbour:
> Can someone pass judgement on including a test on this @GiudGiud @loganharbour @pbehne ? The simplest is a `header()` in `MooseTestApp` like [9e54f63](https://github.com/idaholab/moose/commit/9e54f637e5a2ccc0d6c3a3abd1db5c75539f976d)

I think we should just make another dummy app for it so that we don't get a header in every moose_test-opt run

tophmatthews:
Could also put a header in `MiscApp`

loganharbour:
> Could also put a header in `MiscApp`

I don't like testing framework content outside of the framework. misc testing doesn't contribute to framework coverage for example.

If we don't want another app, we can just do something silly like `--test-set-header="foo"` in `MooseTestApp` and then have `header()` return that result. So the header only shows up when we set it to. Goofy cause we'll basically have

```
--suppress-header --test-set-header="foo"
```

but we've done worse.

tophmatthews:
Pretty easy to put 
```
  params.addCommandLineParam<std::string>(
      "append_header", "--append_header <header>", "String to print at top of console output");
```
in `MooseApp.C` too for someone that may want that optoin...any reason to or just leave it in `MooseTestApp.C`?

tophmatthews:
This works for cascading to subapps @loganharbour ...maybe that previous fix you suggested _did_ work 😬

tophmatthews:
Alright, ready for review!

YaqiWang:
Does this https://github.com/idaholab/moose/pull/29671 fix the missing prefix in the indentation reported in the issue?

tophmatthews:
> Does this #29671 fix the missing prefix in the indentation reported in the issue?

Yeah mostly, but this still adds the capability to turn off the header.

This is good to go for review @loganharbour . Thanks!

framework/src/base/MooseApp.C
```diff
@@ -247,6 +247,10 @@ MooseApp::validParams()
       "Continue the calculation. Without <file base>, the most recent recovery file will be used");
   params.setGlobalCommandLineParam("recover");
 
+  params.addCommandLineParam<bool>(
+      "suppress_header", "--suppress-header", false, "Flag to print the App header");
```
loganharbour, line 251:
```suggestion
  params.addCommandLineParam<bool>(
      "suppress_header", "--suppress-header", "Disables the output of the application header.");
```

test/src/base/MooseTestApp.C
```diff
@@ -52,6 +52,9 @@ MooseTestApp::validParams()
       "--test-check-legacy-params",
       "Check for legacy parameter construction with CheckLegacyParamsAction; for testing");
 
+  params.addCommandLineParam<std::string>(
+      "append_header", "--append-header <header>", "", "String to print at top of console output");
```
loganharbour, line 56:
```suggestion
  params.addCommandLineParam<std::string>(
      "append_header", "--append-header <header>", "String to print at top of console output");
```

test/src/base/MooseTestApp.C
```diff
@@ -141,6 +144,12 @@ MooseTestApp::registerApps()
   registerApp(MooseTestApp);
 }
 
+std::string
+MooseTestApp::header() const
+{
+  return getParam<std::string>("append_header");
```
loganharbour, line 150:
```suggestion
  if (isParamSetByUser("append_header"))
    return getParam<std::string>("append_header");
```

loganharbour (changes requested):

test/src/base/MooseTestApp.C
tophmatthews: Still need to return something. I thought the default was an appropriate way for this?

tophmatthews: Or just `return "";` on a new line.

loganharbour (approved):

tophmatthews:
Thanks @loganharbour !

PR #28114, opened 2024-07-11
End-to-end CUDA container, remove peacock, bump python to 3.13

loganharbour:
- Removes the peacock conda package
- Bumps default python to 3.13, also in apptainer
- Updates python source for 3.13 compatibility 
- Bumps apptainer clang to 19
- Bumps apptainer min gcc to 9
- Supports a full stack CUDA build (mpi and on)
- Builds pytorch within all apptainer builds but min gcc (pytrorch bump to 2.6)
- Updates base images for all apptainer builds (latest rocky 8)
- Updates apptainer openmpi to 5.0.7
- Updates moose-language-server extension in moose-dev apptainer

Closes https://github.com/idaholab/moose/issues/29374 (adds full cuda build)
Closes https://github.com/idaholab/moose/issues/28161 (removes moose-peacock from moose-dev)
Closes https://github.com/idaholab/moose/issues/30382 (removes moose-peacock)
Closes https://github.com/idaholab/moose/issues/30586 (removes extra vtk build that comes from moose-peaacock; confirmed by @hugary1995)

PR #29690, opened 2025-01-14
[WIP] Compiled pytorch within MOOSE-dev container.

### LINKED ISSUE #29661: Add natively compiled torch to moose-dev container.
### Motivation

We run into GLIBC issues due to lagged compilers when fetching precompiled libtorch libraries. Additionally, would be nice to have a consistent pytorch package in there as well.  

### Design

Follow installation instructions and try to fix packages in the pytorch requirements.

### Impact

Will hopefully fix the threading issues we see in containers, and remove the glibc limitations.

grmnptr:
Closes #29661


apptainer/moose-dev.def
```diff
@@ -81,7 +81,8 @@ Fingerprints: 0CFFCAB55E806363601C442D211817B01E0911DB
 
     # Pinned versions
     MINIFORGE_VERSION=23.3.1-1
-    PYTHON_VERSION=3.11
+    PYTHON_VERSION={{ MOOSE_PYTHON_VERSION }}
+    echo $PYTHON_VERSION
```
loganharbour, line 85:
```suggestion
```

apptainer/moose-dev.def
```diff
@@ -156,9 +140,47 @@ Fingerprints: 0CFFCAB55E806363601C442D211817B01E0911DB
...
+    # Fetch the patch that enables the successful compilation on this linux distro
+    curl -L https://github.com/pytorch/pytorch/pull/99468.patch > 99468.patch
+    git apply 99468.patch
+
+    # We have to diasble some errors considering they emerge as some optimization of the compiler on certain distributuins (sadly on this one to)
+    export CFLAGS+=" -Wno-error=maybe-uninitialized -Wno-error=uninitialized -Wno-error=restrict"
+    export CXXFLAGS+=" -Wno-error=maybe-uninitialized -Wno-error=uninitialized -Wno-error=restrict"
+
+    # Build and install
+    export CMAKE_PREFIX_PATH="${CONDA_PREFIX:-'$(dirname $(which conda))/../'}:${CMAKE_PREFIX_PATH}"
```
loganharbour, line 170: Installing something like this directly into CONDA_PREFIX isn't great, because conda doesn't know about it. Is this what they suggest you do?

loganharbour (changes requested):

apptainer/moose-dev.def
grmnptr: Yeah these are the official instructions. So you are suggesting installing this somewhere else and adding it to python path?

hugary1995: I don't think `CMAKE_PREFIX_PATH` has anything to do with "where to install the compiled libtorch libraries". Instead, it is simply used as a hint at CMake configure time.

grmnptr: This PR will be closed, Logan took over this effort. The conclusion was also that the user will not get pytorch from the container, we will only build the libtorch part.

grmnptr:
Reopening this to add this functionality to the CPU containers. The CUDA container will have additional functionality and it is actively being worked on.

grmnptr:
Closing this again, considering https://github.com/idaholab/moose/pull/28114 is almost there.

PR #29706, opened 2025-01-18
Allow material property names to be used in ElementMaterialSampler

### LINKED ISSUE #29705: Allow ElementMaterialSampler to accept material property names
### Motivation

Currently ElementMaterialSampler only takes a single Material name as an input and returns all its declared properties. Need a way handling multiple Materials.

### Design

I would like to specify the material properties directly.

### Impact

The object would take either a Material name or a property name(s).

I have a branch for this and will put up a PR shortly

jessecarterMOOSE:
closes #29705

## Reason
Need a way of sampling properties from multiple Materials

## Design
Allow the names of the property names to be specified directly

## Impact
New feature


jessecarterMOOSE:
@GiudGiud this is what we talked about in #29705 29695. Not sure if there's a good way to pre-check the properties are defined when they are specified directly.

framework/src/vectorpostprocessors/ElementMaterialSampler.C
```diff
@@ -44,35 +46,62 @@ ElementMaterialSampler::ElementMaterialSampler(const InputParameters & parameter
...
+      mooseError(name(), ": boundary materials (i.e. ", mat.name(), ") cannot be used");
 
-    // check requested materials are available
-    for (const auto & id : ids)
+    // Get property names from the Material
+    auto & props = mat.getSuppliedItems(); // returns std::set
+    prop_names = std::vector<MaterialName>(props.begin(), props.end());
+
+    // Check requested materials are available
+    if (_elem_filter)
```
GiudGiud, line 77: I would move this check down a bit for clarity. 

And in the case where we only know the property, you can just poll every single material in the problem to find if it defines a property. If you do it once before the element loop it should be fine. 

It could be good also to turns the element loop into a simple gathering of the block ids. Then once you have those in a set you can test the materials outside of an element loop

test/tests/vectorpostprocessors/element_material_sampler/tests
```diff
@@ -24,9 +24,32 @@
...
+                  'points on all active elements when supplied with the property names defined on different blocks.'
+  []
+  [material_properties_multiple_blocks_elem_ids]
+    type = 'CSVDiff'
+    input = 'property_sampler_multiple_blocks_elem_ids.i'
+    csvdiff = 'property_sampler_multiple_blocks_elem_ids_out_props_0001.csv property_sampler_multiple_blocks_elem_ids_out_props_0002.csv'
+
+    requirement = 'The system shall be able to output material properties calculated at quadrature '
+                  'points on the given elements when supplied with the property names defined on different blocks.'
   []
```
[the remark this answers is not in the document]
jessecarterMOOSE: Do you mean to add a test that purposefully fails? Maybe prop1 only exists on block 1 and prop2 only on block 2?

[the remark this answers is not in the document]
GiudGiud: Yes to make sure we fail as expected

framework/src/vectorpostprocessors/ElementMaterialSampler.C
jessecarterMOOSE: > poll every single material in the problem to find if it defines a property

How do you do that?

> gathering of the block ids

Not sure how to do that either.

jessecarterMOOSE:
Also my tests are diff-ing with distributed mesh. What could be causing that?

GiudGiud:
The samplers sort their output by id and the ids are not always the same with distrivuted meshes

You can set allow_renumbering to false

jessecarterMOOSE:
Let's see if this gets tests to pass, then we can address the checking logic

jessecarterMOOSE:
@GiudGiud the tests pass now. I'd be happy to add more to the initialization checks if you can give me some more detail on how that should be done.

test/tests/vectorpostprocessors/element_material_sampler/tests
```diff
@@ -53,7 +76,16 @@
...
 
-      detail = 'the material does not exist on the mesh.'
+      detail = 'the material does not exist on the mesh, or'
+    []
+
+    [block_restrict_property_sample]
+      type = 'RunException'
+      input = 'property_sampler_multiple_blocks.i'
+      cli_args = ' Materials/mat1/prop_names=prop1 Materials/mat1/prop_values=1 Materials/mat2/prop_names=prop2 Materials/mat2/prop_values=2 VectorPostprocessors/props/property=prop1'
+      expect_err = "Material property 'prop1', requested by 'props' is not defined on block RIGHT_SIDE"
```
GiudGiud, line 86: If we hit this error, this might be all we need. I ll check this out locally and try it out.

framework/src/vectorpostprocessors/ElementMaterialSampler.C
```diff
@@ -44,35 +46,62 @@ ElementMaterialSampler::ElementMaterialSampler(const InputParameters & parameter
...
     _z_coords(declareVector("z"))
 {
-  auto & mat = getMaterialByName(getParam<MaterialName>("material"), true);
-  auto & prop_names = mat.getSuppliedItems();
-  if (mat.isBoundaryMaterial())
-    mooseError(name(), ": boundary materials (i.e. ", mat.name(), ") cannot be used");
+  // Check either "material" or "property" was set but not both
+  if (parameters.isParamSetByUser("material") && parameters.isParamSetByUser("property"))
+    mooseError("Setting both 'material' and 'property' is not allowed. Use one or the other.");
+  if (!parameters.isParamSetByUser("material") && !parameters.isParamSetByUser("property"))
```
GiudGiud, line 52:
```suggestion
  if (parameters.isParamValid("material") && parameters.isParamValid("property"))
    mooseError("Setting both 'material' and 'property' is not allowed. Use one or the other.");
  if (!parameters.isParamValid("material") && !parameters.isParamValid("property"))
```

framework/src/vectorpostprocessors/ElementMaterialSampler.C
GiudGiud: this check still needs to be used in both cases. Maybe dont share the code, just have two different checks for the material and material property cases

instead of that
```
And in the case where we only know the property, you can just poll every single material in the problem to find if it defines a property. If you do it once before the element loop it should be fine.
```
use this, that should be a lot easier
```
  /**
   * Get a vector containing the block ids the material property is defined on.
   */
  virtual std::set<SubdomainID> getMaterialPropertyBlocks(const std::string & prop_name);
```

GiudGiud: nevermind. This is covered

GiudGiud (changes requested):
let's squash this last replcament into the previous commits then we're good

framework/src/vectorpostprocessors/ElementMaterialSampler.C
jessecarterMOOSE: done

jessecarterMOOSE:
@GiudGiud what would it take to extend this object to include vector and tensor material properties? I'm thinking if you ask for a vector property `foo`, it makes columns like `foo0,foo1,foo2,...`. Similar for a tensor. Would this be the place for that or would you make a new class?

GiudGiud:
Either new classes for each type, but preferably a template

jessecarterMOOSE:
Inherit from this one? Otherwise most of the methods would be mostly duplicated. But the constructor of this one isn't compatible with vector/tensor. Maybe the common stuff should go in a base class?

GiudGiud:
Much preferrably actually. We should go even further than what was done for the auxkernels

GiudGiud:
That's where templating works better imo. 

There are dozens of types one would need to support. Real vector array tensor eigen-dense matrix, then the AD versions

jessecarterMOOSE:
Yikes that's a lot. In the meantime, what's with these failures? They look unrelated

GiudGiud:
unrelated

GiudGiud (approved):

jessecarterMOOSE:
Thanks 👍

PR #29773, opened 2025-01-31
Take strain_free_density as initial value for density calculation

### LINKED ISSUE #29556: Allow density to change via strain-free chemistry changes
### Motivation

In some material systems, density can change due to chemistry changes as well as thermal expansion. The material property `Density` currently takes an initial density, given as a input param, and modifies it as a result of any displacements throughout the simulation. This limits density changes to only strain-driven (e.g., thermal expansion) thermal changes.

### Design

Turn:

```params.addRequiredParam<Real>("density", "Density");```

into:

```params.addParam<MaterialPropertyName>("density, "Material property name for strain-free density."```

Really, I would like to turn this into:

```params.addParam<MaterialPropertyName>("strain_free_density, "Material property name for strain-free density."```

and deprecate `density` input parameter altogether, but would take buy in from people like @bwspenc @dschwen 

### Impact

More flexible `Density` material class.

### PR DESCRIPTION
**Pull request by tophmatthews**
Creates new `MaterialPropertyName` `strain_free_density` that will replace the now deprecated `density` param. This enables the use of a material property for density independent of strains (say for example, dependent on chemistry), which can be changed throughout a simulation. Added test for `Density` in `modules/misc/test/tests` as well since `Density` is in the `misc` module, but is only tested in `combined` (almost seems like density should be in `solid_mechanics` but whatevs).

ping maybe @bwspenc and @dschwen ?

Closes #29556

### CODE REVIEW

**Comment by tophmatthews** [comment]
Ready for you to take a look @bwspenc thanks!

**Comment by bwspenc** [comment]
@tophmatthews I think it's in `misc` because it's used by the `heat_transfer` module as well as mechanics.

**Review by bwspenc — COMMENTED** [review]
This looks good to me. My only comment is that to me, it seems like it would be more natural to name the property name and input parameter `initial_density` rather than`strain_free_density`.  Since that's how you chose to name your member variable that points to the input parameter I think you might feel the same way.

**Comment by tophmatthews** [comment]
> I think it's in `misc` because it's used by the `heat_transfer` module as well as mechanics.

Hm, I thought I removed a lot of "Density" at some point. Could just be a `GenericConstantMaterial` anyways. I'll look into this.

**Comment by bwspenc** [comment]
I guess it's up to the user to use whatever material model is most appropriate to define `density`, so they could use either this or `GenericConstantMaterial`.

**Comment by tophmatthews** [comment]
> input parameter `initial_density` rather than`strain_free_density`.

Yeah good suggestion. The bare minimum need for `Density` is an initial density. The use case added here is for evolving density due to everything but strain (e.g., changing chemistry conditions that changes density). Since `strain_free_density` kinda captures both, while using `initial_density` implies it's only the original, I went with `strain_free_density`. Does that make sense?

**Comment by tophmatthews** [comment]
> I guess it's up to the user to use whatever material model is most appropriate to define `density`, so they could use either this or `GenericConstantMaterial`.

Interestingly, there is an old comment in `Density.C`
```
// TODO: We should deprecate the !_is_coupled case and have the
  // user use a GenericConstantMaterial
  if (_is_coupled)
  ```
I didn't touch this at this time, but it does make a good point that unless you are using displacements, you don't need `Density.C`

tophmatthews:
I dont see `Density` in any other module except `Combined` (except for `misc` added in this PR).

tophmatthews:
it's in `gui/tests` and `python/peacock`. Not sure if those are pulling only `misc` and not `solid_mechanics` though.

tophmatthews:
Closed in favor of #29933

Issue #29594, opened 2024-12-20
No longer require single Closures to be supplied to flow channels


joshuahansel:
### Motivation

The flexibility of the Closures system could be increased by not requiring each Closures object to supply all of the necessary closures. For example, one could do this:
```
closures = 'my_friction_closures my_heat_transfer_closures'
```
This allows Closures classes to be more natural, since they don't need to have a combination of types of closure relations.

### Design

The `closures` parameter in `FlowChannelBase` will now correspond to a vector of Closures names, and it will be optional, defaulting to empty. Then all of the places using it will loop over the provided closures.

- [ ] Delete `Closures1PhaseNone`

### Impact

Enhanced flexibility.

joshuahansel:
Re-opening until `Closures1PhaseNone` is removed.

PR #29778, opened 2025-01-31
Deleted deprecated closures member

joshuahansel:
This is a cleanup task for #29594. THM-based applications need to transition to stop using the deprecated singular closures data member.

This will fail at first because application updates are coming in; invalidating after those update should fix.

- [x] Sockeye fix: https://github.inl.gov/ncrc/sockeye/pull/812
- [x] RELAP-7 fix: https://github.inl.gov/ncrc/relap-7/pull/2585
- [x] Update Sabertooth
- [x] Update DireWolf
- [x] Update BlueCRAB

### DIFF
--- a/modules/thermal_hydraulics/include/components/FlowChannelBase.h
+++ b/modules/thermal_hydraulics/include/components/FlowChannelBase.h
@@ -182,11 +182,6 @@ public:
    *
    * @return The closures object(s)
    */
-  std::shared_ptr<ClosuresBase> getClosures() const
-  {
-    mooseDeprecated("getClosures() is deprecated. Use getClosuresObjects() instead.");
-    return _closures;
-  }
   std::vector<std::shared_ptr<ClosuresBase>> getClosuresObjects() const
   {
     return _closures_objects;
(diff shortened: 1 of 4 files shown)


pbehne:
Pls ping me when this is ready for review.

joshuahansel:
@pbehne This is ready, just playing invalidation whack-a-mole at this point

pbehne:
The failing tests look weird in that there are compiler errors. My first guess would be that because this PR is old, not all tests ran at the same time. If this is the case, then moose and some of the apps could have been updated between tests and could cause this screwy behavior. My lazy approach would be to trigger all the tests to run again with
```
git commit --amend --no-edit 
git push --force
```

If tests are still failing in weird ways after that, then further investigation would be required.

The changes themselves look good to me!

joshuahansel:
I think the age of the branch shouldn't matter. I believe the commits get rebased onto devel before testing. You'll see the same build failures all over the MOOSE CIVET page. They can happen occasionally because MOOSE devel isn't compatible with the master branch of an app.

No rush on this though. We can wait for the garbage to pass.

joshuahansel:
@pbehne Invalidation removed the hard fails - just the allowed-to-fails everyone else is seeing right now, so this is ready.

pbehne (approved):

PR #29799, opened 2025-02-03
Make sure linear systems can initialize from exodus as well.

### LINKED ISSUE #29798: Exodus restart not working with linear systems
### Bug Description

The exodus restart segfaults with linear systems

### Steps to Reproduce

Try to restart a linear test with exodus

### Impact

Can't restart linear systems from exodus

### [Optional] Diagnostics

_No response_

grmnptr:
Closes #29798


GiudGiud (approved):
just a note that when we mention this it's best to say "variable initialization" rather than restart because it does not handle high order time integration, stateful variables etc

PR #29802, opened 2025-02-03
Check tangential tolerances in PenetrationThread

### LINKED ISSUE #29777: Spurious Contact/ Incorrect Penetration detection
### Bug Description

Found example of spurious contact when working on explicit contact. MWE below shows incorrect penetration values.

### Steps to Reproduce

```
[Problem]
  solve = false
[]
[Mesh]
  [block_one]
    type = GeneratedMeshGenerator
    dim = 3
    nx = 3
    ny = 3
    nz = 1
    xmin = 4.5
    xmax = 5.5
    ymin = 4.5
    ymax = 5.5
    zmin = 0.001
    zmax = 1.001
    boundary_name_prefix = 'ball'
    elem_type = TET4
  []
  [block_two]
    type = GeneratedMeshGenerator
    dim = 3
    nx = 10
    ny = 10
    nz = 1
    xmin = 3
    xmax = 7
    ymin = 3
    ymax = 7
    zmin = -2
    zmax = 0
    boundary_name_prefix = 'base'
    boundary_id_offset = 10
    elem_type = TET4
  []
  [block_one_id]
    type = SubdomainIDGenerator
    input = block_one
    subdomain_id = 1
  []
  [block_two_id]
    type = SubdomainIDGenerator
    input = block_two
    subdomain_id = 2
  []
  [combine]
    type = MeshCollectionGenerator
    inputs = ' block_one_id block_two_id'
  []
[]

[AuxVariables]
  [penetration]
    order = FIRST
    family = LAGRANGE
  []
  [closest_point_x]
    order = FIRST
    family = LAGRANGE
  []
  [closest_point_y]
    order = FIRST
    family = LAGRANGE
  []
  [closest_point_z]
    order = FIRST
    family = LAGRANGE
  []
[]

[AuxKernels]
  [penetrate]
    type = PenetrationAux
    variable = penetration
    boundary = base_front
    paired_boundary = ball_back
    quantity = distance
  []

  [penetrate11]
    type = PenetrationAux
    variable = closest_point_x
    boundary = base_front
    paired_boundary = ball_back
    quantity = closest_point_x
  []

  [penetrate13]
    type = PenetrationAux
    variable = closest_point_y
    boundary = base_front
    paired_boundary = ball_back
    quantity = closest_point_y
  []

  [penetrate15]
    type = PenetrationAux
    variable = closest_point_z
    boundary = base_front
    paired_boundary = ball_back
    quantity = closest_point_z
  []
[]

[Executioner]
  type = Steady

  solve_type = 'PJFNK'
[]

[Outputs]
  exodus = true
  execute_on = TIMESTEP_END
[]
```
Figure below is from running the above file. 
Image

MWE came from work with explicit contact. Spurious contact below happened with a penalty node face method.
Image

### Impact

Spurious contact prevents users from trusting results.

### [Optional] Diagnostics

_No response_

maxnezdyur:
closes #29777


maxnezdyur:
This fixes a corner case. In which we have 3 penetration infos, all 3 of them are invalid but because the first 2 cause "NEITHER_WINS` causes the third option to win by default without checking if it is a valid info.

test/tests/geomsearch/3d_penetration_locator/tests
```diff
@@ -33,4 +33,11 @@
                   "point search leads to element Jacobian computations too far from the element for "
                   "the Jacobian to be non-negative."
   []
+  [close_tet_mesh]
+    type = Exodiff
+    input = 'close_tet.i'
+    exodiff = 'close_tet_out.e'
+    recover = false # steady solve
+    requirement = "The system will shall be able to ensure contact points are within tangential tolerances."
```
GiudGiud, line 41:
```suggestion
    requirement = "The system shall be able to ensure contact points are within tangential tolerances."
```

bwspenc: Actually, shouldn't it say "The system shall ensure contact points..."? We've been trying to avoid saying "be able to".

GiudGiud:
>  but because the first 2 cause "NEITHER_WINS` causes the third option to win by default without checking if it is a valid info

I see the logic below for the comparisons between two and I m not sure this is the right fix. Seems like we are now discarding  non-winning solutions earlier?

GiudGiud:
looks like you wrote that part of the contact decision making logic @bwspenc

maxnezdyur:
> I see the logic below for the comparisons between two and I m not sure this is the right fix. Seems like we are now discarding non-winning solutions earlier?

With the fix we are discarding non-valid solutions at the last possible moment. When `switch_info` is called, that is when a solution is completely accepted. We have this tangential check in other parts of the code, but there was a corner case where it doesn't get found.

framework/src/geomsearch/PenetrationThread.C
```diff
@@ -427,8 +427,12 @@ PenetrationThread::operator()(const NodeIdRange & range)
           } while (i < p_info.size() && best < p_info.size());
           if (best < p_info.size())
```
bwspenc, line 428:
```suggestion
          if (best < p_info.size() && CIResult != NEITHER_WINS)
```

bwspenc: This would avoid needing to do that comparison again, and might make the code a little clearer since that checking for being within `_tangential_distance` would be done in one place in `competeInteractions()`.

You would also need to put something like `CompeteInteractionResult CIResult = NEITHER_WINS;` before this loop, around line 409, and then don't declare `CIResult` on line 412 anymore.

test/tests/geomsearch/3d_penetration_locator/gold/3d_rings_out_values_0001.csv
```diff
@@ -105,7 +105,7 @@ id,penet,x,y,z
 103,0,0.023321254371987,0.085147669460424,0.52757575757576
 104,0,0.01343582997789,0.080003415385772,0.52757575757576
 105,0,0.001228247892379,0.074916002753794,0.52757575757576
-106,-0.02,-0.012040599039569,0.069254560506297,0.52757575757576
+106,0,-0.012040599039569,0.069254560506297,0.52757575757576
```
bwspenc, line 108: I assume ids 106 and 146 were being erroneously picked up?

test/tests/geomsearch/3d_penetration_locator/tests
```diff
@@ -33,4 +33,11 @@
                   "point search leads to element Jacobian computations too far from the element for "
                   "the Jacobian to be non-negative."
   []
+  [close_tet_mesh]
+    type = Exodiff
+    input = 'close_tet.i'
+    exodiff = 'close_tet_out.e'
+    recover = false # steady solve
+    requirement = "The system shall ensure contact points are within tangential tolerances."
```
bwspenc, line 41:
```suggestion
    requirement = "The system shall ensure all points identified as being in contact are within a distance of the tangential tolerance to the primary surface."
```

bwspenc (commented):
Good catch -- this case was definitely missed in the original logic. I suggested a couple of minor changes that I think would slightly improve readability, but I'm fine with it as you have it if you prefer it that way.

test/tests/geomsearch/3d_penetration_locator/gold/3d_rings_out_values_0001.csv
maxnezdyur: image
Yup, blue is in contact

framework/src/geomsearch/PenetrationThread.C
maxnezdyur: I agree with this fix, but the results showed some problems. I think this should be explored in a different PR.
image

bwspenc (approved):

PR #29803, opened 2025-02-03
Added error check to prevent segfault

### LINKED ISSUE #29788: Segfault when --mesh-only option used with --check-input
### Bug Description

When `--mesh-only` option is used with `--check-input`, a segfault occurs. At a minimum, an error should be thrown instead

### Steps to Reproduce

Run `--check-input --mesh-only` to any existing input file

### Impact

Avoid segfaults and user running into program crash without any additional information

### [Optional] Diagnostics

_No response_

DanielYankura:
## Reason

- Users may not know why it's segfaulting
## Design

- If --check-input is run with any param that creates a final task it results in a segfault
- Added check right after final tasks are set that checks if --check-input is being used as well
## Impact

Will aid in debugging MOOSE inputs

closes #29788

### DIFF
--- a/framework/include/actions/ActionWarehouse.h
+++ b/framework/include/actions/ActionWarehouse.h
@@ -297,6 +297,8 @@ public:
    */
   bool isTaskComplete(const std::string & task) const;
 
+  std::string getFinalTask() { return _final_task; }
+
 protected:
   /**
    * This method auto-builds all Actions that needs to be built and adds them to ActionWarehouse.
--- a/framework/include/actions/ActionWarehouse.h
+++ b/framework/include/actions/ActionWarehouse.h
@@ -297,8 +297,6 @@ public:
    */
   bool isTaskComplete(const std::string & task) const;
 
-  std::string getFinalTask() { return _final_task; }
-
 protected:
   /**
    * This method auto-builds all Actions that needs to be built and adds them to ActionWarehouse.
--- a/framework/include/base/MooseApp.h
+++ b/framework/include/base/MooseApp.h
@@ -1251,7 +1251,8 @@ protected:
 
   /// Indicates whether warnings or errors are displayed when overridden parameters are detected
   bool _error_overridden;
-  std::string _ready_to_exit;
+  /// Indicates if simulation is ready to exit, and keeps track of which param caused it to exit
+  std::string _early_exit_param;
   /// The exit code
   int _exit_code;
 
--- a/framework/include/base/MooseApp.h
+++ b/framework/include/base/MooseApp.h
@@ -1251,7 +1251,7 @@ protected:
 
   /// Indicates whether warnings or errors are displayed when overridden parameters are detected
   bool _error_overridden;
-  bool _ready_to_exit;
+  std::string _ready_to_exit;
   /// The exit code
   int _exit_code;
 
--- a/framework/include/base/MooseApp.h
+++ b/framework/include/base/MooseApp.h
@@ -1253,6 +1253,7 @@ protected:
   bool _error_overridden;
   /// Indicates if simulation is ready to exit, and keeps track of which param caused it to exit
   std::string _early_exit_param;
+  bool _ready_to_exit;
   /// The exit code
   int _exit_code;
 
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -428,6 +428,7 @@ MooseApp::MooseApp(InputParameters parameters)
     _factory(*this),
     _error_overridden(false),
     _early_exit_param(""),
+    _ready_to_exit(false),
     _exit_code(0),
     _initial_from_file(false),
     _distributed_mesh_on_command_line(false),
@@ -815,12 +816,14 @@ MooseApp::setupOptions()
   {
     Moose::out << getPrintableVersion() << std::endl;
     _early_exit_param = "--version";
+    _ready_to_exit = true;
     return;
   }
   else if (getParam<bool>("help"))
   {
     _command_line->printUsage();
     _early_exit_param = "--help";
+    _ready_to_exit = true;
   }
   else if (getParam<bool>("dump") || isParamSetByUser("dump_search"))
   {
@@ -844,6 +847,7 @@ MooseApp::setupOptions()
       Moose::out << "\n### START DUMP DATA ###\n"
                  << formatter.toString(tree.getRoot()) << "\n### END DUMP DATA ###" << std::endl;
       _early_exit_param = "--dump";
+      _ready_to_exit = true;
     }
     else
       mooseError("Search parameter '", search, "' was not found in the registered syntax.");
@@ -868,6 +872,7 @@ MooseApp::setupOptions()
                    << act->_file << "\n";
     }
     _early_exit_param = "--registry";
+    _ready_to_exit = true;
   }
   else if (getParam<bool>("registry_hit"))
   {
@@ -913,6 +918,7 @@ MooseApp::setupOptions()
 
     Moose::out << "\n### END REGISTRY DATA ###\n";
     _early_exit_param = "--registry_hit";
+    _ready_to_exit = true;
   }
   else if (getParam<bool>("definition"))
   {
@@ -924,6 +930,7 @@ MooseApp::setupOptions()
     Moose::out << "%-START-SON-DEFINITION-%\n"
                << formatter.toString(tree.getRoot()) << "\n%-END-SON-DEFINITION-%\n";
     _early_exit_param = "--definition";
+    _ready_to_exit = true;
   }
   else if (getParam<bool>("yaml") || isParamSetByUser("yaml_search"))
   {
@@ -935,6 +942,7 @@ MooseApp::setupOptions()
     _builder.buildFullTree(search);
 
     _early_exit_param = "--yaml";
+    _ready_to_exit = true;
   }
   else if (getParam<bool>("json") || isParamSetByUser("json_search"))
   {
@@ -947,6 +955,7 @@ MooseApp::setupOptions()
 
     Moose::out << "**START JSON DATA**\n" << tree.getRoot().dump(2) << "\n**END JSON DATA**\n";
     _early_exit_param = "--json";
+    _ready_to_exit = true;
   }
   else if (getParam<bool>("syntax"))
   {
@@ -958,6 +967,7 @@ MooseApp::setupOptions()
       Moose::out << it.first << "\n";
     Moose::out << "**END SYNTAX DATA**\n" << std::endl;
     _early_exit_param = "--syntax";
+    _ready_to_exit = true;
   }
   else if (getParam<bool>("show_type"))
   {
@@ -965,6 +975,7 @@ MooseApp::setupOptions()
 
     Moose::out << "MooseApp Type: " << type() << std::endl;
     _early_exit_param = "--show-type";
+    _ready_to_exit = true;
   }
   else if (getInputFileNames().size())
   {
@@ -1054,12 +1065,14 @@ MooseApp::setupOptions()
     moose_server.run();
 
     _early_exit_param = "--language-server";
+    _ready_to_exit = true;
   }
 
   else /* The catch-all case for bad options or missing options, etc. */
   {
     _command_line->printUsage();
     _early_exit_param = "bad or missing";
+    _ready_to_exit = true;
     _exit_code = 1;
   }
 
@@ -1111,19 +1124,26 @@ MooseApp::runInputFile()
   TIME_SECTION("runInputFile", 3);
 
   // If early exit param has been set, then just return
-  if (!_early_exit_param.empty())
+  if (_ready_to_exit)
     return;
 
   _action_warehouse.executeAllActions();
 
   if (isParamSetByUser("mesh_only"))
+  {
     _early_exit_param = "--mesh-only";
+    _ready_to_exit = true;
+  }
   else if (isParamSetByUser("split_mesh"))
+  {
     _early_exit_param = "--split-mesh";
+    _ready_to_exit = true;
+  }
   else if (getParam<bool>("list_constructed_objects"))
   {
     // TODO: ask multiapps for their constructed objects
     _early_exit_param = "--list-constructed-objects";
+    _ready_to_exit = true;
     std::vector<std::string> obj_list = _factory.getConstructedObjects();
     Moose::out << "**START OBJECT DATA**\n";
     for (const auto & name : obj_list)
@@ -1168,7 +1188,7 @@ MooseApp::executeExecutioner()
   TIME_SECTION("executeExecutioner", 3);
 
   // If ready to exit has been set, then just return
-  if (!_early_exit_param.empty())
+  if (_ready_to_exit)
     return;
 
   // run the simulation
@@ -1545,12 +1565,14 @@ MooseApp::run()
 
     Moose::out << docmsg << "\n";
     _early_exit_param = "--docs";
+    _ready_to_exit = true;
     return;
   }
 
   if (showInputs() || copyInputs() || runInputs())
   {
     _early_exit_param = "--show-input, --copy-inputs, or --run";
+    _ready_to_exit = true;
     return;
   }
 
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -1003,6 +1003,12 @@ MooseApp::setupOptions()
     }
     _action_warehouse.build();
 
+    // Check that --check-input was not used with and final task as this will results in a segFault
+    if (_check_input && _action_warehouse.getFinalTask() != "")
+      mooseError("Cannot run --check-input with ",
+                 _action_warehouse.getFinalTask(),
+                 " as this will result in a segfault");
+
     // Setup the AppFileBase for use by the Outputs or other systems that need output file info
     {
       // Extract the CommonOutputAction
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -427,7 +427,7 @@ MooseApp::MooseApp(InputParameters parameters)
     _enable_unused_check(ERROR_UNUSED),
     _factory(*this),
     _error_overridden(false),
-    _ready_to_exit(false),
+    _ready_to_exit(""),
     _exit_code(0),
     _initial_from_file(false),
     _distributed_mesh_on_command_line(false),
@@ -814,13 +814,13 @@ MooseApp::setupOptions()
   else if (getParam<bool>("display_version"))
   {
     Moose::out << getPrintableVersion() << std::endl;
-    _ready_to_exit = true;
+    _ready_to_exit = "display_version";
     return;
   }
   else if (getParam<bool>("help"))
   {
     _command_line->printUsage();
-    _ready_to_exit = true;
+    _ready_to_exit = "help";
   }
   else if (getParam<bool>("dump") || isParamSetByUser("dump_search"))
   {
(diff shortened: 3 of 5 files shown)


pbehne (commented):
While throwing an error is certainly more desirable than allowing a segfault, it would be more useful if we could identify the cause of the segfault and address that directly. Have you looked into this?

DanielYankura:
I looked into that. When using --split-mesh or --mesh-only it sets those as the final task. --check-input isn't a task so it still get's used and calls MooseApp::errorCheck() which tries calling feProblem(). Because not all tasks ran, calling feProblem() results in a segfault.

pbehne:
> I looked into that. When using --split-mesh or --mesh-only it sets those as the final task. --check-input isn't a task so it still get's used and calls MooseApp::errorCheck() which tries calling feProblem(). Because not all tasks ran, calling feProblem() results in a segfault.

Okay, makes sense. To reiterate, the documentation for the `--check-input` option is:

> Check the input file (i.e. requires -i <filename>) and quit

With this in mind, it doesn't make sense to call with `--mesh-only` in the first place. Like you pointed out about some tasks not running, some things aren't getting set up, which causes issues.

framework/include/actions/ActionWarehouse.h
```diff
@@ -297,6 +297,8 @@ class ActionWarehouse : public ConsoleStreamInterface
    */
   bool isTaskComplete(const std::string & task) const;
 
+  std::string getFinalTask() { return _final_task; }
```
pbehne, line 300:
```suggestion
  std::string & getFinalTask() const { return _final_task; }
```

pbehne (changes requested):
When I run moose with options `--check-input` and `--version`, I get a segfault. Could you look into this and some of the other options to make sure our approach is general enough to catch these?

shikhar413:
Would it make sense to check if feProblem() is not null and error out if this is null and _check_inputs is true? This check can be applied right before calling `errorCheck()` in MooseApp.C

This should make the error check more robust regardless of what option is specified

DanielYankura:
I did some more digging and I think I found a way to make it more general. If --check-input is used with any parameter that sets _ready_to_exit=true it will result in a segfault. I'll move the test and rewrite it so that it catches all of these cases.

DanielYankura:
@shikhar413 I just saw your comment. I like that idea. The only issue might be that it'd be hard to give a helpful error message if we don't know why it's null?

shikhar413:
Yeah it might be hard to do that without hard coding which options don’t work with check-inputs. I would say it’s probably fine to just say something like “Command line options are not compatible with —check-inputs as this option requires an FEProblem to be defined in the input file” but someone else might have a different opinion.

The reason that restructuring that logic to prevent ready_to_exit to not be mixed with check-inputs may not be the ideal path forward is that we might in the future want to support mesh-only with check-inputs with a dummy FEProblem instance, and so there’d have to be an exception to that logic if we did that.

shikhar413:
I’ll let someone else make the final call on this, I don’t particularly feel strongly about having the logic be a certain way

DanielYankura:
I feel as well that getting `--check-error` to run with certain other params should be a future issue. The params that cause it to segfault all quit once they're done as well which is why they aren't really compatible. Also keeping track of how many it's interfering with might be difficult because as far as I can tell if there are multiple it will only run one and exit.

I did have a thought for a better way to fix this issue though. If we want to tell the user which command-line argument is interfering I can change `_ready_to_exit` to a string that tells us which param it exited on. That way I can use that info later to know which argument is causing the segfault. This could also come in handy for future issues if we ever need more info for other error checks? I like @shikhar413's idea of checking for a null feproblem instead so I can have it print one error statement if it's null and a different one if it's null and `_ready_to_exit` isn't empty? If that's too much I can just check if it's null and make the error statement more generic.

DanielYankura:
Also I just realized it's not `FEProblem` that's not initialized it's `_executor`. The uninitialized `_executor` calling `feProblem()` is what's causing the segfault. Not a huge difference in how to check for errors but I figured it was important to clarify

shikhar413:
> Also I just realized it's not FEProblem that's not initialized it's _executor. The uninitialized _executor calling feProblem() is what's causing the segfault. Not a huge difference in how to check for errors but I figured it was important to clarify

That's good information to know, thanks for digging into this. I'm planning to work on this issue in the near term, and I think some of these fixes will come into play, so I will keep these changes in mind

https://github.com/idaholab/moose/issues/14550

framework/src/base/MooseApp.C
```diff
@@ -814,13 +814,13 @@ MooseApp::setupOptions()
   else if (getParam<bool>("display_version"))
   {
     Moose::out << getPrintableVersion() << std::endl;
-    _ready_to_exit = true;
+    _ready_to_exit = "display_version";
```
pbehne, line 817:
```suggestion
    _ready_to_exit = "--version";
```

framework/src/base/MooseApp.C
```diff
@@ -814,13 +814,13 @@ MooseApp::setupOptions()
...
     return;
   }
   else if (getParam<bool>("help"))
   {
     _command_line->printUsage();
-    _ready_to_exit = true;
+    _ready_to_exit = "help";
```
pbehne, line 823:
```suggestion
    _ready_to_exit = "--help";
```

framework/src/base/MooseApp.C
```diff
@@ -843,7 +843,7 @@ MooseApp::setupOptions()
       JsonInputFileFormatter formatter;
       Moose::out << "\n### START DUMP DATA ###\n"
                  << formatter.toString(tree.getRoot()) << "\n### END DUMP DATA ###" << std::endl;
-      _ready_to_exit = true;
+      _ready_to_exit = "dump";
```
pbehne, line 846:
```suggestion
      _ready_to_exit = "--dump";
```

framework/src/base/MooseApp.C
```diff
@@ -867,8 +867,7 @@ MooseApp::setupOptions()
         Moose::out << entry.first << "\taction\t" << act->_name << "\t" << act->_classname << "\t"
                    << act->_file << "\n";
     }
-
-    _ready_to_exit = true;
+    _ready_to_exit = "registry";
```
pbehne, line 870:
```suggestion
    _ready_to_exit = "--registry";
```

framework/src/base/MooseApp.C
```diff
@@ -913,7 +912,7 @@ MooseApp::setupOptions()
     Moose::out << root.render();
 
     Moose::out << "\n### END REGISTRY DATA ###\n";
-    _ready_to_exit = true;
+    _ready_to_exit = "registry_hit";
```
pbehne, line 915:
```suggestion
    _ready_to_exit = "--registry_hit";
```

framework/src/base/MooseApp.C
```diff
@@ -924,7 +923,7 @@ MooseApp::setupOptions()
     SONDefinitionFormatter formatter;
     Moose::out << "%-START-SON-DEFINITION-%\n"
                << formatter.toString(tree.getRoot()) << "\n%-END-SON-DEFINITION-%\n";
-    _ready_to_exit = true;
+    _ready_to_exit = "definition";
```
pbehne, line 926:
```suggestion
    _ready_to_exit = "--definition";
```

framework/src/base/MooseApp.C
```diff
@@ -935,7 +934,7 @@ MooseApp::setupOptions()
     _builder.initSyntaxFormatter(Moose::Builder::YAML, true);
     _builder.buildFullTree(search);
 
-    _ready_to_exit = true;
+    _ready_to_exit = "yaml";
```
pbehne, line 937:
```suggestion
    _ready_to_exit = "--yaml";
```

framework/src/base/MooseApp.C
```diff
@@ -947,7 +946,7 @@ MooseApp::setupOptions()
     _builder.buildJsonSyntaxTree(tree);
 
     Moose::out << "**START JSON DATA**\n" << tree.getRoot().dump(2) << "\n**END JSON DATA**\n";
-    _ready_to_exit = true;
+    _ready_to_exit = "json";
```
pbehne, line 949:
```suggestion
    _ready_to_exit = "--json";
```

framework/src/base/MooseApp.C
```diff
@@ -958,14 +957,14 @@ MooseApp::setupOptions()
     for (const auto & it : syntax)
       Moose::out << it.first << "\n";
     Moose::out << "**END SYNTAX DATA**\n" << std::endl;
-    _ready_to_exit = true;
+    _ready_to_exit = "syntax";
```
pbehne, line 960:
```suggestion
    _ready_to_exit = "--syntax";
```

framework/src/base/MooseApp.C
```diff
@@ -958,14 +957,14 @@ MooseApp::setupOptions()
...
   }
   else if (getParam<bool>("show_type"))
   {
     _perf_graph.disableLivePrint();
 
     Moose::out << "MooseApp Type: " << type() << std::endl;
-    _ready_to_exit = true;
+    _ready_to_exit = "show_type";
```
pbehne, line 967:
```suggestion
    _ready_to_exit = "--show_type";
```

framework/src/base/MooseApp.C
```diff
@@ -1054,13 +1053,13 @@ MooseApp::setupOptions()
 
     moose_server.run();
 
-    _ready_to_exit = true;
+    _ready_to_exit = "language_server";
```
pbehne, line 1056:
```suggestion
    _ready_to_exit = "--language_server";
```

framework/src/base/MooseApp.C
```diff
@@ -1112,17 +1111,19 @@ MooseApp::runInputFile()
...
-    _ready_to_exit = true;
+  if (isParamSetByUser("mesh_only"))
+    _ready_to_exit = "mesh_only";
+  else if (isParamSetByUser("split_mesh"))
+    _ready_to_exit = "split_mesh";
   else if (getParam<bool>("list_constructed_objects"))
   {
     // TODO: ask multiapps for their constructed objects
-    _ready_to_exit = true;
+    _ready_to_exit = "list_constructed_objects";
```
pbehne, line 1126:
```suggestion
    _ready_to_exit = "--list-constructed-objects";
```

modules/solid_mechanics/src/base/SolidMechanicsApp.C
```diff
@@ -53,7 +53,7 @@ SolidMechanicsApp::runInputFile()
   MooseApp::runInputFile();
 
   if (getParam<bool>("parse_neml2_only"))
-    _ready_to_exit = true;
+    _ready_to_exit = "parse_neml2_only";
```
pbehne, line 56:
```suggestion
    _ready_to_exit = "--parse-neml2-only";
```

framework/src/base/MooseApp.C
```diff
@@ -1527,13 +1536,13 @@ MooseApp::run()
     }
 
     Moose::out << docmsg << "\n";
-    _ready_to_exit = true;
+    _ready_to_exit = "show_docs";
```
pbehne, line 1539:
```suggestion
    _ready_to_exit = "--docs";
```

framework/src/base/MooseApp.C
```diff
@@ -1139,6 +1140,14 @@ MooseApp::errorCheck()
...
   _builder.errorCheck(*_comm, warn, err);
 
+  if (!_executor.get() && !_executioner.get() && !_ready_to_exit.empty())
+    mooseError("Incompatible command line arguments provided. --check-input cannot be called with ",
+               _ready_to_exit,
+               ".");
+  else if (!_executor.get() && !_executioner.get())
+    mooseError("The Executor is being called without being initialized. This is likely caused by "
+               "incompatible command line arguments");
+
```
pbehne, line 1150: Why `!_executor.get()` AND `!_executioner.get()`? Why not `!_executor.get()` OR `!_executioner.get()`?

In either case, you can make this more efficient with
```
if (!_executor.get() && !_executioner.get())
{
  if (!_ready_to_exit.empty()
  {
    if (_check_input)
      mooseError("Incompatible command line arguments provided. --check-input cannot...");
    else
      mooseError("We should never get here!");
  }
  else
    mooseError("The Executor is being called...");
}
```
Not the added check for `_check_input`. Without it, hardcoding "--check-input" into the error message makes me uneasy. Do you know for sure that the code will only be executed if --check-input is provided?

test/tests/mesh/mesh_only/tests
```diff
@@ -78,4 +78,13 @@
...
     requirement = 'The system shall throw a warning if user-provided extra element id names do not exist on mesh.'
   []
+  [mesh_only_with_incompatible_parameter]
+    type = 'RunException'
+    input = 'mesh_only.i'
+    cli_args = '--check-input --mesh-only'
+    expect_err = "Incompatible command line arguments provided. --check-input cannot be called with mesh_only."
+    method = '!dbg'
+    issues = '#29788'
+    requirement = 'The system shall throw an error is the user provides a parameter that would result in a segfault when used with --mesh-only'
```
pbehne, line 88:
```suggestion
    requirement = 'The system shall throw an error if the user specifies incompatible command line arguments that would cause a segfault.'
```

framework/include/actions/ActionWarehouse.h
pbehne: This is no longer used, and thus no longer needed, correct?

framework/include/base/MooseApp.h
```diff
@@ -1251,7 +1251,7 @@ class MooseApp : public ConsoleStreamInterface,
 
   /// Indicates whether warnings or errors are displayed when overridden parameters are detected
   bool _error_overridden;
-  bool _ready_to_exit;
+  std::string _ready_to_exit;
```
pbehne, line 1254: Let's rename the variable `_ready_to_exit` to something more informative because we have changed it to a string. How about something like `_early_exit_reason`?

pbehne (changes requested):

framework/src/base/MooseApp.C
DanielYankura: The reason I did AND was because MooseApp::feProblem() segfaults only when both are uninitialized. For the params that I checked usually one would be uninitialized and then other initialized. I can definitely go ahead an make it more efficient though.

framework/include/base/MooseApp.h
DanielYankura: What about `_early_exit_param`, or would that be less clear as to what it does?

pbehne: Sounds good to me!

pbehne (changes requested):
Looks good to me after these changes and after all the tests are passing.

shikhar413 (commented):
Took a quick look, some additional comments from me as well

framework/src/base/MooseApp.C
```diff
@@ -1139,6 +1140,24 @@ MooseApp::errorCheck()
...
+  {
+    if (!_early_exit_param.empty())
+    {
+      if (_check_input)
+        mooseError(
+            "Incompatible command line arguments provided. --check-input cannot be called with ",
+            _early_exit_param,
+            ".");
+      else
+        mooseError("We should never get here!");
```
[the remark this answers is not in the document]
pbehne: Good point. I asked that the error be added to be explicit. However, I think that the logic could be changed to this:
```
  if (!_executor.get() && !_executioner.get())
  {
    if (!_early_exit_param.empty() && _check_input)
      mooseError(
          "Incompatible command line arguments provided. --check-input cannot be called with ",
          _early_exit_param,
          ".");
    mooseError("The Executor is being called without being initialized. This is likely "
                "caused by "
                "incompatible command line arguments");
  }
```
or this:
```
  if (!_executor.get() && !_executioner.get())
  {
    if (!_early_exit_param.empty())
      // We only get here if _check_input is true
      mooseError(
          "Incompatible command line arguments provided. --check-input cannot be called with ",
          _early_exit_param,
          ".");
    mooseError("The Executor is being called without being initialized. This is likely "
                "caused by "
                "incompatible command line arguments");
  }
```
or this:
```
  if (!_executor.get() && !_executioner.get())
  {
    if (!_early_exit_param.empty())
    {
      mooseAssert(_check_input, "Something went wrong, we should only get here if _check_input is true.");
      mooseError(
          "Incompatible command line arguments provided. --check-input cannot be called with ",
          _early_exit_param,
          ".");
    }
    mooseError("The Executor is being called without being initialized. This is likely "
                "caused by "
                "incompatible command line arguments");
  }
```

Just so long as we make it explicit that the expectation is for `_check_input` to be true, especially since it is hard-coded in the error message. This will help keep others from having to dig for confirmation like you just did.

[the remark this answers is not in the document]
DanielYankura: I think the last one sounds the best. Having some error get thrown in case `_check_error` is false would be useful for developers who may later try improperly calling `errorCheck()`.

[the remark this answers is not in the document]
pbehne: > I think the last one sounds the best. Having some error get thrown in case `_check_error` is false would be useful for developers who may later try improperly calling `errorCheck()`.

Also useful if the logic in errorCheck or execution is changed down the road.

pbehne:
The changes look good to me. I'm ready to approve once the tests are passing.

pbehne:
The following apps use the variable `_ready_to_exit`:
* Marmot
* Relap-7
* Sabertooth
* SAM

If we want to proceed with this PR as is, we'll have to rename the variable and make a PR for each app. I do not currently have access to Marmot and Sabertooth to do this. I have requested access to Marmot, but Sabertooth is not on NCRC and I haven't yet dug around to find where to get it.

Another option could be to leave `_ready_to_exit` as it was and define `_early_exit_param` as an entirely new variable in addition to `_ready_to_exit`. I'm happy to go either way.

DanielYankura:
Originally I didn't want to make a new variable because I felt it would be redundant. That being said I'm not sure how much `_ready_to_exit` is used in these other apps so it might be wise to leave it alone for now. I can put it back in for now and if we ever want to eliminate it later on I can always open a new PR

DanielYankura:
By the way, I looked at the ones that are still failing. There are two tests that failed because of a timeout. I don't think this is caused by the changes I made.

pbehne:
> By the way, I looked at the ones that are still failing. There are two tests that failed because of a timeout. I don't think this is caused by the changes I made.

I typically invalidate those until they pass :)

shikhar413:
Sorry invalidated the coverage test too early - please ignore

framework/src/base/MooseApp.C
```diff
@@ -1139,6 +1160,22 @@ MooseApp::errorCheck()
...
+    if (!_early_exit_param.empty())
+    {
+      mooseAssert(_check_input,
+                  "Something went wrong, we should only get here if _check_input is true.");
+      mooseError(
+          "Incompatible command line arguments provided. --check-input cannot be called with ",
+          _early_exit_param,
+          ".");
+    }
+    mooseError("The Executor is being called without being initialized. This is likely "
```
pbehne, line 1174: Do we have a way to trigger this error? There is currently no coverage for it.

DanielYankura: As far as I know there is not a way to trigger it right now. There are three cases where `errorCheck()` can get called; if `--check-input` is being used, if `__executor` is initialized, or if `_executioner` is initialized. I left it in there for future developers in case someone makes a change where `errorCheck()` can get called without one of those being true.

pbehne, line 1174:
```suggestion
    // We should never get here
    mooseError("The Executor is being called without being initialized. This is likely "
```

pbehne (approved):
Thanks for the improvement!

PR #29817, opened 2025-02-06
Libmesh update 20250205

roystgnr:
Update libMesh submodule:

- Bug fix: `normals` vectors at the sides of edge elements in 3D, which were previously hard-coded to return `(-1,0,0)` and `(1,0,0)`  at points 0 and 1 respectively (the correct solution only for x-axis-aligned edges), are now correctly calculated for all edges in 2D/3D.  Compatibility with this fix required a corresponding fix in one MOOSE application; developers of any applications not in MOOSE CI who use non-x-axis-aligned edge elements may wish to test their codes for result changes with this update.
- Bug fix: H(curl) and H(div) element calculations are now correct on 2D elements embedded in 3D space and not parallel to the xy plane.
- Bug fix: Some TIMPI communications of "packed-range" data (data with variable-size data types), when given enough input data (typically ~4MB by default) to split between multiple buffers and given an output iterator that requires correct incrementation, were not correctly preserving that iterator from one buffer to the next.  This has been fixed.
- Command line option `--mpi-thread-type` to manually choose between "single", "funneled", "serialized", and "multiple" MPI initialization threading options.  Initialization still defaults to "funneled" (for maximum compatibility with older MPI implementations), but users of slate/strumpack may need to specify "multiple" for compatibility with those packages.
- PetscMatrix support for hash-table-based matrices with newer PETSc versions.  This will allow MOOSE much more flexibility in problems (e.g. contact) where a priori determination of matrix sparsity patterns is difficult.
- Fixes for `HYPRE_SetMemoryLocation` in more complicated (e.g. field split) configurations
- Removed `contrib/boost`, which was a version incompatible with C++17.  libMesh codes (including MOOSE) which require boost now must have an external boost installation to use.
- Support for and compiler flag selection for configuring libMesh with newer Intel compilers
- `MeshTools::n_connected_components()` function to count the number of connected components in a mesh
- Added operator[] for DenseVector.  This is more consistent with other vector/map container types, and enables more compatibility with generic templated algorithms.
- `PetscMatrix::frobenius_norm()` method, and refactoring of other norm calculations
- Improvements and more options in `calculator` app
- Improvements and bug fixes in reduced-basis code
- New grad(div()) problem example code with Raviart-Thomas elements
- Support for "packing" for Cap'n'Proto users
- Bug fix in FParser optimized JIT output
- Bug fix for corrupt time stamps in Exodus outputs with a single time step
- Bug fixes in programmatic setting of PETSc preconditioners
- Workarounds for older PETSc error-checking macros are now removed; raw PETSc code should use `LibmeshPetscCall` for error checking.
- General cleanup in autoconf (`./configure` source) code
- General refactoring in `FEInterface`

Edit: a force push also adds:
- Fix for a multithreading error in applications where multiple threads are encountering elements with negative mapping Jacobians
- Preconditioning option, to scale new raw degrees of freedom, when importing mesh constraint rows from a sparse matrix.  This can greatly improve solver robustness on such constrained systems.
- New `transient_ex3` example application demonstrating a explicit DG/FV formulation of the 2D advection equation

### DIFF
--- a/conda/libmesh/meta.yaml
+++ b/conda/libmesh/meta.yaml
@@ -4,8 +4,8 @@
 #
 # As well as any directions pertaining to modifying those files.
 # ALSO: Follow the directions in scripts/tests/versioner_hashes.yaml
-{% set build = 1 %}
-{% set version = "2024.12.02" %}
+{% set build = 0 %}
+{% set version = "2025.02.25" %}
 
 package:
   name: moose-libmesh
--- a/conda/moose-dev/conda_build_config.yaml
+++ b/conda/moose-dev/conda_build_config.yaml
@@ -3,8 +3,8 @@ mpi:
   - openmpi
 
 moose_libmesh:
-  - moose-libmesh 2024.12.02 mpich_1
-  - moose-libmesh 2024.12.02 openmpi_1
+  - moose-libmesh 2025.02.25 mpich_0
+  - moose-libmesh 2025.02.25 openmpi_0
 
 moose_wasp:
   - moose-wasp 2025.02.25
--- a/conda/moose-dev/meta.yaml
+++ b/conda/moose-dev/meta.yaml
@@ -2,7 +2,7 @@
 # REMEMBER TO UPDATE the .yaml files for the following packages:
 #   moose/conda_build_config.yaml
 # As well as any directions pertaining to modifying those files.
-{% set version = "2025.02.27" %}
+{% set version = "2025.03.03" %}
 
 package:
   name: moose-dev
--- a/conda/moose/conda_build_config.yaml
+++ b/conda/moose/conda_build_config.yaml
@@ -3,7 +3,7 @@ mpi:
   - openmpi
 
 moose_dev:
-  - moose-dev 2025.02.27
+  - moose-dev 2025.03.03
 
 #### Darwin SDK SYSROOT
 CONDA_BUILD_SYSROOT:                                        # [osx]
--- a/modules/doc/content/newsletter/2025/2025_02.md
+++ b/modules/doc/content/newsletter/2025/2025_02.md
@@ -11,6 +11,68 @@ for a complete description of all MOOSE changes.
 
 ## libMesh-level Changes
 
+### `2025.02.14` Update
+
+- Bug fix: `normals` vectors at the sides of edge elements in 3D,
+  which were previously hard-coded to return `(-1,0,0)` and `(1,0,0)`
+  at points 0 and 1 respectively (the correct solution only for
+  x-axis-aligned edges), are now correctly calculated for all edges in
+  2D/3D.  Compatibility with this fix required a corresponding fix in
+  one MOOSE application; developers of any applications not in MOOSE
+  CI who use non-x-axis-aligned edge elements may wish to test their
+  codes for result changes with this update.
+- Bug fix: H(curl) and H(div) element calculations are now correct on
+  2D elements embedded in 3D space and not parallel to the xy plane.
+- Bug fix: Some TIMPI communications of "packed-range" data (data with
+  variable-size data types), when given enough input data (typically
+  ~4MB by default) to split between multiple buffers and given an
+  output iterator that requires correct incrementation, were not
+  correctly preserving that iterator from one buffer to the next.
+  This has been fixed.
+- Fix for a multithreading error in applications where multiple
+  threads are encountering elements with negative mapping Jacobians
+- Preconditioning option, to scale new raw degrees of freedom, when
+  importing mesh constraint rows from a sparse matrix.  This can
+  greatly improve solver robustness on such constrained systems.
+- Command line option `--mpi-thread-type` to manually choose between
+  "single", "funneled", "serialized", and "multiple" MPI
+  initialization threading options.  Initialization still defaults to
+  "funneled" (for maximum compatibility with older MPI
+  implementations), but users of slate/strumpack may need to specify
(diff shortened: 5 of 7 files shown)


lindsayad:
It looks like I may need to go through some apps and fix up the PETSc macros

lindsayad:
- [x] Griffin patch: https://github.inl.gov/ncrc/griffin/pull/2327
- [x] Subchannel patch: https://github.inl.gov/ncrc/subchannel/pull/351

lindsayad:
I think that when https://github.inl.gov/ncrc/pronghorn/pull/1292 and https://github.com/idaholab/virtual_test_bed/pull/520 are merged, then pronghorn devel (which currently has a subchannel module update that fixes the libmesh petsc call macros) can go to master, and then I believe we'll be all set to go for invalidating the failing app tests

lindsayad:
relap-7 and sockeye should be fixed now thanks to @joshuahansel. Looks like we need a rebase here

roystgnr:
I'll do the rebase in a sec.

roystgnr:
Well, I'll rebase the newsletter, anyway; simpler to just redo the other commits.

lindsayad:
I added the other targets. Looks like we need another rebase

milljm:
neat... you'll be one of the first to follow: https://github.com/idaholab/moose/wiki/Updating-packages

lindsayad:
Anyway we can wait to get libMesh/libmesh#4091 in here?

roystgnr:
Okay, most of those latest set of failures were "our CI can't always complete a TCP connection" issues, and I've restarted those, but we've also got a framework test failure in https://civet.inl.gov/job/2712966/ that looks real (related to https://github.com/idaholab/moose/pull/28785 ?  But I don't see how it would have passed there) and a Bison test failure in https://civet.inl.gov/job/2712965/ that just looks bizarre - it's trying to open a `CheckpointIO` file with an `ExodusII_IO` reader??

lindsayad:
I'm pretty sure the latter failure is because of openmpi file system issues that we've seen in many places. I'm invalidating to see what happens

lindsayad (approved):
@milljm do the conda changes look good to you?

milljm (approved):
The output of the verification looks great to me. My only concern is another update that might vie for merge rights affecting moose-dev's version: https://github.com/idaholab/moose/pull/29930

lindsayad:
Oh yea sure enough

lindsayad:
Maybe this means we can get libmesh/libmesh#4091 in 😆

roystgnr:
I'll start another submodule update PR next week, even if all it's got is 4091 and the Prism+subdomain-restriction+AMR fix and the FE+Quadrature API updates ... but at this point I don't want to add any additional uncertainty to this one.  Taking 3 weeks to get a monthly submodule update in is embarrassing enough already.

BWXT is still working with hand-compiled software mixes right now, to work around an IGA conditioning issue with a fix in libMesh.  I'd love to be able to answer Coreform's reasonable question of "do you have a timeline for when we'll be able to run the IGA workflow on MOOSE master?" with something other than "Nope!", but at least I can avoid increasing the delay.
 
In e.g. https://civet.inl.gov/job/2719589/ we're seeing the same ARM Mac test failures (hopefully to be fixed permanently when #29962 hits devel) from other PRs, so at least there doesn't seem to be too much uncertainty on that front.  I just kicked the tests here (before the re-versioning force pushes) and I saw the quicker recipe passing already, so it really must be come kind of parallel race condition.

And ... the re-versioning force pushes didn't actually fix the branch conflicts.  If I rebase on origin/devel, we've still got a conflict, and if I rebase on origin/next then `./scripts/versioner.py --build-templates` thinks I haven't changed `moose-dev` ... which can't be correct, can it?  I'll try pushing from that anyway and let @milljm tell me.

roystgnr:
Yeah, neither rebase works.

roystgnr:
I was hoping next->devel would finish up last night, but no; the API change mixed with our fix in #29962 broke SAM builds in e.g. https://civet.inl.gov/job/2718600/

loganharbour:
@roystgnr I'll look at this shortly. Please keep it as is. I want to make sure our failure checking works.

We've got two separate moose-dev bumps stuck on next, and this has an old date so it's perfect for sanity checking.

loganharbour (approved):

milljm (approved):
bump looks good to me! 👍

lindsayad:
I don't understand why the openmpi tests are taking so long. Can we just merge this bad boy?

roystgnr:
2 crashes with stack traces in PetscInitialize, one timeout just starting a test.

My vote is for "just merge this bad boy" - I could make a long list of all the things I could have theoretically broken with the dozen odd major changes in this update, but "PetscInitialize, but only with OpenMPI" would not be on it.

It's disappointing to see OpenMPI so flaky, though.  Way back when I used to switch between OpenMPI and MPICH frequently, well, neither of them was perfect, but OpenMPI was always the more reliable of the two.

lindsayad:
I would still like to remove the openmpi target, at least on PRs. The sum of time developers have spent wondering about false positives has to at least be on the order of hours at this point

roystgnr:
I like the OpenMPI target as a "sanity check" for weird MPI-looking failures in MPICH, but it really should be going yellow on failure, not red.

What's the verdict for this case, though?  Merge despite OpenMPI, or invalidate and cross our fingers?

roystgnr:
Ah, wait, never mind, we're still branch locked.

Issue #28053, opened 2024-07-02
MOOSE code still triggering PETSc's unused options warning


pbehne:
### Bug Description

MOOSE is setting PETSc options that are not always used. When this occurs, PETSc emits an annoying unused options warning that the user cannot do anything about.

### Steps to Reproduce

So far, I have seen 2 categories of unused options.

The first can be observed by running `test/tests/problems/eigen_problem/eigensolvers/ipm.i`. This results in the following warning:
```
WARNING! There are options you set that were not used!
WARNING! could be spelling mistake, etc!
There are 7 unused database options. They are:
Option left: name:-ksp_converged_reason value: ::failed source: code
Option left: name:-mat_mffd_type value: wp source: code
Option left: name:-snes_converged_reason value: ::failed source: code
Option left: name:-snes_mf_operator (no value) source: code
Option left: name:-st_ksp_atol value: 1e-50 source: code
Option left: name:-st_ksp_max_it value: 10000 source: code
Option left: name:-st_ksp_rtol value: 0.01 source: code
```

The likely solution is not to set these options for `Eigen` problems.

The second category can be observed by running `test/tests/time_integrators/actually_explicit_euler_verification/ee-1d-linear.i` This results in the following warning:
```
WARNING! There are options you set that were not used!
WARNING! could be spelling mistake, etc!
There are 3 unused database options. They are:
Option left: name:-snes_converged_reason value: ::failed source: code
Option left: name:-snes_monitor_cancel (no value) source: code
Option left: name:-snes_type value: ksponly source: code
```

I have not yet investigated this category. All failures can be viewed here: https://civet.inl.gov/event/184918/

### Impact

Users are annoyed with warnings that are not their fault.

pbehne:
I'll work on this one.

lindsayad:
For the latter case, we are likely not doing a nonlinear solve at all, which is why you see the unused options for SNES

pbehne:
I agree. The time integrator used in `ee-1d-linear.i` is `ActuallyExplicitEuler`, which only does a linear solve per time step. This will be the case for any integrator derived from `ExplicitTimeIntegrator`.

Take `-snes_converged_reason`, for example. It is an option listed in `getCommonPetscFlags`. All flags returned by this function are turned on by `FEProblemSolve`.

[...]

PR #29822, opened 2025-02-06
Changed pc type from hypre to none for failing tests. Ref #28053.

pbehne:
This PR should fix the failures occurring on next as a result of PR #27948.

### DIFF
--- a/test/tests/misc/petsc_option_left/2d_diffusion_petsc_option.i
+++ b/test/tests/misc/petsc_option_left/2d_diffusion_petsc_option.i
@@ -53,5 +53,5 @@
   [../]
 
   petsc_options_iname = "-pc_type"
-  petsc_options_value = "hypre"
+  petsc_options_value = "none"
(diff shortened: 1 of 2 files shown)


GiudGiud (approved):

loganharbour (approved):

loganharbour:
Cancelled the two sweeps that were added that aren't used on next.

loganharbour:
@pbehne added additional recipes that aren't on next. Failures/running tests are not needed.

lindsayad:
It turns out that the fundamental issue is that we are performing an explicit solve here and there are no time derivative kernels to create a mass matrix. Shockingly somehow the matrix ended up with some nonzeroes; for the 3 proc case there were 2 procs that had some number of nonzeroes while 1 proc had none at all. This led to seg faults in hypre. We should make sure to error when performing explicit solves if there are no kernels contributing to the matrix

PR #29794, opened 2025-02-02
Update NEML2 submodule

hugary1995:
- Experimental work dispatcher
- Removed torch/linalg.h dependency

ref #29579

PR #29832, opened 2025-02-06
Check if old state exists in neml2 predictor

dschwen:
## Reason
If a NEML2 model does not hav old state on the input axis the `applyPredictor` step in `NEML2ModelExecutor` will throw an exception.

## Design
Use `has_subaxis` to check for the existence of the old state subaxis and bail early if it doesn't exist.

## Impact
Enable models to run.


hugary1995:
I suppose this is superseded by #29794

hugary1995:
or no I mixed the two PRs

PR #29850, opened 2025-02-10
Projection ROMs 

### LINKED ISSUE #27101: Generalized Projection-based ROM System
### Motivation

For design and control problems, the forward problem needs to be run many times. Projection based ROM provide a way to speed up the process.

### Design

Create a projection based system that is easy to use. For non-affine or non-linear problems, consider using a hyper-reduction method that is algebraic in nature. This method will use the Discrete Empirical Interpolation Method (DEIM) and Matrix DEIM, to approximate the Residual and Jacobian. The initial work will only approximate the entire jacobian and residual, which will only allow it to be applicable to affine or non-affine steady linear problems.

In the future, by using vector and matrix tags, users can separate the system into linear steady, linear unsteady, nonlinear steady, nonlinear unsteady, which would give the needed control to solve a greater range of problems. Also a future enhancement would be the ability to compute the jacobian/resdiual on a smaller reduced set of elements that are needed for the DEIM procedures.

### Impact

Allow users to create ROM from MOOSE inputs as simply as possible. Significant speedups can be achieved without new kernels or code changes.

maxnezdyur:
closes #27101


maxnezdyur:
@grmnptr Draft PR

grmnptr:
So far looks good, let me know when it is ready for a detailed review!

PR #29863, opened 2025-02-11
29862 bump apptainer conda instance

milljm:
Lets see what fails when we try to upgrade Conda.


loganharbour:
Can we close this @milljm ?

Issue #29868, opened 2025-02-12
Including AD (Automatic Differentiation) to the Curl and Coupled Curl Operators


csdechant:
### Motivation

As the development and implementation of the electromagnetic module increases, there is a need to include the option to defined the curl and coupled curl operators as AD functions. This would avoid errors involving more complex formulations that include the curl of variables in lieu of using hand-coded Jacobians.

### Design

A function, such as `adCurlSln()`, that calculates the curl of a variable as an AD variable should be included in the MooseVariableData file. Then the `adCurlSln()` function should be properly inherited by the additional MooseVariable files. To ensure that the AD curl operators are working as as intended, tests using `type = 'PetscJacobianTester'` should be added.

### Impact

The inclusion of a AD versions of the curl and coupled curl operator will allow for faster development of additional physics into the MOOSE framework that utilizes the curl of variables.

PR #29869, opened 2025-02-12
Adding AD Curl and Coupled Curl Capability

csdechant:
## Reason

As the development and implementation of the electromagnetic module increases, there is a need to include the option to defined the curl and coupled curl operators as AD functions. This helps to avoid errors involving more complex formulations that include the curl of variables in lieu of using hand-coded Jacobians.

This closes Issue #29868

## Design

A function named `adCurlSln()` that calculates the curl of a variable as an AD variable was included in the MooseVariableData file. This enables the used of the curl operator on AD variables and coupled AD variables. To ensure that the AD curl operators are working as as intended, tests using `type = 'PetscJacobianTester'` were added.

## Impact

The inclusion of a AD versions of the curl and coupled curl operator will allow for faster development of additional physics into the MOOSE framework that utilizes the curl of variables.


cticenhour:
Setting some time aside tomorrow to look at this. In the meantime, looks like one of your tests is diffing slightly on the Linux targets.

test/tests/kernels/vector_fe/gold/ad_coupled_curl_product_out.csv
```diff
@@ -0,0 +1,3 @@
+time,HCurlError,HCurlSemiError,u_L2Error,v_L2Error
+0,0,0,0,0
+1,4.408461264314,4.3468966759226,0.20140570130717,0.7341798198058
```
cticenhour, line 3: These errors seem kinda big...can you explain?

test/tests/kernels/vector_fe/tests
```diff
@@ -20,6 +20,59 @@
...
+    input = 'ad_coupled_curl_product.i'
+    exodiff = 'ad_coupled_curl_product_out.e'
+    custom_cmp = 'ad_coupled_curl_product.cmp'
+    abs_zero = 1e-8
+    issues = '#29868'
+    requirement = "The system shall be able to solve a problem that supplies the product "
+                  "of the curl of a variable vector and a supplied field into a "
+                  "diffusion problem using automatic differentiation."
+  []
+  [ad_coupled_curl_product_errors]
```
cticenhour, line 54: You don't need to run this twice to get the CSV since you're not changing any other params. Have the output generated as part of `ad_coupled_curl_product` with all output turned on and set that test as the `prereq` to this one. Then set `should_execute` to `false` in this test spec to have it skip running the file and just use the output files available from the `prereq` test.

cticenhour (changes requested):

test/tests/kernels/vector_fe/gold/ad_coupled_curl_product_out.csv
csdechant: This is because this test is a 5x5x5 mesh to reduce the size of the gold file. Below is a convergence study showing linear convergence slopes:
ad_coupled_curl_product_convergence

modules/electromagnetics/src/kernels/ADCurlCurlField.C
```diff
@@ -0,0 +1,37 @@
...
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#include "ADCurlCurlField.h"
+#include "Assembly.h"
+
+registerMooseObject("ElectromagneticsApp", ADCurlCurlField);
```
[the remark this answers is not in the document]
csdechant: From our offline conversation:
 `GenericKernelCurl.h` is being included in this PR. Inheriting from this object allows for the templating of AD- and non-AD vector kernels and for the uses of calling `_curl_u` and `_curl_test` without users calling from assembly (i.e. before users had to defined the curl of `u` using `_var.adCurlSln()`). 

For the former, templating of AD- and non-AD vector kernels is not currently present in MOOSE and would require in-depth changes to the current vector kernel object class to facilitate.  For the latter, this process of defining a kernel base for an operation has been done before, such as with `ADKernelGrad`.

test/src/kernels/ADVectorFEWave.C
```diff
@@ -0,0 +1,47 @@
...
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#include "ADVectorFEWave.h"
+#include "Function.h"
+#include "Assembly.h"
+
+registerMooseObject("MooseTestApp", ADVectorFEWave);
```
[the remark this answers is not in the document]
csdechant: Please refer to the previous comment related to `ADCurlCurlField`.

test/tests/kernels/vector_fe/tests
csdechant: Those option settings have been added to the test file.

test/tests/kernels/vector_fe/gold/ad_coupled_curl_product_out.csv
cticenhour: Looks good! Thanks for posting this for future reference.

cticenhour (approved):

Issue #29921, opened 2025-02-20
MeshCut2DFractureUserObject produces incorrect angle of crack propagation


tranm-ansto:
MeshCut2DFractureUserObject produces incorrect angle of crack propagation.

Equation 6 in "Ceramic nuclear fuel fracture modeling with the extended finite element method" was not correctly implemented.

bwspenc:
The 3D variant of this (`CrackMeshCut3DUserObject`) also has the same issue. We should cover both of these with this issue.

PR #29922, opened 2025-02-20
Xfem meshcut2d fracture uo

tranm-ansto:
## Reason

Closes issue #29921.

## Design

Fixes crack propagation angle.

## Impact

Crack propagation angle has been corrected.


tranm-ansto:
@lynnmunday Please review my pull request that fixes the crack propagation angle in `MeshCut2DFractureUserObject`.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,34 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
+        // growth direction in crack front coord (cfc) system based on the max hoop stress criterion
+        // Ref "Ceramic nuclear fuel fracture modeling with the extended finite element method" by
+        // W. Jiang, B. Spencer, and J. Dolbow
+        // Equation 6
         Real ki = _ki_vpp->at(i);
         Real kii = _kii_vpp->at(i);
-        Real sqrt_k = std::sqrt(ki * ki + kii * kii);
-        Real theta = 2 * std::atan((ki - sqrt_k) / (4 * kii));
+        Real sqrt_k = std::sqrt(ki * ki + 8 * kii * kii);
+        Real theta_m = 2 * std::atan((ki - sqrt_k) / (4 * kii)); // a more stable implementation
```
lynnmunday, line 159:
```suggestion
        Real theta_m = 2 * std::atan((ki - sqrt_k) / 4); 
```

lynnmunday: I don't think this 1/kii term should be here or in the below.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,34 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
+        Real theta_p = 2 * std::atan((ki + sqrt_k) / (4 * kii)); // a more stable implementation
```
lynnmunday, line 160:
```suggestion
        Real theta_p = 2 * std::atan((ki + sqrt_k) / 4);
```

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,34 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
+
+        // check sigma_tt
+        // Ref "Ceramic nuclear fuel fracture modeling with the extended finite element method" by
+        // W. Jiang, B. Spencer, and J. Dolbow
+        // Equation 5
+        Real _r = _growth_increment;
```
lynnmunday, line 166:
```suggestion
        Real r = _growth_increment;
```

lynnmunday: The "_" is used to identify something as a member variable.  So remove those from temporary variables.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,34 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
+        Real _sigma_tt_m = (ki / (4 * std::sqrt(2 * M_PI * _r))) *
```
lynnmunday, line 167: remove "_"

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,34 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
+                               (3 * std::cos(theta_m / 2) + std::cos(3 * theta_m / 2)) +
+                           (kii / (4 * std::sqrt(2 * M_PI * _r))) *
+                               (-3 * std::sin(theta_m / 2) - 3 * std::sin(3 * theta_m / 2));
+        Real _sigma_tt_p = (ki / (4 * std::sqrt(2 * M_PI * _r))) *
```
lynnmunday, line 171: same here.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,34 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
+                               (3 * std::cos(theta_p / 2) + std::cos(3 * theta_p / 2)) +
+                           (kii / (4 * std::sqrt(2 * M_PI * _r))) *
+                               (-3 * std::sin(theta_p / 2) - 3 * std::sin(3 * theta_p / 2));
+        Real theta;
+        if (_sigma_tt_m > _sigma_tt_p)
+          theta = theta_m;
+        else
+          theta = theta_p;
```
lynnmunday, line 179:
```suggestion
        Real theta = theta_p;
        if (_sigma_tt_m > _sigma_tt_p)
          theta = theta_m;
```

lynnmunday: I do these if else statements like this.

lynnmunday (changes requested):
Does this now look like the Abaqus criterion?  I noticed they use KII to determine when to choose theta_p or theta_m instead of checking it in the hoop stress equation.  You'll have to regold the output files again.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,34 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
-        // criterion
+        // growth direction in crack front coord (cfc) system based on the max hoop stress criterion
+        // Ref "Ceramic nuclear fuel fracture modeling with the extended finite element method" by
+        // W. Jiang, B. Spencer, and J. Dolbow
+        // Equation 6
         Real ki = _ki_vpp->at(i);
         Real kii = _kii_vpp->at(i);
-        Real sqrt_k = std::sqrt(ki * ki + kii * kii);
-        Real theta = 2 * std::atan((ki - sqrt_k) / (4 * kii));
+        Real sqrt_k = std::sqrt(ki * ki + 8 * kii * kii);
```
lynnmunday, line 158: You could implement this exactly how Wen has it in his paper

lynnmunday, line 158:
```suggestion
        Real sqrt_k = std::sqrt((ki * ki) /(kii*kii)+ 8);
```

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
lynnmunday, line 159:
```suggestion
        Real theta_m = 2 * std::atan(1/4*(ki / kii - sqrt_k));
```

lynnmunday: how it would be if it is the same as Wens paper

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
lynnmunday, line 160: Real theta_m = 2 * std::atan(1/4*(ki / kii + sqrt_k))

lynnmunday: Also how it would be if it is the same as Wens paper

modules/xfem/test/tests/mesh_cut_2D_fracture/gold/kvpp_based_meshCut_uo_xfemcutter_XFEMCutMeshOutput.e-s0004
lynnmunday, line 1: This file still tests what it should

modules/xfem/test/tests/nucleation_uo/gold/nucleate_edge_bulk_crack_2d_out_II_KI_1_0011.csv
lynnmunday, line 1: This file is good and still tests what it should

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -131,7 +131,7 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
                _original_and_current_front_node_ids.size());
 
   if (_k_critical_vpp && ((_k_critical_vpp->size() != _original_and_current_front_node_ids.size())))
-    mooseError("k_critical_vectorpostprocessor must have the same number of crack front points as "
+    mooseError("k_critical_vectorpostprocessor must have the same number of crack front points as ",
```
lynnmunday, line 134: remove this change.

lynnmunday (changes requested):
Thanks Minh, the code looks right.  I had a question about the equation so I would implement it exactly how it is in Wen's paper.  But after making the changes, your implementation and my recommendation below give the same results so your implementation was correct.  For some reason the crack is now jagged.  I'm not sure why.  Maybe the difference in hoop stress between the plus and minus angle is really small.  You need to change on input file to get the tests to correctly test the feature they were made for, I added a comment for that below.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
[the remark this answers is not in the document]
tranm-ansto: > I put it in this format so it is exactly like the equation in Wen's paper. Your implementation might have been right but i didn't know why you had a 1/(4*kii) term in this equation: theta_p = 2 * std::atan((ki + sqrt_k) / (4 * kii)); I thougth it should have been just 1/4

I tested 
```
Real sqrt_k = std::sqrt((ki * ki) / (kii * kii) + 8);
Real theta_m = 2 * std::atan((ki / kii - sqrt_k) / 4);
Real theta_p = 2 * std::atan((ki / kii + sqrt_k) / 4);
```
on my model, and it seems to crash very early. When mode I fracture dominates, the KI term becomes large and the KII term becomes small, `(ki * ki) / (kii * kii)` becomes very large. This makes the 'sqrt_k' unstable. 

Instead, if you recast the equation like the below, you can avoid numerical instability when the KII term becomes really small.
```
Real sqrt_k = std::sqrt(ki * ki + 8 * kii * kii);
Real theta_m = 2 * std::atan((ki - sqrt_k) / (4 * kii));
Real theta_p = 2 * std::atan((ki + sqrt_k) / (4 * kii));
```

tranm-ansto:
> Does this now look like the Abaqus criterion? I noticed they use KII to determine when to choose theta_p or theta_m instead of checking it in the hoop stress equation. You'll have to regold the output files again.

The Equation 6 in Wen's paper does not look like the one in the Abaqus documentation. For consistency, I'm going with Equation 6 and check the hoop stress.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
tranm-ansto: > remove this change.

Removed.

tranm-ansto:
> Thanks Minh, the code looks right. I had a question about the equation so I would implement it exactly how it is in Wen's paper. But after making the changes, your implementation and my recommendation below give the same results so your implementation was correct. For some reason the crack is now jagged. I'm not sure why. Maybe the difference in hoop stress between the plus and minus angle is really small. You need to change on input file to get the tests to correctly test the feature they were made for, I added a comment for that below.

The crack propagation angle is more aggressive now. But, I don't know for sure why the crack is jagged.

jiangwen84:
> Thanks Minh, the code looks right. I had a question about the equation so I would implement it exactly how it is in Wen's paper. But after making the changes, your implementation and my recommendation below give the same results so your implementation was correct. For some reason the crack is now jagged. I'm not sure why. Maybe the difference in hoop stress between the plus and minus angle is really small. You need to change on input file to get the tests to correctly test the feature they were made for, I added a comment for that below.

The codes look correct to me. I am not sure what causes the jagged crack. Have you checked the theta during crack propagation? Does the sign actually change?

jiangwen84:
> > Thanks Minh, the code looks right. I had a question about the equation so I would implement it exactly how it is in Wen's paper. But after making the changes, your implementation and my recommendation below give the same results so your implementation was correct. For some reason the crack is now jagged. I'm not sure why. Maybe the difference in hoop stress between the plus and minus angle is really small. You need to change on input file to get the tests to correctly test the feature they were made for, I added a comment for that below.
> 
> The codes look correct to me. I am not sure what causes the jagged crack. Have you checked the theta during crack propagation? Does the sign actually change?

You can also check https://doi.org/10.1016/S0020-7683(03)00095-7  They provide a simpler way of calculating theta.

jiangwen84:
> > > Thanks Minh, the code looks right. I had a question about the equation so I would implement it exactly how it is in Wen's paper. But after making the changes, your implementation and my recommendation below give the same results so your implementation was correct. For some reason the crack is now jagged. I'm not sure why. Maybe the difference in hoop stress between the plus and minus angle is really small. You need to change on input file to get the tests to correctly test the feature they were made for, I added a comment for that below.
> > 
> > 
> > The codes look correct to me. I am not sure what causes the jagged crack. Have you checked the theta during crack propagation? Does the sign actually change?
> 
> You can also check https://doi.org/10.1016/S0020-7683(03)00095-7 They provide a simpler way of calculating theta.

My old codes are https://github.com/jiangwen84/moose/blob/ae83e8e38d68c11f6a1ef4e80f6e6b014ad6d83e/modules/solid_mechanics/src/userobjects/XFEMMaxHoopStress.C

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
lynnmunday, line 158: we can check that kii is > than libmesh::tol.  Then we don't have to worry about divide by 0.  For that case we should just set theta_m=theta_p=0

tranm-ansto:
> > Thanks Minh, the code looks right. I had a question about the equation so I would implement it exactly how it is in Wen's paper. But after making the changes, your implementation and my recommendation below give the same results so your implementation was correct. For some reason the crack is now jagged. I'm not sure why. Maybe the difference in hoop stress between the plus and minus angle is really small. You need to change on input file to get the tests to correctly test the feature they were made for, I added a comment for that below.
> 
> The codes look correct to me. I am not sure what causes the jagged crack. Have you checked the theta during crack propagation? Does the sign actually change?

I've checked the theta of `kcrit_stress_based_meshCut_uo.i`, and the sign does not change. The theta_m produces higher hoop stress.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
tranm-ansto: I have put in a 'divide by 0' check for kii.

lynnmunday (changes requested):
remove redundant code and it is good.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -149,12 +149,46 @@ MeshCut2DFractureUserObject::findActiveBoundaryGrowth()
...
+        // Ref "Ceramic nuclear fuel fracture modeling with the extended finite element method" by
+        // W. Jiang, B. Spencer, and J. Dolbow
+        // Equation 6
         Real ki = _ki_vpp->at(i);
         Real kii = _kii_vpp->at(i);
-        Real sqrt_k = std::sqrt(ki * ki + kii * kii);
-        Real theta = 2 * std::atan((ki - sqrt_k) / (4 * kii));
+        Real sqrt_k = std::sqrt(ki * ki + 8 * kii * kii);
+        theta_m = 2 * std::atan((ki - sqrt_k) / (4 * kii));
+        theta_p = 2 * std::atan((ki + sqrt_k) / (4 * kii));
```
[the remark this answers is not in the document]
tranm-ansto: Sorry about that. Precheck wasn't working yesterday. I was trying a few things and forgot the redundant code.

jiangwen84:
was the issue with jagged crack resolved?

jiangwen84:
> was the issue with jagged crack resolved?

Also check the calculation of K. Is the choice of radius reasonable?

tranm-ansto:
> was the issue with jagged crack resolved?

it has not been resolved yet. Lynn and I will look into it.

tranm-ansto:
> > was the issue with jagged crack resolved?
> 
> Also check the calculation of K. Is the choice of radius reasonable?

Thank you for the pointer. 

> > was the issue with jagged crack resolved?
> 
> Also check the calculation of K. Is the choice of radius reasonable?

The radius is quite big. q function shows that the contour ring is outside of the domain. With the correction in theta, the crack is now more sensitive to KII.

jiangwen84:
> > > was the issue with jagged crack resolved?
> > 
> > 
> > Also check the calculation of K. Is the choice of radius reasonable?
> 
> Thank you for the pointer.
> 
> > > was the issue with jagged crack resolved?
> > 
> > 
> > Also check the calculation of K. Is the choice of radius reasonable?
> 
> The radius is quite big. q function shows that the contour ring is outside of the domain. With the correction in theta, the crack is now more sensitive to KII.

I do not think it would work if the contour ring is outside the domain.

lynnmunday:
I also noticed jagged crack fronts when the q ring is completely inside the domain.  I changed the size of the growth increment to be smaller and the crack growth is still jagged.   

The test from the figures was made to show the effect of the q ring being outside the domain.  The crack on the right stops before hitting the boundary.  We then implemented a stress criterion to allow grow the crack through the boundary if it the crack front opening stress exceeds a critical value.  

I noticed the tests in mesh_cut_2D_fracture/ have cracks with loading where ki ~ 5*kii and the cracks are jagged.  The cracks in nucleation_uo/ have ki~kii and the cracks are much smoother.  I haven't looked into this at all.  We will need to print out the theta but what Minh has is implemented correctly.  If there is an error it is in something I put in.

lynnmunday:
This python script will plot theta for a range of ki and kii values:
```
import numpy as np
import matplotlib.pyplot as plt
from mpl_toolkits.mplot3d import Axes3D


ki_range = np.linspace(0, 100, 100)
kii_range = np.linspace(-100, 100, 200)
ki, kii = np.meshgrid(ki_range, kii_range)

sqrt_k = np.sqrt(ki**2 + 8 * kii**2)
theta_m = 2 * np.arctan((ki - sqrt_k) / (4 * kii))

fig = plt.figure()
ax = fig.add_subplot(111, projection="3d")
surf = ax.plot_surface(ki, kii, theta_m, cmap="viridis", alpha=0.8)
wire = ax.plot_wireframe(ki, kii, theta_m, color="k", linewidth=0.5)
ax.set_xlabel("ki")
ax.set_ylabel("kii")
ax.set_zlabel("theta_m")
fig.colorbar(surf, ax=ax, shrink=0.5, aspect=5)

ax.grid(True)

plt.show()
```

lynnmunday:
Here is the plot with the ki and kii points show by the dots.  The points are for the mesh_cut_2D_fracture director test [k_growth]
These values of kii results in large angle changes for small changes in kii.

Screenshot 2025-02-25 at 10 58 02 AM

lynnmunday:
For the crack to be jagged like that, it has to be changing between theta_m and theta_p.  The crack curves up which would be theta_p, then it extends straight which would be theta_m from the previous crack tip direction.  Maybe the crack tip direction is not being updated between xfem iterations.

tranm-ansto:
> For the crack to be jagged like that, it has to be changing between theta_m and theta_p. The crack curves up which would be theta_p, then it extends straight which would be theta_m from the previous crack tip direction. Maybe the crack tip direction is not being updated between xfem iterations.

I tried printing out the theta_m and theta_p, it seems that theta_m is being picked. I didn't see switching between theta_m and theta_p.

tranm-ansto:
> Here is the plot with the ki and kii points show by the dots. The points are for the mesh_cut_2D_fracture director test [k_growth] These values of kii results in large angle changes for small changes in kii.
> 
> Screenshot 2025-02-25 at 10 58 02 AM

The angle theta is now much more sensitive to kii.

jiangwen84:
> > > > > was the issue with jagged crack resolved?
> > > > 
> > > > 
> > > > Also check the calculation of K. Is the choice of radius reasonable?
> > > 
> > > 
> > > Thank you for the pointer.
> > > > > was the issue with jagged crack resolved?
> > > > 
> > > > 
> > > > Also check the calculation of K. Is the choice of radius reasonable?
> > > 
> > > 
> > > The radius is quite big. q function shows that the contour ring is outside of the domain. With the correction in theta, the crack is now more sensitive to KII.
> > 
> > 
> > I do not think it would work if the contour ring is outside the domain.
> 
> Sorry. I meant part of the contour ring is outside. image image

This looks reasonable to me.

jiangwen84:
Is this the Mode I crack propagation? We might need to set a tolerance for the small KII. How about refining the mesh?

tranm-ansto:
> Is this the Mode I crack propagation? We might need to set a tolerance for the small KII. How about refining the mesh?

The crack is still jagged even with mesh refinement. The current loading induces mixed-mode crack propagation. I wonder if it is because of the loading and BC's. If I grab more nodes on the top, the left crack becomes smoother.
image

lynnmunday:
@tranm-ansto I have to take care of some other projects and won't be able to work on this for a few weeks.  I see no problem in the code in this PR.  The problem seems to be the code I wrote.  I'll have to figure it out.

tranm-ansto:
> @tranm-ansto I have to take care of some other projects and won't be able to work on this for a few weeks. I see no problem in the code in this PR. The problem seems to be the code I wrote. I'll have to figure it out.

No problem, Lynn. We can reconvene in a few weeks.

lynnmunday:
@bwspenc Will you reopen this?

lynnmunday:
Screenshot 2025-12-03 at 10 45 43 AM
Screenshot 2025-12-03 at 10 46 08 AM
This is the result using Minh's code with a mesh size, crack growth length and q integral radius all reduced by 4.   The crack is smooth.

bwspenc:
This has sat for a while, so there are merge conflicts. I'm going to attempt to force push from my local project, where I have the commit from @tranm-ansto plus a commit from @lynnmunday that updates some of the tests.

bwspenc:
Looks like I don't have access to push to this branch, so that's not going to work. Either @tranm-ansto will need to update this or we'll need to open a new PR.

bwspenc:
Also, we should note that we found that when we decrease the growth increment, we match published mixed-mode fracture propagation results.

framework/contrib/neml2
lynnmunday, line 1: Will you delete this and repush?  I might have accidently added it.

lynnmunday (changes requested):

framework/contrib/neml2
tranm-ansto: How do I delete the neml2 submodule update?

lynnmunday: let me figure that out and push to your branch again.

lynnmunday: You just need to drop this commit that you made:

commit 2c212d2fc4a4771036846fffff686cab3896fb73
Author: ‘tranm-ansto’ <‘tranm@ansto.gov.au’>
Date:   Thu Dec 4 10:50:44 2025 +1100

    update moose;lynn change

lynnmunday: The only thing I see in that commit is a neml2 submodule udpate

lynnmunday: > git reset --hard 640ef857d0ba728b5fdf14db414f88cd614caa06
> git push -f origin xfem_meshcut2d_fracture_uo

bwspenc (approved):

PR #29935, opened 2025-02-21
A feature to add Delaunay mesh for the inner ring of a polygon concentric mesh

### LINKED ISSUE #29934: Add an option to have a Delaunay mesh in the inner ring of polygon concentric meshes
### Motivation

If the number of sectors per side in a mesh generated by polygon concentric mesh generators, the center of the mesh can connect too many elements resulting an undesirable mesh like the following:

Image

### Design

This issue can be resolved by replacing the current inner mesh with a Delaunay mesh generated with Poly2Tri:

Image

### Impact

The picture I showed above is a control drum assembly. The new feature can make the mesh significantly better.

YaqiWang:
This closes #29934.

The mesh without this in the test:

Screenshot 2025-02-21 at 2 32 42 PM

The mesh with Delaunay with a uniform mesh:

Screenshot 2025-02-21 at 2 32 09 PM

The mesh with Delaunay with a varying element size:

Screenshot 2025-02-21 at 2 39 22 PM


miaoyinb:
I can review this one.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,110 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+
+      mesh0->delete_elem(elem);
+    }
+
+    // build the 1d mesh from the new boundary
+    libMesh::ReplicatedMesh poly_mesh(comm());
+    std::vector<dof_id_type> element_id_list;
+    std::vector<unsigned short int> side_list;
+    std::vector<boundary_id_type> bc_id_list;
+    boundary_info.build_side_list(element_id_list, side_list, bc_id_list);
```
miaoyinb, line 988: Need to use `std::vector<BoundaryInfo::BCTuple> BoundaryInfo::build_side_list()` to avert the failed `libMesh disable deprecated` test

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,110 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
   if (flat_side_up)
     MeshTools::Modification::rotate(*mesh0, 180.0 / (Real)_num_sides, 0.0, 0.0);
   mesh0->set_isnt_prepared();
+
+  if (_ring_block_ids.size() > 2 && getParam<bool>("replace_inner_ring_with_delaunay_mesh"))
```
miaoyinb, line 952: So the new feature is only applicable when you have more than two ring blocks?

(just some initial single comments. I will do a full review soon, thanks)

YaqiWang: I actually did not dig this much. I thought the first two elements of `_ring_block_ids` are for the inner ring because the inner ring may contain triangle and quad elements. This check is to insure that there is indeed a ring.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,112 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+  {
+    // remove elements of the inner ring and add the boundary side set
```
GiudGiud, line 954: are there are any EEIDs that need to be copied over?
```suggestion
    // remove elements of the inner ring and create an outer-pin boundary side set
```

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,112 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+
+    const boundary_id_type boundary_id = MooseMeshUtils::getBoundaryIDs(*mesh0, {"_foo"}, true)[0];
+    BoundaryInfo & boundary_info = mesh0->get_boundary_info();
+    for (auto & elem : deleteable_elems)
+    {
+      unsigned int n_sides = elem->n_sides();
+      for (unsigned int n = 0; n != n_sides; ++n)
+      {
+        Elem * neighbor = elem->neighbor_ptr(n);
+        if (!neighbor)
```
GiudGiud, line 968: why would this happen? should there be an assert if we dont expect it it to happen?

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,112 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+    {
+      poly2tri.desired_area() = 0;
+      libMesh::ParsedFunction<Real> area_func{desired_area_func};
+      poly2tri.set_desired_area_function(&area_func);
+    }
+    else
+      poly2tri.desired_area() = min_side * min_side;
+    poly2tri.minimum_angle() = 0;
+    poly2tri.smooth_after_generating() = true;
+    poly2tri.elem_type() = libMesh::ElemType::TRI3;
```
GiudGiud, line 1046: is there a second order option in that generator we should obey?

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: Yes, this looks should not happen unless the pin is cut before reaching this point.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: Possibly yes. We may have ring id and sector id. I will add the code for this.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: Does not seem we have any for this mesh generator.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: eh, I am not so sure. I think this can happen during the process of removing element one by one.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: Added ring_id and sector_id when they are requested.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
miaoyinb: PCCMG supports second order elements.

`poly2tri.elem_type()` can also be set as TRI6 or TRI7.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,135 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
   if (flat_side_up)
     MeshTools::Modification::rotate(*mesh0, 180.0 / (Real)_num_sides, 0.0, 0.0);
   mesh0->set_isnt_prepared();
+
+  if (_ring_block_ids.size() > 2 && getParam<bool>("replace_inner_ring_with_delaunay_mesh"))
+  {
+    // remove elements of the inner ring and create an inner-ring external boundary side set
+    std::set<Elem *> deleteable_elems;
+    for (auto & elem : mesh0->element_ptr_range())
+      if (elem->vertex_average().norm() < _ring_radii[0])
```
miaoyinb, line 957: We should know the block_id of the inner ring at this point. So `active_subdomain_elements_ptr_range()` can be used.

Also, `_ring_radii` might not be safe to use when volume preserve is used.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,135 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+      poly2tri.desired_area() = min_side * min_side;
+    poly2tri.minimum_angle() = 0;
+    poly2tri.smooth_after_generating() = true;
+    poly2tri.elem_type() = libMesh::ElemType::TRI3;
+    // let us keep the center point
+    poly_mesh.add_point(libMesh::Point());
+    poly2tri.triangulate();
+    // keep the old subdomain id
+    for (auto elem : poly_mesh.element_ptr_range())
+      elem->subdomain_id() = _ring_block_ids[0];
```
miaoyinb, line 1052: As mentioned before, `_ring_block_ids` is not necessarily non-empty as it is from an optional input parameter. `block_ids_new` is set.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +948,135 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+
+    if (isParamValid("ring_id_name"))
+    {
+      if (_ring_intervals[0] != 1 && getParam<MooseEnum>("ring_id_assign_type") == "ring_wise")
+        paramError("replace_inner_ring_with_delaunay_mesh",
+                   "Inner ring has multple intervals with each being assigned with a different "
+                   "ring id, replacing inner ring with Delaunay mesh will remove this ring id "
+                   "assign type. Either change 'ring_id_assign_type' to block_wise or set the "
+                   "first element of 'ring_intervals' to 1 or set "
```
miaoyinb, line 1061: I think we should enforce the first element of `ring_intervals` to be 1 when this replacement is used.

miaoyinb (changes requested):
Here is my first round of review. Thanks.

Ideally, we should (1) disable the original generation of inner ring and/or (2) have a general method to replace a "single-piece" XY-block with a XY-Delaunay mesh. These are too much for this PR. I may work on these in the future.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -130,6 +134,19 @@ PolygonConcentricCircleMeshGeneratorBase::validParams()
...
+  params.addParam<bool>("replace_inner_ring_with_delaunay_mesh",
+                        false,
+                        "True to replace the inner ring mesh with a Delaunay unstructured mesh");
+  params.addRangeCheckedParam<Real>("inner_ring_desired_area",
+                                    0,
+                                    "inner_ring_desired_area>=0",
+                                    "Desired (maximum) triangle area, or 0 to use the default area "
+                                    "(square of the smallest side length on the ring circle)");
+  params.addParam<std::string>(
+      "inner_ring_desired_area_func",
```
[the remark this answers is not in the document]
YaqiWang: If the desired area is zero and this function is empty, the code will use a default desired area, evaluated based on the side length of the outer circle of the inner ring.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: Would there possibly two block ids for the elements in the ring?

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
miaoyinb: Yes, not enforcing this would make the input switch convenient. But it would also make the input syntax confusing for new users.

Maybe we can override the user provided input parameters when this replacement is used.

Anyway, it occurred to me that I forgot to mention that this new feature needs to be described in the documentation.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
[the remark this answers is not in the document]
miaoyinb: That is okay. Please provide a description in documentation.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -130,6 +134,19 @@ PolygonConcentricCircleMeshGeneratorBase::validParams()
       false,
       "Whether to rotate the generated polygon mesh to ensure that one flat side faces up.");
 
+  params.addParam<bool>("replace_inner_ring_with_delaunay_mesh",
```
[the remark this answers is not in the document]
YaqiWang: Add a check on `quad_center_elements` to warn users and keep the first element of `ring_intervals` ignored. Please let me know if you have strong opinions on this.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: Changed to `block_ids_new`.

YaqiWang:
Pushed a paragraph in the markdown file. If approved, I will push that png file into large_media.
poly_inner_delaunay

YaqiWang:
Should I go ahead to push the above picture in `large_media`?

modules/reactor/doc/content/source/meshgenerators/PolygonConcentricCircleMeshGenerator.md
```diff
@@ -75,6 +75,15 @@ By default, the mesh generated by this object contains a mixture of QUAD4 and TR
...
...n the number of total sectors is large making the center point connected with too many elements. When this parameter is true, [!param](/Mesh/PolygonConcentricCircleMeshGenerator/quad_center_elements) is ignored and users will get a warning if this parameter is set to true. The first element of [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_intervals) is ignored and the element areas are controlled by [!param](/Mesh/PolygonConcentricCircleMeshGenerator/inner_ring_desired_area) function. The function can be a constant value. The default empty string means that the code will pick a default desired area which is the square of the smallest side length on the ring circle. The unstructured Delaunay mesh of the inner ring will have the triangle element type specified in [!param](/Mesh/PolygonConcentricCircleMeshGenerator/tri_element_type). It will also have the subdomain ID of the original subdomain ID of the inner ring. When [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_id_name) is specified, the element ring IDs will be all equal to 1. When [!param](/Mesh/PolygonConcentricCircleMeshGenerator/sector_id_name) is specified, the element sector IDs will be all equal to 0.
```
miaoyinb, line 85: There are parameters in addition to `quad_center_elements` and `ring_intervals`, that will be ignored. Including the block id and name for the quad elements of the inner ring, as well as the mesh bias and boundary layer parameters

miaoyinb:
> Should I go ahead to push the above picture in `large_media`?

looks good to me.

modules/reactor/doc/content/source/meshgenerators/PolygonConcentricCircleMeshGenerator.md
YaqiWang: Updated. Actually, the code uses the value of the first element in `ring_block_ids` to assign block ids and similarly with `ring_block_names`.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: Updated to let `poly2tri` take `_tri_elem_type`.

YaqiWang:
@miaoyinb and @GiudGiud can you check whether your comments are properly addressed/answered?

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +944,136 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+  {
+    if (getParam<bool>("quad_center_elements"))
+      paramWarning("quad_center_elements",
+                   "ignored because the inner ring will be replaced by a Delaunay mesh with "
+                   "'replace_inner_ring_with_delaunay_mesh=true'");
+
+    // remove elements of the inner ring and create an inner-ring external boundary side set
+    std::set<Elem *> deleteable_elems;
+    for (auto & elem : mesh0->element_ptr_range())
+      if (elem->subdomain_id() == block_ids_new[0] || elem->subdomain_id() == block_ids_new[1])
```
miaoyinb, line 958: the inner ring includes block_ids_new[1] only when `_ring_inner_boundary_layer_params.intervals.front() + _ring_intervals.front() + _ring_outer_boundary_layer_params.intervals.front()>1`,

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +944,136 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+    {
+      poly2tri.desired_area() = 0;
+      libMesh::ParsedFunction<Real> area_func{desired_area};
+      poly2tri.set_desired_area_function(&area_func);
+    }
+    else
+      poly2tri.desired_area() = min_side * min_side;
+    poly2tri.minimum_angle() = 0;
+    poly2tri.smooth_after_generating() = true;
+    poly2tri.elem_type() = libMesh::ElemType::TRI3;
```
miaoyinb, line 1039:
```suggestion
    # poly2tri.elem_type() is TRI3 by default
```

miaoyinb (changes requested):

modules/reactor/doc/content/source/meshgenerators/PolygonConcentricCircleMeshGenerator.md
miaoyinb: if `ring_intervals[0]`>1, the second elements of `ring_block_ids` and `ring_block_names` are ignored. Also, if the new feature is used, the first elements of
`ring_inner_boundary_layer_biases`
`ring_inner_boundary_layer_intervals`
`ring_inner_boundary_layer_widths`
`ring_outer_boundary_layer_biases`
`ring_outer_boundary_layer_intervals`
`ring_outer_boundary_layer_widths`
`ring_radial_biases`
are all ignored.

modules/reactor/doc/content/source/meshgenerators/PolygonConcentricCircleMeshGenerator.md
```diff
@@ -75,6 +75,15 @@ By default, the mesh generated by this object contains a mixture of QUAD4 and TR
...
...[!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_intervals) is ignored and the element areas are controlled by [!param](/Mesh/PolygonConcentricCircleMeshGenerator/inner_ring_desired_area) function. The function can be a constant value. The default empty string means that the code will pick a default desired area which is the square of the smallest side length on the ring circle. The unstructured Delaunay mesh of the inner ring will have the triangle element type specified in [!param](/Mesh/PolygonConcentricCircleMeshGenerator/tri_element_type). It will also have the subdomain ID/name of the original subdomain ID/name of the inner ring, the first value in [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_block_ids) and [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_block_names). When [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_id_name) is specified, the element ring IDs will be all equal to 1. When [!param](/Mesh/PolygonConcentricCircleMeshGenerator/sector_id_name) is specified, the element sector IDs will be all equal to 0. Other parameters such as the mesh bias and boundary layer parameters do not have effect on the unstructured Delaunay mesh.
```
miaoyinb, line 85:
```suggestion
This feature is useful in a situation when the number of total sectors is large making the center point connected with too many elements. When this parameter is true, [!param](/Mesh/PolygonConcentricCircleMeshGenerator/quad_center_elements) is ignored and users will get a warning if this parameter is set to true. The first element of [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_intervals) is ignored and the element areas are controlled by [!param](/Mesh/PolygonConcentricCircleMeshGenerator/inner_ring_desired_area) function. The function can be a constant value. The default empty string means that the code will pick a default desired area which is the square of the smallest side length on the ring circle. The unstructured Delaunay mesh of the inner ring will have the triangle element type specified in [!param](/Mesh/PolygonConcentricCircleMeshGenerator/tri_element_type). It will also have the subdomain ID/name of the original subdomain ID/name of the triangular elements of the inner ring, the first value in [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_block_ids) and [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_block_names). When [!param](/Mesh/PolygonConcentricCircleMeshGenerator/ring_id_name) is specified, the element ring IDs will be all equal to 1. When [!param](/Mesh/PolygonConcentricCircleMeshGenerator/sector_id_name) is specified, the element sector IDs will be all equal to 0. Other parameters such as the mesh bias and boundary layer parameters do not have effect on the unstructured Delaunay mesh.
```

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: The test seems running okay. Is this condition the case for the test?

miaoyinb: The test does run well. If you have `_ring_intervals[0]>1`, this `if` will work fine. Otherwise, this will cause issues.

YaqiWang: Actually hold on, I remembered I choose my original logic because rings may have the same block ids specified by users, even the background region could be the same. I think the original logic is safe typically. Do you have a better solution for this?

miaoyinb: You are right, the block ids might not be unique.

If you want to use the old approach, you need to use `ring_radii_corr` instead of `_ring_radii`.

OR, if the replacement is used, modify the first or first/second elements of `block_ids_new` to max(block_ids_new)+1/+2 before `assignBlockIdsNames()`

YaqiWang: I like the later approach. It can be easily done.

YaqiWang: The last approach will have implication on the block names. It is a lot easier to just check the radius.

GiudGiud:
so in the figures you posted you always replaced the two inner rings with tris. Can you replace a single ring?

YaqiWang:
> so in the figures you posted you always replaced the two inner rings with tris. Can you replace a single ring?

@GiudGiud that is because the number of intervals of the first ring is 2 in this case. The change here is always replacing a single ring.

modules/reactor/doc/content/source/meshgenerators/PolygonConcentricCircleMeshGenerator.md
miaoyinb: I think this is the only unresolved comment of mine. Please add these to the doc.

@GiudGiud Do you want to do an additional round of review?

YaqiWang: Oh, I thought one sentence without this list is fine. I will push this then.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +944,136 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
   if (flat_side_up)
     MeshTools::Modification::rotate(*mesh0, 180.0 / (Real)_num_sides, 0.0, 0.0);
   mesh0->set_isnt_prepared();
+
+  if (_has_rings && getParam<bool>("replace_inner_ring_with_delaunay_mesh"))
+  {
+    if (getParam<bool>("quad_center_elements"))
+      paramWarning("quad_center_elements",
```
GiudGiud, line 951:
```suggestion
      paramError("quad_center_elements",
```

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -130,6 +127,22 @@ PolygonConcentricCircleMeshGeneratorBase::validParams()
...
+  params.addRangeCheckedParam<Real>(
+      "center_quad_factor",
+      "center_quad_factor>0&center_quad_factor<1",
+      "A fractional radius factor used to determine the radial positions of transition nodes in "
+      "the center region meshed by quad elements.");
+  params.addParam<bool>("replace_inner_ring_with_delaunay_mesh",
+                        false,
+                        "True to replace the inner ring mesh with a Delaunay unstructured mesh");
+  params.addParam<std::string>(
+      "inner_ring_desired_area",
```
GiudGiud, line 141: add a check that this parameter is not passed if the one above is false

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +944,136 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+      Elem * elem = mesh0->elem_ptr(std::get<0>(bside));
+      auto side = std::get<1>(bside);
+      std::unique_ptr<Elem> side_elem(elem->build_side_ptr(side));
+      auto copy = side_elem->build(side_elem->type());
+
+      Real l = side_elem->volume();
+      if (l < min_side)
+        min_side = l;
+
+      for (dof_id_type i : side_elem->node_index_range())
```
GiudGiud, line 1004:
```suggestion
      for (const auto i : side_elem->node_index_range())
```

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
```diff
@@ -928,5 +944,136 @@ PolygonConcentricCircleMeshGeneratorBase::generate()
...
+        {
+          neighbor->set_neighbor(return_side, nullptr);
+          boundary_info.add_side(neighbor, return_side, boundary_id);
+        }
+      }
+
+      mesh0->delete_elem(elem);
+    }
+
+    // build the 1d mesh from the new boundary
```
GiudGiud, line 984: the next 40 lines look like they should be a utility instead of being coded here

modules/reactor/test/tests/meshgenerators/polygon_concentric_circle_mesh_generator/tests
```diff
@@ -170,6 +170,37 @@
...
+    cli_args = 'Mesh/gen/background_block_ids="30" Mesh/gen/background_block_names="back"
+                Mesh/gen/ring_radii="2 3" Mesh/gen/ring_intervals="2 2"
+                Mesh/gen/ring_block_ids="10 15 20" Mesh/gen/ring_block_names="center center_quad outer_ring"
+                Mesh/gen/replace_inner_ring_with_delaunay_mesh=true
+                Mesh/gen/uniform_mesh_on_sides=true
+                Mesh/gen/outward_interface_boundary_names="int1 int2"
+                Mesh/gen/ring_id_name=ring_id
+                Mesh/gen/sector_id_name=sector_id
+                --mesh-only "poly_2d_delaunay.e"'
+    requirement = 'The system shall generate a hexagon mesh with rings with a Delaunay mesh in the inner ring.'
```
GiudGiud, line 186:
```suggestion
    requirement = 'The system shall generate a hexagonal pin mesh with rings with a Delaunay mesh in the inner ring.'
```

modules/reactor/test/tests/meshgenerators/polygon_concentric_circle_mesh_generator/tests
```diff
@@ -170,6 +170,37 @@
...
+    exodiff = 'poly_2d_delaunay_func.e'
+    cli_args = 'Mesh/gen/background_block_ids="30" Mesh/gen/background_block_names="back"
+                Mesh/gen/ring_radii="2 3" Mesh/gen/ring_intervals="2 2"
+                Mesh/gen/ring_block_ids="10 15 20" Mesh/gen/ring_block_names="center center_quad outer_ring"
+                Mesh/gen/replace_inner_ring_with_delaunay_mesh=true
+                Mesh/gen/inner_ring_desired_area=(1-(x*x+y*y)/2/2)+(x*x+y*y)/2/2*0.2
+                Mesh/gen/uniform_mesh_on_sides=true
+                Mesh/gen/outward_interface_boundary_names="int1 int2"
+                --mesh-only "poly_2d_delaunay_func.e"'
+    requirement = 'The system shall generate a hexagon mesh with rings with Delaunay mesh with spatially varying element sizes in the inner ring.'
```
GiudGiud, line 201:
```suggestion
    requirement = 'The system shall generate a hexagonal pin mesh with rings with an inner ring consisting of a Delaunay mesh with spatially varying element sizes.'
```

GiudGiud (commented):
looks good almost ready

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
miaoyinb: If we are to make this a utility, these lines will need to be coordinated with the existing code in `LowerDBlockFromSidesetGenerator` to be a single utility.

Same thing could be done the for elements deletion lines above along with the corresponding code in `ElementDeletionGeneratorBase`

framework/src/utils/MooseMeshUtils.C
```diff
@@ -571,4 +571,45 @@ extraElemIntegerSwapParametersProcessor(
                                     elem_integer_swap_pairs.end());
   }
 }
+
+std::unique_ptr<ReplicatedMesh>
+buildBoundaryMesh(const MeshBase & input_mesh, const boundary_id_type boundary_id)
+{
+  auto poly_mesh = std::make_unique<ReplicatedMesh>(input_mesh.comm());
```
GiudGiud, line 578: we should probably take in a ReplicatedMesh reference instead

YaqiWang: Why? I think this is a cleaner interface.

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
miaoyinb: If this is an error instead of a warning, the document also needs to be updated as it now says "ignored".

framework/src/utils/MooseMeshUtils.C
GiudGiud: does this operation serialize the mesh?

YaqiWang: No, because the input mesh is a replicated mesh. I actually changed the input mesh type from `MeshBase` to `ReplicatedMesh` to avoid confusion.

modules/reactor/test/tests/meshgenerators/polygon_concentric_circle_mesh_generator/tests
```diff
@@ -1,12 +1,12 @@
...
   design = 'meshgenerators/PolygonConcentricCircleMeshGenerator.md'
-  issues = '#18628 #20514 #20637 #21420 #23367 #23520 #24025 #26986'
+  issues = '#18628 #20514 #20637 #21420 #23367 #23520 #24025 #26986 #29934'
   [back_only]
     type = 'Exodiff'
     input = 'poly_2d.i'
     exodiff = 'poly_2d_back_only.e'
     cli_args = 'Mesh/gen/background_block_ids="30 35" Mesh/gen/background_block_names="back_in back_out" --mesh-only "poly_2d_back_only.e"'
-    requirement = 'The system shall generate a hexagon mesh without rings and ducts.'
+    requirement = 'The system shall generate a hexagonal mesh without rings and ducts.'
```
GiudGiud, line 9: in order I like:
hexagonal-lattice pin mesh > hexagonal pin mesh > hexagon mesh > hexagonal mesh

modules/reactor/src/meshgenerators/PolygonConcentricCircleMeshGeneratorBase.C
YaqiWang: dang. Updated in the last push.

modules/reactor/test/tests/meshgenerators/polygon_concentric_circle_mesh_generator/tests
YaqiWang: I do not quite like `pin` because it is a little too specific. The mesh may not represent a pin. You want me to revert this back to original `hexagon mesh`? This does not look very different to me.

GiudGiud: hexagonal-lattice concentric circle mesh then?

YaqiWang: I do not quite like including `lattice` word in this name. How about I restore this name to its original one? I do not have a good suggestion. Maybe based on the name of the mesh generator, it should be `polygon concentric circle mesh`. Which one you think better?

GiudGiud: hexagon concentric circle mesh then

GiudGiud (dismissed):

miaoyinb:
@YaqiWang 

It still fails the non-unity build. It seems some header file is not included.
The documentation building also seems to miss some figure.

Please fix these and then I can merge this.

YaqiWang:
I am guessing I do not have the correct syntax in the markdown file to refer the figure? It will be great if you know what the exact issue is and give me a help.

miaoyinb:
You did not update the large_media submodule, so your uploaded figures cannot be seen by the system.

GiudGiud (approved):

miaoyinb:
@GiudGiud Can you check the controlled apps related failed tests? I cannot see them.

Other failed tests seem unrelated.

GiudGiud:
I can confirm. no related failures. SAM build failure is from a transfers PR of mine. NS threads is stochastic

miaoyinb (approved):

PR #29948, opened 2025-02-25
Add framework-level, userfacing API for loading and evaluating torchscript objects

### LINKED ISSUE #30107: Add user-facing objects and interfaces for loading torchscript objects into MOOSE.
### Motivation

This has been requested from multiple projects. The ability to load torch script models into MOOSE enables UQ/ML. We also want to ensure that these models are loadable on the fly at different `execute_on` options. Ideally, every system should have acces to this.

### Design

Use a userobject to store the torchscript object for now.

### Impact

Will enable rapid prototyping with externally trained neural networks/GPs/regression models etc.

grmnptr:
Closes #30107


modules/solid_mechanics/include/materials/TorchPlasticityStressUpdate.h
```diff
@@ -0,0 +1,78 @@
+//* This file is part of the MOOSE framework
```
bwspenc, line 1: As we discussed, this model and the related files should either go into solid_mechanics/test/ or into the repository for the SciDAC. I'd say the SciDAC repository is probably the better place at the moment. Maybe we can make a simple model based on this that is actually physical that could go in the solid_mechanics module.

bwspenc (changes requested):

framework/src/libtorch/materials/TorchScriptMaterial.C
```diff
@@ -0,0 +1,80 @@
...
+
+InputParameters
+TorchScriptMaterial::validParams()
+{
+  InputParameters params = Material::validParams();
+  params.addClassDescription("Material object which relies on the evaluation of a neural network read using the TorchScript format.");
+  params.addRequiredParam<std::vector<std::string>>("prop_names",
+                                            "The names of the properties this material will generate.");
+  params.addRequiredParam<std::vector<PostprocessorName>>("input_names", "The input parameters for the neural network.");
+  params.addRequiredParam<UserObjectName>("torch_script_uo", "The name of the user object which contains the torch script network.");
```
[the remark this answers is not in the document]
grmnptr: I found examples of both `_uo_` and `_user_object_`. I am seeing multiple `_uo`-based parameter names in the solid mechanics, phase field, thermal hydraluics and richards modules so I will go with this. Of course there are other examples too within other modules like xfem.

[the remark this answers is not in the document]
grmnptr: Ahhh, I think I'll spell it out instead. I feel like that is probably better.

framework/src/libtorch/userobjects/TorchScriptUserObject.C
```diff
@@ -0,0 +1,42 @@
...
+
+#include "TorchScriptUserObject.h"
+
+registerMooseObject("MooseApp", TorchScriptUserObject);
+
+InputParameters
+TorchScriptUserObject::validParams()
+{
+  InputParameters params = GeneralUserObject::validParams();
+  params.addRequiredParam<std::string>("filename", "The file name which contains the saved neural network.");
```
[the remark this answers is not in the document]
grmnptr: Yeah I changed this!

framework/src/libtorch/userobjects/TorchScriptUserObject.C
```diff
@@ -0,0 +1,42 @@
...
+}
+
+void TorchScriptUserObject::execute()
+{
+  _nn = std::make_shared<Moose::LibtorchTorchScriptNeuralNet>(_filename);
+}
+
+torch::Tensor TorchScriptUserObject::evaluate(torch::Tensor & input) const
+{
+  mooseAssert(_nn, "We need a neural network to evalute!");
```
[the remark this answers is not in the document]
grmnptr: The function that loads the torch module will throw and error if it cannot load it. This is just protecting against evaluating the neural network before it is read. For example, if the execute_on function of the object is timestep_end and somebody wants to evaluate it at initial.

framework/doc/content/source/libtorch/userobjects/TorchScriptUserObject.md
```diff
@@ -0,0 +1,20 @@
+# TorchScriptUserObject
+
+!syntax description /UserObjects/TorchScriptUserObject
+
+## Description
+
+This user object loads a torch module that has been exported from
+pytorch in a torch script format. The user object loads the modules
+when the user object is execute which can be controlled using the
+`execute_on` flag.
```
hugary1995, line 10: It'll be nice to link to an external documentation explaining what "torch script" is.

framework/include/libtorch/userobjects/TorchScriptUserObject.h
```diff
@@ -0,0 +1,53 @@
+//* This file is part of the MOOSE framework
+//* https://www.mooseframework.org
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#ifdef LIBTORCH_ENABLED
```
hugary1995, line 10: I'm curious about @loganharbour 's opinion on this. If we ifdef guard the entire file, this object won't be available when libtorch is not enabled. This may create several issues:
- user won't be able to autocomplete using language server
- some civet recipe might not like this

framework/include/libtorch/userobjects/TorchScriptUserObject.h
```diff
@@ -0,0 +1,53 @@
...
+  TorchScriptUserObject(const InputParameters & parameters);
+
+  virtual void initialize() override{};
+  virtual void execute() override;
+  virtual void finalize() override{};
+
+  /**
+   * Get access to the torch script module which was loaded in this object.
+   */
+  const std::shared_ptr<Moose::TorchScriptModule> & module() const { return _torchscript_module; }
```
hugary1995, line 36: How about adding two methods with the following signature?
```cpp
const std::shared_ptr<Moose::TorchScriptModule> & modulePtr() const;
const Moose::TorchScriptModule & module() const;
```
the second one dereferences the smart ptr. In my experience this is useful from times to times.

framework/include/libtorch/userobjects/TorchScriptUserObject.h
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Function to evaluate the torch script module at certain input.
+   * @param input The input tensor. Unfortunately, this cannot be const since it creates a graph in
+   * the background.
+   */
+  torch::Tensor evaluate(torch::Tensor & input) const;
```
hugary1995, line 43: Why not `const torch::Tensor & input`?

framework/src/libtorch/userobjects/TorchScriptUserObject.C
```diff
@@ -0,0 +1,57 @@
...
+#include "TorchScriptUserObject.h"
+
+registerMooseObject("MooseApp", TorchScriptUserObject);
+
+InputParameters
+TorchScriptUserObject::validParams()
+{
+  InputParameters params = GeneralUserObject::validParams();
+  params.addClassDescription("User-facing object which loads a torch script module.");
+  params.addRequiredParam<std::string>("filename",
```
hugary1995, line 21: Shouldn't this be `FileName` instead of `std::string`?

framework/src/libtorch/userobjects/TorchScriptUserObject.C
```diff
@@ -0,0 +1,57 @@
...
+      "If we want to load this neural network while we are constructing this object.");
+  return params;
+}
+
+TorchScriptUserObject::TorchScriptUserObject(const InputParameters & parameters)
+  : GeneralUserObject(parameters),
+    _filename(getParam<std::string>("filename")),
+    _torchscript_module(getParam<bool>("load_during_construction")
+                            ? std::make_shared<Moose::TorchScriptModule>(_filename)
+                            : nullptr)
```
hugary1995, line 36: What's the downside of always loading the torch script at construction time? I think if the user uses this object, it is implied that the script module _will_ be used eventually.

framework/src/libtorch/userobjects/TorchScriptUserObject.C
```diff
@@ -0,0 +1,57 @@
...
+{
+}
+
+void
+TorchScriptUserObject::execute()
+{
+  // We load when the user executes this user object
+  _torchscript_module = std::make_shared<Moose::TorchScriptModule>(_filename);
```
hugary1995, line 44: Guard this with `if (!_torchscript_module)`

framework/src/libtorch/userobjects/TorchScriptUserObject.C
```diff
@@ -0,0 +1,57 @@
...
+}
+
+torch::Tensor
+TorchScriptUserObject::evaluate(torch::Tensor & input) const
+{
+  if (!_torchscript_module)
+    mooseError(
+        "The torch script module has not been loaded yet, so it can't be evaluated! Make sure that "
+        "the torch script module is loaded (execute_on flags) before calling this function!");
```
hugary1995, line 53: If we load on construction this can never happen. I guess this could be a mooseAssert.

test/tests/materials/torchscript_material/generate_net.py
hugary1995, line 1: Should we add a license statement at the top?

test/tests/materials/torchscript_material/generate_net.py
```diff
@@ -0,0 +1,20 @@
+import torch
+import torch.nn as nn
+
+torch.manual_seed(42)
```
hugary1995, line 4: In my experience manual_seed isn't reproducible on different architectures. As long as CI doesn't explicitly run this python script it should be fine. But I think in the long term it's better to check in a torch script with saved weights, rather than relying on a python script to generate the network.

hugary1995 (changes requested):

framework/include/libtorch/userobjects/TorchScriptUserObject.h
loganharbour: In general, I do suggest that we have all of the objects (in a bare bones, error-on-construction state) available even when the capabilities do not exist. Same for the material object.

grmnptr: We actually had a design discussion with @loganharbour and @zachmprince when the original implementation went in. The original design was what you are suggesting but based on that discussion we moved to this approach. I am okay changing it back, but in another PR in the future.

framework/include/libtorch/userobjects/TorchScriptUserObject.h
grmnptr: Sounds good, adding it in the next commit!

framework/src/libtorch/userobjects/TorchScriptUserObject.C
grmnptr: Oh yes, great catch!

framework/include/libtorch/userobjects/TorchScriptUserObject.h
hugary1995: The intermediate tensors you create during forward evaluation based on the input tensor will carry over the graph. That's not a problem I think. You could easily verify this.

framework/src/libtorch/userobjects/TorchScriptUserObject.C
grmnptr: Hmmm good point, I will just assume there is already a neural net when we set up the problem. I am thinking about peple using this together with the webserver interface and saving before the start of it. But I assume it is pretty easy to save at the very beginning of the python run before they fire up the moose run.

test/tests/materials/torchscript_material/generate_net.py
grmnptr: Ahhh good point. In general, I think the files that are part of the doco/testharness have it. For the files that are only there for the testing (like MMS scripts, scripts for plotting, scrits for sympying analytic soltuions), we usually don't add them. I'll ask for guidance from more senior guys here.

test/tests/materials/torchscript_material/generate_net.py
grmnptr: Oh wow, that's interesting. Thanks for the warning! Yeah right now we are checking in the script file as well.

framework/src/libtorch/userobjects/TorchScriptUserObject.C
hugary1995: Okay I see what the use case is now. I think there are two ways forward then
1. Add a boolean option to allow "reloading" the script whenever the userobject is executed.
2. Never reload. Then implement another object deriving from this one which always reloads.

Either way is fine with me. If we go with the first approach, I'd suggest renaming the option to "reload" or "always_reload" which might be slightly more intuitive to users.

framework/include/libtorch/userobjects/TorchScriptUserObject.h
grmnptr: Will look into it!

framework/src/libtorch/userobjects/TorchScriptUserObject.C
grmnptr: Or how about an option:
```
addParam<ExecuteFlag>(reload_module, default(will be never), "Something");
```

And then:
```
if (_reload_flags.find(_current_execute_flag) != _reload_flags.end())
  reloadthemodule;
```

grmnptr: You know what, I will just override the `execute_on` flags in here to never. And then the user can control the reload frequency through the `executo_on` flags.

framework/src/libtorch/userobjects/TorchScriptUserObject.C
hugary1995: I'm fine with a boolean knob now that I understand the intended use case.

framework/include/libtorch/userobjects/TorchScriptUserObject.h
grmnptr: Nice, you were right! Will push the fix in the next commit.

test/tests/materials/torchscript_material/generate_net.py
grmnptr: Added!

framework/include/libtorch/userobjects/TorchScriptUserObject.h
dschwen: The problem with fully deactivating is that the documentation generation will error out with "unknown syntax". If we had a fix for this I'd be in favor of fully guarding the objects, because creating an "error on construction" stub can be a _lot_ of tedious work, and leads to files that are chock full of ifdefs.

hugary1995 (approved):
Contingent approval s.t. better code coverage and all tests passing.

framework/include/libtorch/userobjects/TorchScriptUserObject.h
grmnptr: If you are talking about the language server I am not sure about a solution on that front. 
If you are talking about moosedocs, we guard the documentation with:

```
!if function=hasLibtorch()
```

hugary1995: That's not a bad approach IMO. Maybe this could be generalized to the new "capability" system with something like
```
!if function=hasCapability("libtorch")
```

grmnptr: Well, we also had a design discussion with @loganharbour and @zachmprince and decided that most of the documentation can still exist. What we do is something like this:
```
!if function=hasLibtorch()

!syntax parameters /UserObjects/TorchScriptUserObject

!syntax inputs /UserObjects/TorchScriptUserObject

!syntax children /UserObjects/TorchScriptUserObject

!if-end!

!else
!include libtorch/libtorch_warning.md
```

Which will claim that MOOSE was not compiled with libtorch and says that more content will become available if the executable used for the doco has libtorch support.

grmnptr: I am okay doing a sweep if the consensus says that we just want to turn the doco off by default as well. But in a different PR.

test/tests/materials/torchscript_material/tests
```diff
@@ -0,0 +1,38 @@
...
+      cli_args = "Materials/net_material/prop_names=''"
+      detail = "the user does not specify any material properties; "
+      libtorch = true
+    []
+    [noinputs]
+      type = RunException
+      input = 'test.i'
+      expect_err = "Must declare at least one input to the neural net"
+      cli_args = "Materials/net_material/input_names=''"
+      detail = "the user does not specifany input to the module within material property; "
```
bwspenc, line 26:
```suggestion
      detail = "the user does not specify any input to the module within material property; "
```

bwspenc (changes requested):

framework/include/libtorch/userobjects/TorchScriptUserObject.h
dschwen: The problem with the

```
!if function=hasLibtorch()
```

approach is that you have to replicate the exact conditions that you have in the source file to conditionally compile (and those condition can now go out of sync). The only thing the doc system cares about is whether the object exists or not. So why not check _that_?

bwspenc (approved):
Once the other issues raised (about conditional inclusion of documentation) are addressed I'm good with this.

test/tests/materials/torchscript_material/generate_net.py
dschwen: I'm using `manual_seed` in swift tests that run on CPU and CUDA and have not had any issues. In fact, I think this even passes on MPS (I'll double check on monday). Also see https://pytorch.org/docs/stable/notes/randomness.html

grmnptr:
I am not going to change the if-else in this PR, it follows what we have for every other object in the framework/STM. I will add a different PR to change the documentation for everything in one go.

framework/include/libtorch/userobjects/TorchScriptUserObject.h
grmnptr: I will think about it, and then move all the existing documentation in one go (in a different PR) if we decide we need another approach. In my opinion the main question is: do we even want a doc entry for these objects or not? In past conversations I was persuaded that we do even when the objects are not compiled. If that still stands I don't see a difference between the two approaches considering that the only way that object is not registered is if libtorch is not there.

grmnptr:
Failure is unrelated.
Thanks for the reviews!

PR #29953, opened 2025-02-25
Minor cleanups in mesh generators and add a dbg parameter to show how generators are executed

### LINKED ISSUE #29952: Make PeripheralTriangleMeshGenerator take multiple inputs
### Motivation

Currently `PeripheralTriangleMeshGenerator` can take a core mesh as the input and add unstructured mesh surrounding the core. However, in cases, there could be separated components like control drums, instrumentation, etc. outside of the core but disconnected with the core. It is thus desired that the mesh generator can take multiple inputs representing the core and those other components.

### Design

The mesh generator is creating a sub mesh generator in `XYDelaunayGenerator`, which can already accept multiple holes. So adding this feature requires minor changes to the mesh generator. I will submit a PR for this.

### Impact

A new capability.

YaqiWang:
This only contains some minor improvements while working on #29952. #29952 will be addressed separely.

### DIFF
--- a/framework/include/base/MeshGeneratorSystem.h
+++ b/framework/include/base/MeshGeneratorSystem.h
@@ -172,6 +172,9 @@ public:
   /// Get the name of the final mesh generator
   MeshGeneratorName getFinalMeshGeneratorName() const { return _final_generator_name; }
 
+  /// Set the verbose flag
+  void setVerbose(const bool verbose) { _verbose = verbose; }
+
 private:
   /**
    * Gets the MeshGeneratorNames that are referenced in an object's parameters.
@@ -241,4 +244,7 @@ private:
 
   /// Whether any of the mesh generators are a \p BreakMeshByBlockGenerator
   bool _has_bmbb;
+
+  /// Whether to print the names of the mesh generators being executed or not
+  bool _verbose;
 };
--- a/framework/src/actions/SetupDebugAction.C
+++ b/framework/src/actions/SetupDebugAction.C
@@ -47,6 +47,8 @@ SetupDebugAction::validParams()
   params.addParam<bool>("show_mesh_meta_data", false, "Print out the available mesh meta data");
   params.addParam<bool>(
       "show_reporters", false, "Print out information about the declared and requested Reporters");
+  params.addParam<bool>(
+      "show_mesh_generators", false, "Print out the mesh generators being executed");
 
   ExecFlagEnum print_on = MooseUtils::getDefaultExecFlagEnum();
   print_on.addAvailableFlags(EXEC_TRANSFER);
@@ -81,6 +83,7 @@ SetupDebugAction::SetupDebugAction(const InputParameters & parameters) : Action(
   _awh.showActionDependencies(getParam<bool>("show_action_dependencies"));
   _awh.showActions(getParam<bool>("show_actions"));
   _awh.showParser(getParam<bool>("show_parser"));
+  _awh.mooseApp().getMeshGeneratorSystem().setVerbose(getParam<bool>("show_mesh_generators"));
 }
 
 void
--- a/framework/src/base/MeshGeneratorSystem.C
+++ b/framework/src/base/MeshGeneratorSystem.C
@@ -25,7 +25,8 @@ MeshGeneratorSystem::MeshGeneratorSystem(MooseApp & app)
   : PerfGraphInterface(app.perfGraph(), "MeshGeneratorSystem"),
     ParallelObject(app),
     _app(app),
-    _has_bmbb(false)
+    _has_bmbb(false),
+    _verbose(false)
 {
 }
 
@@ -401,6 +402,10 @@ MeshGeneratorSystem::executeMeshGenerators()
     for (const auto & generator : generator_set)
     {
       const auto & name = generator->name();
+      if (_verbose)
+        _app._console << " [DBG] Executing mesh generator (" << COLOR_GREEN << std::setw(20) << name
+                      << COLOR_DEFAULT << ") in type (" << COLOR_GREEN << generator->type()
+                      << COLOR_DEFAULT << ")" << std::endl;
       auto current_mesh = generator->generateInternal();
 
       // Only generating data for this generator
--- a/modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
+++ b/modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
@@ -143,6 +143,7 @@ PeripheralTriangleMeshGenerator::PeripheralTriangleMeshGenerator(const InputPara
     params.set<BoundaryName>("output_boundary") = _external_boundary_name;
     params.set<SubdomainName>("output_subdomain_name") = _peripheral_ring_block_name;
     params.set<MooseEnum>("tri_element_type") = getParam<MooseEnum>("tri_element_type");
+    params.set<bool>("verbose_stitching") = false;
     addMeshSubgenerator("XYDelaunayGenerator", _input_name + "_periphery", params);
     _build_mesh = &getMeshByName(_input_name + "_periphery");
   }
--- a/framework/src/meshgenerators/StitchedMeshGenerator.C
+++ b/framework/src/meshgenerators/StitchedMeshGenerator.C
@@ -43,6 +43,8 @@ StitchedMeshGenerator::validParams()
       "merge_boundaries_with_same_name",
       true,
       "If the input meshes have boundaries with the same name (but different IDs), merge them");
+  params.addParam<bool>(
+      "verbose_stitching", false, "Whether mesh stitching should have verbose output.");
   params.addClassDescription(
       "Allows multiple mesh files to be stitched together to form a single mesh.");
 
@@ -234,7 +236,7 @@ StitchedMeshGenerator::generate()
                         second,
                         TOLERANCE,
                         _clear_stitched_boundary_ids,
-                        /*verbose = */ true,
+                        getParam<bool>("verbose_stitching"),
                         use_binary_search);
 
     if (_merge_boundaries_with_same_name)
--- a/modules/reactor/src/meshgenerators/AdvancedConcentricCircleGenerator.C
+++ b/modules/reactor/src/meshgenerators/AdvancedConcentricCircleGenerator.C
@@ -292,7 +292,7 @@ AdvancedConcentricCircleGenerator::generate()
     // As we rotate the mesh in the negative direction (see "-" before _azimuthal_angles[i]),
     // the order of SLICE_END and SLICE_BEGIN should be reversed compared to the similar call in
     // PolygonConcentricCircleMeshGeneratorBase.C
-    mesh->stitch_meshes(other_mesh, SLICE_END, SLICE_BEGIN, TOLERANCE, true);
+    mesh->stitch_meshes(other_mesh, SLICE_END, SLICE_BEGIN, TOLERANCE, true, false);
     other_mesh.clear();
   }
 
@@ -301,7 +301,7 @@ AdvancedConcentricCircleGenerator::generate()
       mesh->get_boundary_info().remove_id(i + 1 + OUTER_SIDESET_ID_ALT);
 
   // An extra step to stich the first and last slices together
-  mesh->stitch_surfaces(SLICE_END, SLICE_BEGIN, TOLERANCE, true);
+  mesh->stitch_surfaces(SLICE_END, SLICE_BEGIN, TOLERANCE, true, false);
 
   mesh->prepare_for_use();
 
--- a/modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
+++ b/modules/reactor/src/meshgenerators/PeripheralRingMeshGenerator.C
@@ -456,7 +456,7 @@ PeripheralRingMeshGenerator::generate()
     MooseMesh::changeBoundaryId(*input_mesh, _input_mesh_external_bid, OUTER_SIDESET_ID, false);
   mesh->prepare_for_use();
   // Use input_mesh here to retain the subdomain name map
-  input_mesh->stitch_meshes(*mesh, OUTER_SIDESET_ID, OUTER_SIDESET_ID_ALT, TOLERANCE, true);
+  input_mesh->stitch_meshes(*mesh, OUTER_SIDESET_ID, OUTER_SIDESET_ID_ALT, TOLERANCE, true, false);
 
   // Assign subdomain name to the new block if applicable
   if (isParamValid("peripheral_ring_block_name"))


modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
```diff
@@ -21,7 +21,8 @@ InputParameters
 PeripheralTriangleMeshGenerator::validParams()
 {
   InputParameters params = MeshGenerator::validParams();
-  params.addRequiredParam<MeshGeneratorName>("input", "The input mesh to be modified.");
+  params.addRequiredParam<std::vector<MeshGeneratorName>>("input",
+                                                          "The input mesh to be modified.");
```
roystgnr, line 25: If we're potentially taking multiple input meshes, shouldn't the doc string here now reflect that?

miaoyinb:
I think the functionality needed here can be done by https://mooseframework.inl.gov/source/meshgenerators/FlexiblePatternGenerator.html

framework/src/actions/SetupDebugAction.C
```diff
@@ -47,6 +47,8 @@ SetupDebugAction::validParams()
   params.addParam<bool>("show_mesh_meta_data", false, "Print out the available mesh meta data");
   params.addParam<bool>(
       "show_reporters", false, "Print out information about the declared and requested Reporters");
+  params.addParam<bool>(
+      "show_mesh_generators", false, "Print out the mesh generators being executed");
```
GiudGiud, line 51: does not really hurt but also fairly redundant with show_info in the gneerators

YaqiWang:
> I think the functionality needed here can be done by https://mooseframework.inl.gov/source/meshgenerators/FlexiblePatternGenerator.html

Looks like the two are quite similar. But `FlexiblePatternGenerator` requires a pattern to arrange inputs I guess. In the case here, components may placed randomly.

miaoyinb:
> > I think the functionality needed here can be done by https://mooseframework.inl.gov/source/meshgenerators/FlexiblePatternGenerator.html
> 
> Looks like the two are quite similar. But `FlexiblePatternGenerator` requires a pattern to arrange inputs I guess. In the case here, components may placed randomly. Edit: I saw that custom pattern. Let me try. On the other hand, this is a small improvement for this particular mesh generator unless this mesh generator is redundant.

Yes, `extra_positions` can be used for "randomly" placed input mesh. You do not need to transform the input mesh before use when using `FlexiblePatternGenerator`

GiudGiud:
> On the other hand, this is a small improvement for this particular mesh generator unless this mesh generator is redundant.
 
if this entails using transformgenerators on all the inputs before, I would think this is a regression for the workflow overall

YaqiWang:
Unfortunately, we typically have to create individual parts because we need to reassign block ids for the same mesh at different locations.

YaqiWang:
I am feeling insane here. Is there a bug in `AdvancedConcentricCircleGenerator` at
```
  const Real corr_factor =
      _preserve_volumes
          ? PolygonalMeshGenerationUtils::radiusCorrectionFactor(mod_azimuthal_angles, true, _order)
          : 1.0;
```
It is giving a different factor than
```
      const Real corr_factor =
          2 * M_PI / (Real)_boundary_sectors / std::sin(2 * M_PI / (Real)_boundary_sectors);
```
in `FlexiblePatternGenerator`.

miaoyinb:
> radiusCorrectionFactor

Whoops, it's missing a sqrt...

miaoyinb:
Let me fix it

YaqiWang:
I confirmed that `FlexiblePatternGenerator` can do the job, but in my case `PeripheralTriangleMeshGenerator` gives simpler syntax. I agree `PeripheralTriangleMeshGenerator` provides a subset of `FlexiblePatternGenerator` capabilities. However, I think this small improvement is valid if we still want to keep `PeripheralTriangleMeshGenerator`.

miaoyinb:
The comments here actually made me want to rewrite `PeripheralTriangleMeshGenerator` to use `FlexiblePatternGenerator` as a subgenerator.

IMHO `PeripheralTriangleMeshGenerator` should keep its simplest form and `FlexiblePatternGenerator` should be used for all "advanced" applications.

YaqiWang:
Oh, I can change that. Should not take me much time.

YaqiWang:
I attempted `FlexiblePatternGenerator` and pushed. But it is not very successful. I hit an issue which seems a design issue forcing me add a function `addRequestedMeshGenerator`. The subdomain id and name are also giving me headache in the test. @miaoyinb Can you take a look what could go wrong? Note that `FlexiblePatternGenerator` has the radius correction, but current `PeripheralTriangleMeshGenerator` does not, which means we possibly need regolding.

YaqiWang:
... This appears requiring my fix in `XYDelaunayGenerator` which allows us to set a subdomain id and a name.

GiudGiud:
it's a new feature not a fix. i reviewed that PR, we can process it soon

YaqiWang:
The tests will fail. I will rebase this after the other PR gets merged. The failing tests probably need regolding due to the radius correction.

miaoyinb:
> I attempted `FlexiblePatternGenerator` and pushed. But it is not very successful. I hit an issue which seems a design issue forcing me add a function `addRequestedMeshGenerator`. The subdomain id and name are also giving me headache in the test. @miaoyinb Can you take a look what could go wrong? Note that `FlexiblePatternGenerator` has the radius correction, but current `PeripheralTriangleMeshGenerator` does not, which means we possibly need regolding.

So the background id/name part of the `FlexiblePatternGenerator` does not work?

YaqiWang:
> So the background id/name part of the `FlexiblePatternGenerator` does not work?

Not as intended. I have to make some changes there, which has been pushed in here.

YaqiWang:
Yes, it is ready for review.

YaqiWang:
Where is the crash? I tried on my local machine. All tests pass.

miaoyinb:
> Where is the crash? I tried on my local machine. All tests pass.
it's one test in distributed mesh mode
image

YaqiWang:
It crashes in the mesh stitcher:
```
    frame #0: 0x0000000198de7158 libc++abi.dylib`__cxa_throw
    frame #1: 0x00000001209db884 libmesh_dbg.0.dylib`libMesh::DistributedMesh::add_point(libMesh::Point const&, unsigned long long, unsigned int) + 352
  * frame #2: 0x000000012101ce00 libmesh_dbg.0.dylib`libMesh::UnstructuredMesh::copy_nodes_and_elements(libMesh::MeshBase const&, bool, unsigned long long, unsigned long long, unsigned long long, std::__1::unordered_map<unsigned short, unsigned short, std::__1::hash<unsigned short>, std::__1::equal_to<unsigned short>, std::__1::allocator<std::__1::pair<unsigned short const, unsigned short>>>*) + 724
    frame #3: 0x000000012102f904 libmesh_dbg.0.dylib`libMesh::UnstructuredMesh::stitching_helper(libMesh::MeshBase const*, short, short, double, bool, bool, bool, bool, bool, bool, bool) + 13516
    frame #4: 0x000000012102c42c libmesh_dbg.0.dylib`libMesh::UnstructuredMesh::stitch_meshes(libMesh::MeshBase const&, short, short, double, bool, bool, bool, bool, bool, bool) + 184
    frame #5: 0x0000000111f3c3a0 libmoose-dbg.0.dylib`XYDelaunayGenerator::generate(this=0x0000000102726700) at XYDelaunayGenerator.C:587:13
    frame #6: 0x00000001120e3d70 libmoose-dbg.0.dylib`MeshGenerator::generateInternal(this=0x0000000102726700) at MeshGenerator.C:267:15
 ```
I do not have libmesh built myself. Do you have any clue what can go wrong with this change?

**Comment by miaoyinb** [comment]
I think it is also related to https://github.com/libMesh/libmesh/pull/4081.

Before the libmesh submodule is updated, you can swap you two input meshes to dodge the error...

**Comment by YaqiWang** [comment]
I am not sure whether the two are related. The two meshes passed in `FlexiblePatternGenerator` contain multiple elements. I tried change the input by directly using `FlexiblePatternGenerator`. The error with distributed mesh is gone. But `PeripheralTriangleMeshGenerator` in this change is simply passing parameters to `FlexiblePatternGenerator`. I feel this error is revealing an unrelated bug. I attached the mesh generator in `FlexiblePatternGenerator` for the failing test:
```
  [tmg1]
    type = FlexiblePatternGenerator
    inputs = 'hex_in inst'
    extra_positions = '0 0 0 0 0 0'
    extra_positions_mg_indices = '0 1'
    boundary_type = circle
    boundary_size = 300
    boundary_sectors = 50
    desired_area = 50
    background_subdomain_id = 102
  []
```

**Comment by miaoyinb** [comment]
> I am not sure whether the two are related. The two meshes passed in `FlexiblePatternGenerator` contain multiple elements. I tried change the input by directly using `FlexiblePatternGenerator`. The error with distributed mesh is gone. But `PeripheralTriangleMeshGenerator` in this change is simply passing parameters to `FlexiblePatternGenerator`. I feel this error is revealing an unrelated bug. I attached the mesh generator in `FlexiblePatternGenerator` for the failing test:
> 
> ```
>   [tmg1]
>     type = FlexiblePatternGenerator
>     inputs = 'hex_in inst'
>     extra_positions = '0 0 0 0 0 0'
>     extra_positions_mg_indices = '0 1'
>     boundary_type = circle
>     boundary_size = 300
>     boundary_sectors = 50
>     desired_area = 50
>     background_subdomain_id = 102
>   []
> ```

Let me do some checking.

**Comment by roystgnr** [comment]
I can replicate the error with a libMesh install I had laying around; I'm trying a build post-https://github.com/libMesh/libmesh/pull/4081 now to see if that really does fix it.

If not?  It's at least going to prompt *some* kind of work at the libMesh level; at the point where we finally fail an assertion, the stitcher seems to be trying to add a new node with an already-in-use id, and even if bad MOOSE input is somehow the root cause we should still be hitting some kind of more informative assertion earlier.

File: modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C [hunk 3]
```diff
@@ -78,74 +84,54 @@ PeripheralTriangleMeshGenerator::validParams()
...
+        getParam<SubdomainName>("peripheral_ring_block_name");
+  if (isParamValid("external_boundary_id"))
+    params.set<boundary_id_type>("external_boundary_id") =
+        getParam<boundary_id_type>("external_boundary_id");
+  if (isParamValid("external_boundary_name"))
+    params.set<BoundaryName>("external_boundary_name") =
+        getParam<BoundaryName>("external_boundary_name");
+  params.set<MooseEnum>("tri_element_type") = getParam<MooseEnum>("tri_element_type");
+  addMeshSubgenerator("FlexiblePatternGenerator", _input_names[0] + "_periphery", params);
+  _build_mesh = &getMeshByName(_input_names[0] + "_periphery");
```
miaoyinb, line 134:
```suggestion
  addMeshSubgenerator("FlexiblePatternGenerator", name() + "_fpgsub", params);
  _build_mesh = &getMeshByName(name() + "_fpgsub");
```
This seems to solve the issue. I do not exactly know why. It may have something to do with the dependency system...

YaqiWang: That is bizarre. Let me try...

YaqiWang: I can only think that we have a map somewhere using this name as a key, thus changing the name here affecting the ordering in the map.

YaqiWang:
Pushed. But this shows that different path of executing mesh generators can be significant, which should not be the case because the valid paths are not unique.

roystgnr:
That's obviously an ugly workaround, but since it's a libMesh bug at root, the fix is going to have to wait until https://github.com/libMesh/libmesh/pull/4095 makes it downstream.  I'm definitely not adding anything new to the long-delayed "February" libMesh submodule update, but I'll start the next one almost immediately after we both have the current PR merged and have any outstanding bug fixes like 4095 in libMesh master.

miaoyinb:
This might be related to using subgenerator in another subgenerator, which is not officially supported.

A workaround is used here to override the mesh dependency system. We probably need an official or at least consistent way to do so. @loganharbour may have some insights.

miaoyinb:
Wait, no. This also fails...So it is not related to "subgenerator in subgenerator"...

```
[Mesh]
  allow_renumbering = 0
  [hex_in]
    type = FileMeshGenerator
    file = gold/abtr_mesh.e
  []
  [inst]
    type = AdvancedConcentricCircleGenerator
    ring_radii = '3 5'
    ring_intervals = '1 2'
    ring_block_ids = '100 101'
    num_sectors = 20
    ring_block_names = 'inst_inner inst_outer'
  []
  
  [hex_in_pheriphery]
    type = FlexiblePatternGenerator
    inputs = 'hex_in inst'
    extra_positions = '0 0 0 0 130 0'
    extra_positions_mg_indices = '0 1'
    boundary_type = circle
    boundary_size = 300
    boundary_sectors = 50
    desired_area = 50
    background_subdomain_id = 102
  []
[]
```

modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
```diff
@@ -78,74 +84,54 @@ PeripheralTriangleMeshGenerator::validParams()
...
-  for (unsigned int i = 0; i < _peripheral_ring_num_segments; i++)
+  auto params = _app.getFactory().getValidParams("FlexiblePatternGenerator");
+  params.set<std::vector<MeshGeneratorName>>("inputs") = _input_names;
+  params.set<MooseEnum>("boundary_type") = "circle";
+  params.set<Real>("boundary_size") = _peripheral_ring_radius * 2;
+  params.set<unsigned int>("boundary_sectors") = _peripheral_ring_num_segments;
+  auto & pos = params.set<std::vector<Point>>("extra_positions");
+  auto & ids = params.set<std::vector<unsigned int>>("extra_positions_mg_indices");
+  for (const auto i : index_range(_input_names))
   {
```
miaoyinb, line 103:
```suggestion
  params.set<std::vector<Point>>("extra_positions") = std::vector<Point>(_input_names,size(), Point());
  auto & ids = params.set<std::vector<unsigned int>>("extra_positions_mg_indices");
  ids.resize(_input_names,size());
  std::iota(ids.begin(), ids.end(),0);
```

modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
```diff
@@ -78,74 +84,54 @@ PeripheralTriangleMeshGenerator::validParams()
...
-    params.set<std::vector<bool>>("refine_holes") = std::vector<bool>{false};
-    params.set<std::vector<bool>>("stitch_holes") = std::vector<bool>{true};
-    params.set<BoundaryName>("output_boundary") = _external_boundary_name;
-    params.set<SubdomainName>("output_subdomain_name") = _peripheral_ring_block_name;
-    params.set<MooseEnum>("tri_element_type") = getParam<MooseEnum>("tri_element_type");
-    addMeshSubgenerator("XYDelaunayGenerator", _input_name + "_periphery", params);
-    _build_mesh = &getMeshByName(_input_name + "_periphery");
+    pos.push_back(libMesh::Point());
+    ids.push_back(i);
   }
```
miaoyinb, line 106:
```suggestion
```

miaoyinb (changes requested):
Generally looks good.

My main concern is to have a consistent way to use subgenerator in another subgenerator.

framework/src/base/MeshGeneratorSystem.C
```diff
@@ -401,6 +402,10 @@ MeshGeneratorSystem::executeMeshGenerators()
     for (const auto & generator : generator_set)
     {
       const auto & name = generator->name();
+      if (_verbose)
+        _app._console << " [DBG] Executing mesh generator (" << COLOR_GREEN << std::setw(20) << name
```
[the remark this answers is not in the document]
YaqiWang: This
```
 [DBG] Executing mesh generator (              hex_in) in type (FileMeshGenerator)
 [DBG] Executing mesh generator (           inst_hole) in type (AdvancedConcentricCircleGenerator)
 [DBG] Executing mesh generator (                inst) in type (TransformGenerator)
 [DBG] Executing mesh generator (tmg_periphery_circle_boundary) in type (PolyLineMeshGenerator)
 [DBG] Executing mesh generator (hex_in_tmg_periphery_del_ext_bdry) in type (BoundaryDeletionGenerator)
 [DBG] Executing mesh generator ( tmg_periphery_pos_0) in type (TransformGenerator)
 [DBG] Executing mesh generator (inst_tmg_periphery_del_ext_bdry) in type (BoundaryDeletionGenerator)
 [DBG] Executing mesh generator ( tmg_periphery_pos_1) in type (TransformGenerator)
 [DBG] Executing mesh generator (tmg_periphery_pattern) in type (XYDelaunayGenerator)
 [DBG] Executing mesh generator (       tmg_periphery) in type (FlexiblePatternGenerator)
 [DBG] Executing mesh generator (                 tmg) in type (PeripheralTriangleMeshGenerator)
```
does not look too bad. It may not be appealing if we leave too much space for accommodating the longest name.

modules/reactor/src/meshgenerators/FlexiblePatternGenerator.C
```diff
@@ -245,6 +245,8 @@ FlexiblePatternGenerator::FlexiblePatternGenerator(const InputParameters & param
 
 {
   declareMeshesForSub("inputs");
+  for (const auto & input : _input_names)
+    addRequestedMeshGenerator(input);
```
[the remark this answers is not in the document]
YaqiWang: I think it needs to be here because `FlexiblePatternGenerator` is not calling `getMesh` because it is using sub mesh generators to finish the job, thus MOOSE does not know these meshes are its input. This call here to tell moose that this is the case and enables this generator can be used as a sub generator.

modules/reactor/doc/content/source/meshgenerators/PeripheralTriangleMeshGenerator.md
```diff
@@ -4,7 +4,7 @@
...
-The `PeripheralTriangleMeshGenerator` utilizes the general purpose [`XYDelaunayGenerator`](XYDelaunayGenerator.md) triangulation mesh generator and the [`PolyLineMeshGenerator`](PolyLineMeshGenerator.md) to create a circular periphery region around a reactor core. The outer circle properties are specified with the [!param](/Mesh/PeripheralTriangleMeshGenerator/peripheral_ring_radius) and [!param](/Mesh/PeripheralTriangleMeshGenerator/peripheral_ring_num_segments) input parameters.
+The `PeripheralTriangleMeshGenerator` utilizes the general purpose [`XYDelaunayGenerator`](XYDelaunayGenerator.md) triangulation mesh generator and the [`PolyLineMeshGenerator`](PolyLineMeshGenerator.md) to create a circular periphery region around a reactor core. The outer circle properties are specified with the [!param](/Mesh/PeripheralTriangleMeshGenerator/peripheral_ring_radius) and [!param](/Mesh/PeripheralTriangleMeshGenerator/peripheral_ring_num_segments) input parameters. It is noted that [!param](/Mesh/PeripheralTriangleMeshGenerator/input) can accept multiple mesh generators, which represent the core and other components such as control drums, instrumention, etc. disconnected with the core.
```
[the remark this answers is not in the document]
YaqiWang: Added this ` This mesh generator does not perform any translation on these input meshes, thus these meshes must be non-overlapping. This mesh generator uses [FlexiblePatternGenerator.md] internally to perform the mesh generation and only uses a sub-set of its functionalities.`.

modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
YaqiWang: Changed `extra_positions` but kept `extra_positions_mg_indices` because the new way has the same three lines with more letters.

miaoyinb: If we are as serious as counting lines, `push_back` allocates memory every time it is called.

modules/reactor/src/meshgenerators/FlexiblePatternGenerator.C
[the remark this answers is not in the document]
miaoyinb: Thinking about this again, I think both solutions slightly compromise the MG dependency but without catastrophic consequences. So I do not have a strong opinion where these lines should be put.

My understanding is:

For the MG with these lines, the system would think both the MG and its subMG are immediately dependent to the input meshes. In reality, only the subMG is immediately dependent to the input meshes, while the MG is dependent to the input meshes through its subMG. It is not catastrophic because it does not change the fact that the MG is dependent to the input meshes.

modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
YaqiWang: >every time it is called

Not true. I do not know how much it allocates when hitting the limit, but I am sure it is not allocating one more entry.

modules/reactor/src/meshgenerators/PeripheralTriangleMeshGenerator.C
```diff
@@ -78,74 +84,52 @@ PeripheralTriangleMeshGenerator::validParams()
...
-    addMeshSubgenerator("XYDelaunayGenerator", _input_name + "_periphery", params);
-    _build_mesh = &getMeshByName(_input_name + "_periphery");
-  }
+  auto params = _app.getFactory().getValidParams("FlexiblePatternGenerator");
+  params.set<std::vector<MeshGeneratorName>>("inputs") = _input_names;
+  params.set<MooseEnum>("boundary_type") = "circle";
+  params.set<Real>("boundary_size") = _peripheral_ring_radius * 2;
+  params.set<unsigned int>("boundary_sectors") = _peripheral_ring_num_segments;
+  params.set<std::vector<Point>>("extra_positions") =
+      std::vector<Point>(_input_names, size(), Point());
```
miaoyinb, line 101:
```suggestion
      std::vector<Point>(_input_names.size(), Point());
```
sorry for the typo.

YaqiWang: oh?!

miaoyinb: it caused some failed tests.

modules/reactor/src/meshgenerators/FlexiblePatternGenerator.C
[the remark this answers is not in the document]
miaoyinb: That's okay to me.
I can probably work on that base class approach in near future.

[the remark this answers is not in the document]
YaqiWang: Drop the last commit or keep this as it is?

miaoyinb:
Just a reminder that the original code that allows multiple input meshes in `PeripheralTriangleMeshGenerator` is not included in the current two commits after dropping the subgenerator related commit.

YaqiWang:
Can this get approved?

miaoyinb:
> Can this get approved?

The current two commits do not include the support for multiple components input and any related test cases. Please double-check

YaqiWang:
I changed the title of this PR.

framework/include/base/MeshGeneratorSystem.h
```diff
@@ -241,4 +244,7 @@ class MeshGeneratorSystem : public PerfGraphInterface, public libMesh::ParallelO
 
   /// Whether any of the mesh generators are a \p BreakMeshByBlockGenerator
   bool _has_bmbb;
+
+  /// Whether print the names of the mesh generators being executed
```
miaoyinb, line 248:
```suggestion
  /// Whether to print the names of the mesh generators being executed or not
```

miaoyinb (changes requested):
Looks fine to me. Will approve and merge it after this minor editing.

miaoyinb (approved):

YaqiWang:
Thanks Yinbin!

PR #29966, opened 2025-02-26
Pass oversubscribe for the break mesh by block generator test

### LINKED ISSUE #29965: break_mesh_by_block_generator test fails with openmpi because of slots issue
### Bug Description

Error message is
```
--------------------------------------------------------------------------
There are not enough slots available in the system to satisfy the 12
slots that were requested by the application:

  /opt/civet0/build_0/moose/test/moose_test-opt

Either request fewer slots for your application, or make more slots
available for use.

A "slot" is the Open MPI term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which Open MPI processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, Open MPI defaults to the number of processor cores

In all the above cases, if you want Open MPI to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --oversubscribe option to ignore the
number of available slots when deciding the number of processes to
launch.
----------------------------------------------------------
```

### Steps to Reproduce

Run this test in CI with openmpi on the Mac arm architecture

### Impact

Annoyance impeding development

### [Optional] Diagnostics

_No response_

lindsayad:
Closes #29965

### DIFF
--- a/test/tests/meshgenerators/break_mesh_by_block_generator/tests
+++ b/test/tests/meshgenerators/break_mesh_by_block_generator/tests
@@ -224,7 +224,6 @@
     max_parallel = 12
     recover = false
     cli_args = '--oversubscribe'
-    allow_unused = true
   []
 
   [invalid_block_pair]
--- a/test/tests/meshgenerators/break_mesh_by_block_generator/tests
+++ b/test/tests/meshgenerators/break_mesh_by_block_generator/tests
@@ -223,6 +223,8 @@
     min_parallel = 12
     max_parallel = 12
     recover = false
+    cli_args = '--oversubscribe'
+    allow_unused = true
   []
 
   [invalid_block_pair]


roystgnr:
Surely it can't be *this* simple, though, can it?  mpirun/mpiexec don't look at every single argument in front of them, only arguments that precede the executable name.

(and if we *do* get it to precede the executable name, the fact that there's no standardization of the argument means we'll be in the predicament where OpenMPI is now happy but other MPI implementations may scream at us about it ... in fact, yeah, trying it with MPICH gives me "unrecognized argument oversubscribe")

lindsayad:
Oh duh

PR #29968, opened 2025-02-26
Add additional derivatives to PolycrystalDiffusivityTensorBase

### LINKED ISSUE #29963: Add grain order parameter derivatives to PolycrystalDiffusivityTensorBase
### Motivation

The diffusivity tensor calculated by the materials property `PolycrystalDiffusivityTensorBase` can depend on both `c`, which differentiates between the solid and void phase, and order parameters `eta_i`, which differentiate between grains of the solid phase. Currently only derivatives with respect `c` are taken. Adding calculations of derivatives with respect to `eta_i` will enable more accurate calculations of the off-diagonal Jacobians, with hopes of improving performance.

### Design

-Move `_vals_name` from `GrandPotentialTensorMaterial` (derived class) to `PolycrystalDiffusivityTensorBase`
-Add vector of material property derivatives `_dDdeta` and calculate them
-Add contribution from term `_dDdeta * _chi` to `GrandPotentialTensorMaterial`

### Impact

No changes

laagesen:
Closes #29963


dschwen (approved):
Looks ok to me.

Issue #29986, opened 2025-02-27
Automatic reply with "most relevant posts"


MengnanLi91:
This starts from @hugary1995 's brilliant idea and his side project building a MOOSE Discussion database and retrieval method. We want to bring that in MOOSE and build a automatic MOOSE chatbot to show the most relevant posts when there is a new post on MOOSE discussion.

_Originally posted by @hugary1995 in https://github.com/idaholab/moose/discussions/24823_

Mining MOOSE GitHub discussion
https://github.com/hugary1995/moose-gh-mining

PR #29988, opened 2025-02-28
Initial commit for discussion bot

MengnanLi91:
refer to #29986

This project originates from @hugary1995 's idea and his side project of creating a MOOSE Discussion database and retrieval method. Our goal is to integrate this into MOOSE and develop an automated MOOSE chatbot that displays the most relevant posts when a new discussion is initiated on MOOSE.

The current design utilizes purely vector retrieval(use LlamaIndex to build it). The algorithm compares the discussion title with the content of all previous discussions (title + discussions) in our database, and provides the most relevant posts to the user. In future updates, we plan to incorporate Large Language Models (LLMs) into this process to generate summaries of the recommended posts. The database will be updated regularly to include all new posts (potentially on a monthly basis). The scripts for fetching and generating the database, as well as the GitHub bot script, are included in this pull request. Unit tests for each script are also provided.

We are also looking for a fitting name for the discussion bot. If you have any suggestions, please share them here!

Below is a snapshot of how it looks like in discussion page:

Screenshot 2025-03-18 at 9 40 54 AM

link: https://github.com/MengnanLi91/moose/discussions/18


MengnanLi91:
@GiudGiud How do you want to set the threshold? If we don't find similar posts above this threshold, what should we reply to user?

GiudGiud:
> How do you want to set the threshold?

Experimentally? Start too low and raise it over time

> If we don't find similar posts above this threshold, what should we reply to user?

nothing. They can wait for someone to reply

python/MooseDiscussion/db/encoding.npy
GiudGiud, line 1: can this be external? I dont like that we will be refreshing a 15MB file every now and then

MengnanLi91: Yes, I am discussing with @milljm and @loganharbour to move database to another repo. Also I am going to use llamaIndex and vector database rather than .npy and .csv. That will make the database smaller too.

milljm: I wonder if there is a desire for a central storage for AI models in the Idaholab namespace. Just something to consider when we name this repo...

MengnanLi91: This is not AI or LLM, I just use open source vector database tool. I think we are good there.

.github/workflows/discussionbot-workflow.yml
```diff
@@ -0,0 +1,35 @@
+name: Discussion Help Bot
+on:
+  discussion:
+    types:
+    # Use locked as trigger event until this chatbot is ready
+      - created
+    #  - locked
```
hugary1995, line 7:
```suggestion
```

hugary1995, line 7: I wonder if there are other events that should trigger this. Is there an event associated with editing the content?

.github/workflows/discussionbot-workflow.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  run_bot:
+    runs-on: [self-hosted, linux]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
hugary1995, line 14: This is at v4 now.

.github/workflows/discussionbot-workflow.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          ref: MOOSE_discussion_bot
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
hugary1995, line 21: Is hard-coding python version here okay?

.github/workflows/discussionbot-workflow.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Install dependencies
+        run: pip install -r python/MooseDiscussion/requirements.txt
+
+      - name: Run discuss bot
```
hugary1995, line 26:
```suggestion
      - name: Run discussion bot
```

.github/workflows/discussionbot-workflow.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: python python/MooseDiscussion/GitHubBot.py
+
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
hugary1995, line 30: Do you need to explicitly pass this secret? From what I recall github actions use this token by default.

python/MooseDiscussion/GitHubAPI.py
```diff
@@ -0,0 +1,119 @@
+import argparse
```
hugary1995, line 1: We may want a shebang at the top.

python/MooseDiscussion/GitHubBot.py
```diff
@@ -0,0 +1,187 @@
+from pathlib import Path
```
hugary1995, line 1: Consider adding a shebang.

python/MooseDiscussion/IndexGenerator.py
```diff
@@ -0,0 +1,145 @@
+import json
```
hugary1995, line 1: Consider adding a shebang.

hugary1995 (changes requested):
I share similar concerns about database storage. I have some other minor comments.

hugary1995:
> We are also looking for a fitting name for the discussion bot. If you have any suggestions, please share them here!

Not sure if we should try to adhere to our animal naming convention here. But maybe mechamoose?

image

.github/workflows/discussionbot-workflow.yml
MengnanLi91: I don't think I need to. I have modified this in the new version

Issue #29991, opened 2025-02-28
Enhance Grain Orientation Computation in Crystal Plasticity Models


dewenyushu:
### Motivation

In crystal plasticity models, accurately calculating grain orientation, represented by Euler angles, is paramount. Presently, Euler angles are defined at each quadrature point and can evolve differently based on the local deformation of every material point. To facilitate interoperability with phase field simulations at the grain level, a method to compute orientation at the grain level is required.

### Design

- Create a UserObject that accepts Euler angles as a material property and computes the updated grain orientation based on user-defined rules.
- Implement a VectorPostprocessor efficiently extract grain orientations from the user object
- Develop a UserObject to set grain orientations using the extracted values from the VectorPostprocessor

### Impact

Enable grain orientation computation in crystal plasticity models thus facilitates integration with phase-field simulations/

PR #29994, opened 2025-03-01
Grain structure calculation from crystal plasticity

dewenyushu:
Refs #29991


modules/solid_mechanics/doc/content/source/userobjects/ComputeBlockOrientationByRotation.md
```diff
@@ -0,0 +1,19 @@
+# ComputeBlockOrientationByRotation
+
+In a crystal plasticity simulation, the `ComputeBlockOrientationByRotation`  user object calculates the orientation of each grain, represented by material blocks (subdomains), by selecting the most prevalent orientation direction among all material points within each block. It achieves this by:
+
+1. Converting the Euler angles of every element within the block into quaternions.
+2. Computing the distribution of quaternion values.
+3. Choosing the most prevalent orientation direction as the grain orientation.
```
sapitts, line 7: How are subgrain regions, e.g. twinned regions or less than 5 degree grain boundaries, treated with this approach?

Are there recommendations for when to use this class vs. the misorientation class? If yes, can you add those recommendations to the documentation page here?

modules/solid_mechanics/doc/content/source/userobjects/ComputeBlockOrientationByMisorientation.md
```diff
@@ -0,0 +1,17 @@
+# ComputeBlockOrientationByMisorientation
+
+In a crystal plasticity simulation, the `ComputeBlockOrientationByMisorientation` user object determines the orientation of each grain, represented by material blocks (subdomains), by identifying the material point orientation that undergoes the most rotation during the preceding time step. It accomplishes this by calculating the maximum misorientation within each material block and selecting the orientation that yields the highest misorientation value.
+
+The misorientation value, denoted as $M$, is computed in the [ComputeMultipleCrystalPlasticityStress](ComputeMultipleCrystalPlasticityStress.md) class using the formula: $M = \arccos(\left(\text{trace}(\boldsymbol{M}) - 1\right) / 2)$, where $\boldsymbol{M} = \boldsymbol{R} * \boldsymbol{R}\text{old}^{-1}$ represents the misorientation matrix, and $\boldsymbol{R}$ and $\boldsymbol{R}\text{old}$ are the current and previous time step's rotation matrices, respectively.
```
sapitts, line 5: Should there be a citation for this expression? It can be handy to have the citations as sources of additional information for users who want to understand more of the background

sapitts (dismissed):
I think the code looks good; I have a few questions because I don't understand it as well as I'd like. Generally I'd like to request expanded documentation--to help users understand assumptions around units and file formats--and additional regression tests to check for code errors. If there's a way to construct a few more regression tests like with the Diffusion kernel to give an "analytical solution" check, I'd appreciate a few more of those tests too please.

Nice work! I'm looking forward to using this capability!

dewenyushu:
@dschwen here is the PR

modules/solid_mechanics/test/tests/crystal_plasticity/stress_update_material_based/update_euler_angle.i
```diff
@@ -104,7 +104,6 @@
   []
   [updated_euler_angle]
     type = ComputeUpdatedEulerAngle
-    radian_to_degree = true
```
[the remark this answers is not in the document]
dewenyushu: The default is still using degrees. You can see that the option name is now `degree_to_radian` (vs `radian_to_degree`) while the default is set to `false`. I feel that it makes more sense in the conversion that happens in `ComputeUpdatedEulerAngle::computeEulerAngleFromRotationMatrix`.

modules/solid_mechanics/test/tests/block_orientation/grn_8_rand.tex
```diff
@@ -0,0 +1,12 @@
+Texture File
+
+File generated from EXCEL
```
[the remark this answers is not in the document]
dewenyushu: I actually copied this from another example in phase field. Let me make a hyper link instead :)

modules/solid_mechanics/test/tests/block_orientation/block_orientation_calculation.i
```diff
@@ -0,0 +1,163 @@
...
+    type = CrystalPlasticityKalidindiUpdate
+    number_slip_systems = 12
+    slip_sys_file_name = input_slip_sys.txt
+  []
+[]
+
+[UserObjects]
+  [block_orientation]
+    type = ComputeBlockOrientationByRotation
+    # type = ComputeBlockOrientationByMisorientation <- can use this UO as well
```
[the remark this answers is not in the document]
dewenyushu: Yes I have both tested.

modules/solid_mechanics/test/tests/block_orientation/block_orientation_calculation.i
```diff
@@ -0,0 +1,163 @@
...
+    execute_on = 'TIMESTEP_END'
+    execution_order_group = -1
```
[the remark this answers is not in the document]
dewenyushu: Good point - This was originally set to ensure that this object is executed before other UOs, which may rely on the updated grain orientation calculated in this UO. Since we only have one UO in this test case, I will just remove this line to avoid confusion.

modules/solid_mechanics/test/src/vectorpostprocessors/BlockOrientationVectorPostprocessor.C
```diff
@@ -0,0 +1,75 @@
...
+    auto sid = _current_elem->subdomain_id();
+    _sample[0] = sid;
+
+    const EulerAngles & angle = _euler.getEulerAngles(sid);
+    _sample[1] = angle.phi1;
+    _sample[2] = angle.Phi;
+    _sample[3] = angle.phi2;
+    SamplerBase::addSample(_current_elem->vertex_average() /* x,y,z coordinates of elem centroid */,
+                         _current_elem->id(),
+                         _sample);
```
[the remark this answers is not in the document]
dewenyushu: Here the average is to calculate the centroid of the element only. There is no averaging of the Euler angles since this postprocessor is simply grabbing the Euler angle value of the block where this element resides from the `EulerAngleProvider` object.

modules/solid_mechanics/test/include/vectorpostprocessors/BlockOrientationVectorPostprocessor.h
```diff
@@ -0,0 +1,39 @@
...
+#include "ElementVectorPostprocessor.h"
+#include "SamplerBase.h"
+
+// Forward declarations
+class EulerAngleProvider;
+
+/**
+ *  BlockOrientationVectorPostprocessor is a VectorPostprocessor that outputs the
+ *  the coordinates, block id, and Euler Angles associated with each element from a EulerAngleProvider.
+ *  Currently only works with a uniform, structured grid (no mesh adaptivity).
```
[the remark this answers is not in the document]
dewenyushu: I am not sure if it is necessary since this is a test object only and not designed for wider use than existing tests.

However, if this just a few lines of code change with an existing example to refer to, I'd happy to add that.

modules/solid_mechanics/src/userobjects/EulerAngleUpdateFromReporter.C
```diff
@@ -0,0 +1,96 @@
...
+void
+EulerAngleUpdateFromReporter::initialize()
+{
+  // only update Euler angles info from the reporter when the initial Euler angle has been read from
+  // file
+  if (_first_time)
+  {
+    _angles.clear();
+    EulerAngleFileReader::readFile();
+    _first_time = false;
```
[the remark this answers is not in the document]
dewenyushu: This is a good point - I need to double check on this

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByRotation.C
```diff
@@ -0,0 +1,144 @@
...
+  Real total_weight = 0.0;
+
+  for (const auto & q : _quat[sid])
+  {
+    Vector4 v(std::get<0>(q), std::get<1>(q), std::get<2>(q), std::get<3>(q));
+
+    const auto bin = std::make_tuple<int, int, int, int>(std::floor(std::get<0>(q) * 0.5 * _bins),
+                                                         std::floor(std::get<1>(q) * 0.5 * _bins),
+                                                         std::floor(std::get<2>(q) * 0.5 * _bins),
+                                                         std::floor(std::get<3>(q) * 0.5 * _bins));
```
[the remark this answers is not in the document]
dewenyushu: The quaternion values range between -1 to 1. Multiplying by 0.5 make sure that the data is divided into `_bins` number of groups.

modules/solid_mechanics/src/materials/crystal_plasticity/ComputeMultipleCrystalPlasticityStress.C
```diff
@@ -315,6 +317,21 @@ ComputeMultipleCrystalPlasticityStress::postSolveQp(RankTwoTensor & cauchy_stres
...
+  auto cos_val = (_misorientation_mat.trace() - 1.0) / 2.0;
+  // Make sure the value is in feasible range
+  if (cos_val > 1.0 || cos_val < -1.0)
+  {
+    if (MooseUtils::absoluteFuzzyEqual(cos_val, -1.0))
+      cos_val = -1.0;
+    else if (MooseUtils::absoluteFuzzyEqual(cos_val, 1.0))
+      cos_val = 1.0;
+    else
+      mooseError("Misorientation value is undefined. This should not happen.");
```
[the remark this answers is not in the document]
dewenyushu: > Can you include more information in this error message about what might have gone wrong if the misorientation value is out of bounds so that the user knows what to try?

Here is the dilemma: mathematically, the misorientation value would never lie outside of [-1, 1] (because of how it is calculated). If it does, then it indicates numerical error, which should be small enough (and can be guarded by the `absoluteFuzzyEqual` check). However, it fails the `absoluteFuzzyEqual` check, then maybe the tolerance is too tight, or something is really wrong. 

Maybe it is better to set all values less than -1 to -1 and all values larger than 1 to 1 (with the assumption that all out-of-range values are caused by numerical errors)..?

modules/solid_mechanics/include/userobjects/EulerAngleUpdateFromReporter.h
```diff
@@ -0,0 +1,43 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "EulerAngleFileReader.h"
+#include "ReporterInterface.h"
+
+// Forward declaration
```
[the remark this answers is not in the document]
dewenyushu: It is now removed

modules/phase_field/test/tests/polycrystal_elasticity_tensor/tests
```diff
@@ -0,0 +1,10 @@
+[Tests]
+  issues = '#29991'
+  design = 'ComputePolycrystalElasticityTensor.md'
+  [polycrystal_elasticity_tensor]
+    type = 'Exodiff'
+    input = 'polycrystal_elasticity_tensor.i'
+    exodiff = 'polycrystal_elasticity_tensor_out.e'
+    requirement = 'The system shall correctly compute an evolving elasticity tensor coupled to a grain growth phase field model.'
```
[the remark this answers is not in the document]
dewenyushu: I do not have an analytical solution to this test. Do you have a suitable case in mind?

I originally thought this test would be enough for the added few lines for compatibility with CP - Would you mind being more specific about what regression checks would be needed?

modules/solid_mechanics/test/tests/block_orientation/grn_8_rand.tex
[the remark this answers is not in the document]
sapitts: I'd check with Logan about the hyperlinks across modules, just to be safe

modules/solid_mechanics/src/materials/crystal_plasticity/ComputeMultipleCrystalPlasticityStress.C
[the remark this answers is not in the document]
sapitts: ah, now I understand the dilemma. I think I'd rather have the hard error if the misorientation value is out of bounds rather than silently hard-code a correction.  As an example, I could see a case where perhaps a too-loose tolerance setting in a crystal plasticity constitutive model might result in a misorientation value outside of the [-1,1] range--this is a situation where I would want the simulation to fail (as a user).

modules/phase_field/test/tests/polycrystal_elasticity_tensor/tests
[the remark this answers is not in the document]
sapitts: There's a couple of tests that Jesse Carter put in a few years ago to the test the rotation matrix input option for the elasticity tensor that had entries like `11 21 31 12 22 32 31 32 33`, rotated by 90 degrees, and checked for the updated values that was a quick way to make sure the values were rotated correctly (and that was it). Here I was thinking it'd be nice to have a test with a made up simple elasticity tensor to confirm that everything is copied over correctly. 

You may already have such a test in the solid mechanics model and not connected to the phase field module.

modules/solid_mechanics/src/userobjects/EulerAngleUpdateFromReporter.C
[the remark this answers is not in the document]
dewenyushu: Okay, I just realized that during restart the `_first_time ` get reinitialized to `true` and the Euler angles are reset to the initial values provided in the text file. This is not desirable. Any idea around this?

modules/solid_mechanics/src/userobjects/EulerAngleUpdateFromReporter.C
```diff
@@ -0,0 +1,96 @@
...
+    UpdateEulerAngle();
+}
+
+void
+EulerAngleUpdateFromReporter::UpdateEulerAngle()
+{
+  // check sizes of the containers
+  if (_grain_id.size() != _euler_angle_0.size() || _grain_id.size() != _euler_angle_1.size() ||
+      _grain_id.size() != _euler_angle_2.size())
+    paramError("grain_id_name", "Number of reporters' entries do not match.");
```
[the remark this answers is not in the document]
dewenyushu: Test added!

modules/phase_field/test/tests/polycrystal_elasticity_tensor/tests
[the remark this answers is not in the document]
dewenyushu: Humm just to make sure, are you referring to the tests under `modules/solid_mechanics/test/tests/crystal_plasticity/euler_angles` ?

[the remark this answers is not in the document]
sapitts: yes, those ones--are those tests originally yours? I'm sorry if I misremembered the author

modules/phase_field/src/userobjects/EulerAngleUpdater.C
```diff
@@ -158,19 +158,6 @@ EulerAngleUpdater::initialize()
...
-  return _angles.size();
-}
-
-const EulerAngles &
-EulerAngleUpdater::getEulerAngles(unsigned int i) const
-{
-  mooseAssert(i < getGrainNum(), "Requesting Euler angles for an invalid grain id");
-  return _angles[i];
-}
-
```
SudiptaBiswas, line 180: Why are we removing this? Does it not affect any existing phase-field tests?

modules/phase_field/src/userobjects/RandomEulerAngleProvider.C
```diff
@@ -42,16 +41,3 @@ RandomEulerAngleProvider::initialize()
...
-RandomEulerAngleProvider::getGrainNum() const
-{
-  return _angles.size();
-}
-
-const EulerAngles &
-RandomEulerAngleProvider::getEulerAngles(unsigned int i) const
-{
-  mooseAssert(i < getGrainNum(), "Requesting Euler angles for an invalid grain id");
-  return _angles[i];
```
SudiptaBiswas, line 56: Same as before. Why are we removing the grain dependent Euler angles assignment?

modules/solid_mechanics/doc/content/source/userobjects/ComputeBlockOrientationByRotation.md
SudiptaBiswas: I am not sure if this is what we would want. It’s good for getting the grain orientation. But with deformation driven grain rotation we may see orientation changes only in elements near the grain boundaries. This approach would ignore that. Can we add an option to pick the orientation based on what is ‘maximum’? May be we should discuss this further.

modules/solid_mechanics/src/materials/crystal_plasticity/ComputeMultipleCrystalPlasticityStress.C
```diff
@@ -315,6 +317,23 @@ ComputeMultipleCrystalPlasticityStress::postSolveQp(RankTwoTensor & cauchy_stres
   RankTwoTensor rot;
   _elastic_deformation_gradient.getRUDecompositionRotation(rot);
   _updated_rotation[_qp] = rot * _crysrot[_qp];
+
+  // Calculate the misorientation
+  auto _misorientation_mat = _updated_rotation[_qp] * _updated_rotation_old[_qp].inverse();
```
SudiptaBiswas: Looks like this is calculating misorientation between two timesteps. We should rename this variable to reflect that. We should get the grain misorientation from the phase-field.

modules/phase_field/src/userobjects/EulerAngleUpdater.C
dewenyushu: These have been moved to `EulerAngleProvider.h` to avoid code duplication here and in `RandomEulerAngleProvider`

modules/solid_mechanics/src/materials/crystal_plasticity/ComputeMultipleCrystalPlasticityStress.C
dewenyushu: Yeah makes sense - what name would you suggest?

modules/solid_mechanics/doc/content/source/userobjects/ComputeBlockOrientationByRotation.md
dewenyushu: We have the option to choose the orientation of the element that exhibits the "maximum" rotation in the `ComputeBlockOrientationByMisorientation` object.

sapitts:
I'll take this PR over, but it will take me some time before I am able to spend a reasonable amount of time working with through the current state

modules/phase_field/src/userobjects/RandomEulerAngleProvider.C
sapitts: echoing the previous comment to the same question about reducing code duplication

modules/solid_mechanics/src/materials/crystal_plasticity/ComputeMultipleCrystalPlasticityStress.C
SudiptaBiswas: This needs to be updated in the next Commit. Other comments have been resolved.

sapitts: how does `_delta_misorientation_mat` sound as a replacement? I'll note that it's just a local variable, defined inline

sapitts:
@dschwen would you review this PR please? I believe the two failures are unrelated to this PR

sapitts:
@dschwen would you review this PR please? all tests are passing with the latest rebase

modules/solid_mechanics/include/userobjects/ComputeBlockOrientationByMisorientation.h
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * This is called before execute so you can reset any internal data.
+   */
+  virtual void initialize() override;
+
+  /**
+   * Called on every "object" (like every element or node).
+   * In this case, it is called at every quadrature point on every element.
+   */
```
dschwen, line 35: Remove comment, it is generic and also wrong. execute is called once per element here. Rather describe what you are doing in execute!

modules/solid_mechanics/include/userobjects/ComputeBlockOrientationByMisorientation.h
```diff
@@ -0,0 +1,61 @@
...
+  virtual void execute() override;
+
+  /**
+   * Called when using threading.  You need to combine the data from "y"
+   * into _this_ object.
+   */
```
dschwen: Again, this is documented in parent classes. The derived class should have specific docs describing what exactly is combined.

dschwen: same for finalize

modules/solid_mechanics/include/userobjects/ComputeBlockOrientationByRotation.h
```diff
@@ -0,0 +1,63 @@
...
+   */
+  EulerAngles computeSubdomainEulerAngles(const SubdomainID & sid);
+
+  // updated quaternion
+  const MaterialProperty<RankTwoTensor> & _updated_rotation;
+
+  /// number of bins for each quaternion component
+  unsigned int _bins;
+
+  /// L_norm value for averaging
```
dschwen: What is "L norm"? A description beyond just the name of the variable would be helpful.

modules/solid_mechanics/src/materials/crystal_plasticity/ComputeUpdatedEulerAngle.C
```diff
@@ -61,6 +61,6 @@ ComputeUpdatedEulerAngle::computeEulerAngleFromRotationMatrix(const RankTwoTenso
   // convert EulerAngles to RealVectorValue
   euler_angle = (RealVectorValue)ea;
 
-  if (!getParam<bool>("radian_to_degree"))
+  if (getParam<bool>("degree_to_radian"))
```
dschwen: In a _per QP_ call we should not be using `getParam`. Could you please add a bool member here?

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationBase.C
```diff
@@ -0,0 +1,43 @@
...
+EulerAngles
+ComputeBlockOrientationBase::getBlockOrientation(SubdomainID block) const
+{
+  // Note that we can't use operator[] for a std::map in a const function!
+  if (_block_ea_values.find(block) != _block_ea_values.end())
+    return _block_ea_values.find(block)->second;
+
+  mooseError("Unknown block requested for Euler angle values!");
+
+  return EulerAngles(); // To satisfy compilers
```
dschwen:
```suggestion
  return _block_ea_values.at(block);
```

you can use `at`. It will throw if the map element is not found, so for a nicer error message you'd do try/catch.

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByMisorientation.C
```diff
@@ -0,0 +1,114 @@
...
+  _block_ea_values.clear();
+  _grain_misorientation.clear();
+}
+
+void
+ComputeBlockOrientationByMisorientation::execute()
+{
+  // Compute the average of the rotation matrix and misorientation in this element
+  RankTwoTensor rot;
+  MathUtils::mooseSetToZero(rot);
```
dschwen: `rot` is already 0! 

```suggestion
```

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByMisorientation.C
```diff
@@ -0,0 +1,114 @@
...
+  Real misorient = 0.0;
+
+  for (unsigned int _qp = 0; _qp < _qrule->n_points(); _qp++)
+  {
+    rot += _JxW[_qp] * _coord[_qp] * _updated_rotation[_qp];
+    misorient += _JxW[_qp] * _coord[_qp] * _misorient[_qp];
+  }
+  rot /= _current_elem_volume;
+  misorient /= _current_elem_volume;
```
dschwen, line 55: The average of multiple rotation matrices is generally _not_ a rotation matrix. I'm surprised this works, but I suppose you will get "an answer" for the quaternion, it will just not be well defined!

dschwen: It might be good enough if the orientations are really close together. But you should project rot onto SO(3) using SVD to get a better result.

```
(U, S, Vt) = svd(rot)
rot_fixed = U * Vt
```

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByMisorientation.C
```diff
@@ -0,0 +1,114 @@
...
+  // compute Quaternion from rotation matrix
+  Eigen::Quaternion<Real> q(rot_mat);
+
+  // compute EulerAngle from Quaternion
+  EulerAngles ea = EulerAngles(q);
+
+  // store value for the current misorientation in the subdomain
+  // save EulerAngle in tuple so that we can gather the data from all processors
+  _grain_misorientation[_current_elem->subdomain_id()].push_back(
+      std::make_tuple(misorient, ea.phi1, ea.Phi, ea.phi2));
```
dschwen:
```suggestion
  _grain_misorientation[_current_elem->subdomain_id()].emplace_back(misorient, ea.phi1, ea.Phi, ea.phi2);
```

should work here. But that's just a small potential optimization.

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByMisorientation.C
```diff
@@ -0,0 +1,114 @@
...
+{
+  const std::set<SubdomainID> & blocks = _fe_problem.mesh().meshSubdomains();
+
+  for (std::set<SubdomainID>::const_iterator it = blocks.begin(); it != blocks.end(); ++it)
+  {
+    // Sync data from all processors (gather the maximum misorientation and the corresponding
+    // EulerAngle from every processor)
+    _communicator.allgather(_grain_misorientation[*it]);
+    _block_ea_values[*it] = computeSubdomainEulerAngles(*it);
+  }
```
dschwen:
```suggestion
  for (const auto & block : _fe_problem.mesh().meshSubdomains())
  {
    // Sync data from all processors (gather the maximum misorientation and the corresponding
    // EulerAngle from every processor)
    _communicator.allgather(_grain_misorientation[block]);
    _block_ea_values[block] = computeSubdomainEulerAngles(block);
  }
```

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByMisorientation.C
```diff
@@ -0,0 +1,114 @@
...
+  bool has_update = false;
+  for (const auto & orient : _grain_misorientation[sid])
+  {
+    if (std::get<0>(orient) > max_misorientation)
+    {
+      max_misorientation = std::get<0>(orient);
+      ea = EulerAngles(std::get<1>(orient), std::get<2>(orient), std::get<3>(orient));
+      has_update = true;
+    }
+  }
```
dschwen:
```suggestion
  for (const auto & [misorientation, phi1, Phi, phi2] : _grain_misorientation[sid])
  {
    if (misorientation > max_misorientation)
    {
      max_misorientation = misorientation;
      ea = EulerAngles(phi1, Phi, phi2);
      has_update = true;
    }
  }
```

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByRotation.C
```diff
@@ -0,0 +1,144 @@
...
+  // store quaternion values for the current element
+  _quat[_current_elem->subdomain_id()].push_back(std::make_tuple(q.w(), q.x(), q.y(), q.z()));
+}
+
+void
+ComputeBlockOrientationByRotation::finalize()
+{
+  const std::set<SubdomainID> & blocks = _fe_problem.mesh().meshSubdomains();
+
+  for (std::set<SubdomainID>::const_iterator it = blocks.begin(); it != blocks.end(); ++it)
```
dschwen: same

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByRotation.C
```diff
@@ -0,0 +1,144 @@
...
+  // quaternion average matrix Q*w*Q^T
+  typedef Eigen::Matrix<Real, 4, 4> Matrix4x4;
+  Matrix4x4 quat_mat = Matrix4x4::Zero();
+  typedef Eigen::Matrix<Real, 4, 1> Vector4;
+
+  Real total_weight = 0.0;
+
+  for (const auto & q : _quat[sid])
+  {
+    Vector4 v(std::get<0>(q), std::get<1>(q), std::get<2>(q), std::get<3>(q));
```
dschwen: Use structured binding again

modules/solid_mechanics/src/vectorpostprocessors/BlockOrientationFromUserObject.C
```diff
@@ -0,0 +1,108 @@
...
+BlockOrientationFromUserObject::finalize()
+{
+  // parallel communication
+  for (const auto row : make_range(_num_rows))
+  {
+    for (const auto col : make_range(_num_cols))
+    {
+      _communicator.max((*_output_vector[col])[row]);
+    }
+  }
```
dschwen:
```suggestion
  for (const auto row : make_range(_num_rows))
    for (const auto col : make_range(_num_cols))
      _communicator.max((*_output_vector[col])[row]);
```

modules/solid_mechanics/src/vectorpostprocessors/BlockOrientationFromUserObject.C
```diff
@@ -0,0 +1,108 @@
...
+  _output_vector.resize(_num_cols);
+
+  for (const auto j : make_range(_num_cols))
+  {
+    if (j == 0)
+      _output_vector[j] = &declareVector("subdomain_id");
+    else
+    {
+      _output_vector[j] = &declareVector("euler_angle_" + std::to_string(j)); // can change
+    }
```
dschwen:
```suggestion
      _output_vector[j] = &declareVector("euler_angle_" + std::to_string(j)); // can change
```

dschwen (commented):
See comments

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByRotation.C
sapitts: same what please?

modules/solid_mechanics/include/userobjects/ComputeBlockOrientationByRotation.h
sapitts: ~...but who doesn't like recursive self-definition of variables.....?~

Unnecessary comments aside (sorry), it seems this variable is actually a parameter and I am going to have to dig into the cited reference to figure out what this value is besides an exponent for the quaternion weighting expression

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByRotation.C
sapitts: done

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByMisorientation.C
dschwen: Or do Newton-Schulz right here

modules/solid_mechanics/src/userobjects/ComputeBlockOrientationByMisorientation.C
```diff
@@ -0,0 +1,108 @@
...
+
+  // transform RankTwoTensor to Eigen::Matrix
+  Eigen::Matrix<Real, 3, 3> rot_mat;
+
+  for (unsigned int i = 0; i < 3; ++i)
+    for (unsigned int j = 0; j < 3; ++j)
+      rot_mat(i, j) = rot(i, j);
+
+  // compute Quaternion from rotation matrix
+  Eigen::Quaternion<Real> q(rot_mat);
```
dschwen:
```suggestion
  // SVD-based projection to SO(3)
  Eigen::JacobiSVD<Eigen::Matrix<Real, 3, 3>> svd(
      rot_mat, Eigen::ComputeFullU | Eigen::ComputeFullV);

  Eigen::Matrix<Real, 3, 3> U = svd.matrixU();
  Eigen::Matrix<Real, 3, 3> V = svd.matrixV();

  // Enforce proper rotation (det = +1)
  Eigen::Matrix<Real, 3, 3> R = U * V.transpose();
  if (R.determinant() < 0.0)
  {
    Eigen::Matrix<Real, 3, 3> D = Eigen::Matrix<Real, 3, 3>::Identity();
    D(2, 2) = -1.0;
    R = U * D * V.transpose();
  }
  
  // compute Quaternion from rotation matrix
  Eigen::Quaternion<Real> q(R);
```

dschwen (approved):

PR #30014, opened 2025-03-03
Set FParser epsilon to zero

dschwen:
## Reason
Function parser performs fuzzy comparisons with an epsilon value, that by default, is too large (1e-12). This can lead to numerical convergence stagnation. What's even worse is that the epsilon is utilized during the function optimization stage, and can lead to wrong simplifications. 

## Design
It is safer to default the epsilon to zero.

## Impact
Safer parsed material/aux evaluations.

### DIFF
--- a/framework/src/utils/FunctionParserUtils.C
+++ b/framework/src/utils/FunctionParserUtils.C
@@ -42,7 +42,7 @@ FunctionParserUtils<is_ad>::validParams()
   params.addParamNamesToGroup(
       "enable_jit enable_ad_cache enable_auto_optimize disable_fpoptimizer evalerror_behavior",
       "Parsed expression advanced");
-  params.addParam<Real>("epsilon", FunctionParser::epsilon(), "Fuzzy comparison tolerance");
+  params.addParam<Real>("epsilon", 0, "Fuzzy comparison tolerance");
   return params;
 }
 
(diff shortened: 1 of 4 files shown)


dschwen:
Fix for Bison is trivial.

framework/src/utils/FunctionParserUtils.C
```diff
@@ -191,22 +191,34 @@ template <>
 void
 FunctionParserUtils<false>::functionsOptimize(SymFunctionPtr & parsed_function)
 {
+  // set desired epsilon for optimization!
+  auto tmp_eps = parsed_function->epsilon();
+  parsed_function->setEpsilon(_epsilon);
```
lindsayad, line 196: The epsilon documentation string doesn't mention anything about optimization; the documentation string just makes it seem like there's a single epsilon. But this temporary setting of epsilon makes it seem like there's one epsilon for optimization and another epsilon elsewhere. If there are two epsilons at play let's make sure it's indicated in the documentation string, and should there be user ability to set both?

lindsayad (changes requested):

framework/src/utils/FunctionParserUtils.C
dschwen: There is one global epsilon variable. I could add the functionality of setting different epsilons for both stages.

framework/src/utils/FunctionParserUtils.C
```diff
@@ -42,7 +42,12 @@ FunctionParserUtils<is_ad>::validParams()
   params.addParamNamesToGroup(
       "enable_jit enable_ad_cache enable_auto_optimize disable_fpoptimizer evalerror_behavior",
       "Parsed expression advanced");
-  params.addParam<Real>("epsilon", FunctionParser::epsilon(), "Fuzzy comparison tolerance");
+  params.addParam<Real>("epsilon", 0, "Fuzzy comparison tolerance");
+  params.addParam<Real>(
+      "fpoptimizer_epsilon",
+      "Fuzzy comparison tolerance for the FPOptimizer algebraic optimization stage. This defaults "
+      "to teh selected value of the `epsilon` parameter. Use this with care, as it can lead to "
```
lindsayad, line 49:
```suggestion
      "to the selected value of the `epsilon` parameter. Use this with care, as it can lead to "
```

lindsayad (changes requested):
Love the commit title. Note that I just asked whether we should allow the user to set both. I was looking to the expert to make an intelligent decision on that 😄 

Looks like `_fpoptimizer_epsilon` is not used

dschwen:
> Love the commit title. Note that I just asked whether we should allow the user to set both. I was looking to the expert to make an intelligent decision on that 😄
> 
> Looks like `_fpoptimizer_epsilon` is not used

Ha, yeah, I made this commit early in the morning from home, and must have not saved the fixed version. I'll push that later. But this option is not a bad idea. It gives the user the option to tighten the optimizer epsilon, while keeping it lose for eval.

lindsayad:
I feel like we should probably wait until the thread safety PR from libmesh is included in our submodule. What do you think @dschwen ?

dschwen:
> I feel like we should probably wait until the thread safety PR from libmesh is included in our submodule. What do you think @dschwen ?

Agree

lindsayad:
The libmesh code with the mutex will go in with https://github.com/idaholab/moose/pull/30180. It's not in yet

lindsayad (approved):

dschwen:
Hm, @lindsayad approved, but his review still registers as "Requested changes"...

loganharbour:
We can get this in once the package updates have cleared next.

lindsayad (approved):

loganharbour:
Bison is stuck on devel due to VTB; I think the breakage there is valid? we might need to wait for bison to merge to master.

PR #30040, opened 2025-03-06
Correct versioner verify logic for version bump

### LINKED ISSUE #30039: Fix `Versioner` verify
### Motivation

Versioner verify doesn't currently have the correct check for checking a version bump. If the version is bumped, the logic for checking the build number is incorrect.

### Design

Correct this logic.

### Impact

Minor.

loganharbour:
Closes #30039


milljm (approved):
Tested this with an "everything change" with great success.

PR #30122, opened 2025-03-18
[WIP] Add a Fusion Module

cticenhour:
## Reason
In order to coordinate modeling and simulation efforts for fusion blankets (occurring across multiple applications), it was deemed necessary to create a so-called "fusion module" to house simulation objects as well the full-size breeding blanket section model for the Fusion Nuclear Science Facility presented in [Fande's work from 2022](https://doi.org/10.1109/TPS.2022.3173158) as well as more recent additions from @FranklinTrevor's internships

## Design
This module will house objects to take in heat generation information from OpenMC, perform THM and heat transfer simulations in blanket cooling channels and volumes, as well as V&V examples and tests. Finally, it will contain input files associated with the full scale FNSF blanket simulation. 

## Impact
New capability.

> [!WARNING]
> There is still much to do in this PR before it is ready to be merged. At minimum a rebase needs to be performed from December 2023 MOOSE to today; I have already done so in another branch and have found several issues related to test diffs and other deprecations. Posting this branch in its current state to get going toward merging this thing in earnest!


aprilnovak:
What is the scope of the OpenMC tally mapping being added here? (the huge file count is not loading for me properly on github on hotel wifi). Would you be open to putting any OpenMC related capabilities in Cardinal instead, for modularity and clarity? IMO tally extraction from OpenMC is not just a fusion-related activity.

cticenhour:
I would be open to it, if there is interest in what we have here....perhaps you, @simopier, and I should chat. To briefly summarize and point out a file to look at, we apply the independently-generated OpenMC heating output through `FNSFSourceAux`, which can then be used as an IC or a source term in the heat conduction solve. I would welcome your opinion on what might be useful to Cardinal. That would necessitate some reconfiguration (perhaps putting the full-scale FNSF example at the level of the VTB, where Cardinal and this module would both be available), but that's not difficult to coordinate.

cticenhour:
A small portion of this PR has been moved to #31960. The rest of the code is being moved to SALAMANDER, referencing idaholab/salamander#112.

PR #30123, opened 2025-03-18
Moose Syntax Highlighting

### LINKED ISSUE #30112: Syntax highlighting for moose inputs
### Motivation

Currently, MOOSE input examples on webpages are not syntax-highlighted, i.e. plain black-and-white. Having highlighting for our HIT format would make the snippets much more readable. Additionally, since we have object information during the build process, I think we can embed clickable links to the object documentation and hover-over descriptions for parameters.

### Design

The doc build does include syntax highlighting for other languages, e.g. C++ and python, using [PrismJS](https://prismjs.com/index.html). We include a min JavaScript and CSS file from this package in our contrib. The Prism package is quite extendible, so adding a syntax for MOOSE inputs would be tractable. That contrib file is a direct download, but building that file from a forked version of Prism with the MOOSE extension is not difficult.

I think the syntax highlighting in VSCode is really nice, so I think a direct copy of that styling would be good.

For embedding object and parameter information, I think we can use our `pyhit.py` capability to first parse the input file in `listing.py`. Once parsed, each "node" of the input file can be linked to the object information in AppSyntax. We can then directly modify the listing content with the data from AppSyntax. Prism will do it's formatting on this modified content. A relatively simple JS file can then search the meta data to add anchors to doc pages and tooltips of descriptions.

### Impact

The translating step of the doc build would become slower, especially for large applications. Also, the JS for parsing the meta data would make page loading slower. But I think the benefits would outweigh the costs here.

zachmprince:
## Reason

Closes #30112

## Design

The modification of the `prism.min.js` is the implementation of the syntax highlighting. The line for the moose input part was generated from [my prism fork](https://github.com/zachmprince/prism/blob/master/components/prism-moose.js). Since I wanted to look more like the VSCode style, I modified the moose.css to include colors seen in the vscode theme that many of us use. Here is a snippet of what the highlighting looks like:

image

I went a little above-and-beyond, so to speak, and added links and tooltips to page documentation recognized by the syntax. For instance, clicking on `[Functions]` will take you to the Functions syntax page and `ParsedFunction` will take you to ParsedFunction.md. Also, hovering over the parameters and the class name (`ParsedMaterial`) will show the parameter/class description.

image

image

This is my first time writing JavaScript, so please go easy on my ignorance on how to code in that. All the python code and custom JS file is to support the syntax linking and description display. If we don't end up liking that, this is a much simpler PR.

## Impact

The runtime for the doc build seems about the same. Pages with moose input listings do show some confusing text on the initial load, but quickly turn into the right thing once the JS is done. I don't see much other down sides other than we are dependent on a fork prism to get the syntax highlighting regex, but that is well documented in the prism.min.js file.


loganharbour (changes requested):
Not up to date with current packages

python/MooseDocs/common/read.py
```diff
@@ -49,4 +49,6 @@ def get_language(filename):
         return 'cpp'
     elif ext == '.py':
         return 'python'
+    elif ext == '.i':
+        return 'moose'
```
loganharbour, line 53: Does this work for test specs too? Should be able to:

```suggestion
    elif ext in ['.i', 'tests']:
        return 'moose'
```

python/MooseDocs/extensions/listing.py
```diff
@@ -86,6 +120,130 @@ def createToken(self, parent, info, page, settings):
...
+        """
+        if self.extension.syntax is None:
+            return content
+
+        syntax_ext: appsyntax.AppSyntaxExtension = self.extension.syntax # Convenience
+
+        content_lines = content.splitlines()
+        try:
+            root = pyhit.parse(content)
+        except:
```
loganharbour, line 154: Is it worth adding a warning here to see if this is a common issue?

python/MooseDocs/extensions/listing.py
```diff
@@ -86,6 +120,130 @@ def createToken(self, parent, info, page, settings):
...
+
+                # Append content with description
+                ci = node.line(name=param) - 1
+                line = content_lines[ci]
+                ind = line.find(param) + len(param)
+                content_lines[ci] = line[:ind] + '<<<' + json.dumps(attr) + '>>>' + line[ind:]
+
+        return '\n'.join(content_lines)
+
+    def _stringifySyntaxData(
```
loganharbour, line 220: Docstring this please

loganharbour (changes requested):
All minor. Will require another re-do of the package bump

python/MooseDocs/common/read.py
zachmprince: No it doesn't. That would require another "language" to be defined in prism.js. The vscode extension does have a separate language definition for test specs, I believe.

python/MooseDocs/extensions/listing.py
zachmprince: I don't think so. I think it is too common that people put in snippets that are invalid according to pyhit. And example would be something like this:
```
[Mesh]
   [gmg]
       type = GeneratedMeshGenerator
       ...
   []
[]
```

This should probably be allowed without warnings.

python/MooseDocs/common/read.py
loganharbour: Future thing - if you want.

loganharbour (approved):
Coordinate before merging because this is a package update.

loganharbour:
Failure is unrelated.

PR #15597, opened 2020-07-13
Bug fixes for restarting transient multiapp solves from steady multiapp solves

lindsayad:
When "restarting" a transient simulation from a steady-state simulation,
the transient systems will have more vectors than the steady-state
systems. Naively trying to read data corresponding to all the vectors in
the transient simulation will generate garbage results because you're
trying to read-in data that doesn't exist.

Additionally, we need to be careful about how much state we advance
when restarting multiapp problems. If we do too many state advances, then
our stateful material properties will be in a bad place.

Closes #15333

Issue #15333, opened 2020-05-21
Checkpoint restarts and MultiApp causing faulty behavior for coupled simulations


harterj:
## Bug Description
When a transient, coupled (using MultiApp) simulation is restarted from a steady state solution, a few things go wrong.
 
1) The temperature (auxvariable at the subapp level) is not properly restarted at the subapp level:

`bison0: Postprocessor Values:`
`bison0: +----------------+----------------+--------------------+`
`bison0: | time           | avg_temp       | disp_x_max_element | |`
`bison0: +----------------+----------------+--------------------+`
`bison0: |   0.000000e+00 |  -3.158068e+06 |       1.348055e-01 |  `
`bison0: |   1.000000e+01 |   9.598427e+02 |       2.699262e-01 |   `
`bison0: +----------------+----------------+--------------------+`
(I only showed two postprocessors here. The full output can be seen by running the included examples)
In this case, it does not affect the solve because the temperature from master is transferred before the calculation takes place. But that weird initial temperature will undoubtedly confuse analysts and indicate a problem with the restart system.
 
2) The displacements seem properly restarted, but oddly get “doubled” after the first time step. We think the thermal expansion eigenstrain is being reset during the restart causing the temperature change to once again impose a displacement. This behavior should not be occurring.

## Steps to Reproduce
 
To test the proper restart behavior, we run a null-transient, that is we first run a steady-state calculation and restart from it with a Transient executioner, expecting the solution to be time-independent.
 
More specifically, input files and mesh files are included in `examples/doubling_example` [https://github.com/harterj/moose/tree/doubling_bug](https://github.com/harterj/moose/tree/doubling_bug) These are for a simple problem coupling temperature to mechanics. 
Run `master_ss.i` first, and then run `master_tr.i` -- `master_tr.i` restarts from the steady state solution. You will see a negative, large initial value for temperature (problem 1)). You will also see the values for displacement restart, and nearly exactly double. This is present in this pared down example, but also in larger complex simulations.

## Impact
We are trying to couple neutronics and thermal mechanics for a micro reactor project. This prevents us from advancing on our project and we would like as much assistance as possible, please.

Tag @vincentlaboure @elementx54 @YaqiWang @jortensi @snschune

lindsayad:
@harterj in the future can you put your input files in module examples directory corresponding to the modules needed to execute the input file? In this case it looks like it would need to be in `combined/examples`. Then I can just build and run without first looking through the input files and figuring out what modules I need to build

[...]

PR #30135, opened 2025-03-19
Make Transient APIs more atomic

### LINKED ISSUE #30111: Always Populate Old State on Restart
### Motivation

Neutronics applications often restart a transient calculation from a steady-state calculation. In this case, old state is not populated on restart because old state itself doesn't exist in the steady-state calculation. It would be beneficial to copy the current state to the old state in this case.

### Design

Check whether the checkpoint file contains old state and the restarted simulation requires old state. If the checkpoint file does not contain old state while the restarted simulation requires it, copy the current state to the old state.

### Impact

Avoid old state being set to zero.

NamjaeChoi:
Closes #30111.

Related Griffin PR: https://github.inl.gov/ncrc/griffin/pull/2459

### DIFF
--- a/framework/src/executioners/TransientBase.C
+++ b/framework/src/executioners/TransientBase.C
@@ -344,10 +344,10 @@ TransientBase::incrementStepOrReject()
         _problem.adaptMesh();
 #endif
 
+      advanceState();
+
       _time_old = _time;
       _t_step++;
-
-      _problem.advanceState();
 
       if (_t_step == 1)
         return;
@@ -387,6 +387,12 @@ TransientBase::incrementStepOrReject()
     _time_stepper->rejectStep();
     _time = _time_old;
   }
+}
+
+void
+TransientBase::advanceState()
+{
+  _problem.advanceState();
 }
 
 void
--- a/framework/include/executioners/TransientBase.h
+++ b/framework/include/executioners/TransientBase.h
@@ -75,6 +75,8 @@ public:
    */
   virtual void incrementStepOrReject();
 
+  virtual void advanceState();
+
   virtual void endStep(Real input_time = -1.0);
 
   /**


framework/src/restart/RestartableEquationSystems.C
```diff
@@ -19,6 +19,9 @@
 const std::string RestartableEquationSystems::SystemHeader::system_solution_name =
     "SYSTEM_SOLUTION";
 
+const std::string RestartableEquationSystems::SystemHeader::system_old_solution_name =
```
loganharbour, line 22: Is there a static definition of this vector name somewhere else? It would be nice to not hard code it here if we can get it from somewhere else

loganharbour (commented):
This makes sense to me. We do need a test for this, though.

Wouldn't a recover test that has something that depends on old values work?

NamjaeChoi:
> This makes sense to me. We do need a test for this, though.
> 
> Wouldn't a recover test that has something that depends on old values work?

There is a failing test `restart-transient-from-ss-with-stateful`. I'm not yet sure if I messed up something or the test itself was wrong. I need to take a closer look. If it is the latter, this test would be able to cover my change.

YaqiWang:
> @YaqiWang I think I understand restarting systems a bit better now, and I feel like we can easily implement a flexible restart where we can specify variables we want to restart. It's currently looping over all the variables found in the header `sys_header.variables` but we might be able to restrict this list on user's request.

We want to be able to just load variables in a checkpoint file into individual variables like for the weighting functions in IQS with adjoint solution.

framework/src/restart/RestartableEquationSystems.C
```diff
@@ -264,6 +267,9 @@ RestartableEquationSystems::load(std::istream & stream)
 
     bool modified_sys = false;
 
+    bool init_old_solution = !sys_header.vectors.count(SystemHeader::system_old_solution_name) &&
```
YaqiWang, line 270: We will need to do the same for `older`.

loganharbour:
Does the restart happen late enough that all objects have been initialized that request old and older states?

framework/src/restart/RestartableEquationSystems.C
NamjaeChoi: I see `Moose::OLD_SOLUTION_TAG`

framework/src/restart/RestartableEquationSystems.C
NamjaeChoi: If checkpoint only has current and old, which should go into older?

YaqiWang: That can be an error? Otherwise, I would say old go to older.

NamjaeChoi: If checkpoint only has current, current goes to older? That sounds too many conditionals.

lindsayad:
@loganharbour do you want to dig into this? Otherwise I can

loganharbour:
@YaqiWang and @NamjaeChoi have you had more time to think about this?

YaqiWang:
@loganharbour I think Namjae and me are waiting for a help.

lindsayad:
So this test was created by me in https://github.com/idaholab/moose/pull/15597 which was in response to issue https://github.com/idaholab/moose/issues/15333

lindsayad:
The diff is occurring in the sub-app. I'm guessing that your change @NamjaeChoi was not really supposed to be about multiapps? I'm guessing we're running into a difference between main app and subapp restart here

YaqiWang:
Is the code path with this change hit by both main-app and sub-app?

lindsayad:
yes

NamjaeChoi:
I thought it would work regardless of main-app or sub-app.

lindsayad:
Yea the issue is not main vs. sub.

The issue is in the physics. This is an incremental strain calculation. The "steady-state" calculation increments its strain from the zero guess initial state. So you can think I need to increment by strain X amount from 0 in order to have my static equilibrium. Well now with this change, upon restart copying the current solution backwards leads to a base strain of 0 + X. So that means we now must increment by an additional X to return to our static equilibrium. So the final strain must be 2 * X instead of X. So we lose the zero transient

YaqiWang:
The change here assumes we have equilibrium with the steady state solution, i.e. the old solution is equal to the current solution if there is an old solution. Should this `incremental strain calculation` be skipped during restart?

NamjaeChoi:
A reactor transient occurs due to a perturbation (e.g. rod ejection) happened during a normal operation. That's the basis of restarting transient from steady-state in reactor physics applications.

lindsayad:
I do wonder if it makes any sense to have a `Steady` calculation with an incremental strain formulation. I feel like it doesn't

YaqiWang:
If the checkpoint file is generated by a pseudo transient with `Transient` executioner, we will not have this issue. Maybe that is the right fix if the steady-state solution depends on the history approaching the steady state.

lindsayad:
That is a very elegant solution but if I do that then that test no longer satisfies the premise https://github.com/idaholab/moose/issues/15333

lindsayad:
Still investigating this. On a different note, @NamjaeChoi have you looked at `FEProblemBase::copySolutionBackwards`? I wonder if that might be useful given @YaqiWang's comment about older state. Of course this would not be meaningful for restarting transient from steady, since there is no old solution to copy backwards into the older solution

lindsayad:
@loganharbour reading our coverage report, it looks like we never enter the below conditional in our framework testing. Is this code meant to be hit? I do not see any material property storage loading in `FEProblemBase::initialSetup` after `MaterialPropertyStorage::setRestartInPlace`

https://github.com/idaholab/moose/blob/60b56f410427d419c9a7529f7904b5f8bbfabd3f/framework/src/materials/MaterialPropertyStorage.C#L841-L847

NamjaeChoi:
> Still investigating this. On a different note, @NamjaeChoi have you looked at `FEProblemBase::copySolutionBackwards`? I wonder if that might be useful given @YaqiWang's comment about older state. Of course this would not be meaningful for restarting transient from steady, since there is no old solution to copy backwards into the older solution

I haven't but thanks for the suggestion.

lindsayad:
So after investigation, I guess I need further description of your issue. #30111 doesn't give much detail. We already call `FEProblemBase::advanceState` at the beginning of each timestep so this already does what you claim you want in your issue, copying the current solution back into old and shifting stateful properties. This is what allows the current mechanics test to achieve its null-transient like it's supposed to. So what's the problem for you guys? You need this data available during `FEProblemBase::initialSetup`?

lindsayad (changes requested):
Need to understand the motivation for this PR

NamjaeChoi:
Griffin might be abusing the transient executioner, but we have a solve object that executes LINEAR before the first `advanceState`.

lindsayad:
Have you considered a custom executioner that advances the state before this solve?

NamjaeChoi:
@YaqiWang @zachmprince Can you elaborate why we can't advance state before preStep in IQS?

YaqiWang:
I do not think we cannot, it is just no easier interface to allow us to insert the codes in `preStep` after the advance state.

lindsayad:
If we need to make the executioner interfaces more atomic in the framework to help you, I would be open to that

NamjaeChoi:
Okay, do you mind adding a virtual `TransientBase::advanceState` method?

NamjaeChoi:
Well never mind, it doesn't work because our operation should happen after `advanceState` but still when time and time step are zero.

NamjaeChoi:
Well not sure again, can we switch
```
_time_old = _time;
_t_step++;

_problem.advanceState();
```
to
```
_problem.advanceState();

_time_old = _time;
_t_step++;
```
in `TransientBase::incrementStepOrReject`? Then we can hook our operation to a new virtual method that will by default just call `_problem.advanceState()`.

lindsayad:
what would the new virtual method be?

lindsayad:
`advanceState`? 😄

lindsayad:
Let's see what happens with your proposed change

lindsayad:
So that switch is fine. Do you want to push a proposed new virtual method?

NamjaeChoi:
Failures unrelated?

YaqiWang:
> Failures unrelated?

Probably. Otherwise, failures would be more drastic ;-)

NamjaeChoi:
The fix doesn't work as I expected. We will need to take a closer look in Griffin.

NamjaeChoi:
Can we separate `FEProblemBase::advanceState` into `FEProblemBase::advanceSolutionState` and `FEProblemBase::advanceMaterialState`?

lindsayad:
I would be willing to have `FEProblemBase::advanceState` call those two new APIs

NamjaeChoi:
Sure

lindsayad (commented):
So how will you leverage this in Griffin? What advantage does splitting the method give? Is there an issue with advancing the material state at the same time as the solution state?

NamjaeChoi:
> So how will you leverage this in Griffin? What advantage does splitting the method give? Is there an issue with advancing the material state at the same time as the solution state?

The biggest issue is that we need to evaluate residual on time step 0 using the steady-state solution before transient starts, and we have a material property that depends on the solution. But if my understanding is correct, advancing material states is not done by copying current material properties to the old properties but is done by simply swapping the two storages, which will set the current material properties to zero at first `advanceState`. So that initial residual evaluation should be performed after copying current solution to old but before swapping material storages.

lindsayad (dismissed):
Well I'm still pretty nervous about what is happening in Griffin, but the changes in front of my face look pretty reasonable

NamjaeChoi:
> Well not sure again, can we switch
> 
> ```
> _time_old = _time;
> _t_step++;
> 
> _problem.advanceState();
> ```
> 
> to
> 
> ```
> _problem.advanceState();
> 
> _time_old = _time;
> _t_step++;
> ```
> 
> in `TransientBase::incrementStepOrReject`? Then we can hook our operation to a new virtual method that will by default just call `_problem.advanceState()`.

Did this really have no issue? If so, how can this change result in this catastrophic failure?

NamjaeChoi:
This cannot be done because some problems override `FEProblemBase::advanceState`. I need to call `FEProblemBase::advanceState` in `TransientBase` which then doesn't allow me to separate solution and material.

NamjaeChoi:
> This cannot be done because some problems override `FEProblemBase::advanceState()`. I need to call `_problem.advanceState()` in `TransientBase` which then doesn't allow me to separate solution and material.

Let us deal with this in the Griffin side.

lindsayad:
So we want to close this? Let me know if you can't find a solution in Griffin-only

NamjaeChoi:
We still want to hook our code with `advanceState`, so I will just wrap this PR up with adding `TransientBase::advanceState`

lindsayad (approved):
I see. I feel better about not having the option for users to split the advancement of solution and material state 👍

framework/src/executioners/TransientBase.C
```diff
@@ -344,11 +344,11 @@ TransientBase::incrementStepOrReject()
         _problem.adaptMesh();
 #endif
 
+      advanceState();
+
       _time_old = _time;
       _t_step++;
 
-      _problem.advanceState();
```
dschwen, line 350: Why is the order of timestep increase and advance state changed? The commit message just states "Make Transient APIs more atomic", but this does more.

NamjaeChoi: Just revert this PR on your need. This PR failed to resolve our issue in Griffin.

lindsayad: We can revert the whole PR?

NamjaeChoi: Yes

YaqiWang: @dschwen why this order matters to you? I am not suggesting to not revert the change, but having a code depending on this order makes me worried.

PR #30170, opened 2025-03-22
Fix uninitialized memory in UMAT

dschwen:
## Reason
@lindsayad came across a failure in `test:umat/multiple_blocks.multiple_blocks_two_materials_parallel` with valgrind pointing to uninitialized values.

## Design
I went over the code of the fortran UMAT plugin used in this test and found two uninitialized values (some components of the elasticity matrix DDSDDE and a potentially uninitialized direction vector DIR).

## Impact
Added zero initialization to both.


dschwen:
(relatively clean... if we ignore the leaking)

dschwen:
SAM build fail is unrelated

lindsayad (approved):

PR #30175, opened 2025-03-24
Add kernel for distributed loads on shell

### LINKED ISSUE #30174: Add capability to model distributed loads on shell element
### Motivation

Distributed loads on shell elements are commonly encountered in various engineering applications, such as self-weight or wind loads on thin structures, or gas pressure in pressure vessels and tanks. However, the current shell model does not support the functionality to model these distributed loads.

### Design

A new ADKernel (ADDistributedLoadShell) is added to calculate the contribution to the residual from the distributed loads

### Impact

- No impact on the exisiting features, tests ,...
- adds the capability to model distributed loads on shells

Kavan-Khaledi:
## Reason
Distributed loads on shell elements are commonly encountered in various engineering applications, such as self-weight or wind loads on thin structures, or gas pressure in pressure vessels and tanks. However, the current shell model does not support the functionality to model these distributed loads.

## Design
A new ADKernel (ADDistributedLoadShell) is added to calculate the contribution to the residual from the distributed loads

## Impact
- No impact on the exisiting features, tests ,...
- adds the capability to model distributed loads on shells

closes #30174


modules/solid_mechanics/include/kernels/ADDistributedLoadShell.h
```diff
@@ -0,0 +1,79 @@
+//* This file is part of the MOOSE framework
+//* https://www.mooseframework.org
```
milljm, line 2:
```suggestion
//* https://mooseframework.inl.gov
```
I have PR currently that is changing all references of .org to .inl.gov

If you would also change every other occurrence in new files, that would be appreciated! Thanks!

milljm: Easily done with a regex replacement:
```bash
cd moose
git grep -l "www.mooseframework.org" | xargs sed -i "" -e "s|www.mooseframework.org|mooseframework.inl.gov|g"
```

Kavan-Khaledi: @milljm thanks for the hint! I assume I need to wait until your PR is merged and then update all the references. Otherwise, I'll have to commit thousands of changes. Is that correct?

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
```diff
@@ -0,0 +1,107 @@
...
+InputParameters
+ADDistributedLoadShellTempl<is_ad>::validParams()
+{
+  InputParameters params = GenericKernel<is_ad>::validParams();
+  params.addClassDescription(
+      "Applies a distributed load (unit N/m2) on the shell plane in a given "
+      "direction (e.g. self_weight, wind load) or normal to the shell plan (e.g. pressure loads)");
+  params.addRequiredCoupledVar("displacements",
+                               "The string of displacements suitable for the problem statement");
+  params.addParam<Real>(
```
bwspenc, line 30: This parameter isn't really necessary because the user can just put something like `function=1` in their input file to get a constant value. You'll see separate constant parameters like this for older code that we had before we had that capability, but we're going away from that.

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
```diff
@@ -0,0 +1,107 @@
...
+      "factor",
+      1.0,
+      "The magnitude to use in computing a constant distributed load on shells (unit N/m2)");
+  params.addParam<bool>(
+      "project_load_to_normal", false, "True, if the distributed load is normal to the shell");
+  params.addParam<FunctionName>("function",
+                                "The function that describes the ADDistributedLoadShell");
+  params.addParam<PostprocessorName>(
+      "postprocessor", "Postprocessor that will supply the ADDistributedLoadShell value");
```
bwspenc, line 39: This option also isn't necessary because you can use a ParsedFunction to evaluate to the value of a PostProcessor. For example, you can do something like this:
```
[Functions]
  [fn]
    type = ParsedFunction
    expression = 'pp'
    symbol_names = 'pp'
    symbol_values = 'pp'
  []
[]
```
where `pp` is the name of a Postprocessor.

modules/solid_mechanics/doc/content/source/kernels/ADDistributedLoadShell.md
```diff
@@ -0,0 +1,17 @@
+# ADDistributedLoadShell
+
+!syntax description /Kernels/ADDistributedLoadShell
+
+## Description
+
+The `ADDistributedLoadShell` kernel calculates the contribution to the residual from the distributed loads (in units of N/m²) on the shell elements. The load can be applied either in a specified direction within the global x-y-z coordinate system (e.g., self-weight, wind load) or normal to the shell plane (e.g., pressure loads). To determine if the distributed load should be treated as a pressure load (applied normal to the shell), the parameter 'project_load_to_normal=true' must be set. When this parameter is enabled, the load is applied perpendicularly to the shell surface. By default, project_load_to_normal=false, meaning the load is applied in the global coordinate directions.
```
bwspenc, line 7:
```suggestion
The `ADDistributedLoadShell` kernel calculates the contribution to the residual from a distributed load (in units of force per area) on the shell elements. The load can be applied either in a specified direction within the global x-y-z coordinate system (e.g., self-weight, wind load) or normal to the shell plane (e.g., pressure loads). The `project_load_to_normal` parameter defines how the direction of the the load is defined. This parameter defaults to `false`, which indicates that the load is applied in the direction of the variable this is applied to (defined by the `variable` parameter). If `project_load_to_normal=true`, the distributed load is treated as a pressure load applied in the shell normal direction, perpendicular to the shell surface. For this case, an instance of this kernel is required for each of the displacement variables.
```

modules/solid_mechanics/doc/content/source/kernels/ADDistributedLoadShell.md
```diff
@@ -0,0 +1,17 @@
...
+
+The magnitude of the distributed load can be specified as either a scalar (use the input parameter factor, which defaults to 1.0), a function parameter, or a Postprocessor name. If more than one of these are given, they are multiplied by one another.
```
bwspenc, line 9:
```suggestion
The magnitude of the distributed load is specified as a function. A scalar value can be supplied to the function (i.e., `function=1.0`) for constant loading.
```

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
```diff
@@ -0,0 +1,107 @@
...
+    _function(this->isParamValid("function") ? &this->getFunction("function") : NULL),
+    _postprocessor(
+        this->isParamValid("postprocessor") ? &this->getPostprocessorValue("postprocessor") : NULL),
+    _project_load_to_normal(parameters.isParamSetByUser("project_load_to_normal")
+                                ? this->template getParam<bool>("project_load_to_normal")
+                                : false),
+    _nodes(4),
+    _use_displaced_mesh(this->template getParam<bool>("use_displaced_mesh"))
+{
+
```
bwspenc, line 59: Doesn't _ndisp have to be 3 for shells? You should probably check for that here and error out if it isn't 3.

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
```diff
@@ -0,0 +1,107 @@
...
+  }
+  if (_component == libMesh::invalid_uint)
+    mooseError("Problem with displacements in " + _name);
+}
+
+template <bool is_ad>
+GenericReal<is_ad>
+ADDistributedLoadShellTempl<is_ad>::computeQpResidual()
+{
+  for (unsigned int i = 0; i < 4; ++i)
```
bwspenc, line 76: We probably need to verify that the number of nodes for the current element is indeed 4 and generate a mooseError if not.

modules/solid_mechanics/test/tests/shell/static/clamped_plate_flat.i
```diff
@@ -0,0 +1,180 @@
...
+
+# where a = 50 m, q = -10.816 N/m^2 and D = E/(12(1-v^2))
+# The analytical series solution converges to 2.998535904e-03 m
+# when the first 16 terms of the series are considered (i.e., until
+# m & n = 7).
+
+[Mesh]
+  [gmg]
+    type = FileMeshGenerator
+    file = clamped_plate_flat.msh
```
bwspenc, line 26: I think the files that define the gmsh mesh are small text files, right? If so, it would be helpful to also check those in so that someone can modify the mesh if needed in the future. We're not very consistent about doing that in MOOSE tests, but I personally try to always check in the .jou files when the mesh is generated by Cubit.

bwspenc (changes requested):

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
Kavan-Khaledi: you are right! i removed parameters "factor" and "post_processor", and only kept the "function" as a required parameter

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
Kavan-Khaledi: Done!

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
Kavan-Khaledi: I added the check for it, thanks for the hint

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
Kavan-Khaledi: This Kernel can be used with both triangular and quadrilateral shell elements, as it only uses nodes 0, 1, and 2 to compute the normal vector. Therefore, it's better not to throw a mooseError here, since we may support triangular shell elements in the future. Currently, if the user applies this Kernel to any shell element other than a Q4, the ADComputeIncrementalSmallShell will return an error anyway.

modules/solid_mechanics/test/tests/shell/static/clamped_plate_flat.i
Kavan-Khaledi: yes, they are very small files (1KB) in .geo format, i added them to test folder

modules/solid_mechanics/include/kernels/ADDistributedLoadShell.h
```diff
@@ -0,0 +1,80 @@
...
+  std::vector<const Node *> _nodes;
+
+  /// An auxiliary in-plane vector used to calculate the normal vector to the shell
+  RealVectorValue _v1;
+  /// An auxiliary in-plane vector used to calculate the normal vector to the shell
+  RealVectorValue _v2;
+  /// Normal vector to the shell plane calcualted from cross product of _v1 and _v2
+  RealVectorValue _normal;
+
+  const bool _use_displaced_mesh;
```
bwspenc, line 60: This doesn't appear to be used, so it can be deleted.

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
```diff
@@ -0,0 +1,95 @@
...
+ADDistributedLoadShellTempl<is_ad>::ADDistributedLoadShellTempl(const InputParameters & parameters)
+  : GenericKernel<is_ad>(parameters),
+    _ndisp(this->coupledComponents("displacements")),
+    _component(libMesh::invalid_uint),
+    _function(&this->getFunction("function")),
+    _project_load_to_normal(parameters.isParamSetByUser("project_load_to_normal")
+                                ? this->template getParam<bool>("project_load_to_normal")
+                                : false),
+    _nodes(4),
+    _use_displaced_mesh(this->template getParam<bool>("use_displaced_mesh"))
```
bwspenc, line 50: Delete this line since `_use_displaced_mesh` isn't used.

modules/solid_mechanics/src/kernels/ADDistributedLoadShell.C
bwspenc: I see. This does loop through the 4th node, though (i=3), which would be invalid for a triangular element. You can do `_current_elem->n_nodes()` to get the number of nodes, so that loop could be rewritten as:
```
for (unsigned int i = 0; i < _current_elem->n_nodes(); ++i)
```

However, it might be better to just get rid of the local `_nodes` and do something like this below:
```
  _v1 = (*(_current_elem->node_ptr(1)) - *(_current_elem->node_ptr(0)));
  _v2 = (*(_current_elem->node_ptr(2)) - *(_current_elem->node_ptr(0)));
```

It's probably a good idea to put in this assert too. It's probably not really likely that someone would run this on an element with <3 nodes, but this would at least catch that if run in debug mode:
```
mooseAssert(_current_elem->n_nodes() > 2);

File: modules/solid_mechanics/test/tests/shell/static/tests [hunk 11]
```diff
@@ -137,4 +137,28 @@
     exodiff = 'plate_cantilever_out.e'
     requirement = "Local stress and bending moment are calculated correctly on the shell element"
   []
+  [tank_shell]
+    type = 'Exodiff'
+    input = tank_shell.i
+    exodiff = 'tank_shell_out.e'
+    requirement = "pressure loads are applied correctly on the shell element with curved geometry"
```
bwspenc, line 144:
```suggestion
    requirement = "The mechanics system shall correctly apply pressure loads to a curved geometry represented by shell elements"
```

modules/solid_mechanics/test/tests/shell/static/tests
```diff
@@ -137,4 +137,28 @@
...
+  []
+  [tank_shell_rotated]
+    type = 'Exodiff'
+    input = tank_shell_rotated.i
+    exodiff = 'tank_shell_rotated_out.e'
+    requirement = "pressure loads are applied correctly on the shell element with curved geometry"
```
bwspenc, line 150:
```suggestion
    requirement = "The mechanics system shall correctly apply pressure loads to a curved geometry represented by shell elements, in a rotated configuration."
```

modules/solid_mechanics/test/tests/shell/static/tests
```diff
@@ -137,4 +137,28 @@
...
+  []
+  [scordelis_lo_roof_shell]
+    type = 'Exodiff'
+    input = scordelis_lo_roof_shell.i
+    exodiff = 'scordelis_lo_roof_shell_out.e'
+    requirement = "benchmark test for distributed load on shell"
```
bwspenc, line 156:
```suggestion
      requirement = "The mechanics system shall correctly model a benchmark problem consisting of a roof-like geometry with a distributed load, modeled with shell elements"
```

modules/solid_mechanics/test/tests/shell/static/tests
```diff
@@ -137,4 +137,28 @@
...
+  []
+  [clamped_plate_flat]
+    type = 'Exodiff'
+    input = clamped_plate_flat.i
+    exodiff = 'clamped_plate_flat_out.e'
+    requirement = "benchmark test for distributed load on shell"
```
bwspenc, line 162:
```suggestion
    requirement = "The mechanics system shall correctly model a benchmark problem consisting of a flat clamped plate with a distributed load, modeled with shell elements."
```

bwspenc (changes requested):
Almost there! A few more minor suggested changes.

modules/solid_mechanics/include/kernels/ADDistributedLoadShell.h
dschwen: No, just the files you're adding!

bwspenc (approved):

PR #28979, opened 2024-10-31
Use hash table matrices when requested

lindsayad:
This is for problems for which the sparsity pattern is difficult to compute or changing from nonlinear iteration to nonlinear iteration or from timestep to timestep. Refs [petsc/petsc#7939](https://gitlab.com/petsc/petsc/-/merge_requests/7939) and https://github.com/libMesh/libmesh/pull/3983

PR #30177, opened 2025-03-24
[WIP] Debug use hash table

lindsayad:
This PR should never be merged but it's helpful for doing numerical comparisons between matrices that are preallocated vs hash table assembled?


lindsayad:
Consolidating into #28979

Issue #30178, opened 2025-03-24
Refactor Pressure classes and Pressure Action


dschwen:
### Motivation

We have at least two pressure BCs (`Pressure` and `CoupledPressure`) in solid mechanics. Those should have a common base class and we should have actions for both of them.

### Design

Add a common base class that unifies the `component` parameter logic (this parameter is actually deprecated in `Pressure` and is inferred from `displacements` and `variable`). Next we template the Pressure action on the BC class to get the right parameters and create the correct object. 

### Impact

These changes will make adding new actions for derived pressure BCs simpler.

PR #27110, opened 2024-03-15
Generate warning if displaced mesh exists with strain=small

bwspenc:
ref #27079

PR #30181, opened 2025-03-24
Pressure BC/Action refactor

dschwen:
This also has the advantage of making the non-AD Jacobian developed by @jasondhales available in `CouplePressureBC`.

Refs #30178


dschwen:
I'm aware of the test failures and am working on debugging those.

dschwen:
For some reason there are Bison tests that require the action parameter `use_displaced_mesh = false`. Any idea why that could be @gambka ?

gambka:
> For some reason there are Bison tests that require the action parameter `use_displaced_mesh = false`. Any idea why that could be @gambka ?

If I recall correctly it was related to changes to address incorrect use of displaced meshes in simulations with small strain and Pressure BCs. 

Setting `strain = SMALL` in the `Physics/SolidMechanics/QuasiStatic` action would run the strain and stress calculators on the undisplaced meshbut if the BCs, particularly Pressure BCs, were executed on the displaced mesh incorrect comparisons to analytical solutions would be observed. 

One way correct the behavior is to set the `use_displaced_mesh = false` in the Mesh block so the displaced mesh isn't created at all, including for the PressureBC.  However, some analytical solutions when using small strain formulation compare for example to the volume (e.g., a volume postprocessorr), which requires the displaced mesh to compute a new value even in small strain formulations.  In these scenarios, the fix is to set `use_displaced_mesh = false` in the Pressure BC Action itself.

@bwspenc had this PR open for a while to at least warn the users when a displaced mesh was being used with small strain (https://github.com/idaholab/moose/pull/27110).

jessecarterMOOSE:
How can I implement my own pressure class? I assume inherit from PressureBase, but how to get the Action to use it?

jessecarterMOOSE:
@dschwen do i need to write my own Action or can I somehow use the existing PressureAction to call my own PressureBase-derived class?

dschwen:
Jesse, I will rework this a bit to make it easier to create a custom action. Hang on....

dschwen:
Ok, Jesse, the latest version of the PR adds a simple macro to add a custom action.

dschwen:
I have changes ready for Bison.

modules/solid_mechanics/src/bcs/Pressure.C
```diff
@@ -17,96 +17,56 @@
 registerMooseObject("SolidMechanicsApp", Pressure);
 registerMooseObject("SolidMechanicsApp", ADPressure);
 
+registerMoosePressureAction("SolidMechanicsApp", Pressure, PressureAction);
```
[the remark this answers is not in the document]
jessecarterMOOSE: So if I were to do something like `registerMoosePressureAction("MyApp", MyPressure, MyPressureAction);`, what would the corresponding input file syntax be? I'm a bit confused because the Pressure Action uses the `[Pressure]` syntax, but that is also the name of the class that gets called.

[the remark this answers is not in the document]
jessecarterMOOSE: Oh maybe I still need to register my Action syntax

jessecarterMOOSE:
@dschwen this is working well for me

jessecarterMOOSE:
ping @dschwen . I'd still like to use this. Are you just waiting on reviewers?

dschwen:
Oh, I'll get back to this this week!

jessecarterMOOSE:
@dschwen I'd like to put this to use soon!

dschwen:
Bison PR is merged. Waiting for it to make it into the master branch. Then invalidate and (after a review) merge.

modules/solid_mechanics/include/actions/PressureAction.h
```diff
@@ -11,19 +11,104 @@
...
+InputParameters
+PressureActionBase::validParams()
+{
+  InputParameters params = Action::validParams();
+  params += T::actionParams();
+
+  params.addParam<std::vector<BoundaryName>>(
+      "boundary", "The list of boundaries (ids or names) from the mesh where this object applies");
+
+  params.addClassDescription("Set up Pressure boundary conditions");
```
loganharbour, line 80: Shouldn't this be unique?

loganharbour (commented):
Looks good. Minor comment.

Was also going to ask about needing to change tests - looks like they're failing; makes sense

modules/solid_mechanics/include/actions/PressureAction.h
dschwen: Yeah, I could incorporate the pressure bc class name here.

loganharbour:
Will probably still be failing a test, no? Or did you fix a bug? Hard to tell with the squashed commit history...

dschwen:
> Will probably still be failing a test, no? Or did you fix a bug? Hard to tell with the squashed commit history...

Should be fixed.

dschwen:
Failure seems unrelated but persistent.

loganharbour:
> Failure seems unrelated but persistent.

I'll look into this today.

loganharbour (approved):

modules/solid_mechanics/include/actions/PressureAction.h
dschwen: done

jessecarterMOOSE:
🎉

Issue #29579, opened 2024-12-19
Periodic NEML2 submodule update


hugary1995:
### Motivation

Let's (re)use this issue as the persistent ticket for periodically syncing the NEML2 submodule.

### Design

"ref" this issue, not "close" it.

### Impact

Less hassle before creating a PR. Also, this issue can serve as the one-stop-shop for keeping track of all NEML2 submodule updates.

PR #29765, opened 2025-01-30
Pin Python 3.12 and update compiler stack for Sequoia 15.4, update libMesh

milljm:
Initial commit attempting to support Python 3.12. This also updates the conda compiler stack (clang 16 --> clang 18 on MacOS) in order to support Sequoia 15.4.

It was found a libMesh update is also necessary to solve recent Sequoia 15.4.x update.

Closes #29764
Closes #30227
Closes #30349
Closes #29830

PR #30190, opened 2025-03-25
Update NEML2, add NEML2 build script, add NEML2 to moose-dev container

hugary1995:
This update mainly does two things:
- Switch neml2 from gnu-make to update_and_rebuild_neml2.sh
- Add support for work dispatchers

The switch to update_and_rebuild_neml2.sh is needed because
- NEML2 could be bringing in some additional dependencies in the future
- NEML2 may generate files at configure time
- This makes it easier to build binaries and ship NEML2 in containers

While the same can be managed by the MOOSE make-based build system, it is easier to maintain and manage if we just rely on NEML2's own CMake scripts, in the long run.

List of changes
- Added scripts/update_and_rebuild_neml2.sh
- Update neml2.mk
- Update neml2 submodule
- Add support for dispatchers
- Add documentation and tests for dispatchers
- Handle default dtype; we no longer manipulate torch default dtype outside NEML2 region
- Fix --parse-neml2-only cli arg

ref #29579


modules/solid_mechanics/test/include/base/SolidMechanicsTestApp.h
```diff
@@ -9,9 +9,9 @@
 
 #pragma once
 
-#include "MooseApp.h"
+#include "SolidMechanicsApp.h"
 
-class SolidMechanicsTestApp : public MooseApp
+class SolidMechanicsTestApp : public SolidMechanicsApp
```
hugary1995, line 14: Someone tell me I'm not crazy plz.

hugary1995:
@loganharbour This is one of the things we discussed about. I believe we need to make some changes to civet recipe to call `update_and_rebuild_neml2.sh`.

modules/solid_mechanics/include/neml2/interfaces/NEML2ModelInterface.h
```diff
@@ -74,6 +95,14 @@ NEML2ModelInterface<T>::validParams()
       ":<device-index> optionally specifies a device index. For example, device='cpu' sets the "
       "target compute device to be CPU, and device='cuda:1' sets the target compute device to be "
       "CUDA with device ID 1.");
+
+  params.addParam<std::string>(
```
loganharbour, line 99: MooseEnum here for possible schedulers?

modules/solid_mechanics/include/neml2/interfaces/NEML2ModelInterface.h
```diff
@@ -46,12 +52,27 @@ class NEML2ModelInterface : public T
...
 private:
+  /// The device on which to evaluate the NEML2 model
+  const torch::Device _device;
   /// The NEML2 material model
   neml2::Model & _model;
 
-  /// The device on which to evaluate the NEML2 model
-  const torch::Device _device;
+  /// The work scheduler to use
+  std::shared_ptr<neml2::WorkScheduler> _scheduler;
```
loganharbour, line 71: Does this need to be a smart ptr? Isn't just the pointer ok?

If this remains a shared ptr:
```
const std::shared_ptr<neml2::WorkScheduler>
```
otherwise
```
neml2::WorkScheduler * const
```

modules/solid_mechanics/test/include/base/SolidMechanicsTestApp.h
loganharbour: Yep, this is fine

scripts/update_and_rebuild_neml2.sh
```diff
@@ -0,0 +1,243 @@
...
+  echo
+  echo "NEML2 requires TIMPI, WASP, and libtorch as dependencies. They can be "
+  echo "obtained using the following scripts:"
+  echo "  - TIMPI:    scripts/update_and_rebuild_libmesh.sh"
+  echo "  - WASP:     scripts/update_and_rebuild_wasp.sh"
+  echo "  - libtorch: scripts/setup_libtorch.sh"
+  echo
+  echo "Influential environment variables:"
+  echo "  MOOSE_DIR       The path to the MOOSE directory. Default to the parent directory of this script."
+  echo "  LIBMESH_SRC_DIR The path to the libMesh source directory. Default to <MOOSE_DIR>/libmesh."
```
loganharbour, line 36: Why is this and WASP_SRC_DIR needed? Can you not just use the installed libraries?

loganharbour (commented):
First run through. Mostly minor. I can create a step for the cases that use neml2 to run update_and_rebuild_neml2.sh for just this PR

modules/solid_mechanics/include/neml2/interfaces/NEML2ModelInterface.h
hugary1995: This is the name of the scheduler name defined in the neml2 input file. It is reference an object with user-defined options, so it cannot be an enum.

scripts/update_and_rebuild_neml2.sh
hugary1995: This is handling the case of building libraries from source. So for example the installation directory will be different when using update_and_rebuild_libmesh.sh vs e.g. conda.

So for example, I do the following to find timpi (needed by neml2)
```
LIBMESH_SRC_DIR=${LIBMESH_SRC_DIR:-${MOOSE_DIR}/libmesh}
LIBMESH_DIR=${LIBMESH_DIR:-${LIBMESH_SRC_DIR}/installed}
TIMPI_DIR=${TIMPI_DIR:-${LIBMESH_DIR}}
```
The first two lines just help locate timpi, which is inside the installed libmesh.

hugary1995: Maybe I misunderstood, do you mean there's a way for me to directly figure out the installation directory of libmesh without going through all these indirections?

reverendbedford (commented):
Can we get GPU nodes on civet?  If so we could do an "actual" `StaticHybridScheduler` test with a gpu and the cpu.

loganharbour:
> Can we get GPU nodes on civet? If so we could do an "actual" `StaticHybridScheduler` test with a gpu and the cpu.

Yep, we can.

loganharbour (changes requested):
We need to update documentation to include the neml2 build script call (if necessary)

@hugary1995 look at the coverage diff: https://mooseframework.inl.gov/docs/PRs/30190/coverage_diff/solid_mechanics to see if we're missing anything

hugary1995:
I've already updated the relevant documentation in the solid mechanics module to talk about the build script. I can move the instructions up if we see adoption of neml2 in other physics modules in the future.

I reviewed the coverage diffs, they look correct to me. One thing I noticed is that line 337 in file `modules/solid_mechanics/src/neml2/userobjects/NEML2ModelExecutor.C` is missing. This part is not tested in the solid mechanics tests, but it is tested in the combined module in `modules/combined/test/tests/optimization/invOpt_elasticity_modular`. I suppose this line will be shown as "covered" if we include coverage data from the combined module.

hugary1995:
I force pushed to resolve a conflict resulted from lynn's PR.

hugary1995:
@lynnmunday  FYI, I updated your LAROMANCE6DInterpolation class to adapt it to the new neml2 submodule. The changes are mostly superficial. You can review my changes here https://github.com/idaholab/moose/pull/30190/commits/c5b7b29cbcd2d014772a1704b6f02911e815c683

hugary1995:
@loganharbour my last commit should have fixed the only remaining test failure resulted from a bad rebase. If you can do another (hopefully last!) versioner fix for me, this should be ready to go.

loganharbour:
https://github.com/idaholab/moose/pull/29765 is a higher priority for package updates. Will probably wait for it to get in before getting this in (which will require another bump).

hugary1995:
Sure, we are running out of macos upgrade delay here as well.

hugary1995:
So green, finally...

loganharbour:
@dschwen mind reviewing?

modules/solid_mechanics/src/neml2/models/LAROMANCE6DInterpolation.C
```diff
@@ -350,9 +364,8 @@ LAROMANCE6DInterpolation::json_vector_to_torch(const std::string & key) const
   if (!_json.contains(key))
     throw NEMLException("The key '" + std::string(key) + "' is missing from the JSON data file.");
 
-  std::vector<Real> in_data = _json[key].template get<std::vector<Real>>();
-  const int64_t data_dim = in_data.size();
-  return Scalar(torch::from_blob(in_data.data(), {data_dim}).clone());
+  std::vector<Real> in_data = _json[key].get<std::vector<Real>>();
+  return Scalar::create(in_data).clone();
```
lynnmunday, line 368: I didn't know about the libtorch create function, I didn't know there was another torch function for converting an array.  Thanks for fixing all of these.

modules/solid_mechanics/include/neml2/utils/NEML2Utils.h
```diff
@@ -99,42 +113,41 @@ fromBlob(const std::vector<T> & data)
...
 
 /**
- * @brief Directly copy a contiguous chunk of memory of a torch::Tensor to a MOOSE data of type T
+ * @brief Directly copy a contiguous chunk of memory of a at::Tensor to a MOOSE data of type T
  *
- * This assumes the torch::Tensor and T have the same layout, for example both row-major
+ * This assumes the at::Tensor and T have the same layout, for example both row-major
  * with T = RankTwoTensor. If the layouts are different, we may need to reshape/reorder/transpose
- * the torch::Tensor before copying.
+ * the at::Tensor before copying.
```
dschwen, line 127: I think you briefly mentioned this, but why the switch here? I thought `at::Tensor` is not differentiable, while `torch::Tensor` is. Has this changed?

dschwen: That does _net_ necessarily mean that. If there was a separate declaration of `torch::Tensor` that would hide `at::Tensor` (see the A::f lookup example in  https://en.cppreference.com/w/cpp/language/qualified_lookup#Namespace_members )

hugary1995: Yeah, thanks for pointing that out. I was actually aware of the lookup rule. ATen/core/TensorBody.h is the only declaration of at::Tensor. There's no other declaration under other namespaces that hides it.

modules/solid_mechanics/src/neml2/models/LAROMANCE6DInterpolation.C
hugary1995: Yep @dschwen is correct.

loganharbour:
Will do another rebase once https://github.com/idaholab/moose/pull/30362 merges to devel.

dschwen (dismissed):
Please move the submodule and makefiles into the framework

hugary1995:
@loganharbour I'm sure I did the rebase wrong...

hugary1995:
> Please move the submodule and makefiles into the framework

Done.

scripts/apptainer_generator.py
```diff
@@ -619,6 +619,7 @@ def add_definition_vars(self, jinja_data):
                          'libmesh': [{'package': 'libmesh', 'submodule': 'libmesh'}],
                          'moose-dev': [{'package': 'conduit', 'submodule': 'framework/contrib/conduit'},
                                        {'package': 'mfem', 'submodule': 'framework/contrib/mfem'},
+                                       {'package': 'neml2', 'submodule': 'framework/contrib/neml2'},
```
milljm, line 622: Acknowledging change...

loganharbour:
@dschwen we now have a `DiffusionLimitedReaction.h` in moose and bison. See https://civet.inl.gov/job/2844429/. Can you resolve this?

hugary1995:
The include path was polluted by a recent (good) fix for contrib includes. My last commit should have fixed it.

loganharbour (approved):

PR #30195, opened 2025-03-26
A fix to avoid repartitioning when loading checkpoint files

YaqiWang:
This is a followup PR of PR #30052 addressing #30048. Tag @lindsayad


loganharbour (approved):
Failure related, but is fixed now that the griffin patch is in master.

YaqiWang:
Sorry, I forgot mentioning that this is to fix the failures on next branch introduced by the previous PR.

PR #30198, opened 2025-03-27
Add auxvar and quadrature options to ContactAction

### LINKED ISSUE #30208: Allow users to disable auxiliary variable generation using ContactAction
### Motivation

The contact action is allows for greatly simplified input files for contact problems. However, when using a mortar contact formulation, the auxiliary variables necessary for kinematic and penalty contact problems are still generated. Namely these are the `nodal_area` `penetration` and  `contact_pressure` variables. 

From what I can see these variables are not block restricted and can utilise a large amount of memory when the mesh is large. Being able to disable the automatic creation of these variables could allow for larger contact problems to be solved, especially where memory is a limiting factor.

### Design

I would like to see an option within the ContactAction that allows the user to disable the automatic creatopm of these auxiliary variables when a mortar contact formulation is selected.

### Impact

This could allow for larger contact problems to be run, especially where memory is a limiting factor.

TheBEllis:
Added two more input parameters to the ContactAction
- quadrature, which allows the user to change the quadrature order used with Mortar formulations
- add_aux_var, which allows the user to prevent the generation of `contact_pressure`, `penetration`, `nodal_area` and  `mortar_tangent` aux variables when using mortar

## Reason

- Being able to change the quadrature order from the ContactAction is a nice quality of life improvement.
- For large mortar problems, being able to turn off the auxiliary variables required for penalty and kinematic formulations can save a substantial amount of memory. 

## Design

- I have wrapped the generation of the aux variables in an `if` statement that will only be skipped if the formulation is set to `mortar` and add_aux_vars is set to `false`. 
- By default add_aux_vars is set to true, and if the user tries to use it with a formulation that isn't `mortar` they will receive an error.
- The quadrature parameter is passed on to the Mortar constraints that can consume it.
- If the user tries to set quadrature with a formulation that isn't mortar they will receive an error. 

## Impact

- Could help enable larger mortar formulation solves through reduced memory footprint.

Edit: 
Closes #30208.


milljm:
You'll need to adjust the base branch you are attempting to merge into (change to `next` instead of `devel`)

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

milljm:
Looks like [Precheck](https://civet.inl.gov/job/2783064/) failed, for the following:

```pre
##########################################################################
INFO: Your patch contains no trailing whitespace.
INFO: Your patch contains no proprietary or classified keywords.
INFO: Your patch contains no files without newlines before EOF.
INFO: Your patch contains no bad executable files.
INFO: Your patch contains no banned functions.
INFO: Your patch contains no banned keywords.
INFO: Style check disabled
INFO: Your patch contains no unicode characters.
INFO: Your patch contains no old style C++ include guards.
INFO: Your patch contains windows line endings.
##########################################################################


##########################################################################
ERROR: Your patch does not contain a valid ticket reference! (i.e. #1234)
Merge branch 'ContactActionQuadrature' of https://github.com/TheBEllis/moose into test
Added add_aux_vars to initialiser list
Add auxvar and quadrature options to ContactAction

ERROR: MOOSE prefers two spaces instead of tabs. The following files contain tab characters:
	modules/contact/src/actions/ContactAction.C
##########################################################################
```

Seems you'll need to create an Issue (like a feature request), and then amend your commit with that reference ticket.

TheBEllis:
I've created and linked a relevant issue now, hopefully the checks will run fine!

milljm:
You just need to include that in your commit message :) Civet is a real stickler about that:

```bash
git commit --ammend

"""
This adds the following features, etc etc

Closes #30208
"""
```

TheBEllis:
> You just need to include that in your commit message :) Civet is a real stickler about that:
I re-read your first comment and realised I had been foolish:)

milljm:
With Precheck passing, I'll hand this off to one of our developers for approval to run on our clusters.

Cheers!

modules/contact/src/actions/ContactAction.C
```diff
@@ -383,6 +400,14 @@ ContactAction::ContactAction(const InputParameters & params)
...
                  "The 'mortar_dynamics' constraint option can only be used with the 'mortar' "
                  "formulation and in dynamic simulations using Newmark-beta");
+    else if (params.isParamSetByUser("quadrature"))
+      paramError("quadrature",
+                 "The 'quadrature' option can only be used with the 'mortar' formulation.");
+    else if (params.isParamSetByUser("add_aux_vars"))
+      paramError("add_aux_vars",
+                 "The 'add_aux_vars' option can only be used with the 'mortar' formulation. "
+                 "nodal_area, penetration and contact_pressure variables are not optional with "
+                 "kinematic and penalty formulations.x");
```
GiudGiud, line 410:
```suggestion
                 "kinematic and penalty formulations.");
```

modules/contact/src/actions/ContactAction.C
```diff
@@ -273,6 +273,22 @@ ContactAction::validParams()
       "Whether to use the Petrov-Galerkin approach for the mortar-based constraints. If set to "
       "true, we use the standard basis as the test function and dual basis as "
       "the shape function for the interpolation of the Lagrange multiplier variable.");
+
+  params.addParam<MooseEnum>(
+      "quadrature",
+      MooseEnum("DEFAULT FIRST SECOND THIRD FOURTH", "DEFAULT"),
+      "Quadrature rule to use on mortar segments. For 2D mortar DEFAULT is recommended."
```
GiudGiud, line 280: this is the order rather than the rule

let's rename to quadrature_order

modules/contact/src/actions/ContactAction.C
```diff
@@ -273,6 +273,22 @@ ContactAction::validParams()
...
+      "For 3D mortar, QUAD meshes are integrated using triangle mortar segments. "
+      "While DEFAULT quadrature order is typically sufficiently accurate, "
+      "exact integration of QUAD mortar faces requires SECOND order quadrature for FIRST variables "
+      "and FOURTH order quadrature for SECOND order variables.");
+
+  params.addParam<bool>("add_aux_vars",
```
GiudGiud, line 286: I would rename to generate_output and let the developers handle the minimum / maximum levels of outputs each formulation can handle

modules/contact/src/actions/ContactAction.C
```diff
@@ -273,6 +273,22 @@ ContactAction::validParams()
...
+                        true,
+                        "For Mortar contact formulation problems, does the user still wish to "
+                        "generate aux-variables necessary for penalty and kinematic formulations."
```
GiudGiud, line 289:
```suggestion
                        "generate auxiliary variables necessary for penalty and kinematic formulations."
```

modules/contact/src/actions/ContactAction.C
```diff
@@ -273,6 +273,22 @@ ContactAction::validParams()
...
+                        "Turning this to false can be useful if the user is running out of memory "
+                        "on mortar contact solves using a large mesh");
```
GiudGiud, line 291:
```suggestion
);
```
could be other reasons and someone running out of memory will not find this advice anyway

modules/contact/src/actions/ContactAction.C
```diff
@@ -469,165 +494,172 @@ ContactAction::act()
...
-
-  if (_current_task == "add_aux_kernel")
-  { // Add ContactPenetrationAuxAction.
-    if (!_problem->getDisplacedProblem())
-      mooseError("Contact requires updated coordinates.  Use the 'displacements = ...' line in the "
-                 "Mesh block.");
-    // Create auxiliary kernels for each contact pairs
-    for (const auto & contact_pair : _boundary_pairs)
-    {
+  if (!(_formulation == ContactFormulation::MORTAR && !_add_aux_vars))
```
GiudGiud, line 497:
```suggestion
  if (_add_aux_vars || _formulation != ContactFormulation::MORTAR)
```
easier to parse

GiudGiud (changes requested):
@maxnezdyur would you want to review?

modules/contact/src/actions/ContactAction.C
GiudGiud: generate_output is also the name used in the main SM action

maxnezdyur:
@TheBEllis 
Can you add a few more options for mortar contact? The action can't set some options that can be useful.

1. `minimum_projection_angle`  
2. `debug_mesh`

GiudGiud:
Just a note that adding new options is not mandatory.

But now that the additional need has been expressed, let's at least make sure the PR does not preclude meeting it

modules/contact/src/actions/ContactAction.C
```diff
@@ -383,6 +400,14 @@ ContactAction::ContactAction(const InputParameters & params)
       paramError("mortar_dynamics",
                  "The 'mortar_dynamics' constraint option can only be used with the 'mortar' "
                  "formulation and in dynamic simulations using Newmark-beta");
+    else if (params.isParamSetByUser("quadrature"))
+      paramError("quadrature",
+                 "The 'quadrature' option can only be used with the 'mortar' formulation.");
```
maxnezdyur, line 405: To prevent confusion, the quadrature option can also be used with `mortar_penalty`. I would call that out.

maxnezdyur (changes requested):
This adds a QoL improvement that I will be using frequently! Just had a single minor comment.

lindsayad (changes requested):

jbadger95:
Hi Alex, triangle quadratures exactly integrate polynomials with total
order; e.g. x^2 + xy + y^2 + LOT. Quad quadrature rules exactly integrate
polynomials with tenor order; e.g. x^2y^2 + LOT. The quadrature is
evaluating inner products so the 2n+1 is essentially split between the two
functions being integrated. Maybe it’s possible to devise nonstandard
quadrature schemes for triangles and in any case you can often get away
with less since the mortar segment mesh is finer but I think the comment is
correct as written; exact integration with standard quadratures do require
that. Anyway, that’s the motivation on the comment, you probably have more
insight into practice, etc.

lindsayad:
That's clear. Thanks @jbadger95 for your quick reply! Hope you're doing well 😄

modules/contact/src/actions/ContactAction.C
```diff
@@ -273,6 +273,22 @@ ContactAction::validParams()
...
       "the shape function for the interpolation of the Lagrange multiplier variable.");
+
+  params.addParam<MooseEnum>(
+      "quadrature",
+      MooseEnum("DEFAULT FIRST SECOND THIRD FOURTH", "DEFAULT"),
+      "Quadrature rule to use on mortar segments. For 2D mortar DEFAULT is recommended."
+      "For 3D mortar, QUAD meshes are integrated using triangle mortar segments. "
+      "While DEFAULT quadrature order is typically sufficiently accurate, "
+      "exact integration of QUAD mortar faces requires SECOND order quadrature for FIRST variables "
+      "and FOURTH order quadrature for SECOND order variables.");
```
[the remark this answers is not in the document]
roystgnr: Define "decomposed into two tris"?  We're using a composition of two quadrature rules on triangles to integrate biquadratic functions?  The function space is no biquadratic, we've split the element and now it's piecewise quadratic?

Anyway: you can't *exactly* integrate biquadratic polynomial bases on a general quad face with *any* quadrature order: you're wanting to integrate φᵢφⱼ, which is biquartic, so you'd think `FOURTH` would do it, but when you pull it back to the master element where your quadrature rule was defined you tack on a |∂x/∂ξ| term that isn't polynomial at all.  So we content ourselves with integrating, not exactly, but closely enough that the error term is swamped by our numerical approximation error.  For general nonlinear PDEs 2p+1 is usually safe (though I don't recall the conditions guaranteeing that), and with no nonlinearity in the physical-space integral (and an invertible x(ξ); we're hoping to deal with distorted mappings, not singular ones) 2p should be fine.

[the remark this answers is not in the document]
lindsayad: Decomposed into two tris was very inaccurate. We decompose a given face element into `n_sub_elem()`, then we linearize it. We do this for both secondary and primary faces, and then we perform projections with these linearized sub-elements and then triangulate the result. @jbadger95 wrote up some nice documentation (with pictures!) [here](https://mooseframework.inl.gov/source/constraints/AutomaticMortarGeneration.html).

So we apply a quadrature rule on the TRI3 elements which is of order 2*n + 1, where n is the order that the user passes to this parameter, e.g. the user parameter n corresponds to the order of the FE basis, *not* the quadrature.

We take the quadrature points on these TRI3 elements, map them back onto the possibly non-planar sub-elements, and then finally back onto the original possibly quadratic face elements. We then evaluate our finite element bases and variables with the mapped face element quadrature. Is that clear?

So this is why @jbadger95 is saying that you need to square the FE basis order ($n \mapsto n^2$) when your face elements are quads because you will ultimately be mapping a quadrature on TRI3 (linear) back to these quads (bilinear, biquadratic, etc.)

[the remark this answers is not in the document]
lindsayad: Once we're all on the same page, I'm probably going to add some more documentation to the repo capturing this conversation

GiudGiud:
Any update here?

lindsayad:
we could probably finish this if @TheBEllis would prefer us to

GiudGiud:
Yeah we should. @lindsayad want to take it on and I review or the opposite

lindsayad:
Half this PR (the aux var piece) will no longer be relevant after https://github.com/idaholab/moose/pull/31268 which among several other things updates the `ContactAction` to only add auxiliary quantities that the mortar system uses. Instead of outputting penetration it will output the gap which is computed by a mortar aux kernel. Contact pressure will still be output but it will actually be nonzero now because it will use the mortar aux kernel. Nodal area will no longer be output. I think I'll wait until that PR is merged and then update this PR to only be about the quadrature option

lindsayad:
I should update this now that #31268 is merged

lindsayad:
closing in favor of https://github.com/idaholab/moose/pull/32103 for automatically triggered testing

PR #30223, opened 2025-03-30
Treat boundary unrestricted like block unrestricted

### LINKED ISSUE #30221: Need a simple way to apply BCs to _all_ boundaries
### Motivation

Looking at the code for `BoundaryRestrictable` one might get the impression that an empty boundary list, or the special boundary name `ANY_BOUNDARY_ID` would apply the BC simply to all boundaries in the simulation. However that's not the case.

```
[Mesh]
  [gen]
    type = GeneratedMeshGenerator
    dim = 2
    nx = 10
    ny = 10
  []
[]

[Variables]
  [u]
  []
[]

[Kernels]
  [dt]
    type = TimeDerivative
    variable = u
  []
  [diff]
    type = Diffusion
    variable = u
  []
  [body_force]
    type = BodyForce
    variable = u
    function = 1.0
  []
[]

[BCs]
  [zero]
    type = DirichletBC
    variable = u
    value = 0.0
    # boundary = ''
    boundary = 'ANY_BOUNDARY_ID' # This will match any boundary, including node sets and side sets
    # boundary = 'left right top bottom'
  []
[]

[Postprocessors]
  [int]
    type = ElementIntegralVariablePostprocessor
    variable = u
  []
[]

[Executioner]
  type = Transient
  num_steps = 10
  dt = 0.1
[]

[Outputs]
  exodus = true
[]
```

### Design

Let's discuss what's in the code and whether we want to make the half implemented (?) behavior more consistent. Why - for example - wouldn't we make the boundary parameter optional, just like the block parameter, where not specifying it just means "no restriction"...?

### Impact

More consistent application behavior. @jessecarterMOOSE

dschwen:
Closes #30221

Ultimately we might want to store (block|boundary)-unrestricted objects in a separate list rather than copying them into the map entry for each boundary. This seems like unnecessary work and slows map lookups down in cases of many blocks/boundaries.

### DIFF
--- a/framework/include/warehouses/MooseObjectWarehouseBase.h
+++ b/framework/include/warehouses/MooseObjectWarehouseBase.h
@@ -291,9 +291,10 @@ MooseObjectWarehouseBase<T>::addObject(std::shared_ptr<T> object,
   std::shared_ptr<BlockRestrictable> blk = std::dynamic_pointer_cast<BlockRestrictable>(object);
 
   // Boundary Restricted
-  if (bnd && bnd->boundaryRestricted())
+  if (bnd)
   {
-    const std::set<BoundaryID> & ids = bnd->boundaryIDs();
+    const std::set<BoundaryID> & ids =
+        bnd->boundaryRestricted() ? bnd->boundaryIDs() : bnd->meshBoundaryIDs();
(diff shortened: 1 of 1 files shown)


dschwen:
I should add that I have zero hope of this change just passing tests.

dschwen:
I did not, however, expect this level of carnage...

PR #30234, opened 2025-04-02
AD and Eigensolution 

### LINKED ISSUE #30233: EigenSolution decomposition with AD
### Bug Description

When using the ` decomposition_method = EigenSolution ` with AD, problems can occur when the incremental deformation gradient is equal to the identity. This can occur every first nonlinear iteration.

### Steps to Reproduce

Run Below
```
[GlobalParams]
[]

[Mesh]
  [gen]
    type = GeneratedMeshGenerator
    dim = 3
    xmax = 1.0
    ymax = 0.5
    zmax = 0.5
    nx = 2
    ny = 1
    nz = 1

    boundary_name_prefix = flexible
  []
[]

[Modules/TensorMechanics/Master]
  [all]
    strain = FINITE
    decomposition_method = EigenSolution 
    add_variables = true
    use_automatic_differentiation = true
  []
[]

[Materials]
  [elastic_tensor]
    type = ADComputeIsotropicElasticityTensor
    youngs_modulus = 1e+4
    poissons_ratio = 0.4
  []
  [stress]
    type = ADComputeFiniteStrainElasticStress
  []

[]

[BCs]
  [hold_x]
    type = ADDirichletBC
    boundary = 'flexible_bottom'
    variable = disp_x
    value = 0
  []
  [hold_y]
    type = ADDirichletBC
    boundary = 'flexible_bottom'
    variable = disp_y
    value = 0
  []
  [hold_z]
    type = ADDirichletBC
    boundary = 'flexible_bottom'
    variable = disp_z
    value = 0
  []
  [Pressure]
    [push]
      function = 't'
      boundary = 'flexible_left '
      use_automatic_differentiation = true
      use_displaced_mesh = false
    []
  []
[]

[Preconditioning]
  [SMP]
    type = SMP
    full = true
  []
[]

[Executioner]
  type = Transient
  end_time = 2
  dt = 1
  solve_type = 'NEWTON'
  petsc_options = '-snes_converged_reason -ksp_converged_reason '
  petsc_options_value = 'lu'
  # automatic_scaling = true
  line_search = 'none'
  nl_abs_tol = 1e-7
  nl_rel_tol = 1e-12
  nl_forced_its = 1

[]

[Outputs]

  exodus = true
  print_linear_residuals = false
[]

```

### Impact

Prevents AD with EigenSolution method.

### [Optional] Diagnostics

_No response_

maxnezdyur:
closes #30233


hugary1995:
This depends on the particular implementation in Eigen. I was expecting corner cases like this when we switched to AD originally. If you ask me for a fix, I think that would be to write your own 3x3 eigen decomposition routine that fixes this particular corner case.

The issue is more fundamental than implementation details, unfortunately. The derivatives of eigen decomposition is truly undefined in several cases. The most notable case is when you have repeated eigenvalues. Some algorithms are unstable with zero rotation.

maxnezdyur:
In general this fix shouldn't change any converged answer? Since the fix only applies to an incremental deformation gradient when it is the identity. I agree with your assessment though, while unlikely we could get other deformations with zero rotations, but if that is only on an element or two then we wouldn't have the catastrophic failure that happens in the input file I attached.

hugary1995:
Isn't Fhat identity if you hold the entire structure in place, i.e. like a rigid body? This 1e-12 shift will basically expand the rigid body overtime, slightly... I guess that's fine, I'm just trying to think of potential consequences.

maxnezdyur:
I don't think there is any problem of this expansion problem. Only problem we will have is if we need to converge an extremely small residual (~machine precision) and hold the entire structure in place. If Fhat + epsilon caused any type of deformation then on the next residual evaluation it would not hit the "if" statement and the next newton would send it back to Fhat = I and repeat.

modules/solid_mechanics/src/materials/ADComputeFiniteStrain.C
```diff
@@ -209,6 +209,11 @@ ADComputeFiniteStrainTempl<R2, R4>::computeQpIncrements(ADR2 & total_strain_incr
 
     case DecompMethod::EigenSolution:
     {
+      // AD stabilization error
+      if (_fe_problem.currentlyComputingJacobian() && _Fhat[_qp] == ADRankTwoTensor::Identity())
```
hugary1995, line 213: Now I have no complaint :)

hugary1995 (approved):

bwspenc (changes requested):

bwspenc (approved):

PR #30238, opened 2025-04-02
fix location where bottom_p_or_t function is evaluated to well bottom

### LINKED ISSUE #30237: Fix PorousFlowPeacemanBorehole bottom_p_or_t function
### Bug Description

This PR (https://github.com/idaholab/moose/pull/20453) introduced a bug into the porous flow peaceman borehole if the bottom_p_or_t was a function of height along the borehole.  The function is being evaluated at the dirac_kernel location instead of at the bottom of the well.  

### Steps to Reproduce

in module/porous_flow/test/test/bh03.i change the function used for bottom_p_or_t to 
``
  [insitu_pp]
    type = ParsedFunction
    expression = '-2e7*z' #bh_bottom is located at z=-0.5
  []
``
which evaluates to 1e7 at the bore hole bottom where the first injection point is located but will evaluate to -1e7 at the second injection point.  This is wrong, it should produce a constant value of 1e7 along the entire borehole.   

### Impact

The peaceman borehole will prematurely start removing fluid from the borehole if the bottom_p_or_t is a function of height.  

This can be fixed by evaluate bottom_p_or_t at the borehole bottom instead of the dirac kernel location.  I have this fixed and a new test that fails with the existing code.

### [Optional] Diagnostics

_No response_

lynnmunday:
close #30237


modules/porous_flow/test/tests/dirackernels/bh03.i
lynnmunday, line 1: I changed everything in this test back to the original xmin, xmax values.  Then I made the bottom_p_or_t a function of z to catch the error.

lynnmunday:
@cpgr or @WilkAndy Could you review and merge?

cpgr (approved):
Thanks @lynnmunday - looks good!

modules/porous_flow/test/tests/dirackernels/bh03.i
cpgr: 👍

lynnmunday:
@GiudGiud Could you review and merge?  Thanks,

GiudGiud (approved):

PR #30247, opened 2025-04-03
Updated SlepscSupport to support dont_add_these_options

pbehne:
This change ensures that simulations that use SlepscSupport do not add
options if they are in `PetscOptions.dont_add_these_options`.

As proof-of-concept, the `EigenValue` executioner was updated such that the tests in `problems/eigen_problem/eigensolvers` no longer emit petsc's unused option warnings.

Ref #28053

### DIFF
--- a/framework/include/utils/SlepcSupport.h
+++ b/framework/include/utils/SlepcSupport.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include "MultiMooseEnum.h"
 #include "libmesh/libmesh_config.h"
 
 #ifdef LIBMESH_HAVE_SLEPC
@@ -75,12 +76,14 @@ void setSlepcEigenSolverTolerances(EigenProblem & eigen_problem,
 /**
  * Set SLEPc/PETSc options to trigger free power iteration
  */
-void setFreeNonlinearPowerIterations(unsigned int free_power_iterations);
+void setFreeNonlinearPowerIterations(unsigned int free_power_iterations,
+                                     const MultiMooseEnum & dont_add_these_options);
 
 /*
  * Set SLEPc/PETSc options to turn the eigen-solver back to a regular Newton solver
  */
-void clearFreeNonlinearPowerIterations(const InputParameters & params);
+void clearFreeNonlinearPowerIterations(const InputParameters & params,
+                                       const MultiMooseEnum & dont_add_these_options);
 
 /**
  * Form matrix according to tag
--- a/framework/include/utils/SlepcSupport.h
+++ b/framework/include/utils/SlepcSupport.h
@@ -30,6 +30,7 @@
 class EigenProblem;
 class InputParameters;
 class SolverParams;
+class MultiMooseEnum;
 
 namespace Moose
 {
@@ -75,13 +76,11 @@ void setSlepcEigenSolverTolerances(EigenProblem & eigen_problem,
 
 /**
  * Set SLEPc/PETSc options to trigger free power iteration
- * @param dont_add_these_options a list of options that won't be set.
  */
 void setFreeNonlinearPowerIterations(unsigned int free_power_iterations);
 
 /*
  * Set SLEPc/PETSc options to turn the eigen-solver back to a regular Newton solver
- * @param dont_add_these_options a list of options that won't be set.
  */
 void clearFreeNonlinearPowerIterations(const InputParameters & params);
 
--- a/framework/include/utils/SlepcSupport.h
+++ b/framework/include/utils/SlepcSupport.h
@@ -75,12 +75,14 @@ void setSlepcEigenSolverTolerances(EigenProblem & eigen_problem,
 
 /**
  * Set SLEPc/PETSc options to trigger free power iteration
+ * @param dont_add_these_options a list of options that won't be set.
  */
 void setFreeNonlinearPowerIterations(unsigned int free_power_iterations,
                                      const MultiMooseEnum & dont_add_these_options);
 
 /*
  * Set SLEPc/PETSc options to turn the eigen-solver back to a regular Newton solver
+ * @param dont_add_these_options a list of options that won't be set.
  */
 void clearFreeNonlinearPowerIterations(const InputParameters & params,
                                        const MultiMooseEnum & dont_add_these_options);
--- a/framework/include/utils/SlepcSupport.h
+++ b/framework/include/utils/SlepcSupport.h
@@ -77,15 +77,13 @@ void setSlepcEigenSolverTolerances(EigenProblem & eigen_problem,
  * Set SLEPc/PETSc options to trigger free power iteration
  * @param dont_add_these_options a list of options that won't be set.
  */
-void setFreeNonlinearPowerIterations(unsigned int free_power_iterations,
-                                     const MultiMooseEnum & dont_add_these_options);
+void setFreeNonlinearPowerIterations(unsigned int free_power_iterations);
 
 /*
  * Set SLEPc/PETSc options to turn the eigen-solver back to a regular Newton solver
  * @param dont_add_these_options a list of options that won't be set.
  */
-void clearFreeNonlinearPowerIterations(const InputParameters & params,
-                                       const MultiMooseEnum & dont_add_these_options);
+void clearFreeNonlinearPowerIterations(const InputParameters & params);
 
 /**
  * Form matrix according to tag
--- a/framework/src/executioners/Eigenvalue.C
+++ b/framework/src/executioners/Eigenvalue.C
@@ -13,6 +13,7 @@
 #include "Factory.h"
 #include "MooseApp.h"
 #include "NonlinearEigenSystem.h"
+#include "PetscSupport.h"
 #include "SlepcSupport.h"
 #include "UserObject.h"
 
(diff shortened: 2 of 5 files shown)


framework/src/executioners/Eigenvalue.C
```diff
@@ -153,6 +154,16 @@ Eigenvalue::Eigenvalue(const InputParameters & parameters)
...
+
+  // To avoid petsc unused option warnings, ensure we do not set irrelevant options.
+  Moose::PetscSupport::dontAddLinearConvergedReason(_fe_problem);
+  Moose::PetscSupport::dontAddNonlinearConvergedReason(_fe_problem);
+  Moose::PetscSupport::dontAddPetscFlag("-mat_mffd_type", _fe_problem.getPetscOptions());
+  Moose::PetscSupport::dontAddPetscFlag("-st_ksp_atol", _fe_problem.getPetscOptions());
+  Moose::PetscSupport::dontAddPetscFlag("-st_ksp_max_it", _fe_problem.getPetscOptions());
+  Moose::PetscSupport::dontAddPetscFlag("-st_ksp_rtol", _fe_problem.getPetscOptions());
+  if (!_eigen_problem.solverParams()._eigen_matrix_free)
+    Moose::PetscSupport::dontAddPetscFlag("-snes_mf_operator", _fe_problem.getPetscOptions());
```
lindsayad, line 166: Now if a user adds these from their input file or the command line, those options will be respected right?

framework/include/utils/SlepcSupport.h
```diff
@@ -75,12 +76,14 @@ void setSlepcEigenSolverTolerances(EigenProblem & eigen_problem,
 /**
  * Set SLEPc/PETSc options to trigger free power iteration
  */
-void setFreeNonlinearPowerIterations(unsigned int free_power_iterations);
+void setFreeNonlinearPowerIterations(unsigned int free_power_iterations,
+                                     const MultiMooseEnum & dont_add_these_options);
 
 /*
  * Set SLEPc/PETSc options to turn the eigen-solver back to a regular Newton solver
```
lindsayad, line 83: Update doxygen

framework/include/utils/SlepcSupport.h
```diff
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include "MultiMooseEnum.h"
```
lindsayad, line 12: can forward declare instead of including

framework/src/utils/SlepcSupport.C
```diff
@@ -131,56 +131,82 @@ setSlepcEigenSolverTolerances(EigenProblem & eigen_problem,
...
     // linear solver
-    Moose::PetscSupport::setSinglePetscOption(solver_params._prefix + "st_ksp_max_it",
-                                              stringify(params.get<unsigned int>("l_max_its")));
-
-    Moose::PetscSupport::setSinglePetscOption(solver_params._prefix + "st_ksp_rtol",
-                                              stringify(params.get<Real>("l_tol")));
-
-    Moose::PetscSupport::setSinglePetscOption(solver_params._prefix + "st_ksp_atol",
-                                              stringify(params.get<Real>("l_abs_tol")));
+    const auto & dont_add_these_options = eigen_problem.getPetscOptions().dont_add_these_options;
```
lindsayad, line 196: Isn't this shadowing the previous `dont_add_these_options`?

framework/src/utils/SlepcSupport.C
```diff
@@ -361,144 +403,185 @@ setWhichEigenPairsOptions(SolverParams & solver_params)
 }
 
 void
-setFreeNonlinearPowerIterations(unsigned int free_power_iterations)
+setFreeNonlinearPowerIterations(unsigned int free_power_iterations,
+                                const MultiMooseEnum & dont_add_these_options)
```
lindsayad, line 407: Do we actually ever call this method where we don't want to set these options that are getting set?

framework/src/utils/SlepcSupport.C
```diff
@@ -361,144 +403,185 @@ setWhichEigenPairsOptions(SolverParams & solver_params)
...
   // we've satisfied a convergence criterion. Consequently we make this tolerance very tight
-  Moose::PetscSupport::setSinglePetscOption("-eps_tol", "1e-50");
+  Moose::PetscSupport::setSinglePetscOptionIfAppropriate(
+      dont_add_these_options, "-eps_tol", "1e-50");
 }
 
 void
-clearFreeNonlinearPowerIterations(const InputParameters & params)
+clearFreeNonlinearPowerIterations(const InputParameters & params,
+                                  const MultiMooseEnum & dont_add_these_options)
```
lindsayad, line 428: same question

framework/src/utils/SlepcSupport.C
```diff
@@ -361,144 +403,185 @@ setWhichEigenPairsOptions(SolverParams & solver_params)
...
+      dont_add_these_options, "-snes_rtol", stringify(params.get<Real>("nl_rel_tol")));
+  Moose::PetscSupport::setSinglePetscOptionIfAppropriate(
+      dont_add_these_options, "-eps_tol", stringify(params.get<Real>("eigen_tol")));
 }
 
 void
-setNewtonPetscOptions(SolverParams & solver_params, const InputParameters & params)
+setNewtonPetscOptions(SolverParams & solver_params,
+                      const InputParameters & params,
+                      const MultiMooseEnum & dont_add_these_options)
```
lindsayad, line 445: same question

framework/src/utils/SlepcSupport.C
```diff
@@ -361,144 +403,185 @@ setWhichEigenPairsOptions(SolverParams & solver_params)
...
+        dont_add_these_options, "-init_st_type", "sinvert");
 #endif
 #else
   mooseError("Newton-based eigenvalue solver requires SLEPc 3.7.3 or higher");
 #endif
 }
 
 void
-setNonlinearPowerOptions(SolverParams & solver_params)
+setNonlinearPowerOptions(SolverParams & solver_params, EigenProblem & eigen_problem)
```
lindsayad, line 503: same question

framework/src/utils/SlepcSupport.C
```diff
@@ -361,144 +403,185 @@ setWhichEigenPairsOptions(SolverParams & solver_params)
...
 #else
   mooseError("Nonlinear Inverse Power requires SLEPc 3.7.3 or higher");
 #endif
 }
 
 void
-setEigenSolverOptions(SolverParams & solver_params, const InputParameters & params)
+setEigenSolverOptions(SolverParams & solver_params,
+                      const InputParameters & params,
+                      EigenProblem & eigen_problem)
```
lindsayad, line 532: same question. Like why would we not set `-eps_type power` when the `solver_params._eigen_solve_type` is `Moose::EST_POWER`?

lindsayad (changes requested):

framework/src/executioners/Eigenvalue.C
pbehne: Correct. That check is done in `PetscSetOptionsHelper`: https://github.com/idaholab/moose/blob/770678ac586b70c2e51bf1ec2a378080c2ce19fb/framework/src/utils/PetscSupport.C#L259

framework/src/utils/SlepcSupport.C
pbehne: Good catch, I'll take it out.

framework/src/utils/SlepcSupport.C
pbehne: I don't know. If you'd like, I can do some testing to figure this out. Let me know.

Even if it is never the case that we don't want to set some of these options, I think that there is value in allowing for such cases in the future. If we didn't allow for this and someone tried adding one of the options to `dont_add_these_options`, then the option would be added when the developer has requested that it not be added. This is actually why this PR came up: I asked for certain options in `Eigenvalue` to be ignored, but they weren't, and I had to hunt down the reason for it.

lindsayad: How did you ask for them to be ignored?

Taking this to the logical extreme, if I have an API that's like `setFoo()`, then instead of adding an argument like `setFoo(bool possibly_dont_set_foo)`, I would propose that if we don't want to `setFoo()` we simply don't call `setFoo()`

lindsayad: So like for setting up free nonlinear power iterations, the options we are setting in that method seem pretty logical to me. I think I'd want to understand why we wouldn't want to set some of those options in there before we add the ability to not set them

pbehne: > How did you ask for them to be ignored?
I didn't ask for anything in `setFreeNonlinerPowerIterations` to be ignored, but I had asked for some options that are handled by `SlepcSupport` to be ignored:
https://github.com/idaholab/moose/pull/30247/files#diff-0ef7fb9fdd0ad222c148814381eac8b2182a9a79020919879a4cd2a1154e0785R157-R166
When I discovered that these requests weren't being honored, I had to trace the issue to `SlepscSupport` not looking at `dont_add_these_options`. So this isn't really related to your original comment, I was just concerned about the same thing happening down the road IF a developer were to ask for a free nonlinear power iteration option to be ignored.

> I would propose that if we don't want to setFoo() we simply don't call `setFoo()`.
I agree that this is the ideal way to do things. This doesn't seem possible given the current structure of MOOSE because all options are handled in `PetscSupport` and `SlepcSupport`. These two namespaces already have branching logic for particular simulation types to determine what options to set. However, the branches are not complex enough to cover all possible cases; hence we still get unused option warnings. These namespaces *could* be updated to handle all possible cases, but I think this would involve a lot of type checking and more complex logic than what is already there. This is why the current approach is for simulation-specific objects to keep the `dont_add_these_options` list to tell `PetscSupport` and `SlepcSupport` about exceptions that if not handled, will cause unused option warnings.

> I think I'd want to understand why we wouldn't want to set some of those options in there before we add the ability to not set them
Since it does not make sense to have any of these options not set, I'll remove that ability for now. Worst case, we discover unused option warnings related to these options down the road and re-add the `dont_set_these_options` capability.

pbehne:
@lindsayad, this is ready for another review. The force push was to rebase to next to get a test to pass.

framework/include/utils/SlepcSupport.h
lindsayad:
```suggestion
```

lindsayad (approved):

PR #30250, opened 2025-04-03
Add derivatives to GrandPotentialSinteringMaterial

### LINKED ISSUE #30236: Add derivatives to GrandPotentialSinteringMaterial for grand potential model with strict mass conservation
### Motivation

The multi-phase, multi-order parameter grand potential phase-field model has an option that enforces strict mass conservation. However, some material properties added to allow this option do not currently have associated derivatives. This means that kernels that take these material properties as input do not have analytically calculated Jacobians, which may degrade the convergence of the model.

### Design

Add derivatives with respect to phi for _hv_c_min, _hs_c_min, _hv_over_kVa, _hs_over_kVa and with respect to grain order order parameters gr0, gr1, etc. for _hs_c_min

### Impact

None

laagesen:
Closes #30236


sourabhkadambi (commented):
- It might be useful to update the documentation with the mass conservation relation and mention that it can currently be used only with the parabolic free energy model.
- Is there a reason for re-golding the tests that use the dilute, ideal and parabolic free energy models but not the mass conservation? [GrandPotentialStrictMassConservation.i](https://github.com/idaholab/moose/blob/2a85d1c2023b7c0dee252c657b290c819ec248c8/modules/phase_field/test/tests/GrandPotentialPFM/GrandPotentialStrictMassConservation.i#L153) seems to be the test that uses both the parabolic and mass conservation models.

dschwen (approved):

PR #30285, opened 2025-04-09
Fix docs for ElementalVariableValue and replicated, distributed and split meshes

nmnobre:
## Reason
This fixes some descriptions which were just wrong or outdated and expands on other sections.

## Design
For `ElementalVariableValue` I opted to fix the documentation instead of changing the implementation, though I'm open to suggestions. For the mesh docs, I based some of the expanded explanations on what @loganharbour wrote [here](https://github.com/idaholab/moose/discussions/25053#discussioncomment-6578013).

## Impact
Hopefully users will have a better understanding of what's happening in `ElementalVariableValue` and with our mesh reading/generating capabilities.

### DIFF
--- a/framework/doc/content/framework/documenting.md
+++ b/framework/doc/content/framework/documenting.md
@@ -75,7 +75,7 @@ cd doc
 ./moosedocs.py generate SolidMechanicsApp
 ```
 
-To generate pages for the framework, the moose test application can be used as follows.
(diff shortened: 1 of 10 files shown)


loganharbour (commented):
Thanks @nmnobre - this is a great change. We definitely do a poor job of explaining the complexities here

loganharbour (approved):

Issue #30126, opened 2025-03-18
Add testing for rejected timesteps


loganharbour:
### Motivation

We don't currently test at length the testing of rejected timesteps. We want to be sure that when a timestep is rejected, the state at the beginning of the next timestep matches the state at the beginning of the timestep that failed.

### Design

Add a `--restep` test harness option similar to the `--recover` test harness option, along with a `--test-restep` option to `MooseApp` to trigger this.

`--test-restep` will attempt (when a transient executioner is used) to fail one of the ran timesteps and re-run it. We will then make sure that the resulting final output is the same as the case where the timestep did not fail.

### Impact

Improved testing.

lindsayad:
Maybe it would be good to add what motivated this issue. A TH problem right?

regerdavid:
This issue was motivated by the Overlapping Domain Coupling. If on a given time step the coupling became unstable, the friction factor and heat source that we pass to SAM may diverge to huge values (i.e 1e10) throughout the Picard iterations. This typically causes the time step to fail. When the timestep is cut, the values of the friction factor and heat source that are stored in the user object are not reset, causing the userobject to pass the same huge value to SAM again, guaranteeing that the tilmestep will once again fail. We instead would like the friction factor and heat source to be reset to the value that was used at the end of the previous converged time step.

YaqiWang:
Is this turned on for apps like Griffin? I am really glad to see this and hopefully testing Griffin with this.

lindsayad:
@zachmprince can this be closed?

lindsayad:
> Is this turned on for apps like Griffin?

You would have to add it to a Griffin recipe

Issue #13262, opened 2019-04-18
ParsedFunction incorrectly resolves postprocessor dependencies


jack-dawg:
## Bug Description
ParsedFunction sometimes doesn't resolve postprocessor dependencies correctly and this can result in a postprocess value being passed to ParsedFunction before it has the correct value.

## Steps to Reproduce
Run the attached input with debug to produce a division by zero.

## Impact
It can result in ParsedFunction being passed incorrect values.
[parsedfunction_test.txt](https://github.com/idaholab/moose/files/3096099/parsedfunction_test.txt)

dschwen:
You example has a PostProcessor that depends on a Function that depends on a Postprocessor that depends on a Function!

jack-dawg:
Is that a problem?

jack-dawg:
I wanted a simple example and couldn't find a constant value postprocessor.

dschwen:
This would require cross-system dependency resolution, which MOOSE does not (yet) have. As a consequence it would potentially require multiple loops over the mesh (in case your PostProcessors are ElementPostprocessors). In the current architecture for one step all objects of a System are executed together, then the the next system (this is  abit simplified, but mostly correct).

dschwen:
You can take a NumDOFs Postprocessor for example.

jack-dawg:
Okay, I updated the input.

dschwen:
Ok, now we have PostProcessor (A) that depends on a Function that depends on a PostProcessor (B). This could work as we could just run B and then A (where A executes the function). As neither PP needs to loop the mesh that would yield a correct result. The issue is that this will not work in a generic way, as we (MOOSE) does not know that PP B is constant and neither PP needs to loop the mesh.
So really the Dependency resolution never sees the A->B dependence (as there is a Function inbetween). @permcody can probably comment in more detail, but as far as I can see this can only be resolved with general cross-system dependency resolution.

jack-dawg:
I just added a variable and kernel to the test so it has nonzero DOF. @joshuahansel

joshuahansel:
I imagine the full dependency resolution here would be quite difficult. I know this isn't the optimal solution here, but could we add a parameter to all post-processors for users to explicitly specify dependencies on other post-processors?

permcody:
I'm not a big fan of going down this path if we can avoid it. Once we add it to Postprocessors, people will want it in the other systems and it might creep it's way into our input where we really should be able to fix this with a better resolution system. 

[...]

PR #30301, opened 2025-04-10
Restep testing

zachmprince:
## Reason

Refs #30126

## Design

Tests can be run with `./run_tests --restep`. Or an input can be directly run with `moose-opt -i input.i --test-restep`.

The Transient executioner will then attempt to find the approximate half timestep of the simulation (if not it defaults to the first timestep). Once this timestep is reached, the timestep is "failed" in the executioner by manually setting `_last_solve_converged = false` in `TransientBase::takeStep()`, this will trigger a retry of the timestep, notably without any change in the timestep size. Supposedly, this should not change the answer of the simulation when this is done. This all follows very similar logic to `--recover` testing.

It seems most of the infrastructure in the framework is relatively impervious to this type of testing, but there were still many aspects that needed to be addressed. With the changes in this PR, all of the framework tests should be passing. The commits are pretty modular, describing the fix for various capabilities or a reasoning why restep wouldn't be possible/sensible for a given test. I intend on summarizing in this description in a follow-up edit.

## Impact

Better feature testing for new capabilities, ensuring robustness to failed solves.

## App patches

- [x] marmot (breaking): https://github.inl.gov/ncrc/marmot/pull/940
- [x] grizzly: https://github.inl.gov/ncrc/grizzly/pull/794
- [x] griffin: https://github.inl.gov/ncrc/griffin/pull/2532
- [x] TMAP8 (breaking): https://github.com/idaholab/TMAP8/pull/278
- [x] SAM (breaking): https://git-nse.egs.anl.gov/SAM/SAM/-/merge_requests/1086


loganharbour:
@zachmprince let me know when this is ready for review

zachmprince:
@loganharbour This is ready for review

- Currently failing module tests. I can't figure out why these are failing and would need to outsource debugging. Will need to decide whether to pursue fixing or at least explaining, or simply skipping restep for these tests.
    ```
    xfem/test:solid_mechanics_basic.crack_propagation_ave ................................................................................................................ FAILED (EXODIFF)
    contact/test:pdass_problems.cylinder_friction_penalty_al_frictional_action_amg_bussetta_simple ....................................................................... FAILED (CSVDIFF)
    xfem/test:mechanical_constraint.glued_penalty ........................................................................................................................ FAILED (EXODIFF)
    xfem/test:solid_mechanics_basic.crack_propagation_single_point ....................................................................................................... FAILED (EXODIFF)
    contact/test:mortar_aux_kernels.block-dynamics-aux-wear .............................................................................................................. FAILED (EXODIFF)
    phase_field/test:grain_tracker_test.changing_avg_volume .............................................................................................................. FAILED (CSVDIFF)
    combined/test:gap_heat_transfer_jac.gap-heat-transfer-jac ............................................................................................................. FAILED (ERRMSG)
    ```
- Grizzly failures addressed in https://github.inl.gov/ncrc/grizzly/pull/794
- Need help with Marmot since I don't have access
- Griffin failures addressed in https://github.inl.gov/ncrc/griffin/pull/2532
- SAM failures simply require a regold, but I don't have access to ANL repo.

loganharbour:
Let's get this in this week

framework/src/executioners/TransientBase.C
```diff
@@ -428,6 +482,29 @@ TransientBase::takeStep(Real input_dt)
...
+  // We're running with --test-restep and we have just solved
+  // the timestep we are to repeat for the first time
+  if (_test_restep_step && *_test_restep_step == _t_step)
+  {
+    mooseAssert(!_testing_restep, "Should not be set");
+
+    // We shouldn't try to test resolving if the last solve was
+    // a failure. This might actually not be needed, but I'm
+    // not confident
+    // if (!lastSolveConverged())
```
loganharbour: Should this remain?

framework/src/vectorpostprocessors/SamplerBase.C
```diff
@@ -41,6 +44,10 @@ SamplerBase::SamplerBase(const InputParameters & parameters,
   : _sampler_params(parameters),
     _vpp(vpp),
     _comm(comm),
+    _sampler_transient(dynamic_cast<TransientBase *>(
```
loganharbour, line 47: Can this cast fail?

zachmprince: Yes, but it should just return a nullptr, right? That's checked before calling it below.

framework/src/executioners/TransientBase.C
zachmprince: I guess not. I'll remove it.

framework/src/vectorpostprocessors/SamplerBase.C
loganharbour: Ah, yeah. I missed where the parenthesis ended. Thought you were calling `getMooseApp()` on a potentially null cast

framework/include/executioners/SecantSolve.h
```diff
@@ -79,27 +79,30 @@ class SecantSolve : public FixedPointSolve
   /// Print the convergence history of the coupling, at every coupling iteration
   virtual void printFixedPointConvergenceHistory() override final;
 
+  /// Default TagID if storage is not allocated
+  static TagID _default_tag_id;
```
loganharbour: Few things:

- For static variables, don't prefix with an underscore. Underscore implies object state, but this state doesn't live as a member that is unique for each class instantiation
- These should typically be const
- While you can initialize these in the header, we typically prefer to do this in the initializer list

Fortunately - `INVALID_TAG_ID` already exists in `MooseTypes.h` and this should be used instead. So, remove this definition and declaration, use `INVALID_TAG_ID`, and set it in the initializer list instead.

loganharbour (changes requested):

framework/include/executioners/SecantSolve.h
zachmprince: image

loganharbour:
Go ahead and do the rebase - I've got everything as viewed and all looks well.

loganharbour:
SAM devel is currently broken so we can't get any patches into it. Otherwise - this is good.

loganharbour (approved):

framework/include/base/MooseApp.h
```diff
@@ -545,11 +545,16 @@ class MooseApp : public ConsoleStreamInterface,
...
 
   /**
-   *  Whether or not this simulation should only run half its transient (useful for testing
+   * Whether or not this simulation should only run half its transient (useful for testing
    * recovery)
    */
   bool testCheckpointHalfTransient() const { return _test_checkpoint_half_transient; }
 
+  /**
+   * Whether or not this simulation should fail its middle timestep and repeat (for testing)
```
GiudGiud: I m seeing tests with changed stepping behavior to enable re-step testing.
maybe middle is not the right choice. How about "middle" when we know which one and just "second" when we dont?

framework/include/controls/PIDTransientControl.h
```diff
@@ -68,4 +68,8 @@ class PIDTransientControl : public Control
   Real & _old_delta;
   /// whether the app has recovered once, because the logic for setting the value is different after having just recovered
   bool _has_recovered;
+  /// the difference with the target from the previous time step, used if a time step fails
+  Real _delta_prev_tstep = 0.0;
+  /// the difference with the target from the from the second-to-last iteration of previous time step, used if a time step fails
```
GiudGiud:
```suggestion
  /// the difference with the target from the second-to-last iteration of previous time step, used if a time step fails
```

framework/include/executioners/FixedPointSolve.h
```diff
@@ -245,6 +245,7 @@ class FixedPointSolve : public SolveObject
 private:
   /// Postprocessor value for user-defined fixed point convergence check
   const PostprocessorValue * const _fixed_point_custom_pp;
+  const PostprocessorValue * const _fixed_point_custom_pp_old;
```
GiudGiud: isnt PP_old equal to that?

test/tests/time_integrators/convergence/tests
```diff
@@ -10,6 +10,7 @@
       input = 'explicit_convergence.i'
       cli_args = 'Executioner/TimeIntegrator/type=Heun Executioner/dt=0.00390625 Outputs/file_base=heun_0'
       exodiff = 'heun_0.e'
+      restep = false # See mooseError in ExplicitRK2::solve()
```
GiudGiud:
```suggestion
      restep = false # Re-step not implemented for ExplicitRK2
```

GiudGiud: there could be more than 1 mooseError there

test/tests/controls/time_periods/kernels/kernels.i
```diff
@@ -61,6 +61,7 @@
     enable_objects = 'Kernel::diff0'
     disable_objects = '*::diff1'
     start_time = '0'
-    end_time = '0.49'
+    end_time = '0.51'
+    execute_on = 'timestep_begin'
```
GiudGiud, line 65: this kind of changes shows TimePeriod isnt working as expected for re-step
best have an error than make these changes

test/src/vectorpostprocessors/DynamicPointValueSampler.C
```diff
@@ -37,7 +37,7 @@ DynamicPointValueSampler::DynamicPointValueSampler(const InputParameters & param
 void
 DynamicPointValueSampler::initialize()
 {
-  _num_points = _num_points + _adder;
+  _num_points = getParam<unsigned int>("num_points") + _adder * (_t_step + 1);
```
GiudGiud, line 40: should _num_points be restartable instead?

python/TestHarness/testers/RunException.py
```diff
@@ -24,6 +24,9 @@ def validParams():
         # RunException tests executed in parallel need to have their output redirected to a file, and examined individually
         params['redirect_output'] = True
 
+        params['recover'] = False
+        params['restep'] = False
```
GiudGiud, line 28: why would restepping not hit exceptions?

modules/solid_mechanics/test/tests/scalar_material_damage/tests
```diff
@@ -187,6 +187,9 @@
     csvdiff = 'nonlocal_scalar_damage_out.csv'
     abs_zero = 1.e-6
     recover = false
+    # Explicit nature this test is emulating makes it non-restep-able until
```
GiudGiud:
```suggestion
    # Explicit nature that this test is emulating makes it not able to perform a re-step until
```

?

modules/porous_flow/test/tests/sinks/s06.i
```diff
@@ -172,25 +133,37 @@
     execute_on = 'initial timestep_end'
   []
   [m10]
-    type = FunctionValuePostprocessor
-    function = mass10
+    type = ParsedPostprocessor
```
GiudGiud, line 136: I understand that they are not changing the results but why are you making these changes? Functions should work with re-step

modules/rdg/test/tests/advection_1d/tests
```diff
@@ -6,6 +6,7 @@
     abs_zero = 1e-4
     rel_err = 5e-5
     threading = '!pthreads'
+    restep = false # See mooseError in ExplicitRK2::solve()
```
GiudGiud:
```suggestion
    restep = false # ExplicitRK2 does not support re-stepping a failed timestep
```

modules/richards/test/tests/gravity_head_1/gh20.i
```diff
@@ -116,7 +116,7 @@
 [Executioner]
   type = Transient
   solve_type = Newton
-  dt = 1E0
+  num_steps = 25
```
GiudGiud: why is that? we cant determine the middle one?
if so maybe let's either re-step the second step or skip re-step

modules/solid_mechanics/examples/uexternaldb_coupling/tests
```diff
@@ -12,6 +12,7 @@
     input = 'test.i'
     csvdiff = 'UEXTERNALDB_out.csv'
     valgrind = NONE
+    restep = false # Plugin not setup to handle redone timesteps in it's output
```
GiudGiud:
```suggestion
    restep = false # Plugin not setup to handle redone timesteps in its output
```
is there an error?

modules/solid_mechanics/src/userobjects/AbaqusUserElement.C
```diff
@@ -26,7 +26,7 @@ AbaqusUserElement::validParams()
   // execute during residual and Jacobian evaluation
   ExecFlagEnum & exec_enum = params.set<ExecFlagEnum>("execute_on", true);
   exec_enum.addAvailableFlags(EXEC_PRE_KERNELS);
-  exec_enum = {EXEC_PRE_KERNELS, EXEC_TIMESTEP_END};
+  exec_enum = {EXEC_PRE_KERNELS};
```
GiudGiud, line 29: @dschwen can you check this Abaqus UO change?

GiudGiud: you sure we need to remove timestep_end?

rest of changes look fine

modules/porous_flow/test/tests/hysteresis/1phase.i
```diff
@@ -111,7 +111,8 @@
 [Postprocessors]
   [flux]
     type = FunctionValuePostprocessor
-    function = 'if(t <= 9, -10, 10)'
+    function = 'if(t <= 0.5, 0.0, if(t <= 9.5, -10, 10))'
```
GiudGiud: why this change?

test/tests/userobjects/layered_base_restartable/tests
```diff
@@ -3,6 +3,8 @@
     type = 'Exodiff'
     input = 'layered_base_restartable.i'
     exodiff = 'layered_base_restartable_out.e'
+    # With restep, we need to execute UOs between the solution restore and auxkernel evaluation
+    cli_args = "UserObjects/npla/execute_on='multiapp_fixed_point_begin timestep_end'"
```
GiudGiud: use force_aux instead?

referring to multiapps in an inpiut without multiapps wont do

test/tests/transfers/multiapp_userobject_transfer/tests
```diff
@@ -50,6 +50,9 @@
     type = 'Exodiff'
     input = 'restricted_node_parent.i'
     exodiff = 'restricted_node_parent_out.e restricted_node_parent_out_ch00.e restricted_node_parent_out_ch10.e'
+    # With restep, we need to execute UOs between the solution restore and transfer
+    cli_args = "UserObjects/S_avg_front/execute_on='multiapp_fixed_point_begin timestep_end'
+                UserObjects/S_avg_back/execute_on='multiapp_fixed_point_begin timestep_end'"
```
GiudGiud: can you use exec-transfer instead? This is the intended flag here imo

might need to re-gold though :S let me know if so

test/tests/time_integrators/tvdrk2/tests
```diff
@@ -10,6 +10,7 @@
       type = 'Exodiff'
       input = '1d-linear.i'
       exodiff = '1d-linear_out.e'
+      restep = false # See mooseError at the end of ExplicitTVDRK2::solve()
```
GiudGiud, line 13:
```suggestion
      restep = false # ExplicitTVDRK2 does not support re-stepping
```

test/tests/time_integrators/rk-2/tests
```diff
@@ -10,6 +10,7 @@
       type = 'Exodiff'
       input = '1d-linear.i'
       exodiff = '1d-linear_out.e'
+      restep = false # See mooseError in ExplicitRK2::solve()
```
GiudGiud: same

test/tests/tag/tests
```diff
@@ -7,6 +7,8 @@
     requirement = 'The system shall support the ability for a the user to create a tagged vector which is not automatically zeroed.'
     input = 'not_zeroed_tag_vector.i'
     exodiff = 'not_zeroed_tag_vector_out.e'
+    # There is no way to reset a not zeroed vector if a timestep fails and is repeated
```
GiudGiud, line 10: do we have an error? this seems like a problem

test/tests/reporters/perf_graph_reporter/tests
```diff
@@ -23,6 +23,7 @@
     input = 'perf_graph_reporter_recover.i'
     cli_args = '--test-checkpoint-half-transient Outputs/checkpoint=true'
     recover = false
+    restep = false
```
GiudGiud:
```suggestion
    restep = false # half transient testing not compatible with restep
```

test/tests/postprocessors/random_pps/tests
```diff
@@ -18,6 +18,7 @@
       input = random_pps.i
       cli_args = 'Outputs/checkpoint=true --test-checkpoint-half-transient'
       recover = false
+      restep = false
```
GiudGiud:
```suggestion
      restep = false # half-transient not compatible with restep testing
```

test/tests/outputs/recover/tests
```diff
@@ -75,6 +76,7 @@
     max_parallel = 2
     delete_output_before_running = false
     recover = false
+    restep = false
```
GiudGiud:
```suggestion
    restep = false # recovery not compatible with re-step testing
```

test/tests/outputs/iterative/tests
```diff
@@ -79,6 +79,8 @@
     max_threads = 1
     # no. of iterations is not recoverable
     recover = false
+    # Number of iterations will accumulate after a failed timestep
```
GiudGiud:
```suggestion
    # Number of iterations still accumulated after a failed timestep
```

test/tests/outputs/checkpoint/tests
```diff
@@ -241,6 +241,7 @@
       input = checkpoint_block.i
       cli_args = '--test-checkpoint-half-transient'
       recover = false
+      restep = false
```
GiudGiud:
```suggestion
      restep = false # half-transient not compatible with restep testing
```

test/tests/multiapps/steffensen_postprocessor/tests
```diff
@@ -99,7 +109,13 @@
       input = 'transient_main.i'
       csvdiff = 'app_end_transfer_begin_steffensen_sub.csv'
       allow_warnings = true
-      cli_args = "MultiApps/sub/execute_on=timestep_end Transfers/left_from_sub/execute_on=timestep_begin Transfers/right_to_sub/execute_on=timestep_begin Executioner/transformed_postprocessors='' MultiApps/sub/transformed_postprocessors='from_main' Outputs/file_base=app_end_transfer_begin_steffensen_sub"
+      cli_args = "MultiApps/sub/execute_on=timestep_end
+                  Transfers/left_from_sub/execute_on=timestep_begin
+                  Transfers/right_to_sub/execute_on=timestep_begin
+                  Postprocessors/to_sub/execute_on='multiapp_fixed_point_begin timestep_end'
```
GiudGiud: try exec_transfer?

test/tests/multiapps/steffensen_postprocessor/tests
```diff
@@ -30,20 +30,23 @@
       input = 'transient_main.i'
       csvdiff = 'transient_main_out.csv transient_main_out_sub0_csv.csv'
       cli_args = 'Executioner/fixed_point_max_its=5 Executioner/fixed_point_abs_tol=1e-10'
+      restep = false # Test will be repeating it's own time steps
```
GiudGiud:
```suggestion
      restep = false # Test will be repeating its own time steps already
```

test/tests/multiapps/move_and_reset/tests
```diff
@@ -19,6 +19,15 @@
...
       exodiff = 'multilevel_parent_out.e multilevel_parent_out_sub0.e multilevel_parent_out_sub0.e-s002 multilevel_parent_out_sub0_sub0.e multilevel_parent_out_sub0_sub0.e-s002'
       recover = false
+      # Re-doing the time step at the same time as reset and move causes the
+      # sub-sub app to increment the file number once more than intended. This is
+      # due to the reset incrementing the file at the sub-sub app output on
+      # initial call (which doesn't happen in the sub app), then moving the app
+      # causes the exodus object to think it needs to increment again. This
+      # doesn't cause issues without restepping because it doesn't take into
+      # account the mesh change on the subsequent output call, but it does if we
+      # are re-doing that timestep.
```
GiudGiud, line 29: do we have an error for this?

test/tests/multiapps/catch_up/tests
```diff
@@ -7,6 +7,7 @@
     input = 'parent.i'
     exodiff = 'parent_out.e parent_out_sub_app0.e parent_out_sub_app1.e'
     allow_warnings = true
+    restep = false # Test will be repeating it's own time steps
```
GiudGiud:
```suggestion
    restep = false # Test will be repeating its own time steps
```

test/tests/misc/solution_invalid/tests
```diff
@@ -77,6 +77,7 @@
     cli_args = "Executioner/type=Transient Executioner/num_steps=1 Executioner/dtmin=1e-13 Executioner/error_on_dtmin=false Outputs/file_base='solution_invalid_transient' "
     jsondiff = 'solution_invalid_transient.json'
     hpc = false # iteration counts can be MPI dependent
+    restep = false # Test will be repeating it's own time steps
```
GiudGiud:
```suggestion
    restep = false # Test will be repeating its own time steps already
```

test/tests/mesh/add_sideset_ids/tests
```diff
@@ -30,5 +30,8 @@
     csvdiff = remove_sideset_ids_out.csv
     design = MooseMesh.md
     requirement = 'The system shall be able to remove sides from sidesets when tese sides elements are moved to new blocks.'
```
GiudGiud:
```suggestion
    requirement = 'The system shall be able to remove sides from sidesets when these sides elements are moved to new blocks.'
```

test/tests/materials/stateful_prop/tests
```diff
@@ -100,6 +100,9 @@
     input = 'stateful_prop_copy_test.i'
     exodiff = 'out_stateful_copy.e'
     max_parallel = 1
+    # There currently isn't a way to undo a stateful material copy when
+    # trying to re-do the timestep where it happened
+    restep = false
```
GiudGiud, line 105: do we have an error for this?

test/tests/dgkernels/1d_advection_dg/tests
```diff
@@ -5,6 +5,7 @@
     type = 'Exodiff'
     input = '1d_advection_dg.i'
     exodiff = '1d_advection_dg_out.e'
+    restep = false # See mooseError in ExplicitRK2::solve()
```
GiudGiud:
```suggestion
    restep = false # ExplicitRK2 does not support re-stepping at this time
```

test/src/timesteppers/TimeSequenceStepperFailTest.C
```diff
@@ -45,7 +45,4 @@ TimeSequenceStepperFailTest::step()
                    << std::endl);
     _converged = false;
   }
-
-  if (converged())
-    _current_step++;
```
GiudGiud, line 50: where should this be done? do we have a postStep routine?

modules/functional_expansion_tools/test/tests/standard_use/gold/interface_coupled_out.e
GiudGiud, line 1: I would rather you passed exodiff options to ignore one variable than regolded for a missing postprocessor

test/src/functions/TimestepSetupFunction.C
```diff
@@ -19,14 +19,14 @@ TimestepSetupFunction::validParams()
 }
 
 TimestepSetupFunction::TimestepSetupFunction(const InputParameters & parameters)
-  : Function(parameters), _local_timestep(declareRestartableData<unsigned int>("local_timestep", 0))
+  : Function(parameters), _local_timestep(-1)
```
GiudGiud, line 22: why isnt this working? shouldnt restartable data be reset on a failed step?

test/src/kernels/RestartDiffusion.C
```diff
@@ -22,14 +22,20 @@ RestartDiffusion::validParams()
 RestartDiffusion::RestartDiffusion(const InputParameters & parameters)
   : Kernel(parameters),
     _coef(getParam<Real>("coef")),
-    _current_coef(declareRestartableData<Real>("current_coef", 1))
+    _current_coef(declareRestartableData<Real>("current_coef", 1)),
```
GiudGiud, line 25: same question

framework/include/vectorpostprocessors/SamplerBase.h
```diff
@@ -108,6 +109,9 @@ class SamplerBase
   /// The communicator of the child
   const libMesh::Parallel::Communicator & _comm;
 
+  /// Transient executioner used to determine if the last solve converged
+  const TransientBase * const _sampler_transient;
```
GiudGiud, line 113: let's use the executioner base class, that s where it s defined

framework/include/base/MooseApp.h
GiudGiud: just an idea

framework/src/controls/PIDTransientControl.C
```diff
@@ -156,15 +161,10 @@ PIDTransientControl::execute()
     {
       _integral = _integral_old;
       _value = _value_old;
-
-      // Note we do not restore _old_delta because we dont have a way to keep track of it at the
-      // moment. And we really need to, it comes in the time derivative term. In the future, we will
-      // reset restartableData on new fixed point iterations.
```
GiudGiud, line 162: the future has come !

framework/src/executioners/SecantSolve.C
```diff
@@ -105,10 +115,17 @@ SecantSolve::saveVariableValues(const bool primary)
...
   // Before a solve, solution is a sequence term, after a solve, solution is the evaluated term
   xn_m1 = solution;
+
+  // Since we did not update on the 0th iteration, the solution is also the previous evaluated term
+  const unsigned int it = primary ? _fixed_point_it : _main_fixed_point_it;
+  if (it == 1)
+    fxn_m2 = solution;
+  // Otherwise we just advance
+  else
+    fxn_m2 = fxn_m1;
```
GiudGiud, line 128: thanks for catching that!

framework/src/postprocessors/PseudoTimestep.C
```diff
@@ -171,11 +171,23 @@ PseudoTimestep::execute()
   // at the end of each timestep call the postprocessor to set values for dt
   if (_current_execute_flag == EXEC_TIMESTEP_END)
   {
+    // This is all incase a timestep fails and needs to be re-done
```
GiudGiud:
```suggestion
    // This is all in case a timestep fails and needs to be re-done
```

GiudGiud (changes requested):
Absolutely huge PR hitting the whole repo. Thanks for taking care of that, it's an insane amount of work to cover all this. 

More comments:
please add explanation & advice on how to support restepping to doco

in the test objects, it seems that any statefulness needs to be removed by handling the restep explicitly (use a t_old). Why is that? I would hope we can use restartableData instead to get the previous value back

Another check we should do is restep x2. So restepping 2x in a row. This happens in simulations quite often. This would check if the logic added only works once or is robust to this

framework/include/base/MooseApp.h
loganharbour: It's a bit more complex than this.

Tests have changed stepping behavior due to ill-formed tests or literal bugs.

test/src/vectorpostprocessors/DynamicPointValueSampler.C
loganharbour: restartable data isn't loaded on failed timesteps.

that's the purpose of this PR - to work on figuring out where we need to this.

python/TestHarness/testers/RunException.py
loganharbour: same logic with the recover skips. too much complexity to add a full set of new testing, dealing with all of the caveats. it could in the future, but this is enough as is.

test/src/functions/TimestepSetupFunction.C
loganharbour: no

test/src/kernels/RestartDiffusion.C
loganharbour: still no

framework/include/executioners/FixedPointSolve.h
zachmprince: No, `_pp_old` is the value from the previous fixed-point iteration; `_fixed_point_custom_pp_old` is the value from the previous timestep. We want to make sure we set `_pp_old` to the previous timestep on the first iteration after the timestep fails.

GiudGiud: the attributes needs to be renamed to make that clear
and we need a docstring

framework/include/base/MooseApp.h
GiudGiud: A restep needs to work on all steps, with debatedly the exception of the first one
If a test does not work with restep on all steps then either:
- it s due to the test design and we can skip it
- we add an error in the object that does not support restep and we can skip it

modules/richards/test/tests/gravity_head_1/gh20.i
zachmprince: The number of steps is fixed for this test, but we can't programmatically determine it because that function could depend on things like postprocessors. I think is a viable technique to get restep working. Logan and I believe restepping the middle timestep is more robust since it generally allows the system to evolve to their inflection point.

python/TestHarness/testers/RunException.py
GiudGiud: yeah both logic are not correct. Let's create an issue

EDIT: for recover, it depends when the exception is hit. We should hit it, but be able to not perform phase 2 when hitting it. 
no such problem with re-stepping

modules/solid_mechanics/src/userobjects/AbaqusUserElement.C
zachmprince: I forget if we need to remove it, but it isn't necessary with these changes. Previously, `timestep_end` was simply used to do the `swap` operation. Now that it is in `timestepSetup`, there's no reason to have the extra execution.

test/src/vectorpostprocessors/DynamicPointValueSampler.C
GiudGiud: > restartable data isn't loaded on failed timesteps.

I thought we had discussed this was the way forward. 

> that's the purpose of this PR - to work on figuring out where we need to this.

We need restartable data to work. Working around it by using input parameters inside the code (which we usually avoid) in the objects that rely on restartable data is not the way forward. 
At least create an issue if you want to move on with the PR

modules/porous_flow/test/tests/sinks/s06.i
zachmprince: It does work with functions, but the warehouse cannot resolve dependencies between functions and postprocessors, which was manifesting in this test as diffs with restep.

GiudGiud: We need an issue for this

test/tests/controls/time_periods/kernels/kernels.i
zachmprince: TimePeriod works fine with re-step in my opinion. The only problem with this test is that controls should be executed on  `timestep_begin`, which allows them to be re-executed on repeated timesteps at the proper time.

modules/richards/test/tests/gravity_head_1/gh20.i
GiudGiud: > since it generally allows the system to evolve to their inflection point.

I really don't see how

> I think is a viable technique to get restep working

When you can't determine the middle point, use the second step. That's simpler and you don't have to modify multiple tests

test/src/vectorpostprocessors/DynamicPointValueSampler.C
loganharbour: This PR scope is to add in additional testing so that we can systematically determine what restartable data is needed. 

Restartable data will not be added in failed timesteps on this PR.

modules/porous_flow/test/tests/sinks/s06.i
zachmprince: https://github.com/idaholab/moose/issues/13262

test/tests/multiapps/move_and_reset/tests
zachmprince: No, for three reasons: 1) this is a very improbable series of events (move and reset app with its own sub-app + failed timestep at the same time), 2) the logic would be quite complicated to catch, and 3) the result is still valid, it will just have different file names than if the timestep didn't repeat.

test/tests/materials/stateful_prop/tests
zachmprince: No, I think it would be up to the object calling `MaterialData::copy` to throw this error. This test is just using a test object calling that API.

test/src/timesteppers/TimeSequenceStepperFailTest.C
zachmprince: It's done in `acceptStep` now due to the changes in `TimeSequenceStepperBase.C`.

modules/porous_flow/test/tests/sinks/s06.i
GiudGiud: This would not really show up on a search for someone looking at their test re-stepping problem. 
I would recommend a new one linking to this old one

framework/include/vectorpostprocessors/SamplerBase.h
zachmprince: I use the type to determine whether it is a transient simulation. Would you rather see another API be used to determine that?

modules/porous_flow/test/tests/sinks/s06.i
zachmprince: I don't think that type of issue could ever be resolved, though. If there are dependency issues in any form, restep might show a failure. The real fix is to resolve the dependency issue like I did here. I will be adding documentation on the restep capability per your suggestion, what if I mention this as a common cause of failure there?

GiudGiud: >I don't think that type of issue could ever be resolved, though. 

you're not wrong...

>  The real fix is to resolve the dependency issue like I did here.

The problem is that every future developer trying to get restepping to pass will have to do the same thing as you. It's not ideal.

> what if I mention this as a common cause of failure there?

definitely a good step forward!

zachmprince: https://github.com/idaholab/moose/issues/30996

framework/include/vectorpostprocessors/SamplerBase.h
GiudGiud: ah no then keep it

modules/solid_mechanics/examples/uexternaldb_coupling/tests
zachmprince: No, the plugin is built for this test specifically. I don't think it is possible to have a general error.

framework/include/base/MooseApp.h
zachmprince: I ended up changing the logic to find the middle timestep:

- If `num_steps` is provided, get the middle one.
- If `end_time` is provided, find the middle time and repeat the first timestep that goes past this time.
- If `steady_state_detection` then we just use the second timestep.
- If `num_steps` and `end_time` is provided, it triggers restep on either the middle timestep or time, whichever is first.

This made it so that all appropriate tests triggered a restep without modification.

framework/include/executioners/FixedPointSolve.h
zachmprince: New variables and docstrings implemented

test/tests/controls/time_periods/kernels/kernels.i
zachmprince: I added a `paramInfo` in Controls.C that looks at the `execute_on` and whether `restep` is being tested.

modules/richards/test/tests/gravity_head_1/gh20.i
zachmprince: This change is no longer needed

modules/porous_flow/test/tests/hysteresis/1phase.i
zachmprince: No longer needed

test/tests/tag/tests
zachmprince: Added an error in `FEProblemBase::restoreSolutions`, which I believe is the appropriate place.

test/tests/multiapps/move_and_reset/tests
GiudGiud: 3) is a good reason!

modules/solid_mechanics/src/userobjects/AbaqusUserElement.C
GiudGiud: ok works.

test/tests/controls/time_periods/kernels/kernels.i
GiudGiud: works for me

framework/include/executioners/FixedPointSolve.h
GiudGiud: thanks!

framework/include/base/MooseApp.h
GiudGiud: noice logic seems even more robust than I was thinking

framework/src/executioners/TransientBase.C
```diff
@@ -200,14 +204,46 @@ TransientBase::TransientBase(const InputParameters & parameters)
...
+  if (_app.testCheckpointHalfTransient())
   {
+    mooseAssert(!_app.testReStep(), "Cannot use with restep");
     _end_time = (_start_time + _end_time) / 2.0;
     _num_steps /= 2.0;
 
     if (_num_steps == 0) // Always do one step in the first half
       _num_steps = 1;
   }
+  // Retest the middle timestep
```
GiudGiud:
```suggestion
  // Retest a timestep (see options below for which timestep)
```

framework/src/executioners/FixedPointSolve.C
```diff
@@ -239,13 +241,15 @@ FixedPointSolve::solve()
...
+    if (_old_entering_time == _problem.time() &&
+        _fixed_point_status != MooseFixedPointConvergenceReason::UNSOLVED)
     {
       // Keep track of the iteration number of the main app
       _main_fixed_point_it++;
 
       // Save variable values before the solve. Solving will provide new values
-      saveVariableValues(false);
+      if (!_app.isUltimateMaster())
+        saveVariableValues(false);
```
GiudGiud:
```suggestion
        saveVariableValues(/*is parent app of this iteration=*/false);
```

framework/include/base/MooseApp.h
GiudGiud:
```suggestion
   * Whether or not this simulation should fail a timestep and repeat (for testing)
   * Selection rules for which time step to fail in TransientBase.C constructor
```

framework/src/executioners/TransientBase.C
```diff
@@ -200,14 +204,46 @@ TransientBase::TransientBase(const InputParameters & parameters)
...
+  if (_app.testReStep())
```
GiudGiud:
```suggestion
  else if (_app.testReStep())
```

framework/src/multiapps/MultiApp.C
```diff
@@ -845,6 +857,16 @@ MultiApp::restore(bool force)
 
       _end_aux_solutions.clear();
     }
+
+    // Make sure the displaced mesh on the multiapp is up-to-date with displacement variables
+    for (const auto & app_ptr : _apps)
```
GiudGiud, line 862: do we need to do this if keep_solution_during_restore was on?

python/doc/content/python/TestHarness.md
```diff
@@ -45,6 +45,32 @@ through, but some of the important ones are:
...
+  - Unresolved dependencies between objects. Most commonly, this involves
+    function-postprocessor dependencies and cyclic dependencies between user
+    objects.
+  - Stateful class member variables in objects. Typically, this can be resolved
+    by checking if the current timestep is greater than the previous timestep
+    the object was run on. See
+    [Predictor::timestepSetup](framework/src/predictors/Predictor.C) for an
+    example.
+  - Specifying `Executioner/abort_on_solve_fail=true` in the test input or
+    specification `cli_args`.
```
GiudGiud:
```suggestion
    in the command line arguments (`cli_args`).
```

test/tests/multiapps/picard_postprocessor/tests
```diff
@@ -30,6 +30,7 @@
       input = 'transient_main.i'
       csvdiff = 'transient_main_out.csv transient_main_out_sub0_csv.csv'
       cli_args = 'Executioner/fixed_point_max_its=5 Executioner/fixed_point_abs_tol=1e-10'
+      restep = false # Test will be repeating it's own time steps
```
GiudGiud:
```suggestion
      restep = false # Test will be repeating its own time steps
```

GiudGiud (commented):
last round

framework/src/multiapps/MultiApp.C
zachmprince: I'm honestly not sure what the state of the displaced mesh is when it is restored, it could very well be undisplaced. I think it is probably safer to just displace the mesh even if we are keeping solutions. Especially since displacing the mesh is not a heavy operation.

GiudGiud (approved):

GiudGiud:
test failures might be new tests?

zachmprince:
> test failures might be new tests?

Those restep tests failure are some that I could not figure out why they weren't working, nor could I justify having `restep = false`. They will be optionally triggered tests for now until we can recruit experts to debug them.

YaqiWang:
This seems introducing a valgrind issue in Griffin with the custom Richardson executioner. I guess the fix is to set the default for `_last_solve_converged`. Why do we have to this now?

PR #30315, opened 2025-04-11
Remove the used systems from the prefixed logic.

### LINKED ISSUE #30308: Nonlinear assembly-based segregated solvers in NS are in conflict with prefix-based petsc parameter assignment.
### Bug Description

The segregated solvers that rely on the nonlinear system assembly algorithms are in conflict with the prefix-based petsc option assignment currently used for multi-system runs.

### Steps to Reproduce

Run any of the cases with `SIMPLENonlinearAssembly` with `-ksp_view`.

### Impact

Users who are not aware of the prefix might use the default preconditioner which yields poor preformance.

### [Optional] Diagnostics

_No response_

grmnptr:
Closes #30308


lindsayad (approved):

grmnptr:
Thanks for the quick review!

PR #30322, opened 2025-04-14
Import pycapabilities only when used

dschwen:
## Reason
Importing `pycapabilities` with trigger the build of the python binary module. That module depends on framework, so if no executable is present, this build can take a long time. We use `util.py` in cases where no executable is even required. 

## Design
Import `pycapabilities` late. The test harness checks for the existence of an executable anyways, before using capabilities. So if framework hasn't been built we would get a more meaningful error message. If capabilities are not used, then there is no need to do a lengthy compile otherwise.

## Impact
Better user experience.

### DIFF
--- a/python/TestHarness/util.py
+++ b/python/TestHarness/util.py
@@ -14,7 +14,6 @@ from collections import OrderedDict
 import json
 import yaml
 import sys
-import pycapabilities
 
 MOOSE_OPTIONS = {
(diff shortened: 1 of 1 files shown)


loganharbour (approved):
`MooseDocs/test/extensions.all/appsyntax` failure in https://civet.inl.gov/job/2817038/ is unrelated.

Good to merge pending https://civet.inl.gov/job/2817040/

PR #30325, opened 2025-04-14
Revert "Make Transient APIs more atomic"

lindsayad:
Reverts idaholab/moose#30135

Apparently this did not help the Griffin team achieve the capability they want, it decreases encapsulation, and it breaks some of @dschwen's work


dschwen (approved):
Thanks! I have confirmed that this fixes my issue.

loganharbour:
Will merge once https://civet.inl.gov/event/228492/ gets in.

PR #30344, opened 2025-04-16
Add skip_evals option to Terminator UserObject for problems.

### LINKED ISSUE #30343: Add an option to temporarily skip Terminator execution
### Motivation

When working with percentage error on small quantities, initial errors are often more numerical than practically relevant. Disabling a terminator for the first few timesteps seems like a straightforward way around this limitation.

### Design

Add a `skip_evals` parameter to `Terminator` which disables the terminator for the first `n` evaluations, as well as a counter to keep track of evaluations. 

### Impact

Adds something new. Backward compatibility preserved.

ke7kto:
Closes #30343 

## Reason
Handle initial numerical issues for Terminators

## Design
Add a counter and a user-specified cutoff for disabling the first `n` executions of the Terminator.

Technically, I believe there's an off-by-one error here, but with the default behavior, `skip_evals` lines up better with the timestep number, so I thought it might be better to have this behavior.

## Impact
Something new, existing API should be unaffected


dschwen:
The same effect could be achieved by coupling a timestep postprocessor and putting an additional comparison in the evaluated expression. I'd rather not merge this.

dschwen (changes requested):
No changes, just don't merge at a all. The number of evaluations is not a well defined quantity. It depends on convergence, execution flags, etc. The clean way would be to couple a Time or Time Step Postprocessor and explicitly put these in the conditional. This would also be more expressive than a "skip evals" parameter.

PR #30349, opened 2025-04-17
Libmesh submodule update 20250417

roystgnr:
- Autoconf-submodule (ACSM) update 
  - When recent clang versions are detected, the compiler flag `-fno-assume-unique-vtables` is used.  This enables `dynamic_cast` to work reliably on macOS 15 (Sequoia) environments, which fixes a number of libMesh and MOOSE failures there.
  - More thorough detection for code coverage support when `--enable-coverage` builds may require it from multiple languages' compilers.
- TIMPI submodule update
  - Fixes and tests for `timpi_version.h` utilities
  - Unit test fixes for C++14 compilers
  - `timpi_pure` attribute for pure functions; on C++17 and newer compilers this prevents ignoring the function return values
  - Marked `verify()` and `semiverify()` as pure.  This prevents the common usage error of not asserting the result when intended.
  - Removed default `Attributes` values.  Code which attempts parallel reductions on user-defined types without first specifying their attributes now fails to compile, instead of compiling but not working correctly.
- Added support for creating and computing on arbitrary polygons in 2D
- Better tolerance computations when stitching meshes along a boundary with discrete NodeElem elements
- Fix for reading of Gmsh files with physicals of different types but the same id numbers 
- Migrated more print methods of templated classes into headers, to enable user instantiations with unanticipated types
- Added "-V" option to the meshtool utility app to apply the `VariationalSmoother` to the input mesh
- Moved `demangle()` to `libmesh_common.h`.  Any failures in `cast_ptr` or `cast_ref` in dbg or devel modes are now reported using demangled typenames.
- Fix for return values of `FE::n_quadrature_points()`: after using a user-specified list of quadrature points rather than a quadrature-rule-generated list, the length of the specified list is now correctly returned.  This fixes undefined behavior in `JumpErrorEstimator` subclasses with the `integrate_slits` option enabled, which fixes a valgrind-detected error and a potential test failure in libMesh unit tests.
- Fix for undefined behavior in `SimplexRefiner`.  This similarly fixes a libMesh unit test.
- Fix for a Netgen configuration error: configuring the same Netgen-enabled libMesh build directory for a second time no longer leaves Netgen in an "unbuilt, but marked as built" state.
- Refactoring for simplicity and forward compatibility in Gaussian quadrature code
- More assertions

### DIFF
--- a/conda/libmesh/meta.yaml
+++ b/conda/libmesh/meta.yaml
@@ -4,8 +4,8 @@
 #
 # As well as any directions pertaining to modifying those files.
 # ALSO: Follow the directions in scripts/tests/versioner_hashes.yaml
-{% set build = 1 %}
-{% set version = "2025.04.02" %}
+{% set build = 0 %}
+{% set version = "2025.04.17" %}
 
 package:
   name: moose-libmesh
--- a/conda/moose-dev/conda_build_config.yaml
+++ b/conda/moose-dev/conda_build_config.yaml
@@ -3,8 +3,8 @@ mpi:
   - openmpi
 
 moose_libmesh:
-  - moose-libmesh 2025.04.02 mpich_1
-  - moose-libmesh 2025.04.02 openmpi_1
+  - moose-libmesh 2025.04.17 mpich_0
+  - moose-libmesh 2025.04.17 openmpi_0
 
 moose_wasp:
   - moose-wasp 2025.02.25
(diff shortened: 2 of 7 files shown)


roystgnr:
Marking this "do not merge" since it's being combined into #29765, but I'll leave this one open in case there are any CI failures that would be easier to examine against a smaller changeset.

roystgnr:
Aside from an unrelated python failure and HPC hang, CI looks good here.  Closing in favor of #29765.

PR #24063, opened 2023-04-13
Fixup regression in empty collection indicator message

cticenhour:
This addresses S-02 from 2021 Blackbear/Grizzly audit, which had regressed to the prior unsatisfactory state. The test fixups required here should keep us from regressing again. 

Tagging @bwspenc for awareness (as he discovered the regression) and a review if available. 

Refs #18816

PR #30355, opened 2025-04-17
Add default phrase when empty validation and verification sections exist in the VVR

cticenhour:
## Reason
It was discovered, when improving the SQA documentation for a MOOSE-based application, that no default message exists when `!sqa verification` or `!sqa validation` is used *and* no verification or validation cases exist. This could be taken (during an audit) as leaving SQA documentation blank by accident, when some modules or applications may intentionally not have V&V cases, leaving it to their downstream applications to implement them. A reasonable default phrase should be displayed in this scenario.

## Design
This should be implemented similarly to how it was done for requirements in #24063. When no test specs are encountered that are labeled validation or verification, a String token should be produced to display the default phrase. 

## Impact
No impact to MOOSE results. Improved defaults when fields in SQA documentation (most of the time in the VVR, in this case) are empty, indicating intentionality on the part of the developers.

> [!NOTE]
> This PR also improves the test for the requirements default implemented in #24063.

### DIFF
--- a/python/MooseDocs/extensions/sqa.py
+++ b/python/MooseDocs/extensions/sqa.py
@@ -494,10 +494,18 @@ class SQAVerificationCommand(SQARequirementsCommand):
         settings['link-{}'.format(subcommand)] = True
         matrix = SQARequirementMatrix(parent)
         SQARequirementMatrixHeading(matrix, category=category, string=subcommand.title())
+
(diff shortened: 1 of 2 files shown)


loganharbour (approved):

loganharbour:
HPC tests can be disregarded during the HPC outage.

PR #30364, opened 2025-04-21
More platypus migration

lindsayad:
@alexanderianblair hopefully this is the last we need to do? Merge conflicts got pretty tricky with this one


loganharbour (changes requested):
Need to address the unit failures

framework/include/mfem/bcs/MFEMBoundaryCondition.h
```diff
@@ -22,17 +22,16 @@ class MFEMBoundaryCondition : public MFEMGeneralUserObject
   // Get name of the test variable labelling the weak form this kernel is added to
   const VariableName & getTestVariableName() const { return _test_var_name; }
 
+  bool isBoundaryRestricted() { return !(_bdr_attributes.Size() == 1 && _bdr_attributes[0] == -1); }
```
loganharbour, line 25:
```suggestion
  bool isBoundaryRestricted() const { return !(_bdr_attributes.Size() == 1 && _bdr_attributes[0] == -1); }
```

framework/src/mfem/bcs/MFEMBoundaryCondition.C
```diff
@@ -14,7 +15,9 @@ MFEMBoundaryCondition::validParams()
   // Create user-facing 'boundary' input for restricting inheriting object to boundaries
   params.addParam<std::vector<BoundaryName>>(
       "boundary",
-      "The list of boundaries (ids or names) from the mesh where this boundary condition applies");
+      {"-1"},
```
loganharbour, line 18: Does `-1` represent something to MFEM?

loganharbour (changes requested):

framework/src/mfem/bcs/MFEMBoundaryCondition.C
alexanderianblair: Yes - it's the ID used to signify the boundary condition should be applied to every sideset on the mesh (although adding a comment that this is the default would be desirable here).

framework/src/mfem/bcs/MFEMBoundaryCondition.C
```diff
@@ -14,7 +15,9 @@ MFEMBoundaryCondition::validParams()
   // Create user-facing 'boundary' input for restricting inheriting object to boundaries
```
loganharbour, line 15:
```suggestion
  // Create user-facing 'boundary' input for restricting inheriting object to boundaries
  // MFEM uses the boundary -1 to signify every sideset
```

framework/src/mfem/bcs/MFEMBoundaryCondition.C
loganharbour: Perfect, went ahead and added this

loganharbour (approved):
LGTM.

Pending a second set of eyes from @Heinrich-BR

framework/src/mfem/equation_systems/EquationSystem.C
```diff
@@ -123,7 +145,7 @@ EquationSystem::FormLinearSystem(mfem::OperatorHandle & op,
     default:
       MFEM_VERIFY(_test_var_names.size() == 1,
                   "Non-legacy assembly is only supported for single-variable systems");
-      MFEM_VERIFY(!_mblf_kernels_map_map.size(),
+      MFEM_VERIFY(_test_var_names.size() == _trial_var_names.size(),
```
Heinrich-BR, line 148: I can see in line 113 of this same file we use `mooseAssert`, while here we use `MFEM_VERIFY`. Have we established a rule for when to use each, or is it mostly random?

Heinrich-BR: Also, unless I'm missing something, checking that the sizes of `_test_var_names` and `_trial_var_names` are the same and equal to one isn't exactly the same as checking that the system is square. For instance you could have a system with one test variable and another, different trial variable. I think such a system could still work with non-legacy assembly as it is written here, but maybe then we should change the warning messages to "Non-legacy assembly is only supported for single trial variable systems", and "Non-legacy assembly is only supported for single test variable systems", or something of the sort.

framework/src/mfem/equation_systems/EquationSystem.C
Heinrich-BR, line 1: This is not one of the changes made for this PR, but I've just noticed that there seems to be an error on line 205. We have       
`auto trial_var_name = _test_var_names.at(j);`. 
Shouldn't it be      
`auto trial_var_name = _trial_var_names.at(j);`?

framework/src/mfem/equation_systems/EquationSystem.C
```diff
@@ -350,20 +353,17 @@ EquationSystem::BuildMixedBilinearForms()
...
       // kernels
-      if (_mblf_kernels_map_map.Has(test_var_name) &&
-          _mblf_kernels_map_map.Get(test_var_name)->Has(trial_var_name))
+      if (_kernels_map.Has(test_var_name) && _kernels_map.Get(test_var_name)->Has(trial_var_name) &&
+          test_var_name != trial_var_name)
       {
-        auto mblf_kernels = _mblf_kernels_map_map.GetRef(test_var_name).GetRef(trial_var_name);
-        auto mblf = std::make_shared<mfem::ParMixedBilinearForm>(_test_pfespaces.at(j),
-                                                                 _test_pfespaces.at(i));
+        mblf->SetAssemblyLevel(_assembly_level);
```
Heinrich-BR, line 363: We just need to check that non-legacy assembly here works with single test and single trial mixed variable systems. This hasn't been tested because it didn't use to be allowed (it would fail an MFEM_VERIFY).

alexanderianblair: The upcoming Darcy example that's being ported over from [Platypus PR #82](https://github.com/aurora-multiphysics/platypus/pull/82) should cover that, but for the purposes of this PR, this should be no worse than the current state (since previously any `MixedBilinearForm` present was ignoring the `AssemblyLevel`.

framework/src/mfem/equation_systems/EquationSystem.C
alexanderianblair: It should indeed!

framework/src/mfem/equation_systems/EquationSystem.C
alexanderianblair: We should use `mooseAssert` going forward, since it provides more context to the user for the error. And yes - you're right, this check won't catch cases where the same number of (different) test and trial spaces have been added to the system

Heinrich-BR (commented):
Thank you for the changes @lindsayad! It's looking good to me now. Apparently the code is failing a couple of tests on CIVET. Once that is addressed, we should be good to go.

loganharbour (approved):
Failure here is unrelated https://civet.inl.gov/job/2844151/

PR #30376, opened 2025-04-22
Leaner build for capabilities without MOOSE dependency

dschwen:
No more linking against moose framework lib -> faster build time for `capabilities.so`.

Let's hope this gets me off @lindsayad's and @roystgnr's naughty list.

Refs #26470

### DIFF
--- a/framework/moose.mk
+++ b/framework/moose.mk
@@ -75,7 +75,7 @@ pyhit_srcfiles  := $(hit_srcdir)/hit.cpp $(hit_srcdir)/lex.cc $(hit_srcdir)/pars
 # capabilities python bindings
 #
 CAPABILITIES_DIR ?= $(MOOSE_DIR)/framework/contrib/capabilities
-capabilities_srcfiles := $(CAPABILITIES_DIR)/capabilities.C
+capabilities_srcfiles := $(CAPABILITIES_DIR)/capabilities.C $(FRAMEWORK_DIR)/src/utils/CapabilityUtils.C
 
 # Making a .la object instead.  This is what you make out of .lo objects...
 moose_LIB := $(FRAMEWORK_DIR)/libmoose-$(METHOD).la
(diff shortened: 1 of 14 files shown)


framework/moose.mk
```diff
@@ -581,10 +578,10 @@ install_python: $(pyhit_LIB) $(capabilities_LIB)
 	@cp -f $(capabilities_LIB) $(python_install_dir)/
 	@echo "Installing python library $(capabilities_LIB)"
 	@$(call remove_rpath,$(python_install_dir)/$(capabilities_LIBNAME),$(FRAMEWORK_DIR))
-	@$(call remove_rpath,$(python_install_dir)/$(capabilities_LIBNAME),$(HIT_DIR))
+	# @$(call remove_rpath,$(python_install_dir)/$(capabilities_LIBNAME),$(HIT_DIR))
```
loganharbour, line 581:
```suggestion
```

roystgnr:
Ha!  I'll try it out locally now.

framework/moose.mk
```diff
@@ -581,10 +578,10 @@ install_python: $(pyhit_LIB) $(capabilities_LIB)
...
 	@$(call patch_rpath,$(python_install_dir)/$(capabilities_LIBNAME),$(lib_install_dir))
-	@$(call patch_relink,$(python_install_dir)/$(capabilities_LIBNAME),$(shell realpath $(moose_DYLIB)),$(shell readlink $(moose_DYLIB)))
-	@$(call patch_relink,$(python_install_dir)/$(capabilities_LIBNAME),$(shell realpath $(hit_DYLIB)),$(shell readlink $(hit_DYLIB)))
+	# @$(call patch_relink,$(python_install_dir)/$(capabilities_LIBNAME),$(shell realpath $(moose_DYLIB)),$(shell readlink $(moose_DYLIB)))
+	# @$(call patch_relink,$(python_install_dir)/$(capabilities_LIBNAME),$(shell realpath $(hit_DYLIB)),$(shell readlink $(hit_DYLIB)))
```
loganharbour, line 584:
```suggestion
```

loganharbour (changes requested):

roystgnr:
I'm still seeing a loop of
```
Building and linking capabilities...
Building and linking capabilities...
Building and linking capabilities...
```
when I invoke `run_tests`.

roystgnr:
The loop does appear to be interleaved with actual running tests, eventually - not sure if I had to hit Ctrl+C first or if that was just coincidence.

loganharbour:
This means that `import pycapabilities` is failing multiple times.

@dschwen we should do the build once up front (if capabilities are enabled)

dschwen:
Disappointment intensifies...

dschwen:
This happens behind the scenes

```
>>> import pycapabilities
Building and linking capabilities...
Traceback (most recent call last):
  File "/home/daniel/Programs/moose/python/pycapabilities/__init__.py", line 13, in <module>
    import capabilities
ModuleNotFoundError: No module named 'capabilities'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/daniel/Programs/moose/python/pycapabilities/__init__.py", line 18, in <module>
    import capabilities
ImportError: /home/daniel/Programs/moose/python/pycapabilities/../../framework/contrib/capabilities/capabilities.so: file too short

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/daniel/Programs/moose/python/pycapabilities/__init__.py", line 22, in <module>
    import capabilities
ImportError: /home/daniel/Programs/moose/python/pycapabilities/../../framework/contrib/capabilities/capabilities.so: undefined symbol: _ZN15CapabilityUtils5checkENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERKSt3mapIS5_St4pairISt7variantIJbiS5_EES5_ESt4lessIS5_ESaIS7_IKS5_SA_EEE
```

Basically forgot to link the implementation of `CapabilityUtils::check` (which used to be linked along with all of MOOSE). I'll fix that tomorrow.

dschwen:
Ok, I just pushed a version that works locally for me.

dschwen:
This is bs. Again clang-format changes for lambda formatting are biting me.

lindsayad:
clang-format is getting closer to idempotent. We shouldn't be seeing too much more of this as we progress in versions

roystgnr:
e9c4ac9b appears to be running tests fine for me.  Thanks to both of you!

loganharbour:
@roystgnr can you give this a try now?

lindsayad:
I still observe 10 seconds before I get any console output with `run_tests`. Not very long ago I feel like console output began almost immediately. I don't know if the whole 10 seconds is attributable to capabilities stuff. Either way it would be nice to get something like a timed print to know what is happening for that 10 seconds

loganharbour:
> I still observe 10 seconds before I get any console output with `run_tests`. Not very long ago I feel like console output began almost immediately. I don't know if the whole 10 seconds is attributable to capabilities stuff. Either way it would be nice to get something like a timed print to know what is happening for that 10 seconds
> 
> I'm not saying that's the responsibility of this PR

INL machine or HPC?

lindsayad:
rod in `/data/lindad/projects/moose/test`

lindsayad:
Hmm well then I'll stop complaining and accept 😄

lindsayad:
Accept the 10 seconds that is. Doing a print of what we're doing still would be of value I think

loganharbour:
@lindsayad do you mind reviewing now that both Daniel and I have contributed

framework/src/utils/CapabilityUtils.C
lindsayad, line 1: Should this file go in contrib/capabilities instead?

lindsayad: along with the header

lindsayad: I guess maybe not since it relies on MooseStringUtils

lindsayad (approved):

loganharbour (approved):
Approving changes that are not my own

roystgnr:
I'm not sure if it was supposed to be fixed by this PR or if it's completely unrelated, but I do still have the problem (which cropped up around the same time and had assumed was related) where I can't Ctrl+C or Ctrl+\ to kill a MOOSE `make` anymore.  I'm somewhat impressed by that bug - it can't be my environment, since Ctrl+C still works on `make` in the libMesh directory, yet I can't imagine what someone could have put in a Makefile to cause it to resist being killed either.  Does GNU Make have a `--go-full-skynet` option it can pass to a sub-make?

PR #30384, opened 2025-04-23
Tidy up initial batch of MFEM integration docs/tests

nmnobre:
Replaces aurora-multiphysics/platypus#87

## Reason
Redundancies in test specifications and their input meshes.
Lack of proper copyright/license for the MFEM meshes in use.
Need for requirement specifications that better align with their respective tests.

## Design
Code refactoring for extended reuse of existing files.

## Impact
Testing against the same inputs/outputs where applicable provides stronger guarantees for the functionality being tested.
Attribute the correct copyright/license where appropriate.


test/tests/mfem/kernels/curlcurl.i
```diff
@@ -3,7 +3,7 @@
 
 [Mesh]
   type = MFEMMesh
-  file = gold/small_fichera.mesh
+  file = ../../../../unit/data/small_fichera.mesh
```
loganharbour, line 6: This will not work with installed builds (unless it's a symlink).

When we install the tests, you only get the data from the root (./test) in this case. However, if you create a symlink for

```
ln -s ../../../../unit/data/small_fichera.mesh small_fichera.mesh
```

and then use
```
file = small_fichera.mesh
```

during the install process we will follow the symlinks and duplicate the data

nmnobre: Thanks for chiming in!

> This will not work with installed builds

That with conda?

> However, if you create a symlink

Okay, in that case, I'd perhaps create a `test/data/mfem` folder that symlinks `unit/data/mfem`? The latter doesn't exist yet, but I think it'd probably make sense to dump the meshes there instead of directly in `unit/data`?
This made me think of something crazier though, what do you think of symlinking directly the mfem meshes folder? For users who don't clone mfem, the symlinks would be broken, but if they are hidden in the folders I just mentioned, would this be problematic?

nmnobre: I've moved all meshes to `test/tests/mfem/mesh` so it's the unit tests instead that rely on reading the meshes from outside their directory (which I presume is safer?). No symlinks, though that `mug.e` _is_ big...

PS: The bot complains when the repo size increases a lot but doesn't bring you cake when it reduces by 12MiB? Shame. xD

loganharbour: Perfect, that'll do. The whole `test/tests` directory is copied, so any relative paths beneath that folder will work.

> PS: The bot complains when the repo size increases a lot but doesn't bring you cake when it reduces by 12MiB? Shame. xD

Well... it still exists in history, so it doesn't get removed 😢

test/tests/mfem/kernels/tests
lindsayad, line 1: usually I harp on not having implementation details such as "MFEM" in requirements ... but as I've helped push having multiple finite element backends in MOOSE, I must accept that sometimes the only way to avoid conceptually duplicate requirements at this point is to specify that implementation detail

test/tests/mfem/outputs/tests
```diff
@@ -1,15 +1,18 @@
...
+    input = ../kernels/diffusion.i
+    cli_args = 'Outputs/active="ParaViewDataCollection VisItDataCollection ConduitDataCollection" '
+               'Outputs/ParaViewDataCollection/file_base=OutputData/ParaViewDataCollection'
+    check_files = 'OutputData/ParaViewDataCollection/Run0/Run0.pvd '
+                  'OutputData/ParaViewDataCollection/Run0/Cycle000001/proc000000.vtu '
+                  'OutputData/VisItDataCollection/Run0_000000/mesh.000000 '
+                  'OutputData/VisItDataCollection/Run0_000001/concentration.000000 '
+                  'OutputData/VisItDataCollection/Run0_000001.mfem_root'
+    requirement = 'The system shall have the ability to write out MFEM solutions to '
+                  'ParaView, VisIt and Conduit mfem::DataCollections.'
```
lindsayad, line 15:
```suggestion
                  'ParaView, VisIt and Conduit data collections.'
```

test/tests/mfem/variables/tests
```diff
@@ -1,12 +1,12 @@
...
   design = 'syntax/Problem/index.md'
   issues = '#29632'
-  [./MFEMVariableSetupFromMOOSEVariables]
+  [MFEMVariableSetupFromMOOSEVariables]
     type = XMLDiff
     input = mfem_variables_from_moose.i
     xmldiff = 'OutputData/VariableSetupTest/Run0/Cycle000000/data.pvtu
                 OutputData/VariableSetupTest/Run0/Cycle000001/data.pvtu'
-    requirement = 'MOOSE shall have the ability to add an MFEMVariable from a user specification of a MOOSE variable, inferring the FESpace.'
+    requirement = 'The system shall have the ability to add an MFEMVariable from a user specification of a MOOSE variable, inferring the FESpace.'
```
lindsayad, line 9:
```suggestion
    requirement = 'The system shall have the ability to add an MFEM variable from a user specification of a MOOSE variable, inferring the finite element space.'
```

lindsayad (changes requested):

test/tests/mfem/kernels/tests
nmnobre: > usually I harp on not having implementation details such as "MFEM" in requirements ...

I understand the reasoning, but is that a mandated requisite?

"The software requirements or other life-cycle documentation shall identify operating systems, functions, interfaces, performance requirements, installation considerations, design inputs, and any design constraints of the computer program; specify technical and software-engineering requirements; identify applicable reference drawings, specifications, codes, standards, regulations, procedures, or instructions that establish software design requirement test, inspection, and acceptance criteria, as applicable."

test/tests/mfem/kernels/curlcurl.i
nmnobre: Well... it still exists in history, so it doesn't get removed 😢

> Indeed, `git clone --filter=blob:none` to the rescue perhaps.

test/tests/mfem/kernels/tests
lindsayad: If it was a mandated requisite then I would have to reject instead of reluctantly accept

lindsayad (approved):

test/tests/mfem/kernels/curlcurl.i
lindsayad: this is resolved?

test/tests/mfem/kernels/tests
nmnobre: Fair. I thought of "alternative finite element backend" which would read okay on the kernel tests but I feel doesn't accurately reflect the motivation and, thus, the requirement. If we say "GPU-capable/ready finite element backend" that's probably more accurate and doesn't imply the test itself requires GPU capability but might be a bit cumbersome? We can drop "finite element", so just "GPU-ready backend"? Dunno, some ideas :)

lindsayad: I think mentioning MFEM is the best option

PR #30412, opened 2025-04-28
Ensuring the displaced mesh is re-displaced if adaptivity is run, but mesh is unchanged

### LINKED ISSUE #30390: Displaced problem contains undisplaced mesh on timestep_begin when adaptivity is on
### Bug Description

When performing adaptivity, the displaced mesh is undisplaced so that refinement/coarsening is one-to-one with the reference mesh. However, the next time the mesh is displaced is during residual and Jacobian evaluations. As such, objects executing between the adaptivity step and the nonlinear solve are not taking into account any mesh displacement. This includes `timestepSetup` hooks and user objects, auxkernels, etc. on `timestep_begin`.

### Steps to Reproduce

Here is a very simple input file. Essentially it computes the volume of a 1D mesh with a fixed displacement, which should result in a value of 2. Note that without the displacement, the value is 1.
```
[Mesh]
  [gmg]
    type = GeneratedMeshGenerator
    dim = 1
  []
[]

[Problem]
  solve = false
[]

[GlobalParams]
  displacements = 'disp_x'
[]

[AuxVariables]
  [disp_x]
    [InitialCondition]
      type = FunctionIC
      function = 'x'
    []
  []
[]

[Postprocessors]
  [vol]
    type = VolumePostprocessor
    use_displaced_mesh = true
    execute_on = 'timestep_begin'
  []
[]

[Executioner]
  type = Transient
  num_steps = 5
[]
```

Indeed, the resulting value is 2:
```
Postprocessor Values:
+----------------+----------------+
| time           | vol            |
+----------------+----------------+
|   0.000000e+00 |   0.000000e+00 |
|   1.000000e+00 |   2.000000e+00 |
|   2.000000e+00 |   2.000000e+00 |
|   3.000000e+00 |   2.000000e+00 |
|   4.000000e+00 |   2.000000e+00 |
|   5.000000e+00 |   2.000000e+00 |
+----------------+----------------+
```

However, by simply adding an adaptivity block, which actually isn't doing anything:
```
[Adaptivity]
  steps = 0
  marker = uniform
  [Markers]
    [uniform]
      type = UniformMarker
      mark = DO_NOTHING
    []
  []
[]
```

The volume is erroneously calculated as 1:
```
Postprocessor Values:
+----------------+----------------+
| time           | vol            |
+----------------+----------------+
|   0.000000e+00 |   0.000000e+00 |
|   1.000000e+00 |   2.000000e+00 |
|   2.000000e+00 |   1.000000e+00 |
|   3.000000e+00 |   1.000000e+00 |
|   4.000000e+00 |   1.000000e+00 |
|   5.000000e+00 |   1.000000e+00 |
+----------------+----------------+
```

### Impact

This is especially impactful for contact simulations, which utilize the `NodalArea` user object executing on `timestep_begin`.

I believe the correct solution is to call `DisplacedProblem::updateMesh` after the adaptivity step is complete. This has a slight performance penalty, but it will give the correct displaced mesh for object executed before a timestep's solve. This would also undoubtedly cause test failures that are golded on the current schema.

### [Optional] Diagnostics

_No response_

zachmprince:
## Reason

Closes #30390

## Design

Simple call to to `DisplacedProblem::updateMesh` in `FEProblemBase::adaptMesh` during adaptivity cycle loop when mesh was unchanged. This was necessary since `Adaptivity::adaptMesh` undisplaces the mesh. When a mesh is changed, it goes through the `meshChanged` process, which includes a call to `updateMesh`. But no such call happens when the mesh is unchanged, keeping the mesh in an undisplaced state. This call solves this discrepancy, which was noticed in a contact test that is regolded in this PR.

## Impact

See #30390


lindsayad:
You can mention me once this is ready for review. Unsubscribing for now

zachmprince:
@lindsayad Ready for review

framework/src/problems/FEProblemBase.C
```diff
@@ -7900,6 +7900,11 @@ FEProblemBase::adaptMesh()
     }
     else
     {
+      // If the mesh didn't change, we still need to update the displaced mesh
+      // to undo the undisplacement performed in Adaptivity::adaptMesh
+      if (_displaced_problem)
+        _displaced_problem->updateMesh();
```
lindsayad, line 7906: Did you try putting this in `Adaptivity::adaptMesh` and removing the call to `updateMesh()` in `DisplacedProblem::meshChanged()`? If that didn't break anything, I think that would be my preferred change. A developer who looks only at this code and the lines immediately surrounding it will wonder if/where the re-displacement happens when `mesh_changed_this_step`. It would also conceptually make good sense to undisplace and re-displace in the same method

lindsayad (changes requested):

framework/src/problems/FEProblemBase.C
zachmprince: I tried that at first, but the solutions need to be projected first, which I believe happens in `meshChangedHelper`. I can add it it in `Adaptivity::adaptMesh` for the case when no refinement happens, but the call in `meshChangedHelper()` would need to remain.

lindsayad: Ok that's fine. Let's make sure what you just told me is included in a comment somewhere in the code though

zachmprince: Sounds good. I added an explanation in the doxygen of Adaptivity::adaptMesh.

lindsayad (approved):

PR #30427, opened 2025-04-30
Fix VTK nlohmann json workaround for VTK with external nhlomann json

### LINKED ISSUE #30425: Compilation broken for VTK built with external nlohmann json
### Bug Description

#30362 added a patch for `VTK` built with vendored `nlohmann/json`. The macros added in https://github.com/loganharbour/moose/commit/f47a89cdbc0ea3f454352e2f86bd0fd5703447c2 `#undef` header guard macros which allows VTK to re-include `nlohmann/json`. When VTK is built with vendored `nlohmann/json` it applies horrible macro magic to put symbols in a different namespace, however when built with external  `nlohmann/json` there is no magic to put symbols in a different namespace and hence when `nlohmann/json` is re-included redefinitions occur, causing compilation errors.

I think first importing [`ThirdParty/nlohmannjson/vtk_nlohmannjson.h.in`](https://gitlab.kitware.com/vtk/vtk/-/blob/master/ThirdParty/nlohmannjson/vtk_nlohmannjson.h.in?ref_type=heads) if it exists and then only `#undef`ing the header guard macros if `VTK_MODULE_USE_EXTERNAL_vtknlohmannjson` is  not `1` should fix this problem.

Relevant issue:
- #30372 

### Steps to Reproduce

Attempt to compile MOOSE with VTK built with external `nlohmann/json`.

### Impact

I currently use a workaround:

```bash
export ADDITIONAL_CPPFLAGS="-DMOOSE_VTK_NLOHMANN_INCLUDED"
```

### [Optional] Diagnostics

_No response_

k-collie:
Enables compilation when VTK is built with external nhlomann json.

Closes #30425 

@milljm sorry I saw you assigned yourself to the issue but I thought I'd submit a PR since I think I have it worked out.

I'm currently compiling this on my system, will un-draft when done.

### DIFF
--- a/framework/include/utils/ImageSampler.h
+++ b/framework/include/utils/ImageSampler.h
@@ -30,6 +30,9 @@
 // because OF COURSE NOT, but now we need to take care not to let the
 // include guards prevent them from including their copy with their
 // different namespace.
+#if __has_include("vtk_nlohmannjson.h")
+#include "vtk_nlohmannjson.h"
+#if !(VTK_MODULE_USE_EXTERNAL_vtknlohmannjson == 1)
 #ifndef MOOSE_VTK_NLOHMANN_INCLUDED
 #ifdef INCLUDE_NLOHMANN_JSON_HPP_
 #define MOOSE_ALREADY_INCLUDED_NLOHMANN_JSON_HPP_
@@ -40,6 +43,8 @@
 #undef INCLUDE_NLOHMANN_JSON_FWD_HPP_
 #endif
 #define MOOSE_VTK_NLOHMANN_INCLUDED
+#endif
+#endif
 #endif
 
 #include "vtkSmartPointer.h"
--- a/framework/include/utils/ImageSampler.h
+++ b/framework/include/utils/ImageSampler.h
@@ -40,10 +40,10 @@
 #else
 #define MOOSE_VTK_UNDEF_NLOHMANNJSON_HEADER_GUARDS 1
 #endif // (VTK_MODULE_USE_EXTERNAL_vtknlohmannjson == 1)
-#else // __has_include("vtk_nlohmannjson.h")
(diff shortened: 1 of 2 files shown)


lindsayad (commented):
@roystgnr I believe you were also looking into this?

roystgnr:
I added the `MOOSE_VTK_NLOHMANN_INCLUDED` hackery at the same place as the changes here, but that just handled the case where VTK *doesn't* have an external nlohmann and uses its internal modified copy.  I see how I must have screwed up the case where it does have an unmodified external nlohmann.

roystgnr (approved):

nmnobre:
@k-collie, would you mind fencing this (just one more time xD) with `#if defined __has_include` as recommended [here](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html)? You never know, there are odd compilers.

k-collie:
@nmnobre thanks for the catch, I've tried to handle your exotic compilers without `__has_include` so that they must define `MOOSE_VTK_UNDEF_NLOHMANNJSON_HEADER_GUARDS=0/1`. Just adding the extra fence would've worked for VTK with external nlohmann but not for vendored nlohmann.

I'm definitely not an expert with the preprocessor so happy to hear feedback.

roystgnr (approved):
This is increasingly hideous, so I'd love other ideas on how to improve it, but it still seems like a step forward from the previous incompatible-with-one-VTK-configuration-or-another setups.

loganharbour (approved):
🤮

nmnobre:
Apologies, I was away for a few days.

> Just adding the extra fence would've worked for VTK with external nlohmann but not for vendored nlohmann.

Yes, my idea was simply not to break the existing functionality for those hypothetical compilers, not to provide an alternative flag to workaround their limitations. Thanks a lot for going the extra mile!

Issue #11751, opened 2018-06-26
Extend MOOSE_MPI_COMMAND to other test objects


tophmatthews:
## Rationale

Some cluster setups require the use of `srun` or special `mpirun` to handle executions. `MOOSE_MPI_COMMAND` was implemented way back when to handle this for `RunApp` when back when (https://github.com/idaholab/moose/issues/2676). It should be extended to `test_json.py`, `analyzejacobian.py`, and `bench.py`.

## Description

If `srun` is not run on some machines, moose cannot execute. `MOOSE_MPI_COMMAND` fixes this for `RunApp`, but not all testers.

## Impact

Makes the TestHarness applicable to more machines.

permcody:
Tag @milljm

permcody:
I'm not sure that the PostCommand stuff needs to worry about the mpi command. We already run "exodiff" after many of our tests and don't need to do anything special. Are you saying that for slurm, we always need to worry about srun?

milljm:
Looks like I was tagged, but never looked into this... sorry!

I'll try and address this... I also want to link in an older googlegroups thread. Its a bit lengthy, and I think we narrowed it down to a rather nasty issue I am not sure the TestHarness is prepared to solve:

[Installation on Linux Clusters, wanting to use Phase Field module](https://groups.google.com/forum/#!searchin/moose-users/Installation$20on$20Linux$20Clusters$2C$20wanting$20to$20use$20Phase$20Field$20module%7Csort:date/moose-users/tnLwKx_n7PQ/Km50SMn5AwAJ)

Long story short, we found that:

```bash
srun --nodes 1 --mincpus 4 -p branicio -t 2:00:00 mpiexec -n 2 /bin/false
echo $?
```

returns 0

So asking the TestHarness to `srun` any application is not going to be 100% accurate. Everything passes. The tests that found this, were none-other than the tests the TestHarness knows is supposed to fail (way to go RunException).

Maybe you can verify if perhaps, that email thread is an isolated incident and the above srun command indeed returns a non-zero?

jessecarterMOOSE:
@milljm I just ran into this issue on our cluster. FWIW I'm trying to run with `MOOSE_MPI_COMMAND=mpiexec`. For me, not including this causes the application to hang. So any test that launches a process which in turns runs MOOSE times out.

milljm:
Going to link #27561, as I think this PR, once merged, will fix issues involving Slurm :)

[...]

PR #30457, opened 2025-05-06
Mpi exec test 11751

tophmatthews:
ref #11751


tophmatthews:
@milljm do I have to remove `pyhit/mooseutils/mooseutils.py` from `wasp/meta.yaml' too?

PR #30472, opened 2025-05-08
Modify the GradientOperator.h for SolidMechanics App

hdt5kt:
## Reason

Bug fix for the GradientOperator.h file

## Design

Change GradOp and addGradOp in GradientOperator.h to be public and shared across all instances.

## Impact

No


hugary1995:
I think the precheck will yell at you for not referencing an existing issue. You'll need to create an issue describing the bug then modify the commit message to reference that issue.

hugary1995:
This isn't needed.

PR #30502, opened 2025-05-13
templated AD and nonAD nodeElemConstraint

### LINKED ISSUE #30499: AD version of node element constraints
### Motivation

We have a Node element constraint for rebar pullout from concrete.  The jacobian for this model is not exactly correct.  It would be nice to test this against an AD version of the rebar pullout model.

### Design

Follow the design used in IntegratedBC to create an AD version of nodeElemConstraint that can be templated on is_ad.

### Impact

We should be able to take larger time steps with bond slip constraints.

lynnmunday:
closes #30499

### DIFF
--- a/framework/doc/content/source/constraints/ADEqualValueEmbeddedConstraint.md
+++ b/framework/doc/content/source/constraints/ADEqualValueEmbeddedConstraint.md
@@ -5,7 +5,7 @@
 This version of `EqualValueEmbeddedConstraint` uses automatic differentiation, see [EqualValueEmbeddedConstraint.md] for information on formulation.
 
 !alert warning title=`ADEqualValueEmbeddedConstraint` only works for PENALTY formulation
-Automatic differentation cannot be used to compute Jacobian terms for [!param](/Constraints/ADEqualValueEmbeddedConstraint/formulation) = KINEMATIC because the KINEMATIC implementation enforces the constraint by enforcing the residual value on the secondary node prior to application of the constarint, $r_{s,copy}$ in [!eqref](EqualValueEmbeddedConstraint.md#eqn:kinematic).  This secondary node residual value is a nonAD copy of the residual and cannot be used to compute derivatives for the Jacobian.
+Automatic differentation cannot be used to compute Jacobian terms for [!param](/Constraints/ADEqualValueEmbeddedConstraint/formulation) = KINEMATIC because the KINEMATIC implementation enforces the constraint by enforcing the residual value on the secondary node prior to application of the constraint, $r_{s,copy}$ in [!eqref](EqualValueEmbeddedConstraint.md#eqn:kinematic).  This secondary node residual value is a nonAD copy of the residual and cannot be used to compute derivatives for the Jacobian.
 
 !syntax parameters /Constraints/ADEqualValueEmbeddedConstraint
 
--- a/framework/doc/content/source/constraints/ADEqualValueEmbeddedConstraint.md
+++ b/framework/doc/content/source/constraints/ADEqualValueEmbeddedConstraint.md
@@ -1,0 +1,14 @@
+# ADEqualValueEmbeddedConstraint
+
+!syntax description /Constraints/ADEqualValueEmbeddedConstraint
+
+This version of `EqualValueEmbeddedConstraint` uses automatic differentiation, see [EqualValueEmbeddedConstraint.md] for information on formulation.
+
+!alert warning title=`ADEqualValueEmbeddedConstraint` only works for PENALTY formulation
+Automatic differentation cannot be used to compute Jacobian terms for [!param](/Constraints/ADEqualValueEmbeddedConstraint/formulation) = KINEMATIC because the KINEMATIC implementation enforces the constraint by enforcing the residual value on the secondary node prior to application of the constarint, $r_{s,copy}$ in [!eqref](EqualValueEmbeddedConstraint.md#eqn:kinematic).  This secondary node residual value is a nonAD copy of the residual and cannot be used to compute derivatives for the Jacobian.
+
+!syntax parameters /Constraints/ADEqualValueEmbeddedConstraint
+
+!syntax inputs /Constraints/ADEqualValueEmbeddedConstraint
+
+!syntax children /Constraints/ADEqualValueEmbeddedConstraint
--- a/framework/include/constraints/ADNodeElemConstraint.h
+++ b/framework/include/constraints/ADNodeElemConstraint.h
@@ -14,9 +14,9 @@
 #include "NeighborCoupleableMooseVariableDependencyIntermediateInterface.h"
 
 /**
- * A ADNodeElemConstraint is used when you need to create constraints between
- * a secondary node and a primary element. It works by allowing you to modify the
- * residual and jacobian entries on the secondary node and the primary element.
+ * An ADNodeElemConstraint is used when you need to create constraints between a subdomain of
+ * secondary nodes and a subdomain of primary elements. It works by allowing you to modify the
+ * residual and jacobian entries on the secondary nodes and the primary elements.
  */
 class ADNodeElemConstraint : public NodeElemConstraintBase
 {
@@ -25,28 +25,25 @@ public:
 
   ADNodeElemConstraint(const InputParameters & parameters);
 
-  /// Computes the residual Nodal residual.
+  /// Computes the Nodal residual.
   virtual void computeResidual() override;
 
   /// Computes the jacobian for the current element.
   virtual void computeJacobian() override;
 
 protected:
-  /// prepare the _secondary_to_primary_map
+  /// prepare the _secondary_to_primary_map (see NodeElemConstraintBase)
   virtual void prepareSecondaryToPrimaryMap() = 0;
 
-  /// This is the virtual that derived classes should override for computing the residual.
+  /// This is the virtual method that derived classes should override for computing the residual.
   virtual ADReal computeQpResidual(Moose::ConstraintType type) = 0;
 
   /// This is the virtual that derived classes should override for computing the Jacobian.
-  virtual Real computeQpJacobian(Moose::ConstraintJacobianType /*type*/) { return 0; }
+  virtual Real computeQpJacobian(Moose::ConstraintJacobianType /*type*/);
 
   /// This is the virtual that derived classes should override for computing the off-diag Jacobian.
   virtual Real computeQpOffDiagJacobian(Moose::ConstraintJacobianType /*type*/,
-                                        unsigned int /*jvar*/)
-  {
-    return 0;
-  }
+                                        unsigned int /*jvar*/);
 
   /// Holds the current solution at the current quadrature point
   const ADVariableValue & _u_primary;
--- a/framework/include/constraints/ADNodeElemConstraint.h
+++ b/framework/include/constraints/ADNodeElemConstraint.h
@@ -10,33 +10,21 @@
 #pragma once
 
 // MOOSE includes
-#include "Constraint.h"
+#include "NodeElemConstraintBase.h"
 #include "NeighborCoupleableMooseVariableDependencyIntermediateInterface.h"
-
-// libMesh forward declarations
-namespace libMesh
-{
-template <typename T>
-class SparseMatrix;
-}
 
 /**
  * A ADNodeElemConstraint is used when you need to create constraints between
  * a secondary node and a primary element. It works by allowing you to modify the
  * residual and jacobian entries on the secondary node and the primary element.
  */
-class ADNodeElemConstraint : public Constraint,
-                           public NeighborCoupleableMooseVariableDependencyIntermediateInterface,
-                           public NeighborMooseVariableInterface<Real>
+class ADNodeElemConstraint : public NodeElemConstraintBase
 {
 public:
   static InputParameters validParams();
 
   ADNodeElemConstraint(const InputParameters & parameters);
   virtual ~ADNodeElemConstraint();
-
-  /// Compute the value the secondary node should have at the beginning of a timestep.
-  void computeSecondaryValue(NumericVector<Number> & current_solution);
 
   /// Computes the residual Nodal residual.
   virtual void computeResidual() override;
@@ -50,43 +38,9 @@ public:
   /// Gets the indices for all dofs connected to the constraint
   virtual void getConnectedDofIndices(unsigned int var_num);
 
-  /**
-   * Whether or not this constraint should be applied.
-   * @return bool true if this constraint is active, false otherwise
-   */
-  virtual bool shouldApply() { return true; }
-
-  /**
-   * Whether or not the secondary's residual should be overwritten.
-   * @return bool When this returns true the secondary's residual as computed by the constraint will
-   * _replace_ the residual previously at that node for that variable.
-   */
-  virtual bool overwriteSecondaryResidual();
-
-  /**
-   * Whether or not the secondary's Jacobian row should be overwritten.
-   * @return bool When this returns true the secondary's Jacobian row as computed by the constraint
-   * will _replace_ the residual previously at that node for that variable.
-   */
-  virtual bool overwriteSecondaryJacobian() { return overwriteSecondaryResidual(); };
-
-  /**
-   * The variable on the primary elem.
-   * @return MooseVariable & a reference to the primary variable
-   */
-  virtual MooseVariable & primaryVariable() { return _primary_var; }
-
-  /**
-   * The variable number that this object operates on.
-   */
-  const MooseVariable & variable() const override { return _var; }
-
 protected:
   /// prepare the _secondary_to_primary_map
   virtual void prepareSecondaryToPrimaryMap() = 0;
-
-  /// Compute the value the secondary node should have at the beginning of a timestep.
-  virtual Real computeQpSecondaryValue() = 0;
 
   /// This is the virtual that derived classes should override for computing the residual.
   virtual ADReal computeQpResidual(Moose::ConstraintType type) = 0;
@@ -183,8 +137,6 @@ protected:
   /// primary block id
   unsigned short _primary;
 
-  MooseVariable & _var;
-
   const MooseArray<ADPoint> & _primary_q_point;
   const QBase * const & _primary_qrule;
 
@@ -200,9 +152,6 @@ protected:
   VariablePhiValue _phi_secondary;
   /// Shape function on the secondary side.  This will always only have one entry and that entry will always be "1"
   VariableTestValue _test_secondary;
-
-  /// Primary side variable
-  MooseVariable & _primary_var;
 
   /// Number for the primary variable
   unsigned int _primary_var_num;
@@ -232,24 +181,7 @@ protected:
   /// maps secondary node ids to primary element ids
   std::map<dof_id_type, dof_id_type> _secondary_to_primary_map;
 
-  /**
-   * Whether or not the secondary's residual should be overwritten.
-   *
-   * When this is true the secondary's residual as computed by the constraint will _replace_
-   * the residual previously at that node for that variable.
-   */
-  bool _overwrite_secondary_residual;
-
   /// Data members for holding residuals
   ADReal _r;
   std::vector<Real> _residuals;
-
-public:
-  SparseMatrix<Number> * _jacobian;
-  /// dofs connected to the secondary node
-  std::vector<dof_id_type> _connected_dof_indices;
-  /// stiffness matrix holding primary-secondary jacobian
-  DenseMatrix<Number> _Kne;
-  /// stiffness matrix holding secondary-secondary jacobian
-  DenseMatrix<Number> _Kee;
 };
--- a/framework/include/constraints/ADNodeElemConstraint.h
+++ b/framework/include/constraints/ADNodeElemConstraint.h
@@ -91,6 +91,16 @@ protected:
   /// This is the virtual that derived classes should override for computing the residual.
   virtual ADReal computeQpResidual(Moose::ConstraintType type) = 0;
 
+  /// This is the virtual that derived classes should override for computing the Jacobian.
+  virtual Real computeQpJacobian(Moose::ConstraintJacobianType /*type*/) { return 0; }
+
+  /// This is the virtual that derived classes should override for computing the off-diag Jacobian.
+  virtual Real computeQpOffDiagJacobian(Moose::ConstraintJacobianType /*type*/,
+                                        unsigned int /*jvar*/)
+  {
+    return 0;
+  }
+
   // coupling interface:
   virtual const VariableValue & coupledSecondaryValue(const std::string & var_name,
                                                       unsigned int comp = 0)
--- a/framework/include/constraints/ADNodeElemConstraint.h
+++ b/framework/include/constraints/ADNodeElemConstraint.h
@@ -38,10 +38,10 @@ protected:
   /// This is the virtual method that derived classes should override for computing the residual.
   virtual ADReal computeQpResidual(Moose::ConstraintType type) = 0;
 
-  /// This is the virtual that derived classes should override for computing the Jacobian.
+  /// This is the virtual method that derived classes should override for computing the Jacobian.
   virtual Real computeQpJacobian(Moose::ConstraintJacobianType /*type*/);
 
-  /// This is the virtual that derived classes should override for computing the off-diag Jacobian.
+  /// This is the virtual method that derived classes should override for computing the off-diag Jacobian.
   virtual Real computeQpOffDiagJacobian(Moose::ConstraintJacobianType /*type*/,
                                         unsigned int /*jvar*/);
 
--- a/framework/include/constraints/ADNodeElemConstraint.h
+++ b/framework/include/constraints/ADNodeElemConstraint.h
@@ -24,7 +24,6 @@ public:
   static InputParameters validParams();
 
   ADNodeElemConstraint(const InputParameters & parameters);
-  virtual ~ADNodeElemConstraint();
 
   /// Computes the residual Nodal residual.
   virtual void computeResidual() override;
@@ -33,10 +32,7 @@ public:
   virtual void computeJacobian() override;
 
   /// Computes d-residual / d-jvar...
-  virtual void computeOffDiagJacobian(unsigned int jvar) override;
-
-  /// Gets the indices for all dofs connected to the constraint
-  virtual void getConnectedDofIndices(unsigned int var_num);
+  virtual void computeOffDiagJacobian(unsigned int /*jvar*/) override {};
 
 protected:
   /// prepare the _secondary_to_primary_map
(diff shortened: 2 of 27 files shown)


framework/src/constraints/EqualValueEmbeddedConstraint.C
```diff
@@ -41,20 +44,23 @@ EqualValueEmbeddedConstraint::validParams()
...
-  : NodeElemConstraint(parameters),
+template <bool is_ad>
+EqualValueEmbeddedConstraintTempl<is_ad>::EqualValueEmbeddedConstraintTempl(
+    const InputParameters & parameters)
+  : GenericNodeElemConstraint<is_ad>(parameters),
     _displaced_problem(parameters.get<FEProblemBase *>("_fe_problem_base")->getDisplacedProblem()),
     _fe_problem(*parameters.get<FEProblem *>("_fe_problem")),
-    _formulation(getParam<MooseEnum>("formulation").getEnum<Formulation>()),
-    _penalty(getParam<Real>("penalty")),
+    _formulation(this->template getParam<MooseEnum>("formulation").template getEnum<Formulation>()),
```
maxnezdyur, line 53: Might have missed it, but for AD do you have an error for the kinematic formulation? Or am I wrong and the kinematic form works with AD?

maxnezdyur (changes requested):
Have a few notes of things I'd like I'm not sure about. But in general, seems like this will make it easy to have templated nodeelem constaints for the future. Not for this PR, but I think this would be useful for other constraint classes too.

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,62 @@
...
+
+  /// This is the virtual that derived classes should override for computing the Jacobian.
+  virtual Real computeQpJacobian(Moose::ConstraintJacobianType /*type*/) { return 0; }
+
+  /// This is the virtual that derived classes should override for computing the off-diag Jacobian.
+  virtual Real computeQpOffDiagJacobian(Moose::ConstraintJacobianType /*type*/,
+                                        unsigned int /*jvar*/)
+  {
+    return 0;
+  }
```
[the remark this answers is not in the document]
lynnmunday: Yes, EqualValueEmbeddedConstraint won't build without this.  @GiudGiud said I would need some kind of partial specialization to only implement the nonAD version of these.

framework/src/constraints/EqualValueEmbeddedConstraint.C
lynnmunday: It works with AD.  i'll add the ad kinematic tests.

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,62 @@
...
+  ADNodeElemConstraint(const InputParameters & parameters);
+
+  /// Computes the residual Nodal residual.
+  virtual void computeResidual() override;
+
+  /// Computes the jacobian for the current element.
+  virtual void computeJacobian() override;
+
+  /// Computes d-residual / d-jvar...
+  virtual void computeOffDiagJacobian(unsigned int /*jvar*/) override {};
```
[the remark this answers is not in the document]
lynnmunday: You're right, I removed it.

framework/src/constraints/EqualValueEmbeddedConstraint.C
lynnmunday: I was wrong, AD with Kinematic constraints do not work.  The test is using PJFNK so it didn't need an accurate jacobian to pass.  I changed it to Newton and it fails.  I'll add an error for using kinematic with AD.

framework/doc/content/source/constraints/ADEqualValueEmbeddedConstraint.md
```diff
@@ -0,0 +1,14 @@
+# ADEqualValueEmbeddedConstraint
+
+!syntax description /Constraints/ADEqualValueEmbeddedConstraint
+
+This version of `EqualValueEmbeddedConstraint` uses automatic differentiation, see [EqualValueEmbeddedConstraint.md] for information on formulation.
+
+!alert warning title=`ADEqualValueEmbeddedConstraint` only works for PENALTY formulation
+Automatic differentation cannot be used to compute Jacobian terms for [!param](/Constraints/ADEqualValueEmbeddedConstraint/formulation) = KINEMATIC because the KINEMATIC implementation enforces the constraint by enforcing the residual value on the secondary node prior to application of the constarint, $r_{s,copy}$ in [!eqref](EqualValueEmbeddedConstraint.md#eqn:kinematic).  This secondary node residual value is a nonAD copy of the residual and cannot be used to compute derivatives for the Jacobian.
```
GiudGiud, line 8:
```suggestion
Automatic differentation cannot be used to compute Jacobian terms for [!param](/Constraints/ADEqualValueEmbeddedConstraint/formulation) = KINEMATIC because the KINEMATIC implementation enforces the constraint by enforcing the residual value on the secondary node prior to application of the constraint, $r_{s,copy}$ in [!eqref](EqualValueEmbeddedConstraint.md#eqn:kinematic).  This secondary node residual value is a nonAD copy of the residual and cannot be used to compute derivatives for the Jacobian.
```

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,59 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+// MOOSE includes
+#include "NodeElemConstraintBase.h"
+#include "NeighborCoupleableMooseVariableDependencyIntermediateInterface.h"
+
+/**
+ * A ADNodeElemConstraint is used when you need to create constraints between
```
GiudGiud, line 17:
```suggestion
 * An ADNodeElemConstraint is used when you need to create constraints between
```

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,59 @@
...
+ * a secondary node and a primary element. It works by allowing you to modify the
```
GiudGiud, line 18:
```suggestion
 * a subdomain of secondary nodes and a subdomain of primary elements. It works by allowing you to modify the
```

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,59 @@
...
+ * residual and jacobian entries on the secondary node and the primary element.
```
GiudGiud, line 19:
```suggestion
 * residual and jacobian entries on the secondary nodes and the primary elements.
```

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,59 @@
...
+ */
+class ADNodeElemConstraint : public NodeElemConstraintBase
+{
+public:
+  static InputParameters validParams();
+
+  ADNodeElemConstraint(const InputParameters & parameters);
+
+  /// Computes the residual Nodal residual.
```
GiudGiud, line 28: ?

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,59 @@
...
+  virtual void computeResidual() override;
+
+  /// Computes the jacobian for the current element.
+  virtual void computeJacobian() override;
+
+protected:
+  /// prepare the _secondary_to_primary_map
+  virtual void prepareSecondaryToPrimaryMap() = 0;
+
+  /// This is the virtual that derived classes should override for computing the residual.
```
GiudGiud, line 38:
```suggestion
  /// This is the virtual method that derived classes should override for computing the residual.
```

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,59 @@
...
+  virtual ADReal computeQpResidual(Moose::ConstraintType type) = 0;
+
+  /// This is the virtual that derived classes should override for computing the Jacobian.
+  virtual Real computeQpJacobian(Moose::ConstraintJacobianType /*type*/) { return 0; }
```
GiudGiud, line 42:
```suggestion
  virtual Real computeQpJacobian(Moose::ConstraintJacobianType /*type*/) { mooseAssert(false, "Should not be used"); }
```

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,59 @@
...
+
+  /// This is the virtual that derived classes should override for computing the off-diag Jacobian.
+  virtual Real computeQpOffDiagJacobian(Moose::ConstraintJacobianType /*type*/,
+                                        unsigned int /*jvar*/)
+  {
+    return 0;
```
GiudGiud, line 48:
```suggestion
    mooseAssert(false, "Should not be used");
```

framework/include/constraints/EqualValueEmbeddedConstraint.h
```diff
@@ -54,9 +55,14 @@ class EqualValueEmbeddedConstraint : public NodeElemConstraint
   /// Formulations, currently only supports KINEMATIC and PENALTY
   const enum class Formulation { KINEMATIC, PENALTY } _formulation;
   /// Penalty parameter used in constraint enforcement for kinematic and penalty formulations
-  const Real _penalty;
+  const GenericReal<is_ad> _penalty;
```
GiudGiud, line 58: if ADEqual...Constaint does not work for kinematic, isnt this always a constant?

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+namespace libMesh
+{
+template <typename T>
+class SparseMatrix;
+}
+
+/**
+ * A NodeElemConstraintBase is used when you need to create constraints between
+ * a secondary node and a primary element. It works by allowing you to modify the
+ * residual and jacobian entries on the secondary node and the primary element.
```
GiudGiud, line 25:
```suggestion
 * A NodeElemConstraintBase is a base class for constraints between a subdomain with
 * secondary nodes and a subdomain with primary elements. It works by allowing you to modify the
 * residual and jacobian entries on the secondary nodes and the primary elements.
```

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+  std::map<dof_id_type, dof_id_type> _secondary_to_primary_map;
+
+  /**
+   * Whether or not the secondary's residual should be overwritten.
+   *
+   * When this is true the secondary's residual as computed by the constraint will _replace_
+   * the residual previously at that node for that variable.
+   */
+  bool _overwrite_secondary_residual;
+  // fixme move
```
GiudGiud, line 112: let s move it

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+public:
+  SparseMatrix<Number> * _jacobian;
```
GiudGiud, line 114: docstirng

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+  /// Shape function on the secondary side.  This will always only have one entry and that entry will always be "1"
+  VariableTestValue _test_secondary;
+
+  /// Side shape function.
+  const VariablePhiValue & _phi_primary;
+
+  /// Side test function
+  const VariableTestValue & _test_primary;
+
+  const std::map<dof_id_type, std::vector<dof_id_type>> & _node_to_elem_map;
```
GiudGiud, line 100: docstring
is this a local one or is this the one from the MooseMesh

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+   * @return bool true if this constraint is active, false otherwise
+   */
+  virtual bool shouldApply() { return true; }
+
+  /**
+   * Whether or not the secondary's residual should be overwritten.
+   * @return bool When this returns true the secondary's residual as computed by the constraint will
+   * _replace_ the residual previously at that node for that variable.
+   */
+  virtual bool overwriteSecondaryResidual();
```
GiudGiud, line 52:
```suggestion
  virtual bool overwriteSecondaryResidual() const;
```

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+   * @return bool When this returns true the secondary's Jacobian row as computed by the constraint
+   * will _replace_ the residual previously at that node for that variable.
+   */
+  virtual bool overwriteSecondaryJacobian() { return overwriteSecondaryResidual(); };
+
+  /**
+   * The variable on the primary elem.
+   * @return MooseVariable & a reference to the primary variable
+   */
+  virtual MooseVariable & primaryVariable() { return _primary_var; }
```
GiudGiud, line 65:
```suggestion
  virtual const MooseVariable & primaryVariable() const { return _primary_var; }
```

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * The variable number that this object operates on.
+   */
+  const MooseVariable & variable() const override { return _var; }
+
+protected:
+  /// Compute the value the secondary node should have at the beginning of a timestep.
+  virtual Real computeQpSecondaryValue() = 0;
+
+  MooseVariable & _var;
```
GiudGiud, line 76: docstring

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,121 @@
...
+  MooseVariable & _primary_var;
+
+  /// secondary block id
+  unsigned short _secondary;
+  /// primary block id
+  unsigned short _primary;
+
+  /// current node being processed
+  const Node * const & _current_node;
+  const Elem * const & _current_elem;
```
GiudGiud, line 87:
```suggestion
  /// current element being processed
  const Elem * const & _current_elem;
```

framework/src/constraints/ADNodeElemConstraint.C
```diff
@@ -0,0 +1,78 @@
...
+#include "MooseMesh.h"
+#include "MooseTypes.h"
+
+#include "libmesh/string_to_enum.h"
+
+InputParameters
+ADNodeElemConstraint::validParams()
+{
+  InputParameters params = NodeElemConstraintBase::validParams();
+  return params;
```
GiudGiud, line 24: missing class description?

framework/src/constraints/ADNodeElemConstraint.C
```diff
@@ -0,0 +1,78 @@
...
+ADNodeElemConstraint::computeJacobian()
+{
+  _qp = 0;
+
+  std::vector<ADReal> primary_residual(_test_primary.size(), 0);
+
+  for (_i = 0; _i < _test_primary.size(); _i++)
+    primary_residual[_i] += computeQpResidual(Moose::Primary);
+
+  addJacobian(_assembly, primary_residual, _primary_var.dofIndicesNeighbor(), _var.scalingFactor());
```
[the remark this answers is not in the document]
GiudGiud: which equation is it? it s the scaling factor for that equation

since it s the dofs of the primary, should be the scaling factor for the primary

framework/src/constraints/NodeElemConstraintBase.C
```diff
@@ -0,0 +1,76 @@
...
+void
+NodeElemConstraintBase::computeSecondaryValue(NumericVector<Number> & current_solution)
+{
+  const dof_id_type & dof_idx = _var.nodalDofIndex();
+  _qp = 0;
+  current_solution.set(dof_idx, computeQpSecondaryValue());
+}
+
+bool
+NodeElemConstraintBase::overwriteSecondaryResidual()
```
GiudGiud, line 73:
```suggestion
NodeElemConstraintBase::overwriteSecondaryResidual() const
```

GiudGiud (changes requested):
most comments apply to things prior to this PR but it would be nice to get this class up to code

framework/src/constraints/ADNodeElemConstraint.C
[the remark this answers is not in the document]
maxnezdyur: This is how the contributions are added before the AD change. It uses the _var for both. But I don't know if that's incorrect.
https://github.com/idaholab/moose/blob/122840c1fdf80ef78a179bb019702c45924ec46d/framework/src/systems/NonlinearSystemBase.C#L2601-L2613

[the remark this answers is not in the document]
GiudGiud: it's wrong there too imo.
And i see the same-ish error 300 lines above for NodeFaceConstraints.

@lindsayad thoughts? Are we wrong?

framework/include/constraints/NodeElemConstraintBase.h
lynnmunday: MooseMesh

framework/src/constraints/ADNodeElemConstraint.C
lynnmunday: Do I need one if this class can't be in the input file?

GiudGiud: ah no you're good

framework/src/constraints/ADNodeElemConstraint.C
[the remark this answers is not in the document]
lindsayad: No, you're not wrong. Looks like a bug to me for sure

framework/include/constraints/EqualValueEmbeddedConstraint.h
lynnmunday: penalty is constant.  NonAD will work for kinematic, AD won't.

framework/src/constraints/ADNodeElemConstraint.C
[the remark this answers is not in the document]
lynnmunday: https://github.com/idaholab/moose/issues/30530

I made a two variable test and it doesn't work for this constraint so I added an error to this PR and an issue:  https://github.com/idaholab/moose/issues/30530

I didn't try nodeFaceConsraint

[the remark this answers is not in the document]
GiudGiud: @lynnmunday can you please also fix the mistake in the NonlinearSystem that Max pointed out? and same mistake about 300 lines above for NodeFace constratins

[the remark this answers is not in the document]
lynnmunday: @GiudGiud I changed those two lines in NonlinearSystem base but they do not fix the convergence problem in issue #30530    

I did add scaling to the two variable nodeElem constraint problem and it will fail without fixing the scaling in ADNodeElemConstraint so that is working.

lynnmunday:
@GiudGiud I addressed all of the comments in your review.  Could you take another look and merge if its good?

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,56 @@
...
+public:
+  static InputParameters validParams();
+
+  ADNodeElemConstraint(const InputParameters & parameters);
+
+  /// Computes the Nodal residual.
+  virtual void computeResidual() override;
+
+  /// Computes the jacobian for the current element.
+  virtual void computeJacobian() override;
```
GiudGiud, line 32: can we check this
Nodal residual and element Jacobian?

maxnezdyur (approved):
Fixed all my comments. Like I said in one of the comments, would be nice to get this done for the other constraints but now we have a model of how to do that efficiently!

framework/include/constraints/ADNodeElemConstraint.h
```diff
@@ -0,0 +1,56 @@
...
+
+protected:
+  /// prepare the _secondary_to_primary_map
```
GiudGiud, line 35:
```suggestion
  /// prepare the _secondary_to_primary_map (see NodeElemConstraintBase)
```

framework/include/constraints/NodeElemConstraintBase.h
```diff
@@ -0,0 +1,125 @@
...
+  unsigned short _primary;
+
+  /// current node being processed
+  const Node * const & _current_node;
+  /// current element being processed
+  const Elem * const & _current_elem;
+
+  /// Shape function on the secondary side.
+  VariablePhiValue _phi_secondary;
+  /// Shape function on the secondary side.  This will always only have one entry and that entry will always be "1"
```
GiudGiud, line 93: I think this is only true for lagrange variables?

framework/src/constraints/EqualValueEmbeddedConstraint.C
```diff
@@ -41,20 +43,25 @@ EqualValueEmbeddedConstraint::validParams()
...
+    _penalty(this->template getParam<Real>("penalty")),
     _residual_copy(_sys.residualGhosted())
 {
   _overwrite_secondary_residual = false;
   prepareSecondaryToPrimaryMap();
+  if (is_ad && (_formulation == Formulation::KINEMATIC))
```
GiudGiud, line 58:
```suggestion
  if (constexpr is_ad && (_formulation == Formulation::KINEMATIC))
```

test/tests/constraints/equal_value_embedded_constraint/tests
```diff
@@ -119,8 +110,95 @@
...
+                  "differentiation and penalty constraint and it will produce the same "
+                  "results as a single variable simulation."
+    issues = '#30499'
+    type = RunApp
+    input = 'two_var_embedded_constraint.i'
+    mesh_mode = REPLICATED
+  []
+  [error_two_var_AD]
+    design = 'source/constraints/ADEqualValueEmbeddedConstraint.md'
+    requirement = "The system shall error when the nonAD constraint is used for coupling two variables."
```
GiudGiud, line 196: does it converge with PJFNK ? is the Jacobian is not 100% correct witohut AD, makes sense that it does not converge with Newton

lynnmunday: I didn't try PJFNK.  The Jacobian is wrong but I didn't look at how wrong.  The residual diverges.

GiudGiud: Let's try it. These constraints were meant to be used with PJFNK since the Jacobians were not correct

framework/include/constraints/NodeElemConstraintBase.h
lynnmunday: It is the shape function for a single node that is the secondary "surface".  It is not element connected to the node.

lynnmunday:
@GiudGiud Its ready for another review.

test/tests/constraints/equal_value_embedded_constraint/tests
lynnmunday: We had assumed the nonAD version of these constraints had a correct Jacobian.  We didn't know it was incorrect until I added the test in this PR that coupled two different variables.  The nonAD version of this constraint works with Newton when this constraint is coupling the same variable.

GiudGiud: >  It does not work for kinematic and PJFNK

It worked fine on the test case. you just need to adjust the scaling.
But you're right, the jacobian is terribly wrong. 

We need an issue for this. 
I m fine with the error

framework/include/constraints/NodeElemConstraintBase.h
GiudGiud: yes but that shape function is set to be 1 because that's the value of the lagrange shape function at that node

lynnmunday: This is hardcoding the shape function for a node.  I think this has more to do with setting the secondary_primary lagrange multiplier than the type of shape function.  I don't think this is right for elements with dofs that aren't located at physical coordinates, I don't know what that would do.  This was already in the code before I started this PR to add AD so this is the behavior people are currently using.

test/tests/constraints/equal_value_embedded_constraint/tests
lynnmunday: You are right, I purposefully put in bad manual scaling to highlight the issue.  It would converge without the bad scaling.  I added an issue that links to this PR and test.

lynnmunday:
@GiudGiud I addressed all of your comments, could you merge this?

test/tests/constraints/equal_value_embedded_constraint/tests
GiudGiud: ok found one error which prevents getting the correct Jacobian
for off-diag on primary side, we are actually using the wrong dofs (sending to secondary side)
should be
```

                  // Cache the jacobian block for the primary side
                  nec->addJacobian(_fe_problem.assembly(0, number()),
                                   nec->_Kne,
                                   nec->primaryVariable().dofIndicesNeighbor(),
                                   nec->_connected_dof_indices,
                                   nec->primaryVariable().scalingFactor());
```

but it s not the only problem.
the getConnectedDofs is not finding any connected DOFs for the primary variables

lynnmunday: how did you find that, are stepping through this in gdb?

GiudGiud: the node_to_elem maps, is not going to get elements that live on the primary side? 
I think we need more logic in getConnectedDofs to go grab those

framework/include/constraints/NodeElemConstraintBase.h
GiudGiud: yeah I m thinking we should be limiting the FE families allowed for the variables here. ANother PR

test/tests/constraints/equal_value_embedded_constraint/tests
```diff
@@ -119,8 +110,95 @@
...
+    issues = '#30499'
```
GiudGiud, line 197: @lynnmunday just adding the new issue there is the last item

test/tests/constraints/equal_value_embedded_constraint/tests
lindsayad: why is this test problem so big 😦

GiudGiud (approved):
Thanks for the contribution!

test/tests/constraints/equal_value_embedded_constraint/tests
GiudGiud: done

hugary1995:
Could you leave me with one more day to review this? If I don't get back just ignore me.

GiudGiud:
sure! no rush

lindsayad:
@GiudGiud I'm pretty sure you just overwrote some of @lynnmunday's commits

lindsayad:
It looks like you dropped

* lynnmunday/moose@5228f6326d 2025-05-20 Lynn Munday fixing constexpr if statement
* lynnmunday/moose@63d53b18cb 2025-05-20 Lynn Munday Apply suggestions from code review

GiudGiud:
I squashed the "address review comment" and the "fix constexpr" ones with the commit they pertained to

hugary1995:
I have a high-level question before I start my review. What's the motivation of the AD constraint?

By reading the issue description, it seems to me that the motivation was that the non-AD version has wrong Jacobian if we try to tie two different variables. But in the last commits by @lindsayad , it seems like that problem has been resolved -- correct me if I interpreted it wrong.

In this comment https://github.com/idaholab/moose/pull/30502#issuecomment-2894659091, @GiudGiud mentioned that AD might be useful for complicated cases. But I fail to come up with such a case where the non-AD constraint would not work (assuming its Jacobian has been fixed). The variables being constrained are independent, nonlinear variables, so I don't see much value in adding an AD version here. Am I missing something?

hugary1995:
I see. There you compute the bond stress for the constraint, which may have dual numbers from arbitrary variables. I agree AD would be useful then. Let me finish my review real quick.

hugary1995:
My elem-elem constraint stuff is just another way of enforcing the constraint. I haven't implemented bond-slip constraint using it. Speaking of Jacobian, I expect it to face the exact same issues as those you are facing here.

lynnmunday:
@GiudGiud and @lindsayad Thanks for fixing this.

lynnmunday:
@GiudGiud Can you merge this?

hugary1995 (approved):

Issue #4325, opened 2014-12-02
abnormal termination return codes


alfoa:
It looks like that the "return codes" that MOOSE (or maybe LibMesh/PTSC) returns when an abnormal termination happens (no convergence, etc) are often == 0. So fare 3 different situations have been proved to not return a code =/ that 0:
- failure for reaching "dt" min;
- failure for pivot zero detected;
- when the code exits using the method "MooseError";
  Hopefully it is an easy fix.
  Thank you all.

lindsayad:
We actually explicitly test that a failed solve yields a 0 exit code in https://github.com/idaholab/moose/blob/f346f789ca40a343cbf1fa61eb225a3d491e98b6/test/tests/misc/check_error/tests#L5-L14

IMO we shouldn't return 0 when the executioner/solver fails. Pinging @YaqiWang who just triggered my search through the issue tracker with his slack post

PR #30515, opened 2025-05-14
Return code one when solve fails

YaqiWang:
Refer to #4325. This is not an ideal solution but give a nonzero flag if solve fails which bash scripts may expect.


roystgnr:
I'm not surprised this trips up a ton of tests, and I think the solution is to just fix the tests ... but I am kind of surprised that Civet reports exit code 1 as `CRASH` rather than something else.  IMHO a crash is always a bug whereas returning failure is often a standard expected+required behavior.

YaqiWang:
The extra changes in tests fixes all tests in framework on my mac. Let us see how many modules and apps are impacted by this.

YaqiWang:
More fixes in modules. Someone needs to check whether those unconverged solves in the NS module are intended.

YaqiWang:
I think this is ready for review. I had a patch in Griffin. Other failures seem random. The two executioners `SIMPLE` and `Optimize` need you more attention.

modules/fluid_properties/test/tests/interfaces/nan_interface/tests
```diff
@@ -55,6 +56,7 @@
     type = 'RunApp'
     input = 'nan_interface.i'
     cli_args = 'FluidProperties/fp/emit_on_nan=warning'
+    should_crash = true
```
GiudGiud, line 59: why is it crashing? we allow warnings

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/segregated-comparison/tests
```diff
@@ -4,6 +4,7 @@
   [nonlinear]
     type = 'Exodiff'
     input = segregated-nonlinear.i
+    should_crash = true
```
GiudGiud, line 7:
```suggestion
    # only testing two iterations
    should_crash = true
```
same for the others

GiudGiud (commented):
should we catch this error code 1 in the test harness and allow it rather than set should_crash on every test that does not exit with 0?

YaqiWang:
> should we catch this error code 1 in the test harness and allow it rather than set should_crash on every test that does not exit with 0?

I think yes because this can prevent a potential convergence behavior change from happening and `should_crash` indicates the test will have a non-zero exit code.

framework/src/executioners/InversePowerMethod.C
```diff
@@ -81,7 +81,10 @@ void
 InversePowerMethod::execute()
 {
   if (_app.isRecovering())
+  {
+    _last_solve_converged = true;
```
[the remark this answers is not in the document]
YaqiWang: There is a message in `InversePowerMethod::init()`
```
_console << "\nCannot recover InversePowerMethod solves!\nExiting...\n" << std::endl;
```
I did not try making things consistent for this to-be-removed executioner.

framework/src/executioners/NonlinearEigen.C
```diff
@@ -101,7 +101,10 @@ void
 NonlinearEigen::execute()
 {
   if (_app.isRecovering())
+  {
+    _last_solve_converged = true;
```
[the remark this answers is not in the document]
YaqiWang: Same comment as the one for `InversePowerMethod` above.

modules/contact/test/tests/bouncing-block-contact/tests
```diff
@@ -165,6 +165,7 @@
     issues = '#21329'
     type = 'Exodiff'
     input = frictional-penalty-weighted-vel.i
+    should_crash = true
```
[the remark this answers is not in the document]
YaqiWang: Added `this test does not converge`. I think it is not intentional but do not know how to fix it.

modules/fluid_properties/test/tests/interfaces/nan_interface/tests
YaqiWang: Even it seems we allow this as a warning, but we still get this `Nonlinear solve did not converge due to DIVERGED_FNORM_NAN iterations 0`.

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/segregated-comparison/tests
YaqiWang: Added as suggested.

modules/contact/test/tests/bouncing-block-contact/tests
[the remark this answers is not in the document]
GiudGiud: min_parallel = 13 makes me think it s intentional

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/segregated-comparison/tests
GiudGiud: these tests should be doing this instead
```suggestion
    cli_args = 'Executioner/continue_on_max_its=true'
```

modules/navier_stokes/test/tests/finite_volume/limiters/lid-driven-segregated/tests
```diff
@@ -12,10 +12,12 @@
       heavy = true
       recover = false # we don't support recovery for SIMPLE yet
       max_threads = 1 # see libmesh issue #3808
```
GiudGiud, line 14: all these segregated tests should use continue_on_max_its

modules/thermal_hydraulics/test/tests/components/heat_transfer_from_heat_structure_1phase/tests
```diff
@@ -38,6 +38,7 @@
   [phy:T_wall_transfer_3eqn_z]
     type = Exodiff
     input = 'phy.T_wall_transfer_3eqn_z.i'
+    should_crash = true
```
GiudGiud, line 41: @joshuahansel want to take a look at this one?

GiudGiud (commented):
good after this for me

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/segregated/2d/tests
```diff
@@ -89,6 +89,7 @@
     cli_args = "FVKernels/inactive='scalar_1_src scalar_2_src scalar_1_coupled_source "
                "scalar_2_coupled_source' Postprocessors/inactive=''"
     recover = false # we don't support recovery for SIMPLE yet
+    should_crash = true
```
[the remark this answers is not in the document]
YaqiWang: If we do this, we will have a trivial regolding here because the time changed from 1 to 2. I am not sure whether this is intended.

modules/navier_stokes/test/tests/finite_volume/limiters/lid-driven-segregated/tests
YaqiWang: This will result trivial regolding as well because now the output will contain two time steps.

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/segregated/2d/tests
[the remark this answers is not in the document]
YaqiWang: I will push the regolding.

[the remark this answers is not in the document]
GiudGiud: Before we only output initial?

Or are outputting the accepted twice now?

GiudGiud:
I merged something too early it s my fault

modules/contact/test/tests/bouncing-block-contact/tests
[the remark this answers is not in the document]
GiudGiud: but not with the end time specified. Created an issue

GiudGiud:
@joshuahansel can I please get an approval on the TH regold?
@lindsayad can you please sign off on the other items in my commits?

This PR is uncovering a few badly created tests, where the gold file is empty, hence the regolds.

joshuahansel:
> @joshuahansel can I please get an approval on the TH regold?

Looks fine

modules/thermal_hydraulics/test/tests/components/heat_transfer_from_heat_structure_1phase/phy.T_wall_transfer_3eqn_z.i
```diff
@@ -51,6 +51,7 @@
     orientation = '0 0 1'
     length = 1
     n_elems = 50
+    scaling_factor_1phase = '1 1 1e-1'
 
     A   = 9.6858407346e-01
```
GiudGiud, line 56: @lindsayad check this out, no formatting :D

test/tests/time_integrators/abort/tests
```diff
@@ -6,6 +6,8 @@
     input = 'abort.i'
     exodiff = 'out.e'
     recover = false
+    expect_out = 'Aborting as solve did not converge and input selected to abort'
+    should_crash = true
```
GiudGiud, line 10: since this is an exodiff, should_crash is the right one here imo

test/tests/outputs/debug/tests
```diff
@@ -133,6 +133,7 @@
       type = RunApp
       input = 'show_execution_kernels_bcs.i'
       expect_out = '\[DBG\] limit_v limit_u'
+      should_crash = true
```
GiudGiud, line 136: I kept these as should_crash because the intent of the test is not to hit an exception. It's to get the desired log, the fact that it crashes is kind of irrelevant.

modules/navier_stokes/test/tests/finite_volume/limiters/lid-driven-segregated/tests
GiudGiud: as it should. This was not trivial, the previous gold file was missing the solution

modules/thermal_hydraulics/test/tests/components/heat_transfer_from_heat_structure_1phase/phy.T_wall_transfer_3eqn_z.i
lindsayad: haha beautiful save!

lindsayad:
> @lindsayad can you please sign off on the other items in my commits?

I approve once the failing test is fixed

GiudGiud:
failing test is fixed! I misread the requirement

GiudGiud:
@YaqiWang there are still Griffin test failures

YaqiWang:
> @YaqiWang there are still Griffin test failures

Yes, the patch is in a PR, but that requires this to be merged first and then has a moose submodule update.

GiudGiud:
If should-crash=true and it does not crash, you get an error?

YaqiWang:
> If should-crash=true and it does not crash, you get an error?

Yes. Because the test harness will be expecting a crash but it does not see.

GiudGiud:
ok I checked and you wont be able to make it forward compatible. I ll request the next break

YaqiWang:
I am double checking my PR. It is possible all tests in the PR with `should_crash` can possibly turn into `RunException` tests, thus can be merged separately.

GiudGiud:
some of them are exodiff and csvdiff. in that case I would make sure the gold file is covered by another test.

YaqiWang:
Yeah. It will take me some time because looks like I need update conda and recompile everything.

YaqiWang:
Unfortunately, we have to merge this first. The `RunException` test expects a crash caused by `mooseError`, but those diverged calculations won't be considered as a crash because it returns normally without this change.

framework/src/base/MooseApp.C
```diff
@@ -1678,6 +1678,10 @@ MooseApp::executeExecutioner()
     _executioner->init();
     errorCheck();
     _executioner->execute();
+    if (_executioner->lastSolveConverged())
+      setExitCode(0);
+    else
+      setExitCode(1);
```
loganharbour:
```suggestion
    if (!_executioner->lastSolveConverged())
      setExitCode(1);
```

Something else could set a nonzero exit code. We don't want to overwrite it based on this.

loganharbour (dismissed):

GiudGiud:
please use a meaningful commit message

framework/src/base/MooseApp.C
GiudGiud: you undid Logan's change.

I ll fix this

GiudGiud (approved):
looks good

framework/src/base/MooseApp.C
YaqiWang: I was on a detached head before force pushing. Now this is fixed.

loganharbour:
Reminder that explicit approval is required for breaking merges

loganharbour:
Griffin valgrind is failing and the patch will not get in.

PR #30518, opened 2025-05-14
Add validation framework to test harness

### LINKED ISSUE #30517: Add python-based, flexible infrastructure for validation test
### Motivation

We would like to have a standardized system that can:
- Collect code validation-based data and save it into a database (so that we can collect results for different commit hashes)
- Gives information on the result (pass,fail+reason) and can interface with a github PR report
- Make python-based postprocessing of the data possible
- Collect runtime information in a database

### Design

Add new validation class to TestHarness.

### Impact

Would make it easy for the applications to manage the validation efforts.

loganharbour:
Closes #30517


loganharbour:
@zachmprince I don't need a full review for this, but a glance would be nice

test/tests/test_harness/validation_ok.py
```diff
@@ -0,0 +1,56 @@
...
+        params.addRequiredParam('validation_upper_bound', 'The lower bound for the data')
+        return params
+
+    # Specific method that is ran once before all "testXXX" methods. Should
+    # primarily be used to load any data from the case that was ran,
+    # especially when the same data can be utilizied across multiple tests.
+    def initialize(self):
+        # Get the CSV file that the CSVDiff tester produces. We could manually
+        # search for the file, but instead we use getTesterOutputs() to get
+        # the paths to theoutput files that the TestHarness Tester was reading
```
grmnptr:
```suggestion
        # the paths to the output files that the TestHarness Tester was reading
```

test/tests/test_harness/validation_ok.py
```diff
@@ -0,0 +1,56 @@
...
+        # Store the single value from the csv output that we will
+        # use within the tests. In this case, just the last value
+        # in the MOOSE CSV output.
+        self.value = float(df['value'].iloc[-1])
+
+        # Load the lower and upper bounds from the extended parameters.
+        # We store these in parameters so that this single validation
+        # test case can be utilized across multiple test specifications
+        # without needing to define new test scripts.
+        self.lower_bound = float(self.getParam('validation_lower_bound'))
```
grmnptr, line 43: I think it would be cool to create specific classes already available with these parameters for the most general cases like this. This is just a not for future applications, not required for this PR.

test/tests/test_harness/validation_ok.py
```diff
@@ -0,0 +1,56 @@
...
+    # If you're only utilizing this validation case for a single test case,
+    # you can statically define the data within this script instead of
+    # loading it externally from the parameters. In this case, this method
+    # does not need to be overridden.
+    @staticmethod
+    def validParams():
+        params = ValidationCase.validParams()
+        params.addRequiredParam('validation_lower_bound', 'The upper bound for the data')
+        params.addRequiredParam('validation_upper_bound', 'The lower bound for the data')
+        return params
```
grmnptr, line 21: Is there a way to type check these or is it too complicated? Or should we trust the user to do the assert themselves?

grmnptr (commented):
Nice!

test/tests/test_harness/validation_ok.py
loganharbour: Yep, I agree.

It's way too much to do here, there's already enough complexity here. I think that was we produce more validation tests, we'll have a better idea of what else we can generalize

test/tests/test_harness/validation_ok.py
loganharbour: Not currently. But we could extend the parameter system to add something like:

```
params.addRequiredFloatParam()...
```

grmnptr: Would be cool, thanks!

python/FactorySystem/Parser.py
```diff
@@ -69,54 +68,68 @@ def error(self, msg, node=None, param=None):
...
+        for key, _ in node.params():
+            if key not in params:
+                self.error(f'unused parameter "{key}"', node=node, param=key)
 
-        params['have_errors'] = have_err
+        # Check for missing required parameters
+        params_parsed = params['_params_parsed']
+        missing = params.required_keys() - params_parsed
+        if len(missing) > 0:
+            self.error('required missing parameter(s): ' + ', '.join(missing))
```
zachmprince:
```suggestion
            self.error('missing required parameter(s): ' + ', '.join(missing))
```

python/TestHarness/schedulers/Job.py
```diff
@@ -486,15 +505,88 @@ def try_catch(do, exception_name, timer_name):
...
+        run_exception = False
+        for test_case in self.validation_cases:
+            name = type(test_case).__name__
+            message = f'Running validation case(s) in {path}:{name}\n\n'
+            output.appendOutput(message)
+
+            # Do the actual execution; here we lazily mock around print()
+            # in order to capture all output. This isn't exhaustive, but
+            # is probably good enough for now
+            run_exception = False
```
zachmprince: I'm pretty sure you don't want this guy
```suggestion
```

python/TestHarness/schedulers/Job.py
```diff
@@ -486,15 +505,88 @@ def try_catch(do, exception_name, timer_name):
...
+        """
+        Initilizes (constructs) the validation cases, if any
+        """
+        init_kwargs = {'params': self.__tester.parameters(),
+                       'tester_outputs': self.getOutputFiles(self.options)}
+        self.validation_cases = [c(**init_kwargs) for c in self.__tester._validation_classes]
+
+    def runValidation(self):
+        """
+        Runs the validation cases, if any
```
zachmprince, line 544: It would be really nice to include some documentation of the data that is added to the class from running this. Mainly, it would be nice know the types of errors that could come from the running tests and generally what they mean and how it happened..

python/TestHarness/schedulers/Job.py
```diff
@@ -486,15 +505,88 @@ def try_catch(do, exception_name, timer_name):
...
+            run_exception = False
+            def wrapped_print(*args, **kwargs):
+                end = kwargs.get('end', '\n')
+                values = [f'{v}' for v in args]
+                output.appendOutput(" ".join(values) + end)
+            with patch("builtins.print", wraps=wrapped_print):
+                try:
+                    test_case.run()
+                except test_case.TestRunException:
+                    run_exception = True
```
zachmprince: Do we want to `break` if one of the tests has an exception?

python/TestHarness/testers/Tester.py
```diff
@@ -114,6 +114,58 @@ def validParams():
...
+    def augmentParams(params):
+        # Augment our parameters with parameters from the validation test, if we
+        # have any validation tests
+        script = params['validation_test']
+        validation_classes = []
+        if script:
+            # Sanity checks
+            if '..' in script:
+                message = f'validation_test={script} out of test directory'
+                raise Exception(message)
```
zachmprince: It's good to be a little more descriptive with the type of exception.
```suggestion
                raise ValueError(message)
```

python/TestHarness/testers/Tester.py
```diff
@@ -114,6 +114,58 @@ def validParams():
...
+            path = os.path.abspath(script)
+            if not os.path.exists(script):
+                message = f'validation_test={path} not found'
+                raise Exception(message)
```
zachmprince:
```suggestion
                raise FileNotFoundError(message)
```

python/TestHarness/testers/Tester.py
```diff
@@ -114,6 +114,58 @@ def validParams():
...
+
+            # Load the script; throw an exception here if it fails
+            # so that the Parser can report a reasonable error
+            spec = importlib.util.spec_from_file_location(f'validation', path)
```
zachmprince:
```suggestion
            spec = importlib.util.spec_from_file_location('validation', path)
```

python/TestHarness/testers/Tester.py
```diff
@@ -114,6 +114,58 @@ def validParams():
...
+            module = importlib.util.module_from_spec(spec)
+            try:
+                spec.loader.exec_module(module)
+            except Exception as e:
+                raise Exception(f'In validation_test={path}:\n{e}')
```
zachmprince:
```suggestion
                raise ImportError(f'In validation_test={path}:\n') from e
```

zachmprince: This allows for the original exception to retain its context, like what type of exception it is.

python/TestHarness/testers/Tester.py
```diff
@@ -114,6 +114,58 @@ def validParams():
...
+                    if key in params:
+                        raise Exception(f'Duplicate parameter "{key}" from validation test')
+
+                # Extend this Tester's parameters
+                params += subclass.validParams()
+                # Store the class so that it can be used later
+                validation_classes.append(subclass)
+
+        # Store the classes that are used in validation so that they
+        # can be constructed within the Job at a leter time
```
zachmprince:
```suggestion
        # can be constructed within the Job at a later time
```

python/TestHarness/tests/TestHarnessTestCase.py
```diff
@@ -20,25 +20,26 @@ class TestHarnessTestCase(unittest.TestCase):
...
+    def runTests(self, *args, tmp_output=True, as_json=False, check=True):
         cmd = ['./run_tests'] + list(args) + ['--term-format', 'njCst']
         sp_kwargs = {'cwd': os.path.join(os.getenv('MOOSE_DIR'), 'test'),
-                     'text': True}
+                     'stdout': subprocess.PIPE,
+                     'stderr': subprocess.STDOUT,
+                     'text': True,
+                     'check': check}
+        if as_json:
+            sp_kwargs['check'] = False
```
zachmprince:
```suggestion
                     'check': check and not as_json}
```

python/TestHarness/tests/test_Replay.py
```diff
@@ -42,9 +42,9 @@ def reCompile(self):
     def testReplay(self):
         """ Test ability to replay back previous run results """
         with tempfile.TemporaryDirectory() as output_dir:
-            base_args = ['--verbose', '-c', '--timing', '--results-file', 'unittest_Replay', '-o', output_dir]
+            base_args = ['--verbose', '-c', '--results-file', 'unittest_Replay', '-o', output_dir]
             base_kwargs = {'tmp_output': False}
-            output_a = self.runTests(*base_args, '-i', 'always_ok', **base_kwargs)
+            output_a = self.runTests(*base_args, '-i', 'always_ok_simple', **base_kwargs)
```
zachmprince: What is the purpose of this?

python/TestHarness/tests/test_Validation.py
```diff
@@ -0,0 +1,94 @@
...
+#* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+#*
+#* Licensed under LGPL 2.1, please see LICENSE for details
+#* https://www.gnu.org/licenses/lgpl-2.1.html
+
+from TestHarnessTestCase import TestHarnessTestCase
+
+class TestHarnessTester(TestHarnessTestCase):
+    def test(self):
+        out = self.runTests('-i', 'validation', '--re', 'ok', as_json=True)
```
zachmprince: Let's add some doco and whitespace in this file. Looking like a lot of fettuccini right now.

python/TestHarness/tests/test_ValidationCase.py
```diff
@@ -0,0 +1,188 @@
...
+# pylint: disable
+import unittest
+from TestHarness import ValidationCase
+from FactorySystem.InputParameters import InputParameters
+
+class TestValidationCase(unittest.TestCase):
+    def testNoTests(self):
+        test = ValidationCase()
+        with self.assertRaises(ValidationCase.NoTestsDefined):
+            test.run()
```
zachmprince: You're probably aware, but just in case, there is a `self.assertRaisesRegex` which allows testing that the message from the exception is expected.

python/TestHarness/tests/test_ValidationCase.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def testAddResult(self):
+        test_message = 'What a cool message'
+
+        for validation in [True, False]:
+            for status in ValidationCase.Status.list():
+                class Test(ValidationCase):
+                    def test(self):
+                        self.addResult(status, test_message, validation=validation)
```
zachmprince, line 29: Would it be better to just add the `status` and `validation` as constructor arguments so we don't have to put the class definition inside the loop? Or is that not a good test of the actual workflow? For some reason, this gives me some anxiety.

python/TestHarness/testers/Tester.py
```diff
@@ -114,6 +114,58 @@ def validParams():
...
+            validation_classes = []
+            subclasses = module.ValidationCase._subclasses.copy()
+            module.ValidationCase._subclasses = []
+            for subclass in subclasses:
+                validation_params = subclass.validParams()
+
+                # Don't allow duplicates
+                for key in validation_params.keys():
+                    if key in params:
+                        raise Exception(f'Duplicate parameter "{key}" from validation test')
```
zachmprince, line 163: I think it should be valid if multiple validation cases have the same parameters. Obviously, they shouldn't be duplicative with tester parameters. But I think this inhibits some polymorphism with validation cases.

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+                                   test=self._current_test, **kwargs)
+
+        prefix = ''
+        if status_value.data_key is not None:
+            data = self._data[status_value.data_key]
+            prefix = f'{data.key} = {data.printableValue()} '
+        print(f'[{status.value:>4}] {prefix}{message}')
+        self._results.append(status_value)
+
+    class DataKeyAlreadyExists(Exception):
```
zachmprince: I'm starting to think all of these classes shouldn't be sub-classes, this class is getting difficult to read and I think having them reside in the same module achieves the same locality of their definition.

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+
+        if data.bounds is not None:
+            assert len(data.bounds) == 2
+            min, max = data.bounds
+            assert max > min
+            success = value >= min and value <= max
+            status = self.Status.OK if success else self.Status.FAIL
+            message = [('within' if success else 'out of') + ' bounds;']
+            message += [f'min = {min:.5E}{units},']
+            message += [f'max = {max:.5E}{units}']
```
zachmprince, line 246: Just a suggestion
```suggestion
            message += [f'min = {min:.5E} {units},']
            message += [f'max = {max:.5E} {units}']
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+                         'validation': kwargs.pop('validation', True)}
+        units = f' {data.units}' if data.units is not None else ''
+
+        if data.bounds is not None:
+            assert len(data.bounds) == 2
+            min, max = data.bounds
+            assert max > min
+            success = value >= min and value <= max
+            status = self.Status.OK if success else self.Status.FAIL
+            message = [('within' if success else 'out of') + ' bounds;']
```
zachmprince, line 244: Just a grammatical thing
```suggestion
            message = [('within' if success else 'out of') + ' bounds:']
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        Get all of the data
+        """
+        return self._data
+
+    def getNumResults(self) -> int:
+        """
+        Get the number of results populated
+        """
+        return len(self._results)
```
zachmprince: I think it would be nice if these were just properties
```suggestion
    @property
    def results(self) -> list['Result']:
        """
        Get all of the results
        """
        return self._results

    @property
    def data(self) -> dict[str, 'Data']:
        """
        Get all of the data
        """
        return self._data

    @property
    def numResults(self) -> int:
        """
        Get the number of results populated
        """
        return len(self._results)
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+        Exception for when an exception was found when running a test
+        """
+        pass
+
+    def run(self):
+        """
+        Runs all of the tests. Stores results and data.
+        """
+        all_functions = [v[0] for v in inspect.getmembers(self.__class__,
+                                                          predicate=inspect.isfunction)]
```
zachmprince:
```suggestion
        all_functions = [v[0] for v in inspect.getmembers(self.__class__,
                                                          predicate=inspect.isfunction) if len(v)]
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+        if not test_functions:
+            raise ValidationCase.NoTestsDefined(self)
+
+        prefix = '-' * 2
+        name = self.__class__.__name__
+        print(f'{prefix} Running {len(test_functions)} test case(s) in {name}')
+
+        if 'initialize' in all_functions:
+            print(f'{prefix} Running {name}.initialize')
+            getattr(self, 'initialize')()
```
zachmprince: Why the fanciness with `getattr`?
```suggestion
            self.initialize()
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        all_functions = [v[0] for v in inspect.getmembers(self.__class__,
+                                                          predicate=inspect.isfunction)]
+        test_functions = [v for v in all_functions if v.startswith('test')]
+        if not test_functions:
+            raise ValidationCase.NoTestsDefined(self)
+
+        prefix = '-' * 2
+        name = self.__class__.__name__
+        print(f'{prefix} Running {len(test_functions)} test case(s) in {name}')
```
zachmprince: Instead of adding `{prefix}` to every print statement, I recommend either defining a lambda:
```python
print_with_prefix = lambda msg: print('--',  msg)
```

Or just defining a method:
```python
def _print(msg: str) -> None:
    print('--', msg)
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+
+            print(f'{prefix} Running {self._current_test}')
+            try:
+                getattr(self, function)()
+            except:
+                traceback.print_exc()
+                run_exceptions += 1
+            else:
+                results = self.getResults()
+                if not [r for r in results if r.test == self._current_test]:
```
zachmprince:
```suggestion
                if not any(r.test == self._current_test for r in results):
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+                    raise self.TestMissingResults(self, function)
+
+            self._current_test = None
+            self._current_not_validation = True
+
+        if 'finalize' in all_functions:
+            print(f'{prefix} Running {name}.finalize')
+            getattr(self, 'finalize')()
```
zachmprince:
```suggestion
            self.finalize()
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+
+        summary = f'Acquired {len(self.getData())} data value(s), '
+        summary += f'{self.getNumResults()} result(s): '
+        results = []
+        for status in ValidationCase.Status.list():
+            results.append(f'{self.getNumResultsByStatus(status)} {status.name.lower()}')
+        if run_exceptions:
+            results.append(f'{run_exceptions} exception')
```
zachmprince:
```suggestion
            results.append(f'{run_exceptions} exception(s)')
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,342 @@
...
+        summary += ", ".join(results)
```
zachmprince:
```suggestion
        summary += ', '.join(results)
```

python/TestHarness/tests/test_ValidationCase.py
```diff
@@ -0,0 +1,188 @@
...
+                'description': 'Peak temperature'}
+
+        test = ValidationCase()
+        test.addFloatData(*args.values())
+        all_data = test.getData()
+        self.assertEqual(len(all_data), 1)
+        data = test.getData()[args['key']]
+        for key, value in args.items():
+            self.assertEqual(getattr(data, key), value)
+        self.assertEqual(None, data.test)
```
zachmprince:
```suggestion
        self.assertIsNone(data.test)
```

python/doc/content/python/TestHarness.md
```diff
@@ -234,4 +234,13 @@ Ran 2 tests in 2.9 seconds.
 ```
 
 Caveats of MOOSE_TERM_COLS; If you specify too low a MOOSE_TERM_COLS, the TestHarness will only drop printing of the justification filler (see MOOSE_TERM_FORMAT above).
-   
\ No newline at end of file
+
+# Validation
```
zachmprince, line 238: I'm not sure where this would be put. But can we put a `!pysyntax` breakdown somewhere?

zachmprince (changes requested):
Review is primarily for python syntax and low-level doco, I'll rely on Peter for the review of the capability.

python/TestHarness/schedulers/Job.py
loganharbour: It's not fatal if one of these breaks. So why not run em all

python/TestHarness/tests/test_Replay.py
loganharbour: It's just faster (no need to run a full moose solve for this). `always_ok` runs simple diffusion, `always_ok_simple` just runs `echo foo`

python/TestHarness/validation/ValidationCase.py
loganharbour: Units isn't necessarily set, so the space is captured in `units`

python/TestHarness/tests/test_ValidationCase.py
loganharbour: My goal is just to test all of the statuses. Maybe I don't understand fully what you're trying to say

zachmprince: I mean like this:

```python
class Test(ValidationCase):
    def __init__(self, status, validation):
        self.status = status
        self.validation = validation
        super().__init()
    def test(self):
        self.addResult(self.status, test_message, validation=self.validation)

for validation in [True, False]:
    for status in ValidationCase.Status.list():
        test = Test(status, validation)
        test.run()
        ...
```

But I'm realizing now this suggestion is a bit pedantic for a test. Feel free to ignore and resolve.

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,343 @@
...
+            assert max > min
+            success = value >= min and value <= max
+            status = self.Status.OK if success else self.Status.FAIL
+            message = [('within' if success else 'out of') + ' bounds;']
+            message += [f'min = {min:.5E}{units},']
+            message += [f'max = {max:.5E}{units}']
+            self.addResult(status, ' '.join(message), **result_kwargs)
+
+    @property
+    def results(self) -> list['Result']:
```
zachmprince: Is this what you meant?
```suggestion
    def results(self) -> list[Result]:
```

python/TestHarness/validation/ValidationCase.py
```diff
@@ -0,0 +1,343 @@
...
+        """
+        Get all of the results
+        """
+        return self._results
+
+    @property
+    def data(self) -> dict[str, 'Data']:
```
zachmprince: Same here?
```suggestion
    def data(self) -> dict[str, Data]:
```

python/TestHarness/schedulers/Job.py
zachmprince: Reading the method more carefully, I would just like to see some explanation of what all `self.setStatus` calls mean. When would you see a `ERROR` vs. `EXCEPTION` vs. `FAILED` vs. nothing.

loganharbour: This is the same logic we use above when running the tester.

"Exception" literally means python exception of sorts. Error means some developer error. Fail means a diff (or something like it).

python/doc/content/python/TestHarness.md
loganharbour: I tried this, but pysyntax looks gross.

python/TestHarness/schedulers/Job.py
loganharbour: So - what I'm trying to say is that there's nothing new here, I suppose. The only difference is that we don't specifically have a status for a validation "failure", so it fits in with the standard error.

zachmprince (approved):
Now let's see what crazy feature requests we'll get after this is adopted.

PR #30544, opened 2025-05-20
Fix logic for forcing MPI

loganharbour:
refs #30512

Fixes the failures in https://civet.inl.gov/event/233357/ brought about by https://github.com/idaholab/moose/pull/30516

FYI @jessecarterMOOSE looks like your change didn't work within some of our apptainer builds. This only sets it if MOOSE_MPI_COMMAND is set, as you had before.


milljm (approved):

loganharbour:
Failures are unrelated.

PR #30565, opened 2025-05-21
Make Gaussian Process Libtorch Compatible

DanielYankura:
refs #30564

## Reason

Utilizing Libtorch would add more capabilities to Gaussian Process
## Design

Replace all Eigen::Matrices with torch::Tensors
Replace methods that have a Libtorch equivalent.

## Impact

This PR alone should not require re-golding of tests. However it should open up possibilities of future feature additions using libtorch.


framework/include/restart/DataIO.h
```diff
@@ -24,6 +24,10 @@
 #include "libmesh/parameters.h"
 #include "libmesh/numeric_vector.h"
 
+#ifdef LIBTORCH_ENABLED
+#include "LibtorchUtils.h"
+#endif
```
grmnptr: No need to guard this, it is already guarded in the other header.

framework/src/restart/DataIO.C
```diff
@@ -223,6 +223,25 @@ dataStore(std::ostream & stream, RealEigenMatrix & v, void * context)
...
+  stream.write((char *)&m, sizeof(m));
+  unsigned int n = t.sizes()[1];
+  stream.write((char *)&n, sizeof(n));
+  auto t_accessor = t.accessor<Real, 2>();
+  for (unsigned int i = 0; i < m; i++)
+    for (unsigned int j = 0; j < n; j++)
+    {
+      Real r = t_accessor[i][j];
+      dataStore(stream, r, context);
+    }
```
grmnptr: Libtorch tensors can be of arbitrary dimensions. The way I would do this is: first get the array of the dimensions (sizes) and output that. Then you can take the raw-data (dat pointer of the tensor) and just output that. Then in the read you just do it the same order and construct the same tensor.

modules/stochastic_tools/examples/surrogates/cross_validation/all_trainers_uniform_cv_no_gp.i
```diff
@@ -0,0 +1,142 @@
+[StochasticTools]
```
grmnptr, line 1: Maybe we could call this notorch. And then we do a torch/notorch run.

modules/stochastic_tools/src/covariances/ExponentialCovariance.C
```diff
@@ -40,54 +41,57 @@ ExponentialCovariance::ExponentialCovariance(const InputParameters & parameters)
...
                                            const Real sigma_n_squared,
                                            const Real gamma,
                                            const bool is_self_covariance)
 {
-  unsigned int num_samples_x = x.rows();
-  unsigned int num_samples_xp = xp.rows();
-  unsigned int num_params_x = x.cols();
+  unsigned int num_samples_x = x.sizes()[0];
+  unsigned int num_samples_xp = xp.sizes()[0];
+  unsigned int num_params_x = x.sizes()[1];
```
grmnptr:
```suggestion
  const unsigned int num_samples_x = x.sizes()[0];
  const unsigned int num_samples_xp = xp.sizes()[0];
  const unsigned int num_params_x = x.sizes()[1];
```

modules/stochastic_tools/src/covariances/ExponentialCovariance.C
```diff
@@ -40,54 +41,57 @@ ExponentialCovariance::ExponentialCovariance(const InputParameters & parameters)
...
+        r_scaled += pow((x_accessor[ii][kk] - xp_accessor[jj][kk]) / length_factor[kk], 2);
       r_scaled = sqrt(r_scaled);
-      K(ii, jj) = sigma_f_squared * std::exp(-pow(r_scaled, gamma));
+      K_accessor[ii][jj] = sigma_f_squared * std::exp(-pow(r_scaled, gamma));
     }
     if (is_self_covariance)
-      K(ii, ii) += sigma_n_squared;
+      K_accessor[ii][ii] += sigma_n_squared;
   }
 }
```
grmnptr: This can be a potential cause of slowdown. Libtorch has optimized routines to compute these matrix operations.

modules/stochastic_tools/src/covariances/LMC.C
```diff
@@ -63,31 +64,32 @@ LMC::LMC(const InputParameters & parameters)
...
+  torch::Tensor K_working =
+      torch::zeros({x.sizes()[0] * _num_outputs, xp.sizes()[0] * _num_outputs}, at::kDouble);
 
   // For every expansion term we add the contribution to the covariance matrix
   for (const auto exp_i : make_range(_num_expansion_terms))
   {
     _covariance_functions[exp_i]->computeCovarianceMatrix(K_params, x, xp, is_self_covariance);
     computeBMatrix(B, exp_i);
-    MathUtils::kron(K_working, B, K_params);
+    // MathUtils::kron(K_working, B, K_params);
```
grmnptr:
```suggestion
```

modules/stochastic_tools/src/covariances/LMC.C
```diff
@@ -123,15 +125,16 @@ LMC::computedKdhyper(RealEigenMatrix & dKdhp,
       computeLambdaGradient(dBdhp, number, ind);
       _covariance_functions[number]->computeCovarianceMatrix(K_params, x, x, true);
     }
-    MathUtils::kron(dKdhp, dBdhp, K_params);
+    // MathUtils::kron(dKdhp, dBdhp, K_params);
```
grmnptr:
```suggestion
```

modules/stochastic_tools/src/covariances/LMC.C
```diff
@@ -158,39 +162,45 @@ LMC::computedKdhyper(RealEigenMatrix & dKdhp,
...
   const auto & lambda_coeffs = *_lambdas[exp_i];
 
   for (const auto row_i : make_range(_num_outputs))
     for (const auto col_i : make_range(_num_outputs))
     {
-      Bmat(row_i, col_i) = a_coeffs[row_i] * a_coeffs[col_i];
+      Bmat_accessor[row_i][col_i] = a_coeffs[row_i] * a_coeffs[col_i];
       if (row_i == col_i)
-        Bmat(row_i, col_i) += lambda_coeffs[col_i];
+        Bmat_accessor[row_i][col_i] += lambda_coeffs[col_i];
```
grmnptr: this is just a*a^T+lambda*I, can be defnitely accelerated with matrix operations in torch.

modules/stochastic_tools/src/covariances/LMC.C
```diff
@@ -158,39 +162,45 @@ LMC::computedKdhyper(RealEigenMatrix & dKdhp,
...
 
 void
-LMC::computeLambdaGradient(RealEigenMatrix & grad,
+LMC::computeLambdaGradient(torch::Tensor & grad,
                            const unsigned int /*exp_i*/,
                            const unsigned int index) const
 {
-  grad.setZero();
-  grad(index, index) = 1.0;
+  // grad.setZero();
```
grmnptr:
```suggestion
```

modules/stochastic_tools/src/covariances/MaternHalfIntCovariance.C
```diff
@@ -41,33 +42,37 @@ MaternHalfIntCovariance::MaternHalfIntCovariance(const InputParameters & paramet
...
-  unsigned int num_samples_xp = xp.rows();
-  unsigned int num_params_x = x.cols();
+  auto K_accessor = K.accessor<Real, 2>();
+  auto x_accessor = x.accessor<Real, 2>();
+  auto xp_accessor = xp.accessor<Real, 2>();
 
-  mooseAssert(num_params_x == xp.cols(),
+  unsigned int num_samples_x = x.sizes()[0];
+  unsigned int num_samples_xp = xp.sizes()[0];
+  unsigned int num_params_x = x.sizes()[1];
```
grmnptr:
```suggestion
  const unsigned int num_samples_x = x.sizes()[0];
  const unsigned int num_samples_xp = xp.sizes()[0];
  const unsigned int num_params_x = x.sizes()[1];
```

modules/stochastic_tools/src/covariances/SquaredExponentialCovariance.C
```diff
@@ -73,17 +78,18 @@ SquaredExponentialCovariance::SquaredExponentialFunction(RealEigenMatrix & K,
       // Compute distance per parameter, scaled by length factor
       Real r_squared_scaled = 0;
       for (unsigned int kk = 0; kk < num_params_x; ++kk)
-        r_squared_scaled += std::pow((x(ii, kk) - xp(jj, kk)) / length_factor[kk], 2);
-      K(ii, jj) = sigma_f_squared * std::exp(-r_squared_scaled / 2.0);
+        r_squared_scaled +=
+            std::pow((x_accessor[ii][kk] - xp_accessor[jj][kk]) / length_factor[kk], 2);
+      K_accessor[ii][jj] = sigma_f_squared * std::exp(-r_squared_scaled / 2.0);
```
grmnptr: Yeah there is an optimized way to do this one, might help. You can use ||x-y||^2 = ||x||^2 + ||y||^2 + 2 x*y.

modules/stochastic_tools/src/libtorch/surrogates/LibtorchANNSurrogate.C
```diff
@@ -44,12 +44,18 @@ LibtorchANNSurrogate::evaluate(const std::vector<Real> & x) const
   const auto & input_mean = _input_standardizer.getMean();
   const auto & input_std = _input_standardizer.getStdDev();
 
-  mooseAssert(input_mean.size() == converted_input.size() &&
-                  input_std.size() == converted_input.size(),
+  auto input_mean_accessor = input_mean.accessor<Real, 2>();
+  auto input_std_accessor = input_std.accessor<Real, 2>();
+
+  // unsigned long int i_mean = (unsigned long int)torch::size(input_mean, 0);
+  // unsigned long int i_std = (unsigned long int)torch::size(input_std, 0);
```
grmnptr:
```suggestion
```

modules/stochastic_tools/src/utils/GaussianProcess.C
```diff
@@ -67,35 +68,39 @@ GaussianProcess::linkCovarianceFunction(CovarianceFunctionBase * covariance_func
...
 
 void
-GaussianProcess::setupStoredMatrices(const RealEigenMatrix & input)
+GaussianProcess::setupStoredMatrices(const torch::Tensor & input)
 {
-  _K_cho_decomp = _K.llt();
-  _K_results_solve = _K_cho_decomp.solve(input);
+  // llt does the llt decomposition as part of the Cholesky module
+  _K_cho_decomp = torch::linalg_cholesky(_K);
+  _K_results_solve = torch::cholesky_solve(input, _K_cho_decomp);
```
grmnptr, line 192: I think this will do another cholesky decom of the cholesky (itself) in the background. Would be interesting to see if we could do the two without the double decomposition.

modules/stochastic_tools/src/utils/GaussianProcess.C
```diff
@@ -188,11 +197,17 @@ GaussianProcess::tuneHyperParamsAdam(const RealEigenMatrix & training_params,
     MooseUtils::shuffle<unsigned int>(v_sequence, generator, 0);
     for (unsigned int ii = 0; ii < _batch_size; ++ii)
     {
-      for (unsigned int jj = 0; jj < training_params.cols(); ++jj)
-        inputs(ii, jj) = training_params(v_sequence[ii], jj);
+      for (unsigned int jj = 0; jj < training_params.sizes()[1]; ++jj)
+      {
+        // inputs(ii, jj) = training_params(v_sequence[ii], jj);
```
grmnptr:
```suggestion
```

modules/stochastic_tools/src/utils/GaussianProcess.C
```diff
@@ -188,11 +197,17 @@ GaussianProcess::tuneHyperParamsAdam(const RealEigenMatrix & training_params,
...
+        inputs_accessor[ii][jj] = params_accessor[v_sequence[ii]][jj];
+      }
 
-      for (unsigned int jj = 0; jj < training_data.cols(); ++jj)
-        outputs(ii, jj) = training_data(v_sequence[ii], jj);
+      for (unsigned int jj = 0; jj < training_data.sizes()[1]; ++jj)
+      {
+        // outputs(ii, jj) = training_data(v_sequence[ii], jj);
```
grmnptr:
```suggestion
```

modules/stochastic_tools/src/utils/GaussianProcess.C
```diff
@@ -263,8 +279,10 @@ GaussianProcess::getGradient(RealEigenMatrix & inputs) const
     {
       const auto global_index = first_index + ii;
       _covariance_function->computedKdhyper(dKdhp, inputs, hyper_param_name, ii);
-      RealEigenMatrix tmp = alpha * dKdhp - _K_cho_decomp.solve(dKdhp);
-      grad_vec[global_index] = -tmp.trace() / 2.0;
+      // This is just temporary. Convert dKhp_matrix to tensor
+      auto tmp = torch::trace(torch::mm(alpha, dKdhp) - torch::cholesky_solve(dKdhp, _K_cho_decomp))
```
grmnptr: This can be expensive if the database is large. We are doing another solve in the background I think.

modules/stochastic_tools/src/utils/GaussianProcess.C
```diff
@@ -263,8 +279,10 @@ GaussianProcess::getGradient(RealEigenMatrix & inputs) const
...
+                     .item<Real>();
+      grad_vec[global_index] = -1 * (tmp / 2.0);
```
grmnptr:
```suggestion
      grad_vec[global_index] = tmp / -2.0;
```

grmnptr: Would this work?

modules/stochastic_tools/test/tests/surrogates/gaussian_process/GP_squared_exponential_torched.i
```diff
@@ -0,0 +1,123 @@
+[StochasticTools]
```
grmnptr: It think you can remove this at this point.

modules/stochastic_tools/src/surrogates/GaussianProcessSurrogate.C
```diff
@@ -79,34 +80,43 @@ GaussianProcessSurrogate::evaluate(const std::vector<Real> & x,
...
+      torch::mm(torch::transpose(K_train_test, 0, 1), _gp.getKResultsSolve()), 0, 1);
+
+  //   De-center/scale the value and store for return
   _gp.getDataStandardizer().getDestandardized(pred_value);
 
-  RealEigenMatrix pred_var =
-      K_test - (K_train_test.transpose() * _gp.getKCholeskyDecomp().solve(K_train_test));
+  torch::Tensor pred_var =
+      K_test - torch::mm(torch::transpose(K_train_test, 0, 1),
+                         torch::cholesky_solve(K_train_test, _gp.getKCholeskyDecomp()));
```
grmnptr, line 110: I think this line causes most of the slowdown.

grmnptr (changes requested):
Super nice! Let's talk about how to make it faster!

grmnptr:
Ooops, this got forgotten.

grmnptr:
Okay, I investigated the increase in the memory consumption. Well AI investigate it for me:

The heavy part is PyTorch’s CPU runtime and dependency tree, not the tiny `libtorch.dylib` shim and not the GP matrix data for these tests.

What I measured locally:

Probe / case |	Max RSS
minimal C++ binary linked to libtorch.dylib, does no tensor work | ~86 MiB
same, but does small randn, mm, linalg_cholesky, cholesky_solve, backward | ~95 MiB
stochastic_tools-opt --version | ~146 MiB
ActiveLearningGP COV, num_random_seeds=64	| ~167 MiB
ActiveLearningGP COV, num_random_seeds=10000 | ~217 MiB
ActiveLearningGP COV, default 100000 seeds | ~670 MiB

So the libtorch/tensor part is roughly:

~80-90 MiB RSS fixed per process just from loading the libtorch runtime.
~10 MiB more when small ATen/autograd/linalg tensor paths are first exercised.
The GP tensors in these tests are small; they are not hundreds of MiB.
The concrete heavy library is not libtorch.dylib itself. That file is only ~20 KiB. It loads:

libtorch_cpu.dylib      ~170 MiB on disk
libc10.dylib            ~716 KiB
libomp.dylib            ~828 KiB
libtorch_cpu.dylib is the big one.

A vmmap snapshot of a sleeping no-op libtorch-linked process showed:

Physical footprint: ~49 MiB
ps RSS:             ~76-78 MiB
Default malloc:     ~41 MiB allocated/dirty
libtorch_cpu text:  ~22 MiB resident
libtorch_cpu data:  ~2.6 MiB resident/dirty

So the expensive fixed part is:

1) PyTorch CPU operator/runtime registration
The no-op linked process already has ~41 MiB dirty malloc before any tensor work. I cannot attribute every allocation to a symbol without malloc stack logging, but this probe has no MOOSE and no GP. That memory is from the libtorch dependency tree and static/runtime initialization.

2) libtorch_cpu.dylib code and metadata
The binary maps a ~170 MiB PyTorch CPU library. Not all of that is private or even resident, but enough code/metadata pages become resident to show up in RSS. On macOS, the harness falls back to RSS, not PSS, so shared library pages can be counted per process: monitor_processes.py (line 22), monitor_processes.py (line 54).

3) Apple framework dependencies pulled by this libtorch build
libtorch_cpu.dylib links/weak-links Foundation, Metal/MPS, Accelerate/BLAS/LAPACK, CoreFoundation, IOKit, etc. These are mostly shared/read-only, but they still contribute to RSS accounting.

4) First tensor/linalg/autograd use
Small tensor work adds ~10 MiB in the probe. In the GP path, that corresponds to calls like torch::cdist, torch::mm, torch::linalg_cholesky, torch::cholesky_solve, and Adam setup

So if one libtorch-linked process has an extra ~80-100 MiB RSS baseline, and MPI launches multiple processes. That is why the same test shape can suddenly need more slots after being moved into the libtorch lane.

grmnptr:
I decreesed the number of seeds for PSS and AIS from 100k to 1k. That will keep the tests at reasonable memory level, yielding the same results.

framework/include/libtorch/utils/LibtorchUtils.h
```diff
@@ -25,8 +25,47 @@ namespace LibtorchUtils
...
+/**
+ * Utility function that creates a non-owning tensor view of a standard vector.
+ * The returned tensor shares the mutable storage of the provided vector, so the
+ * vector must outlive the tensor and may be modified through the tensor.
+ * @tparam DataType The vector element type
+ * @param vector The vector that needs to be wrapped
+ * @param sizes The desired tensor shape
+ */
+template <typename DataType>
+torch::Tensor vectorToTensorView(std::vector<DataType> & vector, c10::IntArrayRef sizes);
```
zachmprince, line 50: This should probably added to LibtorchUtilsTest.C

framework/include/restart/DataIO.h
```diff
@@ -809,6 +815,10 @@ template <>
 void dataLoad(std::istream & stream, std::stringstream & s, void * context);
 template <>
 void dataLoad(std::istream & stream, ADReal & dn, void * context);
+#ifdef MOOSE_LIBTORCH_ENABLED
+template <>
+void dataLoad(std::istream & stream, torch::Tensor & t, void * context);
```
zachmprince, line 820: Let's add some unit tests for the dataLoad and dataStore methods here.

modules/stochastic_tools/examples/surrogates/cross_validation/all_trainers_uniform_cv_no_gp.i
zachmprince, line 1: Can we just use inactive in the test spec instead of create a whole new file? I worry about having an input in the example directory that isn't referenced and is just used for testing.

modules/stochastic_tools/examples/surrogates/cross_validation/gold/all_trainers_no_libtorch.json
```diff
@@ -21,18 +18,11 @@
...
-                    ]
-                ],
                 "np_surr": [
                     [
-                        7.354381697450355,
-                        6.213144835306278,
-                        7.6440625444211845
+                        8.083719028212807,
+                        9.333672230328466,
+                        7.841425702031651
```
zachmprince: Why did these change so signficantly?

modules/stochastic_tools/src/covariances/LMC.C
```diff
@@ -149,7 +148,8 @@ LMC::computedKdhyper(RealEigenMatrix & dKdhp,
       B += B_tmp;
     }
 
-    MathUtils::kron(dKdhp, B, dKdhp_sub);
+    // MathUtils::kron(dKdhp, B, dKdhp_sub);
```
zachmprince: Remove

modules/stochastic_tools/src/libtorch/surrogates/LibtorchANNSurrogate.C
zachmprince: Remove

modules/stochastic_tools/src/libtorch/surrogates/LibtorchANNSurrogate.C
```diff
@@ -44,12 +44,18 @@ LibtorchANNSurrogate::evaluate(const std::vector<Real> & x) const
   const auto & input_mean = _input_standardizer.getMean();
   const auto & input_std = _input_standardizer.getStdDev();
 
-  mooseAssert(input_mean.size() == converted_input.size() &&
-                  input_std.size() == converted_input.size(),
+  auto input_mean_accessor = input_mean.accessor<Real, 2>();
+  auto input_std_accessor = input_std.accessor<Real, 2>();
```
zachmprince: What does the `2` do here? Is that the dimension? If so, shouldn't it be `1`?

modules/stochastic_tools/src/surrogates/ActiveLearningGaussianProcess.C
```diff
@@ -79,29 +90,37 @@ ActiveLearningGaussianProcess::reTrain(const std::vector<std::vector<Real>> & in
...
-    for (unsigned int j = 0; j < inputs[i].size(); ++j)
-      _training_params(i, j) = inputs[i][j];
+    if (input.size() != input_size)
+      mooseError("All active learning retraining inputs must have the same dimension.");
+    flat_inputs.insert(flat_inputs.end(), input.begin(), input.end());
   }
 
+  _training_params =
+      LibtorchUtils::vectorToTensorCopy(flat_inputs, {long(outputs.size()), long(input_size)});
+  _training_data = LibtorchUtils::vectorToTensorCopy(outputs, {long(outputs.size()), 1});
```
zachmprince: I'm curious why it's necessary to have the `long` conversion. Does that not happen automatically?

modules/stochastic_tools/src/utils/GaussianProcess.C
```diff
@@ -67,35 +150,49 @@ GaussianProcess::linkCovarianceFunction(CovarianceFunctionBase * covariance_func
...
+
+  _covariance_function->computeCovarianceMatrix(_K, params, params, true);
+  const auto flattened_tensor = flattenOutputData(data);
 
-  _covariance_function->buildHyperParamMap(_hyperparam_map, _hyperparam_vec_map);
+  // Compute the Cholesky decomposition and inverse action of the covariance matrix.
+  setupStoredMatrices(flattened_tensor);
+
+  _hyperparam_map.clear();
+  _covariance_function->buildHyperParamMap(_hyperparam_map);
```
zachmprince: This seems duplicative to line 173-174

modules/stochastic_tools/src/utils/GaussianProcess.C
```diff
@@ -133,136 +230,163 @@ GaussianProcess::generateTuningMap(const std::vector<std::string> & params_to_tu
...
-
-  // Internal params for Adam; set to the recommended values in the paper
-  Real b1 = opts.b1;
-  Real b2 = opts.b2;
-  Real eps = opts.eps;
-  static constexpr Real lambda = 1e-4;
+  const auto options = doubleOptionsLike(training_params);
+  std::vector<Real> theta_values(_num_tunable, 0.0);
+  _hyperparam_map.clear();
+  _covariance_function->buildHyperParamMap(_hyperparam_map);
```
zachmprince: And again here?

modules/stochastic_tools/src/actions/LoadCovarianceDataAction.C
```diff
@@ -7,13 +7,67 @@
...
+      mooseError("Expected scalar hyperparameter for ", param_name, ".");
+    params.set<unsigned int>(param_name) = cast_int<unsigned int>(tensor.item<Real>());
+  }
+  else if (params.have_parameter<std::vector<Real>>(param_name))
+  {
+    if (!CovarianceFunctionBase::isVectorHyperParameter(tensor))
+      mooseError("Expected vector hyperparameter for ", param_name, ".");
+    params.set<std::vector<Real>>(param_name) = exportHyperParameter(tensor);
+  }
+}
```
zachmprince, line 57: Let's add an error if the parameter is of unsupported type.

framework/src/restart/DataIO.C
```diff
@@ -555,6 +582,34 @@ dataLoad(std::istream & stream, std::stringstream & s, void * /* context */)
   s.write(s_s.get(), s_size);
 }
 
+#ifdef MOOSE_LIBTORCH_ENABLED
+template <>
+void
+dataLoad(std::istream & stream, torch::Tensor & t, void * context)
```
zachmprince, line 588: Do we need to make sure to load the tensor on the same device it was stored on?

modules/stochastic_tools/test/tests/reporters/ActiveLearningGP/tests
```diff
@@ -9,6 +9,9 @@
       input = main_adam.i
       jsondiff = 'single_proc_single_row_ufunction.json'
       detail = 'a single row in the sampler using the U learning function;'
+      capabilities = 'libtorch'
+      compute_devices = 'cpu cuda'
+      min_slots = 3 # Reserve extra CI memory for libtorch runtime in GP tests
```
zachmprince, line 14: Can you check whether this is still necessary?

modules/stochastic_tools/test/tests/reporters/BFActiveLearning/gold/multiple_proc_multiple_row_ufunction.json
```diff
@@ -1230,14 +1230,122 @@
...
+            "constant": {
+                "reporter_transfer_lf:avg:value": [
+                    348.54618910457225
+                ],
+                "reporter_transfer_lf:converged": [
+                    true
+                ]
+            },
+            "time": 23.0,
+            "time_step": 23
```
zachmprince, line 1348: I wanted to bring this re-gold to @somu15's attention. The std toward the end seem pretty different, causing the extra time steps.

zachmprince (approved):
Great work @DanielYankura  and @grmnptr! I don't think I have anything big here, mainly clean-up and some testing suggestions.

zachmprince (changes requested):
Dismissing my erroneous approval above

modules/stochastic_tools/examples/surrogates/cross_validation/gold/all_trainers_no_libtorch.json
grmnptr: I regolded this, but it looks like the code still gives the original results. I will reset the file.

modules/stochastic_tools/examples/surrogates/cross_validation/all_trainers_uniform_cv_no_gp.i
grmnptr: Great idea!

framework/include/libtorch/utils/LibtorchUtils.h
grmnptr: Great catch, will be added in the next push!

framework/include/restart/DataIO.h
grmnptr: Oh yes, roger that, will be in the next push!

modules/stochastic_tools/src/libtorch/surrogates/LibtorchANNSurrogate.C
grmnptr: The standardizer stores these in a [n, 1] tensor. Why not just [n]? That is a great question, I will modify this.

modules/stochastic_tools/src/surrogates/ActiveLearningGaussianProcess.C
grmnptr: Well, I was thinking: size_t is unsigned so we could run into issues where it goes over the signed representation limits.  But I suspect the libmesh convert int functions could be better here. Will add that!

modules/stochastic_tools/src/utils/GaussianProcess.C
grmnptr: Well, if you are training then:
We build the map -> Convert it to double/move it to device -> Reload it into the covariance objects
then we train (which loads the updated values into the covariance)
Then we rebuild to make sure we have the latest map.

I think the last one is not needed because the tune function already updates the map within the covariance objects.

Will test this guy.

grmnptr: You were right, these are not needed. Will remove them.

modules/stochastic_tools/src/utils/GaussianProcess.C
grmnptr: Removed

modules/stochastic_tools/src/actions/LoadCovarianceDataAction.C
grmnptr: Will do!

framework/src/restart/DataIO.C
grmnptr: that is a great question, I never thought about it. I suppose if we run with the same device setting then it is probably doable. Let me check a few things.

grmnptr: I mean right now we are moving everything to host in the save and loading back onto the host. I think that is a safe option for now. I suppose we could do it, but I am not sure how big the differences are in terms of device and device and what we need to store to be able to make sure we are loading back oxactly the same thing. Because device number might not be enough on its own.

grmnptr: If it is okay with you, I wouldn't do it in this PR.

modules/stochastic_tools/test/tests/reporters/ActiveLearningGP/tests
grmnptr: Definitely hitting these with regular settings.
I got these last time and nothing changed since then. I actually tried to use the least amount of slots.
```
JOB KILLED (OVER MEMORY): Memory/slot 311.96 MB > allowed 300.00 MB
```
We might have to increase the allowed limits in STM if we want to reduce the slots in the tests.

modules/stochastic_tools/examples/surrogates/cross_validation/all_trainers_uniform_cv_no_gp.i
grmnptr: The issue with this one is that Covariance syntax currently is only registered if libtorch is enabled. So we can't make it active/inactive. We could still register an empty shell to save this input file, not sure which one is a better solution. I will trust your call on this.

modules/stochastic_tools/examples/surrogates/cross_validation/gold/all_trainers_no_libtorch.json
grmnptr: Well, this was not a change. We just removed the gp from this and github thinks it is smart.

modules/stochastic_tools/examples/surrogates/cross_validation/all_trainers_uniform_cv_no_gp.i
zachmprince: Oh, I missed that the syntax is gone without libtorch. This makes sense then.

framework/src/restart/DataIO.C
zachmprince: I think this is fine. The store technique is copying to host, so load should follow the same pattern. We will probably have to think carefully when we start playing with recover. Which I'm doing in #33295

modules/stochastic_tools/test/tests/reporters/ActiveLearningGP/tests
zachmprince: That limit is useful to see if framework changes cause memory spikes. If we need this many slots for this test now, then we need it. I was mainly just surprised, even with reducing the number of random number seeds.

zachmprince (approved):

grmnptr (approved):

grmnptr:
Thanks @DanielYankura for the great work and thanks @zachmprince for the great review!

PR #30579, opened 2025-05-22
Stop throwing errors about libtorch/libmfm for clean targets.

### LINKED ISSUE #30578: Makefile throws error on clobber if it can't fine libtorch library
### Bug Description

When we configure with libtorch and it can't find the libary we throw an error. This error is wrong in two ways:
- We throw it with every target, whereas we should not trow it for targets like clobber.
- We should not say that users should configure with libtorch considering that this error only occures when it has already happened.

### Steps to Reproduce

Configure with libtorch, remove the libtorch libraries, try to clobber.

### Impact

Makes targets like clobber awkward to use.

### [Optional] Diagnostics

_No response_

grmnptr:
Closes #30578

### DIFF
--- a/framework/moose.mk
+++ b/framework/moose.mk
@@ -105,6 +105,7 @@ include $(FRAMEWORK_DIR)/contrib/neml2.mk
 #
 # Conditional parts if the user wants to compile MOOSE with torchlib
 #
+SKIP_EXTERNAL_LIBRARY_CHECK_TARGETS := clean clobber cleanall clobberall echo_include echo_library install_make_dir libmesh_submodule_status
 ifeq ($(ENABLE_LIBTORCH),true)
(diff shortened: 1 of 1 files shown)


GiudGiud (approved):

grmnptr:
@lindsayad , this might influence the mfem workflow as well. Are you fine with this?

lindsayad:
@grmnptr I've modified the logic slightly. I believe your previous logic would skip the error check if *any* of the make target goals were non-compilation goals. The logic I just pushed checks whether *all* of the make target goals are non-compilation goals. Because if someone typed something like `make all echo_include` then we would want to do the error check

lindsayad (approved):

grmnptr:
Ahhh yes, great catch, thanks!

PR #30622, opened 2025-05-30
Allow MFEM coefficients to take numeric values

cmacmackin:
## Reason
MOOSE functors will accept numerical values, in addition to names of functions etc. Previously MFEM coefficients (which are used analogously to functors) did not do this. Allowing them to accept numeric values also means we no longer need the constant versions of materials and boundary conditions, reducing the amount of code that needs to be maintained.

## Design
When the CoefficientManager class tries to look up a coefficient (e.g., in `getXXXCoefficient` or when declaring an alias for a coefficient) which it finds has not been declared, it will now attempt to interpret the name of the coefficient as a numeric value. For scalar coefficients, it will attempt to read it as a single float. For vectors it will try to parse the name as a space-separated list of floats. If this succeeds then it will create and return a Constant(Vector)Coefficient with the corresponding value. This behaviour is loosely modeled after how functors are looked up in the main part of MOOSE.

Currently there is no support for using numeric values with matrix coefficients, as it was not clear what would be the appropriate syntax for that. However, matrix coefficients are not actually used anywhere yet anyway.

## Impact
This PR removes the following components, as their functionality is now redundant. All tests have been updated to reflect these changes.

- MFEMScalarDirichletBC (use MFEMScalarFunctorDirichletBC)
- MFEMVectorBoundaryIntegratedBC (use MFEMVectorFunctorBoundaryIntegratedBC)
- MFEMVectorDirichletBC (use MFEMVectorFunctorDirichletBC)
- MFEMVectorDirichletBCBase (use MFEMVectorFunctorDirichletBCBase)
- MFEMVectorNormalDirichletBC (use MFEMVectorFunctorNormalDirichletBC)
- MFEMVectorNormalIntegratedBC (use MFEMVectorFunctorNormalIntegratedBC)
- MFEMGenericConstantFunctorMaterial (use MFEMGenericFunctorMaterial)
- MFEMGenericConstantVectorFunctorMaterial (use MFEMGenericFunctorVectorMaterial)

One tricky aspect of these changes were how to allow MFEMGenericFunctorVectorMaterial to take a mix of function names (etc.) and numeric vector values for `prop_values`. The numeric values will typically require multiple space-separated components to describe the vector, e.g.:
```
prop_names = mag_field gravity
prop_values = func1 0 0 9.8
```
The HIT parser will interpret prop_values as `std::vector<MFEMVectorCoefficientName>({"func1", "0", "0", "9.8"})` and there is no general way to associate each item with the appropriate `prop_name`. (E.g., Consider the case of multiple numeric properties  being defined, with vectors of different dimensions). The solution adopted was to add a preprocessing step for this parameter, that will group items enclosed in curly braces. So the input 
```
prop_names = mag_field gravity
prop_values = func1 {0 0 9.8}
```
will be processed into `std::vector<MFEMVectorCoefficientName>({"func1", "0 0 9.8"})`. This appears to work well, but does have the disadvantage of introducing new syntactic conventions.


cmacmackin:
@alexanderianblair @loganharbour @lindsayad You would probably all be good people to take a look at this.

One question I have is whether we should consider changing the names of things like `MFEMScalarFunctorDirichletBC`, now that we no longer have a non-functor version. Should we rename it to `MFEMScalarDirichletBC`? That would be a bit shorter and less wordy. The main downside is that we previously had a different object with that name and this might cause confusion and more subtle breaking changes for users.

alexanderianblair:
> @alexanderianblair @loganharbour @lindsayad You would probably all be good people to take a look at this.
> 
> One question I have is whether we should consider changing the names of things like `MFEMScalarFunctorDirichletBC`, now that we no longer have a non-functor version. Should we rename it to `MFEMScalarDirichletBC`? That would be a bit shorter and less wordy. The main downside is that we previously had a different object with that name and this might cause confusion and more subtle breaking changes for users.

I'd personally prefer the 'functor-less' naming convention to take precedence here now we no longer have duplication of BCs/kernels for different input types, since it's shorter/more readable, and IMO easier for MOOSE users to deduce which MFEM kernel corresponds closest to an existing MOOSE kernel. Plus, it's a little clearer from the name that the MFEM BCs/Kernels don't accept MOOSE functors.

The main drawback will come when we start adding dFEM-like AD kernels, as we'll need to distinguish those from the standard MFEM integrators - but it's probably premature to let that determine the naming scheme here before we have a dFEM example running.

> The HIT parser will interpret prop_values as std::vector<MFEMVectorCoefficientName>({"func1", "0", "0", "9.8"}) and there is no general way to associate each item with the appropriate prop_name. (E.g., Consider the case of multiple numeric properties being defined, with vectors of different dimensions). The solution adopted was to add a preprocessing step for this parameter, that will group items enclosed in curly braces. 

I like this solution of using curly braces for vectors, since it makes the groupings of components clear. It also could nicely generalise for MatrixCoefficients when we get around add support for them, too.

loganharbour:
> I'd personally prefer the 'functor-less' naming convention to take precedence here now we no longer have duplication of BCs/kernels for different input types, since it's shorter/more readable, and IMO easier for MOOSE users to deduce which MFEM kernel corresponds closest to an existing MOOSE kernel.

I agree with this, however...

> The main drawback will come when we start adding dFEM-like AD kernels, as we'll need to distinguish those from the standard MFEM integrators 

Can we at least come up with a reasonable naming scheme (even though we don't know if it'll fit the design yet) that would align with this change?

alexanderianblair:
> Can we at least come up with a reasonable naming scheme (even though we don't know if it'll fit the design yet) that would align with this change?

The simplest here would be `MFEMADXXXKernel` / `MFEMADXXXBC`, since that'll be our method of AD support, and would avoid a name clash with either of these options.

However, for context - the dFEM-like method of specifying kernel contributions appears to be much closer in philosophy to how a regular MOOSE kernel is written (with the additional ability to support the full range of GPU assembly levels), so I anticipate/hope that we'll eventually use those as our preferred kernel/BCs over our current `mfem::XXXFormIntegrator`-based ones, at which point the simpler `MFEMXXXKernel` designation would better describe them instead, and the current batch migrate to using the `MFEMXXXFunctorKernel` naming convention. 

So pros/cons to both options - using `MFEMXXXKernel` for the current integrator-type batch and `MFEMADXXXKernel` for the dFEM ones is clearer for users based on the current state IMO, but using `MFEMXXXFunctorKernel` for the current integrator-type batch and `MFEMXXXKernel` for future dFEM ones aligns closer to the design philosophy of the analagous MOOSE objects, if things pan out as expected.

lindsayad:
> However, for context - the dFEM-like method of specifying kernel contributions appears to be much closer in philosophy to how a regular MOOSE kernel is written

Can you elaborate on this? I believe when we've talked in the past you've said that MOOSE is at too high of an assembly level (element/E level in libCEED terms) vs. being at the quadrature-point/Q level. Is this what you're talking about?

cmacmackin:
There is still one failing CI job. Looking at the results, I'm not sure why that would happen as it doesn't involve any of the code I've changed. The particular test that failed also appears to have passed in other jobs. Should I just try marking that job as invalid and running again?

alexanderianblair:
> > However, for context - the dFEM-like method of specifying kernel contributions appears to be much closer in philosophy to how a regular MOOSE kernel is written
> 
> Can you elaborate on this? I believe when we've talked in the past you've said that MOOSE is at too high of an assembly level (element/E level in libCEED terms) vs. being at the quadrature-point/Q level. Is this what you're talking about?

Yes, exactly this - they can support partial (or matrix free) assembly - eg. [here](https://github.com/mfem/mfem/pull/4804/files#diff-dfa3af7086de566b4682999b0c9447fa2ccbb104680d3c9e5ba210554475b1f7R39-R69)

loganharbour:
> There is still one failing CI job. Looking at the results, I'm not sure why that would happen as it doesn't involve any of the code I've changed. The particular test that failed also appears to have passed in other jobs. Should I just try marking that job as invalid and running again?

The failure in https://civet.inl.gov/job/2917216/ is unrelated. Fix is in next and it can be ignored.

cmacmackin:
@lindsayad @loganharbour What are your feelings on the renaming issue? It would be good to get this decided so we can merge the PR.

lindsayad:
I like the idea to remove `Functor` from the names. And since dFEM capability is not even merged in MFEM, I don't think we should change the current prefix of `MFEM`. Future `dFEM`-based MFEM kernels can use `MFEMAD`. This follows MOOSE convention in terms of not-AD vs. AD. Even if the dFEM-based kernels are more like traditional MOOSE kernels in terms of using element-level assembly, I don't think that matters much.

In the far future if no-one uses the MFEM kernels based on the integrators then we can remove those and consider renaming the MFEM AD kernels, but that's a far in the distance decision I think. I believe we have plenty of `AD` prefixed kernels that do not have hand-coded Jacobian analogs. In fact I can't really think of any AD-based objects that don't use an AD prefix other than our FV objects, but those are all AD; there is no hand-coded Jacobian support for FV

cmacmackin:
Now something is broken in one of the modules, with `test:finite_volume/two_phase/mixture_model.mixture_lid_driven_two_phase_physics` failing. Is this unrelated issue that can be ignored? Why do these sort of issues keep happening?

lindsayad:
We have a PR up to fix that test at https://github.com/idaholab/moose/pull/30748

framework/include/mfem/bcs/MFEMVectorDirichletBC.h
```diff
@@ -1,14 +1,14 @@
 #ifdef MFEM_ENABLED
 
 #pragma once
-
 #include "MFEMVectorDirichletBCBase.h"
 
 class MFEMVectorDirichletBC : public MFEMVectorDirichletBCBase
 {
+
```
lindsayad:
```suggestion
```

framework/include/mfem/bcs/MFEMVectorDirichletBC.h
```diff
@@ -1,14 +1,14 @@
 #ifdef MFEM_ENABLED
 
 #pragma once
-
```
lindsayad: Let's keep this new-line

framework/src/mfem/bcs/MFEMVectorDirichletBC.C
```diff
@@ -4,16 +4,6 @@
...
 registerMooseObject("MooseApp", MFEMVectorDirichletBC);
 
-InputParameters
-MFEMVectorDirichletBC::validParams()
-{
-  InputParameters params = MFEMVectorDirichletBCBase::validParams();
-  params.addClassDescription(
-      "Applies a Dirichlet condition to all components of a vector variable.");
-  return params;
-}
```
lindsayad: As this is a registered object, we should have a class description

framework/include/mfem/bcs/MFEMVectorDirichletBC.h
```diff
@@ -1,14 +1,14 @@
...
 #include "MFEMVectorDirichletBCBase.h"
 
 class MFEMVectorDirichletBC : public MFEMVectorDirichletBCBase
 {
+
 public:
-  static InputParameters validParams();
```
lindsayad, line 19: keep

lindsayad (changes requested):
37/78 through

cmacmackin:
I've implemented the requested changes. Note that these issues predate this PR; they just showed up in the diff because git got confused about the fact I renamed a file.

framework/include/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.h
```diff
@@ -19,7 +19,7 @@ class MFEMGenericFunctorVectorMaterial : public MFEMFunctorMaterial
 
 protected:
   const std::vector<std::string> & _prop_names;
-  const std::vector<MFEMVectorCoefficientName> & _prop_values;
+  const std::vector<MFEMVectorCoefficientName> _prop_values;
```
lindsayad, line 31: I'd revert this change. No reason to store this twice, once in the input parameters warehouse and once here

lindsayad: Oops I was looking at the wrong local repo. I see that this does indeed have to be its own storage

framework/include/mfem/utils/CoefficientManager.h
```diff
@@ -129,9 +145,22 @@ class CoefficientManager
     return this->declareMatrixProperty(name, blocks, _matrix_coeffs.make<P>(args...));
   }
 
+  /// Return a scalar coefficient with the given name or, if that
+  /// doesn't exists, try interpreting the name as a number with which
+  /// to build a new constant coefficient.
   mfem::Coefficient & getScalarCoefficient(const std::string name);
```
lindsayad, line 160:
```suggestion
  mfem::Coefficient & getScalarCoefficient(const std::string & name);
```
I know this isn't in your PR but would be good to get these all at once

framework/include/mfem/utils/CoefficientManager.h
```diff
@@ -129,9 +145,22 @@ class CoefficientManager
...
+
+  /// Return a vector coefficient with the given name or, if that
+  /// doesn't exists, try interpreting the name as a vector of number with which
+  /// to build a new constant vector coefficient.
   mfem::VectorCoefficient & getVectorCoefficient(const std::string name);
```
lindsayad, line 165:
```suggestion
  mfem::VectorCoefficient & getVectorCoefficient(const std::string & name);
```

framework/include/mfem/utils/CoefficientManager.h
```diff
@@ -129,9 +145,22 @@ class CoefficientManager
...
+
+  /// Return scalar coefficient with the given name. Unlike for
+  /// the scalar and vector counterparts, there is currently no way to
+  /// try interpreting the name as numbers with which to construct a
+  /// constant matrix coefficient.
   mfem::MatrixCoefficient & getMatrixCoefficient(const std::string name);
```
lindsayad, line 171:
```suggestion
  mfem::MatrixCoefficient & getMatrixCoefficient(const std::string & name);
```

framework/include/mfem/utils/CoefficientManager.h
```diff
@@ -157,6 +186,9 @@ class CoefficientManager
   mfem::MatrixCoefficient & declareMatrixProperty(const std::string & name,
                                                   const std::vector<std::string> & blocks,
                                                   std::shared_ptr<mfem::MatrixCoefficient> coef);
+  std::shared_ptr<mfem::Coefficient> getScalarCoefficientPtr(const std::string name);
+  std::shared_ptr<mfem::VectorCoefficient> getVectorCoefficientPtr(const std::string name);
+  std::shared_ptr<mfem::MatrixCoefficient> getMatrixCoefficientPtr(const std::string name);
```
lindsayad, line 200:
```suggestion
  std::shared_ptr<mfem::Coefficient> getScalarCoefficientPtr(const std::string & name);
  std::shared_ptr<mfem::VectorCoefficient> getVectorCoefficientPtr(const std::string & name);
  std::shared_ptr<mfem::MatrixCoefficient> getMatrixCoefficientPtr(const std::string & name);
```

framework/include/utils/MooseStringUtils.h
```diff
@@ -10,6 +10,7 @@
 #pragma once
 
 #include <algorithm>
+#include <sstream>
```
lindsayad, line 13: without any other changes in the header this seems not needed
```suggestion
```

framework/src/mfem/bcs/MFEMVectorTangentialDirichletBC.C
```diff
@@ -4,15 +4,6 @@
...
 registerMooseObject("MooseApp", MFEMVectorTangentialDirichletBC);
 
-InputParameters
-MFEMVectorTangentialDirichletBC::validParams()
-{
-  InputParameters params = MFEMVectorDirichletBCBase::validParams();
-  params.addClassDescription(
-      "Applies a Dirichlet condition to the tangential components of a vector variable.");
-  return params;
-}
```
lindsayad: This should also stay

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
```diff
@@ -5,6 +5,59 @@
...
+  MFEMVectorCoefficientName literal;
+  for (const auto & item : input)
+  {
+    if (in_literal)
+    {
+      if (item.front() == '{')
+      {
+        mooseError("Nested numeric vector values are not permitted in "
+                   "MFEMGenericFunctorVectoMaterial prop_values.");
+      }
```
lindsayad, line 32:
```suggestion
        mooseError("Nested numeric vector values are not permitted in "
                   "MFEMGenericFunctorVectoMaterial prop_values.");
```

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
```diff
@@ -5,6 +5,59 @@
...
+      else if (item.back() == '}')
+      {
+        in_literal = false;
+        literal += " " + item.substr(0, item.size() - 1);
+        result.push_back(literal);
+      }
+      else
+      {
+        literal += " " + item;
+      }
```
lindsayad, line 42:
```suggestion
        literal += " " + item;
```

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
```diff
@@ -5,6 +5,59 @@
...
+    }
+    else if (item.front() == '{')
+    {
+      if (item.back() == '}')
+      {
+        result.push_back(item.substr(1, item.size() - 2));
+      }
```
lindsayad, line 49:
```suggestion
        result.push_back(item.substr(1, item.size() - 2));
```

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
```diff
@@ -5,6 +5,59 @@
...
+      else
+      {
+        literal += " " + item;
+      }
+    }
+    else if (item.front() == '{')
+    {
+      if (item.back() == '}')
+      {
+        result.push_back(item.substr(1, item.size() - 2));
```
lindsayad, line 48: Why is it -2 here instead of -1 ?

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
```diff
@@ -5,6 +5,59 @@
...
+    {
+      result.push_back(item);
+    }
+  }
+  if (in_literal)
+  {
+    mooseError("No closing curly brace for vector value in "
+               "MFEMGenericFunctorVectorMaterial prop_values: '{" +
+               literal + "'");
+  }
```
lindsayad, line 66:
```suggestion
      result.push_back(item);
  }
  if (in_literal)
    mooseError("No closing curly brace for vector value in "
               "MFEMGenericFunctorVectorMaterial prop_values: '{" +
               literal + "'");
```

framework/src/mfem/utils/CoefficientManager.C
```diff
@@ -102,29 +103,67 @@ CoefficientManager::declareMatrixProperty(const std::string & name,
...
-  std::shared_ptr<mfem::MatrixCoefficient> coef =
-      this->_matrix_coeffs.getCoefficientPtr(existing_coef);
+  std::shared_ptr<mfem::MatrixCoefficient> coef = this->getMatrixCoefficientPtr(existing_coef);
   if (std::dynamic_pointer_cast<mfem::PWMatrixCoefficient>(coef))
     mooseError("Properties must not be defined out of other properties or piecewise coefficients.");
   return this->declareMatrixProperty(name, blocks, coef);
 }
 
+std::shared_ptr<mfem::Coefficient>
+CoefficientManager::getScalarCoefficientPtr(const std::string name)
```
lindsayad, line 122:
```suggestion
CoefficientManager::getScalarCoefficientPtr(const std::string & name)
```

unit/src/MFEMMaterialTest.C
```diff
@@ -371,10 +217,12 @@ TEST_F(MFEMFunctorScalarMaterialTest, MFEMGenericFunctorMaterial_Exception)
 TEST_F(MFEMFunctorVectorMaterialTest, MFEMGenericFunctorVectorMaterial)
 {
   InputParameters coef_params = _factory.getValidParams("MFEMGenericFunctorVectorMaterial");
-  coef_params.set<std::vector<std::string>>("prop_names") = {"coef1", "coef2"};
-  coef_params.set<std::vector<MFEMVectorCoefficientName>>("prop_values") = {"func1", "func2"};
+  coef_params.set<std::vector<std::string>>("prop_names") = {
+      "coef1", "coef2", "constant1", "constant2"};
+  coef_params.set<std::vector<MFEMVectorCoefficientName>>("prop_values") = {
+      "func1", "func2", "{0.", "1.", "2.}", "{1.5}"};
```
lindsayad, line 223: What happens when only a single value is supplied like with `"{1.5}"`?

lindsayad: Are all three components set to 1.5 or just the x-component?

lindsayad (changes requested):

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
cmacmackin: We don't want to include the trailing close bracket.

framework/include/utils/MooseStringUtils.h
cmacmackin: At some point I'd had an issue with a non-unity build. I can't remember exactly why I concluded this was the best place to put the include though.

unit/src/MFEMMaterialTest.C
cmacmackin: `{1.5}` is interpreted as a single-component (1D) vector. If it gets used along with vectors with more components then the missing components of this one will be treated as 0. So, in effect, we are just setting the x-component.

framework/include/utils/MooseStringUtils.h
cmacmackin: OK, I've looked this up in the CI logs. The templated [MooseUtils::tokenizeAndConvert](https://mooseframework.inl.gov/docs/doxygen/moose/namespaceMooseUtils.html#aadb0b7c436697b8d7ded35fb34109bc9) function creates a local `std::stringstream` variable, which requires the `sstream` header to have been included. It seems that previously this template was only invoked in files that had done this, but it's fragile to assume that other people will know to do so.

cmacmackin:
We have some more apparently unrelated test failures:
- `transfers/multiapp_nearest_node_transfer.invalid_source_boundary_from_app`: "The boundary 'abc' was not found in the mesh", Clang debug version only.
- `heat_transfer/test:gap_heat_transfer_balance.large_gap_heat_transfer_test_rz_cylinder`: timeout, debug build only
- `solid_mechanics/test:neml2/laromance.radial_eturn`: timeout, debug build only

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
lindsayad: But earlier in the function you use `item.size() - 1` to remove a closing bracket (presumably?)

cmacmackin: Ah, I remember now. The second argument is not the final index but the length of the substring being taken. In the previous case we started from index 0 and took all but the last character of the string, so we needed `item.size() - 1` characters. In this case the first character is an open brace so we skip that and start from index 1. We still want to leave out the last character, meaning our substring has length `item.size() - 2`.

lindsayad: Lol I even looked up the c++ reference for substr and didn't register this
> The second argument is not the final index but the length

lindsayad (approved):

lindsayad:
Looks like we have some conflicts now. Do you want to resolve them @cmacmackin ? If not, I can do it

cmacmackin:
I won't be able to get to it until tomorrow. If you have time today then go
ahead and resolve the conflicts.

framework/src/mfem/functormaterials/MFEMGenericFunctorVectorMaterial.C
lindsayad: It seems like a lot of these suggested changes were not made but marked as resolved. It's traditional MOOSE style to not use wrapping braces for single statements, so I'd like the change to happen. If you disagree, I'm happy to hear differing opinions

cmacmackin: That's odd. I was sure I'd clicked to commit the changes. I'll try again.

cmacmackin: Actually, seeing as you've now committed these in #30803 we might as well just merge that.

lindsayad: Yea I don't think this is your fault at all. I also tried clicking it here and it didn't work. I'm guessing this is due to the PR being from an organizational repo? E.g. the same reason why I can't push to this branch? But I don't know for sure

nmnobre:
Closing as completed via #30803.

PR #30626, opened 2025-05-30
AddVariableAction flexibility

### LINKED ISSUE #30625: Pass name and values to createInitialConditionAction in AddVariableAction
### Motivation

There are times where names and ic need to be modified (e.g., log transform). If the name and value is passed to `AddVariableAction::createInitialConditionAction`, then all complexities in that method can just be inherited.

### Design

Turn `AddVariableAction::createInitialConditionAction()` into `AddVariableAction::createInitialConditionAction(const std::string var_name, const std::vector<Real> value)`

### Impact

more flexibility in `AddVariableAction`

tophmatthews:
Closes #30625

### DIFF
--- a/framework/include/actions/AddVariableAction.h
+++ b/framework/include/actions/AddVariableAction.h
@@ -77,7 +77,7 @@ protected:
    * If the user supplies a value for 'initial_condition' in the input file this
    * method will create the proper InitialCondition object.
    */
-  void createInitialConditionAction();
+  void createInitialConditionAction(const std::string var_name, const std::vector<Real> value);
 
   /**
    * Get the block ids from the input parameters
--- a/framework/include/actions/AddVariableAction.h
+++ b/framework/include/actions/AddVariableAction.h
@@ -72,12 +72,17 @@ protected:
   void addVariable(const std::string & var_name);
 
   /**
+   * Return the name of the nonlinear variable to be created
+   */
+  virtual std::string varName() const { return _name; }
+
+  /**
    * Create the action to generate the InitialCondition object
    *
    * If the user supplies a value for 'initial_condition' in the input file this
    * method will create the proper InitialCondition object.
    */
-  void createInitialConditionAction(const std::string var_name, const std::vector<Real> & value);
+  void createInitialConditionAction(const std::vector<Real> & value);
 
   /**
    * Get the block ids from the input parameters
--- a/framework/include/actions/AddVariableAction.h
+++ b/framework/include/actions/AddVariableAction.h
@@ -77,7 +77,7 @@ protected:
    * If the user supplies a value for 'initial_condition' in the input file this
    * method will create the proper InitialCondition object.
    */
-  void createInitialConditionAction(const std::string var_name, const std::vector<Real> value);
+  void createInitialConditionAction(const std::string var_name, const std::vector<Real> & value);
 
   /**
    * Get the block ids from the input parameters
--- a/framework/src/actions/AddVariableAction.C
+++ b/framework/src/actions/AddVariableAction.C
@@ -160,7 +160,7 @@ AddVariableAction::act()
   // Set the initial condition
   if (isParamValid("initial_condition"))
   {
-    const auto value = getParam<std::vector<Real>>("initial_condition");
+    const auto & value = getParam<std::vector<Real>>("initial_condition");
     createInitialConditionAction(value);
   }
 }
--- a/framework/src/actions/AddVariableAction.C
+++ b/framework/src/actions/AddVariableAction.C
@@ -168,18 +168,15 @@ AddVariableAction::act()
   addVariable(var_name);
 
   // Set the initial condition
+  auto value = _pars.get<std::vector<Real>>("initial_condition");
   if (_pars.isParamValid("initial_condition"))
-    createInitialConditionAction();
+    createInitialConditionAction(var_name, value);
 }
 
 void
-AddVariableAction::createInitialConditionAction()
+AddVariableAction::createInitialConditionAction(const std::string var_name,
+                                                const std::vector<Real> value)
 {
-  // Variable name
-  std::string var_name = name();
-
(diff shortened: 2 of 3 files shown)


framework/include/actions/AddVariableAction.h
```diff
@@ -77,7 +77,7 @@ class AddVariableAction : public MooseObjectAction
    * If the user supplies a value for 'initial_condition' in the input file this
    * method will create the proper InitialCondition object.
    */
-  void createInitialConditionAction();
+  void createInitialConditionAction(const std::string var_name, const std::vector<Real> value);
```
lindsayad:
```suggestion
  void createInitialConditionAction(const std::string var_name, const std::vector<Real> & value);
```

framework/src/actions/AddVariableAction.C
```diff
@@ -168,18 +168,15 @@ AddVariableAction::act()
...
+  auto value = _pars.get<std::vector<Real>>("initial_condition");
   if (_pars.isParamValid("initial_condition"))
-    createInitialConditionAction();
+    createInitialConditionAction(var_name, value);
 }
 
 void
-AddVariableAction::createInitialConditionAction()
+AddVariableAction::createInitialConditionAction(const std::string var_name,
+                                                const std::vector<Real> value)
```
lindsayad:
```suggestion
                                                const std::vector<Real> & value)
```

lindsayad (changes requested):
I don't yet understand why this is useful. A test case might help. When would you want to create an initial condition for a variable other than the variable being created by this action?

tophmatthews:
Use case in new class `AddLogVariableAction` that inherits from `AddVariableAction`:
```
void
AddLogVariableAction::act()
{
  // If we've been called that means that current_task == "add_variable"
  init();

  // Get necessary data for creating a variable
  std::string var_name = name();
  if (getParam<bool>("log_transform"))
    var_name = var_name + "_log";
  addVariable(var_name);

  // Set the initial condition
  if (_pars.isParamValid("initial_condition"))
  {
    auto value = _pars.get<std::vector<Real>>("initial_condition");
    if (getParam<bool>("log_transform"))
    {
      for (unsigned int i = 0; i < value.size(); ++i)
      {
        mooseAssert(value[i] > 0.0,
                    "If doing log transforms, initial conditions must be greater than zero.");
        value[i] = std::log(value[i]);
      }
    }
    createInitialConditionAction(var_name, value);
  }
}
```
Essentially, allows me to only override `act`, and nothing else. This is for a separate app, not planned for MOOSE.

tophmatthews:
not sure what is going on with coverage, but this is ready for re-review @lindsayad

lindsayad:
Users want the variable name to be changed? `foo_log` instead of `foo`?

tophmatthews:
> Users want the variable name to be changed? `foo_log` instead of `foo`?

Not users really, I want it to be a silent name change (it's wrapped up in actions). The bigger thing is the initial condition transformation which is enabled here by passing it via the `act` method. Alternatively, the initial condition application method could be templated, but it seems really unnecessary to have to duplicate it all.

lindsayad:
Need a test

tophmatthews:
> Need a test

I'm not sure what test would be relevant beyond the many many tests on `TimeDerivative` already. I'll have my own tests for the log transform which inherits from `TimeDerivative`, but this PR is focused only on making `TimeDerivative` more flexible, right?

lindsayad:
`TimeDerivative`? You mean `AddVariableAction`?

lindsayad:
My preference would be to make `createInitialConditionAction` virtual so that you can do the same rename that you show in your `act` override

lindsayad:
Or better yet perhaps would be to introduce a `virtual std::string varName() const` method whose default implementation is `return name();`

tophmatthews:
> `TimeDerivative`? You mean `AddVariableAction`?

Whoops, yes of course, `AddVariableAction`.

tophmatthews:
> My preference would be to make createInitialConditionAction virtual so that you can do the same rename that you show in your act override

That's fine and would work, but I have to just copy and paste all the complexity in `createInitialConditionAction` to only change the initial condition value. Makes me worried if it changes in the future. By passing it in, I don't have to ever touch  `createInitialConditionAction`.

lindsayad:
> Or better yet perhaps would be to introduce a `virtual std::string varName() const` method whose default implementation is `return name();`

what about this suggestion?

framework/src/actions/AddVariableAction.C
```diff
@@ -168,15 +168,19 @@ AddVariableAction::act()
...
+  }
 }
 
 void
-AddVariableAction::createInitialConditionAction(const std::string var_name,
-                                                const std::vector<Real> & value)
+AddVariableAction::createInitialConditionAction(const std::vector<Real> & value)
 {
+  // Variable name
+  std::string var_name = name();
```
lindsayad: You don't want the name of the action though right? I was expecting a new `virtual std::string varName() const`

tophmatthews: I'm fighting ` symbol not found in flat namespace` locally, trying to see if civet has the same problem, so was trying to simplify.

But I think `name` might actually work since it's a virtual class in `MooseBase.h`? Still trying to test...

tophmatthews: Oh I see, can't do name(). Added varName back.

tophmatthews:
Looks like coverage errors went away...ready for you to review again @lindsayad , thanks!!

lindsayad (approved):

lindsayad:
@grmnptr could you review the small patch I put up for the libtorch error/info when libtorch libs aren't found?

grmnptr (commented):
The changes in our make file look good to me, thanks!
Just a quick question: shouldn't we change MFEM too? Or that is a different beast in this sense?

lindsayad:
I'll do that in a follow-on

PR #30629, opened 2025-05-30
Make Node::children const

### LINKED ISSUE #30628: Mark `hit::Node::children` as `const`
### Motivation

This method returns a list of child nodes. It does not require write access to the current node and should be marked `const`. This will allow objects to traverse the hit tree even if they have a pointer to a const hit node.

### Design

Add `const`

### Impact

Added functionality.

dschwen:
Closes #30628


loganharbour:
Why? The resulting children aren't const...

loganharbour:
Constification to hit will happen in https://github.com/idaholab/moose/pull/30204

PR #30657, opened 2025-06-03
Array variable custom name support

### LINKED ISSUE #30656: Add ability to specify naming convention to array variables
### Motivation

Currently a user specifies array variables like

```
[Variables]
  [density]
    order = FIRST
    family = MONOMIAL
    components = 10
  []
[]
```
And the names of each array variable component will automatically be set as `density_1`, `density_2`, ..., `density_10`. This behavior is fine for most applications as users interact with the array variable based on the name `density` in an input file. However, for post processing outputs, in many cases it could be very useful to be able to have custom names that represent what the components physically are. For example, one might want exodus output variables to be named `density_O16`, `density_C12`, ..., `density_B10`.

### Design

Add a new parameter that allows custom naming for array variable components. This feature could be used in an input file in different ways, for example referring to an array variable component in other places in the input file by its custom name instead of its component number. But at first, the support will only be for adding a custom array variable name so that when it is output to exodus, it has a meaningful name.

### Impact

Users will be able to add custom names to array variable components so that when output to exodus, they could have some physical name.

jthano:
closes #30656


framework/include/variables/MooseVariableBase.h
```diff
@@ -178,7 +181,7 @@ class MooseVariableBase : public MooseObject,
   /**
    * @return whether this is an array variable
    */
-  bool isArray() const { return _is_array; }
+  virtual bool isArray() const { return _is_array; }
```
loganharbour: Why? And can't this just use `_array_var_component_names.size()` now?

framework/src/variables/MooseVariableBase.C
```diff
@@ -119,7 +128,10 @@ MooseVariableBase::MooseVariableBase(const InputParameters & parameters)
     _count(getParam<unsigned int>("components")),
     _scaling_factor(_count, 1.0),
     _use_dual(getParam<bool>("use_dual")),
-    _is_array(getParam<bool>("array"))
+    _is_array(getParam<bool>("array")),
+    _array_var_component_names(isParamValid("array_var_component_names")
```
loganharbour: Yeah, it seems redundant now having both of these variables.

framework/src/systems/SystemBase.C
```diff
@@ -738,10 +738,34 @@ SystemBase::addVariable(const std::string & var_type,
     if (fe_field_type == TYPE_VECTOR)
       mooseError("Vector family type cannot be used in an array variable");
 
+    std::vector<std::string> array_var_component_names;
+    const bool has_array_names = parameters.isParamValid("array_var_component_names");
+    if (has_array_names)
+    {
+      array_var_component_names =
+          parameters.get<std::vector<std::string>>("array_var_component_names");
+      if (array_var_component_names.size() != components)
```
loganharbour, line 747: Can't you keep this error within `MooseVariableBase`? That way we can use `paramError()` there and get context by line to the offending parameter?

framework/src/variables/MooseVariableField.C
```diff
@@ -85,7 +85,11 @@ template <typename OutputType>
 bool
 MooseVariableField<OutputType>::isArray() const
 {
-  return std::is_same<OutputType, RealEigenVector>::value;
+  const auto is_array = MooseVariableBase::isArray();
+  if (std::is_same<OutputType, RealEigenVector>::value != is_array)
```
loganharbour, line 89:
```suggestion
  if constexpr (std::is_same<OutputType, RealEigenVector>::value != is_array)
```

WIll make this a compile time check

framework/include/variables/MooseVariableBase.h
```diff
@@ -42,6 +42,9 @@ class MooseVariableBase : public MooseObject,
 
   MooseVariableBase(const InputParameters & parameters);
 
+  /// Returns the variable name of a component of an array variable
+  std::string arrayVariableComponent(const unsigned int i) const;
```
loganharbour: Can't this be a const reference? Since we're now accessing it from storage

framework/src/systems/SystemBase.C
```diff
@@ -1186,7 +1216,7 @@ SystemBase::copyVars(ExodusII_IO & io)
         const auto & array_var = getFieldVariable<RealEigenVector>(0, vci._dest_name);
         for (MooseIndex(var.count()) i = 0; i < var.count(); ++i)
         {
-          const auto exodus_var = _subproblem.arrayVariableComponent(vci._source_name, i);
+          const auto exodus_var = var.arrayVariableComponent(i);
```
loganharbour: Should be able to be `const auto & exodus_var` now as well

loganharbour (changes requested):

framework/src/systems/SystemBase.C
jthano: The difficulty with how this works is that SystemBase needs the parameters before the MooseVariableBase object is constructed. That is why the check of the `array` parameter was also here in SystemBase instead of MooseVariableBase. But I could of course find some way to indicate to MooseVariableBase that `array_var_component_names` was invalid and it could throw a paramError instead of SystemBase throwing a mooseError. I think the current code is a little cleaner even though it is using a mooseError where a paramError would be more appropriate. If you look at the error message, it does at least specify which variable has the incorrect parameter so it should be easy for a user to fix the issue.

jthano: Perhaps we could pass `MooseVariableBase` a system pointer and let it actually add the variable so that it can also check its own parameters? That would be a bigger change, but might make more sense than the way things are now where all the parameters are needed before the `MooseVariableBase` object exists.

framework/include/variables/MooseVariableBase.h
YaqiWang: No because an array variable is allowed to have only one component.

jthano: In this case, I can actually just check if `_array_var_component_names` is empty or not. If the array variable has a single component, `_array_var_component_names` will still have a single name in it. It is only empty if the `array` parameter was false

YaqiWang: Oh, sorry, I see.

framework/src/systems/SystemBase.C
jthano: @loganharbour what is your opinion about this, should I do some more refactoring beyond the current changes so `paramError` can be called from `MooseVariableBase`?

framework/src/variables/MooseVariableField.C
jthano: I get a compiler error when trying this suggested code change because 
```
 note: initializer of 'is_array' is not a constant expression
 ```
 I don't think this can work as suggested, but I can do something similar. This part, `std::is_same<OutputType, RealEigenVector>::value`, at least can have `constexpr`.

loganharbour: Oh wait, hold on... why do you need a runtime check now? Is this related to why you have to add the virtual method?

framework/src/systems/SystemBase.C
loganharbour: Just leave it as is. We can add more context to this error in the future

framework/src/variables/MooseVariableField.C
jthano: `MooseVariableBase` had a non-virtual method `isArray()` and the derived class `MooseVariableFieldBase` had a method of the same name, `isArray()`, that was marked virtual. There wasn't any connection between whether `MooseVariableBase` and the derived class `MooseVariableFieldBase` agreed about whether the variable was an array variable or not. This could easily cause an error because `MooseVariableFieldBase` was calling its own version of `isArray()` and then calling `arrayVariableComponent` from `MooseVariableBase` without knowing if `MooseVariableBase` agreed this was an array variable. There is definitely something wrong with the original design and this `isArray` method. Maybe the code can be refactored in the future, but for now, I made these `isArray` related changes to make sure all classes, bases and derived classes, agree that a variable is an array variable.

loganharbour: Okay, that's fine. I follow.

loganharbour:
Failures are related and require an app patch.

loganharbour (approved):

PR #30591, opened 2025-05-26
LOR Solvers rebased

lindsayad:
Rebased version of #30428 dropping merge commits

PR #30660, opened 2025-06-03
Documentation and preconditioner fixes for MFEM backend

nmnobre:
## Reason
Incomplete or incorrectly rendering documentation.
A bug introduced in #30591 prevents the correct selection of the solver's preconditioner.

## Design
Added or fixed documentation. Had to improve `extractInputBlocks()` in `listing.py` to strip trailing slashes only.
Fixed preconditioning selection bug by essentially rolling back to the original design whereby we get the respective user object from the solver constructor, not an `MFEMProblem` method.

## Impact
Beautiful documentation.
Most notably, this should mean we now use the preconditioner we intended to: some tests marked LOR were in fact running without LOR.

### DIFF
--- a/framework/include/mfem/solvers/MFEMSolverBase.h
+++ b/framework/include/mfem/solvers/MFEMSolverBase.h
@@ -29,10 +29,10 @@ public:
   /// Retrieves the preconditioner userobject if present, sets the member pointer to
   /// said object if still unset, and sets the solver to use this preconditioner.
   template <typename T>
-  void setPreconditioner(const std::shared_ptr<T> & solver);
+  void setPreconditioner(T & solver);
 
-  /// Returns a shared pointer to the instance of the Solver derived-class.
-  virtual std::shared_ptr<mfem::Solver> getSolver() { return _solver; }
+  /// Returns the wrapped MFEM solver
+  mfem::Solver & getSolver();
 
   /// Updates the solver with the given bilinear form and essential dof list, in case an LOR or algebraic solver is needed.
   virtual void updateSolver(mfem::ParBilinearForm & a, mfem::Array<int> & tdofs) = 0;
@@ -48,8 +48,15 @@ protected:
   bool _lor;
 
   // Solver and preconditioner to be used for the problem
-  std::shared_ptr<mfem::Solver> _solver{nullptr};
-  MFEMSolverBase * _preconditioner{nullptr};
+  std::unique_ptr<mfem::Solver> _solver;
+  MFEMSolverBase * _preconditioner;
 };
+
+inline mfem::Solver &
+MFEMSolverBase::getSolver()
+{
+  mooseAssert(_solver, "Attempting to retrieve solver before it's been constructed");
+  return *_solver;
+}
 
 #endif
--- a/framework/include/mfem/solvers/MFEMSolverBase.h
+++ b/framework/include/mfem/solvers/MFEMSolverBase.h
@@ -26,6 +26,11 @@ public:
 
   MFEMSolverBase(const InputParameters & parameters);
 
+  /// Retrieves the preconditioner userobject if present, sets the member pointer to
+  /// said object if still unset, and sets the solver to use this preconditioner.
+  template <typename T>
+  void setPreconditioner(const std::shared_ptr<T> & solver);
+
   /// Returns a shared pointer to the instance of the Solver derived-class.
   virtual std::shared_ptr<mfem::Solver> getSolver() { return _solver; }
 
--- a/framework/include/mfem/solvers/MFEMSolverBase.h
+++ b/framework/include/mfem/solvers/MFEMSolverBase.h
@@ -32,7 +32,8 @@ public:
   /// Updates the solver with the given bilinear form and essential dof list, in case an LOR or algebraic solver is needed.
   virtual void updateSolver(mfem::ParBilinearForm & a, mfem::Array<int> & tdofs) = 0;
 
-  bool isLOR() const { return _lor; }
+  /// Returns whether or not this solver (or its preconditioner) uses LOR
+  bool isLOR() const { return _lor || (_preconditioner && _preconditioner->isLOR()); }
 
 protected:
   /// Override in derived classes to construct and set the solver options.
@@ -41,8 +42,9 @@ protected:
   // Variable defining whether to use LOR solver
   bool _lor;
 
-  // Solver to be used for the problem
+  // Solver and preconditioner to be used for the problem
   std::shared_ptr<mfem::Solver> _solver{nullptr};
+  MFEMSolverBase * _preconditioner{nullptr};
 };
 
 #endif
--- a/framework/src/mfem/solvers/MFEMCGSolver.C
+++ b/framework/src/mfem/solvers/MFEMCGSolver.C
@@ -39,26 +39,25 @@ void
 MFEMCGSolver::constructSolver(const InputParameters &)
 {
   auto solver =
-      std::make_shared<mfem::CGSolver>(getMFEMProblem().mesh().getMFEMParMesh().GetComm());
+      std::make_unique<mfem::CGSolver>(getMFEMProblem().mesh().getMFEMParMesh().GetComm());
   solver->SetRelTol(getParam<mfem::real_t>("l_tol"));
   solver->SetAbsTol(getParam<mfem::real_t>("l_abs_tol"));
   solver->SetMaxIter(getParam<int>("l_max_its"));
   solver->SetPrintLevel(getParam<int>("print_level"));
-  setPreconditioner(solver);
-  _solver = solver;
+  setPreconditioner(*solver);
+  _solver = std::move(solver);
 }
 
 void
 MFEMCGSolver::updateSolver(mfem::ParBilinearForm & a, mfem::Array<int> & tdofs)
 {
-
   if (_lor && _preconditioner)
     mooseError("LOR solver cannot take a preconditioner");
 
   if (_preconditioner)
   {
     _preconditioner->updateSolver(a, tdofs);
-    setPreconditioner(std::dynamic_pointer_cast<mfem::CGSolver>(_solver));
+    setPreconditioner(static_cast<mfem::CGSolver &>(*_solver));
   }
   else if (_lor)
   {
--- a/framework/src/mfem/solvers/MFEMCGSolver.C
+++ b/framework/src/mfem/solvers/MFEMCGSolver.C
@@ -30,11 +30,7 @@ MFEMCGSolver::validParams()
   return params;
 }
 
-MFEMCGSolver::MFEMCGSolver(const InputParameters & parameters)
-  : MFEMSolverBase(parameters),
-    _preconditioner(isParamSetByUser("preconditioner")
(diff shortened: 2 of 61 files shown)


nmnobre:
@cticenhour, you only need to look at the first commit please, [cd45264](https://github.com/idaholab/moose/pull/30660/commits/cd45264cace00c8e943e21f694bcfc553e9dedb2), which I needed so the second commit, [eb9a1f0](https://github.com/idaholab/moose/pull/30660/commits/eb9a1f06e1b8923ba665f8ca00f8b563d4361a75), would work. @alexanderianblair, I moved all the fixes here, of course most notably the preconditioner refactor we discussed elsewhere.

nmnobre:
> ### Versioner influential files
> Found 57 influential files, 1 changed, 0 added, 0 removed
> 
> package	status	file
> wasp	CHANGE	python/pyhit/pyhit.py

For clarity, I'm no longer modifying this file. It was just a typo fix in a docstring, and I can't be bothered to bump all the conda stuff just for that...

python/MooseDocs/extensions/listing.py
```diff
@@ -403,7 +403,7 @@ def extractInputBlocks(filename, blocks):
         hit = pyhit.load(filename)
         out = []
         for block in blocks.split():
-            node = moosetree.find(hit, lambda n: n.fullpath.endswith(block.strip('/')))
+            node = moosetree.find(hit, lambda n: n.fullpath.endswith(block.rstrip('/')))
```
cticenhour, line 409: I am fine with this extension adjustment. 👍🏻 Will leave other aspects to @lindsayad since he is more familiar with the MFEM integration.

framework/include/mfem/solvers/MFEMSolverBase.h
```diff
@@ -23,7 +23,7 @@ class MFEMSolverBase : public MFEMGeneralUserObject
   /// Updates the solver with the given bilinear form and essential dof list, in case an LOR or algebraic solver is needed.
   virtual void updateSolver(mfem::ParBilinearForm & a, mfem::Array<int> & tdofs) = 0;
 
-  bool isLOR() const { return _lor; }
+  bool isLOR() const { return _lor || (_preconditioner ? _preconditioner->isLOR() : false); }
```
lindsayad:
```suggestion
  bool isLOR() const { return _lor || (_preconditioner && _preconditioner->isLOR()); 
```

lindsayad: Some doxygen for why we need this new `||` would be great

framework/src/mfem/solvers/MFEMCGSolver.C
```diff
@@ -40,8 +36,12 @@ MFEMCGSolver::constructSolver(const InputParameters &)
   solver->SetMaxIter(getParam<int>("l_max_its"));
   solver->SetPrintLevel(getParam<int>("print_level"));
 
-  if (_preconditioner)
+  if (isParamSetByUser("preconditioner"))
+  {
+    auto & pre_ref = const_cast<MFEMSolverBase &>(getUserObject<MFEMSolverBase>("preconditioner"));
+    _preconditioner = std::dynamic_pointer_cast<MFEMSolverBase>(pre_ref.getSharedPtr());
```
lindsayad: Do we need `_preconditioner` to be a shared pointer? Would just a regular pointer suffice? Then we can avoid bouncing up and down the hierarchy. MFEMProblem will definitely destruct after this object

framework/src/mfem/solvers/MFEMHyprePCG.C
```diff
@@ -41,8 +37,10 @@ MFEMHyprePCG::constructSolver(const InputParameters &)
   solver->SetMaxIter(getParam<int>("l_max_its"));
   solver->SetPrintLevel(getParam<int>("print_level"));
 
-  if (_preconditioner)
+  if (isParamSetByUser("preconditioner"))
   {
+    auto & pre_ref = const_cast<MFEMSolverBase &>(getUserObject<MFEMSolverBase>("preconditioner"));
+    _preconditioner = std::dynamic_pointer_cast<MFEMSolverBase>(pre_ref.getSharedPtr());
```
lindsayad: This pattern repeats over and over for these Solvers. You get the twist of the following four lines for Hypre solvers but that doesn't actually change the implementation. Those are just for error checking (which is a good thing). It seems like maybe we could add something like this:
```c++
void
MFEMSolverBase::finishSolverConstruction(mfem::Solver & solver, const bool check_for_hypre)
{
  if (isParamSetByUser("preconditioner"))
  {
    _preconditioner = &const_cast<MFEMSolverBase &>(getUserObject<MFEMSolverBase>("preconditioner"));
    if (check_for_hypre)
    {
      auto * const hypre_pre = dynamic_cast<mfem::HypreSolver *>(_preconditioner->getSolver())
      if (!hypre_pre)
        mooseError("Preconditioner must be a hypre solver);
    }

    solver.SetPreconditioner(*_preconditioner->getSolver());
  }

  _solver = &solver;
}
```
This method assumes that both `_preconditioner` and `_solver` become regular pointers instead of shared pointers

lindsayad (changes requested):

framework/include/mfem/solvers/MFEMSolverBase.h
nmnobre: Oh that's embarrassing, good catch!

framework/src/mfem/solvers/MFEMCGSolver.C
nmnobre: I just kept it as it was originally, but I agree it seems unnecessary for `_preconditioner`, so that's now a raw pointer. I've left `_solver` as a shared pointer though, as changing that in the same way isn't (unless I'm missing something) quite as straightforward, we'd need to manage that memory ourselves.

framework/src/mfem/solvers/MFEMHyprePCG.C
nmnobre: "Seems" being the keyword. :) Unfortunately the MFEM API isn't quite as nice as to allow that.
`SetPreconditioner()` is not a member of `mfem::Solver`, but of `mfem::IterativeSolver` (base class of `mfem::CGSolver` and `mfem::GMRESSolver`) and `mfem::HyprePCG`, `mfem::HypreGMRES` and `mfem::HypreFGMRES`.
I've given it a try nonetheless, the best I could come up with is in 3d6b21b.
Feel free to amend and push if you come up with better ideas.

lindsayad:
An important commit relevant for LOR is lindsayad/moose@c5e1de6 which is included in #28114 in order to address [failures](https://civet.inl.gov/job/2925959/) on our upcoming cuda recipe
```
[43.32s]     FAIL mfem/kernels.MFEMCurlCurlLOR FAILED (CRASH)
[10.50s]     FAIL mfem/kernels.MFEMGradDivLOR FAILED (CRASH)
```

lindsayad:
@nmnobre let me know how you feel about the most recent commit

framework/src/mfem/solvers/MFEMSolverBase.C
```diff
@@ -14,8 +14,44 @@ MFEMSolverBase::validParams()
...
+      if (auto * const hypre_pre = dynamic_cast<mfem::HypreSolver *>(&mfem_pre))
+        solver.SetPreconditioner(*hypre_pre);
+      else
+        mooseError("hypre solver preconditioners must themselves be hypre solvers");
+    else if constexpr (std::is_base_of_v<mfem::IterativeSolver, T>)
+      solver.SetPreconditioner(mfem_pre);
+    else
+      mooseAssert(
+          true,
+          "We should never get here because we only instantiate this method for iterative solvers");
```
nmnobre:
```suggestion
    else
      solver.SetPreconditioner(mfem_pre);
```

Since, as you note, we're explicitly instantiating anyway, maybe we could just get rid of both assertions in this method...? The compiler will complain anyway if `SetPreconditioner` isn't a member of `T`.

The suggestion above should actually cover all cases, but one... (sigh)
`mfem::HypreAME` which takes an `mfem::HypreSolver` for preconditioner, but is not itself an `mfem::HypreSolver`. But again, if we ever implement that solver and instantiate this function, the compiler will complain `mfem_pre` is of the wrong type, at which point we can amend line 37 with `|| T == mfem::HypreAME`.

nmnobre:
> @nmnobre let me know how you feel about the most recent commit

I like the clever use of `unique_ptr` so we don't need to manage the memory ourselves while at the same time avoiding the overhead of referencing counting from `shared_ptr`. But that's not a "regular pointer" :P

lindsayad (dismissed):

nmnobre:
So there's another bug we could probably fix here, _we never initialise hypre_...
I've added an option for AMG yesterday (f1d2aed1ecf) to ignore hypre errors which works around this issue, but now I know what's causing it, we should probably pick a place to call `mfem::Hypre::Init()`. Since we use `mfem::HypreParMatrix` in `EquationSystem`, we are already effectively assuming hypre is present at all times. Discussing with @alexanderianblair...

lindsayad:
Ok, I'll wait to merge then. Your comment showed up just in time

loganharbour:
I'm planning on adding MFEM tests to our main valgrind target. Need to test it first. And because this is pending @nmnobre, I'm gonna add it here

nmnobre:
> I'm planning on adding MFEM tests to our main valgrind target. Need to test it first. And because this is pending @nmnobre, I'm gonna add it here

Seems like it caught some leaks already :P

loganharbour:
> > I'm planning on adding MFEM tests to our main valgrind target. Need to test it first. And because this is pending @nmnobre, I'm gonna add it here
> 
> Seems like it caught some leaks already :P

We can ignore them. I'll just put up a separate PR to test this. I think that @lindsayad might have added some fixes in https://github.com/idaholab/moose/pull/28114? Does that sound right?

lindsayad:
> I think that @lindsayad might have added some fixes in #28114? Does that sound right?

The only fix I've included in #28114 is for a segmentation fault attributable to bad destruction ordering which only shows up when using cuda

nmnobre:
@loganharbour, can we lift both the timeout and 5 maximum failures limits on the new valgrind test, just for the time being, 'cause it currently feels like a cat and mouse game xD

@lindsayad Feel free to add that segfault fix here if you'd like, we're going to beat the record for the most bugs fixed in a single PR haha

nmnobre:
I think this is done! Just 3 spurious errors: 2 timeouts and 1 due to too big a stack frame (~2MiB stack allocations in [framework/contrib/mfem/fem/qinterp/det.cpp:252](https://github.com/mfem/mfem/blob/9f945a5863407bdb5e1212465687ba09a49bbd7e/fem/qinterp/det.cpp#L252-L254) which confuse Valgrind). Can add some `valgrind = none`, or `valgrind = heavy`, or `max_time = ...`, but I'll let you decide.

giphy

lindsayad:
I think `valgrind = none` for the stack frame failure and `valgrind = heavy` for the timeouts. At some point we'll probably want to start splitting out tests from the `mfem/kernels `directory because it just keeps growing

lindsayad (approved):

PR #30464, opened 2025-05-07
Tidy up VTK/nlohmann hackery

nmnobre:
Refs #30362.

## Reason
"This is increasingly hideous, so I'd love other ideas on how to improve it, but it still seems like a step forward from the previous incompatible-with-one-VTK-configuration-or-another setups."

_Originally posted by @roystgnr in https://github.com/idaholab/moose/pull/30427#pullrequestreview-2816370760_

## Design
Simpler logic.

## Impact
None beyond cleaner code.

PR #30663, opened 2025-06-04
Fix another VTK nlohmann json workaround

k-collie:
Same problem as issue #30425

Related: #30427, #30464

## Reason
To fix compilation when VTK built with external nlohmann json (avoiding redefinitions).

vtk-9.5.0 is hopefully coming soon which I believe removes their hackery from their public headers, so we can delete all this `#undef` mess soon.

### DIFF
--- a/framework/include/utils/ImageSampler.h
+++ b/framework/include/utils/ImageSampler.h
@@ -23,31 +23,8 @@
 // loves to warn about it...
 #include "libmesh/ignore_warnings.h"
 
-// If VTK is built without an external nlohmann, then it assumes it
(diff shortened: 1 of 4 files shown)


k-collie:
VTK 9.5 has been released, and [this problem is fixed](https://docs.vtk.org/en/latest/release_details/9.5.html#api). I'll close this in favour of telling people to avoid VTK 9.4.

milljm:
We should also be good moving forwards, as we no longer include `moose-peacock` within the `moose-dev` family of packages (no more Conda VTK package).

PR #30677, opened 2025-06-05
Cache calls to runExe in moosedocs tests

### LINKED ISSUE #30676: Reduce application calls in appsyntax tests
### Motivation

The appsyntax tests within MooseDocs repeatedly call the application with the same arguments. This leads to a slow test.

### Design

Cache the callbacks to the application and mock the calling of the application within the appsyntax tests.

### Impact

Faster tests, fewer timeouts.

loganharbour:
Closes #30676

This caches (most) calls to runExe so that we don't call the app with the same arguments more than once within a single set of tests.

MooseDocs tests before:

```
Ran 66 tests in 316.6 seconds. Average test time 8.7 seconds, maximum test time 219.2 seconds.
```

MooseDocs tests after:

```
Ran 66 tests in 165.9 seconds. Average test time 6.0 seconds, maximum test time 65.1 seconds.
```

### DIFF
--- a/python/MooseDocs/test/__init__.py
+++ b/python/MooseDocs/test/__init__.py
@@ -8,10 +8,11 @@
 #* https://www.gnu.org/licenses/lgpl-2.1.html
 
 import unittest
-import collections
+import mooseutils
+from mock import patch
 from MooseDocs.common import load_extensions
 from MooseDocs import base
-from MooseDocs.tree import tokens, pages, html, latex
+from MooseDocs.tree import pages, html, latex
 from MooseDocs.extensions import command
 
 class MooseDocsTestCase(unittest.TestCase):
@@ -22,6 +23,9 @@ class MooseDocsTestCase(unittest.TestCase):
     READER = base.MarkdownReader
     RENDERER = base.HTMLRenderer
     EXECUTIONER = base.Serial
+
+    RUN_EXE_CACHE = {}
+    ORIG_RUN_EXE = mooseutils.runExe
 
     def __init__(self, *args, **kwargs):
         super(MooseDocsTestCase, self).__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


python/MooseDocs/test/__init__.py
```diff
@@ -53,6 +57,22 @@ def __setup(self, reader=None, renderer=None, extensions=None, executioner=None)
...
+        # a given set of arguments, it'll run the app. Every time after
+        # that, it'll use the cache
+        def run_exe_cached(app_path, args):
+            key = app_path + ''.join(args)
+            cache = MooseDocsTestCase.RUN_EXE_CACHE
+            if key not in cache:
+                result = MooseDocsTestCase.ORIG_RUN_EXE(app_path, args)
+                cache[key] = result
+            return cache[key]
+        patcher = patch.object(mooseutils, 'runExe', wraps=run_exe_cached)
```
dschwen, line 72: This is some next level stuff

python/MooseDocs/test/__init__.py
```diff
@@ -8,10 +8,11 @@
 #* https://www.gnu.org/licenses/lgpl-2.1.html
 
 import unittest
-import collections
+import mooseutils
+from mock import patch
```
dschwen, line 12: Is `mock` always available?

python/MooseDocs/test/__init__.py
```diff
@@ -53,6 +57,22 @@ def __setup(self, reader=None, renderer=None, extensions=None, executioner=None)
         self.__translator.init()
         self.__translator.execute() # This is required to setup the meta data
 
+    def setUp(self):
+        # Wrap mooseutils.runExe so that we don't call the application
+        # with the same arguments more than once. The first time with
+        # a given set of arguments, it'll run the app. Every time after
+        # that, it'll use the cache
+        def run_exe_cached(app_path, args):
+            key = app_path + ''.join(args)
```
dschwen, line 66: Not 

```suggestion
            key = app_path + ' '.join(args)
```

?

loganharbour: What??

I think you mean:
```
key = app_path + ' '.join(args)
```
?

This isn't ever read by a human. It's just a squash of all of the things to make a hashable key based on the app and args

python/MooseDocs/test/__init__.py
loganharbour: Yes (it's a system module)

python/MooseDocs/test/__init__.py
loganharbour: Python mock is both wonderful and horrifying

zachmprince (approved):
Very clever!

loganharbour:
Failure in https://civet.inl.gov/job/2922285/ is unrelated and will be fixed in https://github.com/idaholab/moose/pull/30679.

python/MooseDocs/test/__init__.py
dschwen: I know, but space separation removes a potential ambiguity. Squashing arguments together would make different argument sets that have spaces in different spots equivalent. I cannot see a potential combo for only our options, but if you add in hit parameter changes, it could lead to obscure bugs that are super easy to avoid by adding that single space.

```
./moose-opt -i input.i xy=y xz=y
./moose-opt -i input.i xy=yx z=y
```

loganharbour: This is ran for:

- /path/to/moose_test-opt --json --allow-test-objects
- /path/to/moose_test-opt --show-type

You're definitely right. But... I'm not particularly concerned and am a bit checked out after four python fixes in a day

PR #30683, opened 2025-06-06
Minimal NEML2 action mode

### LINKED ISSUE #30682: Minimal NEML2 action mode if no inputs or outputs are specified
### Motivation

To use NEML2 in Swift I'd like to keep using the `NEML2Action`, however in case of a pure FFT run we do not need an index or executor UO as we'll have a tensor compute call NEML2 and we won't need data transfer from a mesh.
Not adding a Swift specific action will facilitate mixed FEM FFT runs with NEML2 in the future.

### Design

Skip index and executor of both outputs and inputs are empty.

### Impact

Added capability.

dschwen:
Closes #30682

### DIFF
--- a/modules/solid_mechanics/src/neml2/actions/NEML2Action.C
+++ b/modules/solid_mechanics/src/neml2/actions/NEML2Action.C
@@ -251,23 +251,27 @@ NEML2Action::act()
                    param.moose.name);
     }
 
-    // The index generator UO
+    // do not add user objects if no MOOSE input or output mappings are defined
+    if (!_inputs.empty() || !_outputs.empty())
     {
-      auto type = "NEML2BatchIndexGenerator";
-      auto params = _factory.getValidParams(type);
-      params.applyParameters(parameters());
-      _problem->addUserObject(type, _idx_generator_name, params);
-    }
+      // The index generator UO
+      {
+        auto type = "NEML2BatchIndexGenerator";
+        auto params = _factory.getValidParams(type);
+        params.applyParameters(parameters());
+        _problem->addUserObject(type, _idx_generator_name, params);
+      }
 
-    // The Executor UO
-    {
-      auto type = "NEML2ModelExecutor";
-      auto params = _factory.getValidParams(type);
-      params.applyParameters(parameters());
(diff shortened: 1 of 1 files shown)


dschwen:
Ugh, I should probably warn if a model is specified in this case :-/

modules/solid_mechanics/src/neml2/actions/NEML2Action.C
```diff
@@ -251,23 +251,27 @@ NEML2Action::act()
                    param.moose.name);
     }
 
-    // The index generator UO
+    // do not add user objects if no MOOSE input or output mappings are defined
```
hugary1995, line 254: I'm afraid the logic needed here might be a bit more complicated than this.

The four quadrants are
1. Have both input mapping & output mapping: definitely add UOs
2. Have input mapping but no output mapping: this might be a user error if the neml2 model has outputs
3. Have output mapping but no input mapping: similarly this might be a user error if the neml2 model has inputs; another possibility is that the neml2 model simply provides output without any input -- I think that's possible.
4. Neither: could be user error

hugary1995 (changes requested):

modules/solid_mechanics/src/neml2/actions/NEML2Action.C
hugary1995: Oh, as you mentioned above, I guess this also depends on whether a model is specified...

hugary1995:
I'll try to take care of this in #30799

hugary1995:
Actually, with #30799 you no longer need this change. Because #30799 moves away from a global neml2 factory singleton, so you can load models from (possibly different) input files wherever you want, as many times as you want.

In swift, don't use the NEML2Action, and in your TensorCompute object, you could just do something like
```cpp
neml2::set_default_dtype(neml2::kFloat64);
auto factory = neml2::load_input("my_input.i");
_model = factory->get_model("my_model");
_model->to(neml2::kCPU);
```
where `_model` is of type `std::shared_ptr<neml2::Model>`.

dschwen:
Perfect. I'll close this, and hopefully remember to look here when the NEML2 submodule gets updated.

PR #30693, opened 2025-06-07
Do not use subprocess in test harness unit tests

### LINKED ISSUE #30692: Do not use subprocess to test the test harness
### Motivation

Using a subprocess to test the test harness is slow and also makes it impossible to mock functions or inspect variables within the executed test harness.

### Design

Execute test harness tests directly using python.

### Impact

More versatile testing.

loganharbour:
Closes #30692

### DIFF
--- a/python/TestHarness/TestHarness.py
+++ b/python/TestHarness/TestHarness.py
@@ -186,8 +186,8 @@ def findDepApps(dep_names, use_current_only=False):
 
 class TestHarness:
     @staticmethod
-    def buildAndRun(argv: list, app_name: str, moose_dir: str, moose_python: str = None,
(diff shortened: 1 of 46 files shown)


loganharbour:
@milljm this is ready for review.

The tests themselves haven't changed - but this means we can do cooler stuff like mock things in the test harness and actually query the testharness in python after a run, instead of fully relying on on screen output and json output.

For example, in https://github.com/loganharbour/moose/blob/136455158f4342cc6bc0734efb07a75bb4bacc69/python/TestHarness/tests/TestHarnessTestCase.py#L77 we no longer need to parse from the output the number of tests. We just get the variable from the harness itself.

milljm (approved):
noice!

PR #30707, opened 2025-06-10
Use hash table matrix assembly for another constraint test

lindsayad:
Refs #28979


loganharbour (approved):
Navier stokes failure is unrelated

PR #30709, opened 2025-06-10
Re-enable \deprecated etc. Doxygen commands

### LINKED ISSUE #30708: MOOSE-generated Doxygen should include important Doxygen command output
### Motivation

Compare
https://libmesh.github.io/doxygen/classlibMesh_1_1BoundaryInfo.html#aaf1175bed1e0ec9cfb27b7b7762fc80e

to:
 
https://mooseframework.inl.gov/docs/doxygen/libmesh/classlibMesh_1_1BoundaryInfo.html#aaf1175bed1e0ec9cfb27b7b7762fc80e

The former generates a nice "Deprecated" block from the `\deprecated` Doxygen command; the latter turns that command into an empty string and thereby **omits all mention of deprecation**.

This (with a more recently deprecated method not yet on the manually-updated libmesh.github.io docs) burned @elementx54 today.

### Design

The MOOSE `Doxyfile` currently sets `GENERATE_DEPRECATEDLIST = NO`.  We'll change that back to the default `YES`, and do the same to the TODO, TEST, and BUG options as well while we're at it.  I'll take another look for any similarly questionable settings in our configuration, but I think it's just those 4.

### Impact

Users of `\deprecated`, `\todo`, etc. in Doxygen comments will no longer have those omitted from Doxygen output.

Half of the settings changes (but none of the layout changes) from commit https://github.com/idaholab/moose/commit/178bab4cd53210006a0305cb4e52a0467f711c71 will be reverted, probably with a snarky commit message.

No API changes.

roystgnr:
Commit 178bab4, "greatly simplify doxygen", may have violated the old aphorism that "Everything should be as simple as it can be, but not simpler".  When a method is deprecated, for instance, our users do need the message explaining why it's been deprecated and what the non-deprecated alternative is.

This should fix #30708

### DIFF
--- a/framework/doc/doxygen/Doxyfile
+++ b/framework/doc/doxygen/Doxyfile
@@ -597,26 +597,26 @@ STRICT_PROTO_MATCHING  = NO
 # list. This list is created by putting \todo commands in the documentation.
 # The default value is: YES.
 
-GENERATE_TODOLIST      = NO
+GENERATE_TODOLIST      = YES
 
 # The GENERATE_TESTLIST tag can be used to enable (YES) or disable (NO) the test
 # list. This list is created by putting \test commands in the documentation.
 # The default value is: YES.
 
-GENERATE_TESTLIST      = NO
+GENERATE_TESTLIST      = YES
 
 # The GENERATE_BUGLIST tag can be used to enable (YES) or disable (NO) the bug
 # list. This list is created by putting \bug commands in the documentation.
 # The default value is: YES.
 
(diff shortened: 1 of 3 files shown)


roystgnr:
https://mooseframework.inl.gov/docs/PRs/30709/site/framework_development/index.html in the docs built for this PR links to the non-PR-specific Doxygen sites like https://mooseframework.inl.gov/docs/doxygen/libmesh/classes.html so I can't verify the effects here...

loganharbour (approved):

PR #30738, opened 2025-06-12
Torch visc uq

Crashprn:
## Reason

- This feature revises the TorchScriptTurbulentViscosityMaterial to incorporate the G1 value into the turbulent viscosity. Also adds parameterization of k and an optional debug flag.
- Fixes bouyant_channel/channel_test_1.i to allow the Neural Network predictions to be used in the simulation.

## Design

- Creates Aux variable to convert scalar NN output to a functor through AuxKernel.
- Added k and debug members to class in order to parameterize them in input file.

## Impact

- Adds parameterization


grmnptr:
Very interesting feature! 
Feel free to keep it open if you want to merge it at some point. You can just convert it to a draft PR.

PR #30761, opened 2025-06-14
Get MOOSE to work on Aurora

hugary1995:
Aurora is ANL's new HPC: https://docs.alcf.anl.gov/aurora/

@reverendbedford suffered through the compilation process to get the entire moose stack to work. The minor changes in this PR are required to support MOOSE on Aurora.

### DIFF
--- a/framework/include/userobjects/NearestPointBase.h
+++ b/framework/include/userobjects/NearestPointBase.h
@@ -121,7 +121,7 @@ NearestPointBase<UserObjectType, BaseType>::NearestPointBase(const InputParamete
 {
   if (this->template getParam<MooseEnum>("dist_norm") != "radius" &&
       parameters.isParamSetByUser("axis"))
-    this->template paramError("axis",
+    this->template paramError<>("axis",
                               "'axis' should only be set if 'dist_norm' is set to 'radius'");
 
   fillPoints();
--- a/framework/include/userobjects/NearestPointBase.h
+++ b/framework/include/userobjects/NearestPointBase.h
@@ -122,7 +122,7 @@ NearestPointBase<UserObjectType, BaseType>::NearestPointBase(const InputParamete
   if (this->template getParam<MooseEnum>("dist_norm") != "radius" &&
       parameters.isParamSetByUser("axis"))
     this->template paramError<>("axis",
-                              "'axis' should only be set if 'dist_norm' is set to 'radius'");
+                                "'axis' should only be set if 'dist_norm' is set to 'radius'");
 
   fillPoints();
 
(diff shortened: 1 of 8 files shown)


framework/include/userobjects/NearestPointBase.h
```diff
@@ -121,7 +121,7 @@ NearestPointBase<UserObjectType, BaseType>::NearestPointBase(const InputParamete
 {
   if (this->template getParam<MooseEnum>("dist_norm") != "radius" &&
       parameters.isParamSetByUser("axis"))
-    this->template paramError("axis",
+    this->template paramError<>("axis",
```
hugary1995, line 124: @reverendbedford now that I look at this more carefully, I see that all changes are similar to this one. You only made these changes in the solid mechanics module, however. Do you think we need to make similar changes in other modules, too? If so, I wonder if there's another, less intrusive way of fixing this issue...

hugary1995:
@reverendbedford the issue with EigenProblem.C was already fixed in https://github.com/idaholab/moose/pull/30469, so I reverted your change.

framework/include/userobjects/NearestPointBase.h
dschwen: Which compiler requires the `<>`? I've been trying to find this in the standard, but no luck so far.

Most compilers I tried don't seem to require the `<>`. Or am I missing something peculiar about the `paramError` template?

https://godbolt.org/z/6rG365vK8

dschwen: Actually... check this out:
https://godbolt.org/z/8csW6vWe8

Can you try if 
```suggestion
    this->paramError("axis",
```

works for you? I believe this is standard compliant and sufficient because the syntax isn't ambiguous! The `template` keyword is only required to tell the compiler that the `<` and `>` aren't comparison operators.  But we shouldn't even supply a `<>` in the first place.

dschwen (commented):
Let's see if the more compact syntax works for you.

framework/include/userobjects/NearestPointBase.h
dschwen: This will obviously work, as we already have that variant all over our code...

reverendbedford:
I’ll check when I get back but it will be a week.

dschwen:
This is already in next. Closed by #28114 in a stealth move by @loganharbour

hugary1995:
LOL

loganharbour:
Pew pew. Found these with clang 19

hugary1995:
Okay, I shouldn't have blindly blamed intel compilers then. I bet it's because the intel compilers on Aurora are wrapping clang 19. Perhaps now is the right time to bring intel compilers back to the moose ecosystem.

PR #30799, opened 2025-06-18
Parallel exception handling and factory improvements in NEML2

hugary1995:
* Update NEML2 submodule
* Use the new factory design to allow loading models from different input files
* Handle exceptions properly in parallel
* Add a test case for multiple input files
* Fix LibtorchModel::to
* Make the laromance regression test lighter
* Handle the mesh change event

close #30793


hugary1995:
@lynnmunday please take a look at the changes I made to the laromance regression test for return mapping. I reduced the number of elements and adjusted tolerances.

hugary1995:
@dschwen @chaibhave I made a small change in LibtorchModel.C to handle optional metadata. Maybe see if you need to make similar changes in any downstream app that does similar things with torch script.

hugary1995:
@Wendy-Ji @hdt5kt This should fix the parallel hang you encountered.

hugary1995:
@zaheen1258 with these changes you should be able to use different neml2 models defined in different neml2 input files, even in a multiapp.

modules/solid_mechanics/test/tests/neml2/laromance/laromance_return_mapping.i
lynnmunday, line 1: These changes to the laromance model are good.

lynnmunday (approved):

framework/include/neml2/interfaces/NEML2ModelInterface.h
```diff
@@ -58,22 +59,26 @@ class NEML2ModelInterface : public T
...
 
   /// The work scheduler to use
-  neml2::WorkScheduler * const _scheduler;
+  std::shared_ptr<neml2::WorkScheduler> _scheduler;
   /// Work dispatcher
   std::unique_ptr<DispatcherType> _dispatcher;
   /// Whether to dispatch work asynchronously
   const bool _async_dispatch;
+  /// Models for each thread
+  std::unordered_map<std::thread::id, std::shared_ptr<neml2::Model>> _model_pool;
```
lindsayad, line 79: Is this a possibility?
```suggestion
  thread_local std::shared_ptr<neml2::Model> _model_pool;
```

framework/include/neml2/interfaces/NEML2ModelInterface.h
```diff
@@ -156,22 +179,24 @@ NEML2ModelInterface<T>::NEML2ModelInterface(const InputParameters & params, P &&
...
+      _model_pool[std::this_thread::get_id()] = std::move(model);
     };
 
     _dispatcher = std::make_unique<DispatcherType>(
         *_scheduler,
         _async_dispatch,
         [&](neml2::ValueMap && x, neml2::Device device) -> RJType
         {
-          auto & model = NEML2Utils::getModel(_model.name());
+          auto & model = _async_dispatch ? _model_pool[std::this_thread::get_id()] : _model;
```
lindsayad: If you're sticking with the map
```suggestion
          auto & model = _async_dispatch ? libmesh_map_find(_model_pool, std::this_thread::get_id()) : _model;
```

framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -208,80 +210,92 @@ NEML2ModelExecutor::initialize()
...
-  _model_params.clear();
-
-  // Request gradient for the model parameters that we request AD for
-  for (const auto & [y, dy] : _retrieved_parameter_derivatives)
-    for (const auto & [p, tensor] : dy)
-      model().get_parameter(p).requires_grad_(true);
+  try
+  {
+    for (const auto & uo : _gatherers)
+      uo->insertInto(_in, _model_params);
```
lindsayad, line 250: Is `_model_params` used outside of this method? Just wondering if it makes sense to use a local variable instead

framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -208,80 +210,92 @@ NEML2ModelExecutor::initialize()
...
+    if (!model().input_axis().has_state())
+      return;
+    if (!model().input_axis().has_old_state())
+      return;
+
+    // Set trial state variables (i.e., initial guesses).
+    // Right now we hard-code to use the old state as the trial state.
+    // TODO: implement other predictors
+    const auto input_state = model().input_axis().subaxis(neml2::STATE);
+    const auto input_old_state = model().input_axis().subaxis(neml2::OLD_STATE);
```
lindsayad: Is there a reason you're catching this by value instead of by reference?

framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -299,64 +313,111 @@ NEML2ModelExecutor::expandInputs()
...
+               e.what(),
+               NEML2Utils::NEML2_help_message);
+  }
+}
+
+void
+NEML2ModelExecutor::finalize()
+{
+  // See if any rank failed
+  unsigned int pid;
```
lindsayad:
```suggestion
  processor_id_type pid;
```

framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -299,64 +313,111 @@ NEML2ModelExecutor::expandInputs()
...
+               "step.";
+      _console << COLOR_YELLOW << msg << COLOR_DEFAULT << std::endl;
     }
+    _fe_problem.setFailNextNonlinearConvergenceCheck();
+  }
 
-  // clear derivatives
-  _dout_din.clear();
+  if (_t_step > 0)
+    _output_ready = true;
```
lindsayad, line 423: Is output ready if we had an `_error`?

framework/include/neml2/interfaces/NEML2ModelInterface.h
hugary1995: We are spawning threads so probably not.

framework/src/neml2/userobjects/NEML2ModelExecutor.C
hugary1995: Yes, there will be user objects gathering data from moose and fill them into this member. They need a reference to this member.

framework/src/neml2/userobjects/NEML2ModelExecutor.C
hugary1995: Not at all. Good catch.

framework/src/neml2/userobjects/NEML2ModelExecutor.C
hugary1995: Nice catch. Previously I was throwing an exception, and when I switched to `_fe_problem.setFailNextNonlinearConvergenceCheck();` I forgot to also change this logic.

framework/include/neml2/interfaces/NEML2ModelInterface.h
lindsayad: But I noticed that you populate the map with `std::this_thread::get_id()` in the constructor

lindsayad:
I'm curious about the branch name 😄

hugary1995:
I was about to continue on this branch to stitch the two libraries. That should be possible with Intel's oneAPI + Intel's accelerator because they support the so-called unified memory management. That is, I should not need to copy anything when transferring from a mfem data structure to a neml2 data structure, in theory. That's attractive to me because then I can do the full solid mechanics solve with our material models on GPU. With other accelerators such as CUDA a copy will be required -- which could still be fine.

That being said, I read through the mfem-moose source code a few days ago, I think we are not quite ready for that yet on the moose side. See the question I asked a few months back: https://github.com/orgs/mfem/discussions/4749, the mfem developers said a new capability called DifferentiableOperator would make things much easier.

Another possibility is to directly interface with SYCL. Again, that's easier with Intel's extension to pytorch because it's also using SYCL (and DPC++) behind the scenes. I heard @NamjaeChoi have been working on a SYCL moose, but I really don't know where to start looking.

lindsayad:
He's been working on a [kokkos MOOSE](https://github.com/namjaechoi/moose/tree/kokkos)

framework/include/neml2/interfaces/NEML2ModelInterface.h
hugary1995: In the constructor, that's used inside a functor which is passed to the constructor of the neml2 dispatcher. The dispatcher spawns thread and runs the functor in each spawned thread.

hugary1995:
@lindsayad do you happen to know what portable programming models (if any) mfem is using? Is it also kokkos, or maybe something else, RAJA or OCCA?

lindsayad:
> https://github.com/orgs/mfem/discussions/4749

So if MFEM is the finite element library and it provides your strain vector and NEML2 provides the stress-strain relationship, what role do you see MOOSE playing, if any?

lindsayad:
From the MFEM device doxygen

> The MFEM [Device](https://docs.mfem.org/4.1/classmfem_1_1Device.html) class abstracts hardware devices such as GPUs, as well as programming models such as CUDA, OCCA, RAJA and OpenMP

NamjaeChoi:
SYCL is one of the backends of Kokkos, so me working on SYCL is not completely incorrect, but I'm not sure how well SYCL is supported. It's technically a standard programming model but only actively supported by Intel GPUs.

hugary1995:
> So if MFEM is the finite element library and it provides your strain vector and NEML2 provides the stress-strain relationship, what role do you see MOOSE playing, if any?

I'd like to keep using MOOSE for its declarative input file support. That's really the only practical way for me to get a new student (without FEM or even C++ experience) up to speed within a reasonable amount of time. It's also the only practical way for me to package an application as a deliverable for industry to use.

The reason I'm exploring MFEM-NEML2 integration is that you told us the MOOSE team is unsure about what approach(s) to use to take the framework onto GPUs. You mentioned both approaches, MFEM and kokkos, will be explored. At the end of the day, more likely than not, when the MOOSE community favors one over the other, I will add the corresponding support in NEML2, i.e., as an alternative tensor backend to libtorch. But due to the current uncertainty, I need to explore both.

hugary1995:
A third option is to migrate NEML2 entirely into MOOSE (not just as a submodule, but embracing the framework as a whole). I'm just not a big fan of the MOOSE material system for solid mechanics.

lindsayad:
> I will add the corresponding support in NEML2, i.e., as an alternative tensor backend to libtorch

My understanding is that NEML2 can use torch at the backend, so should this read

>  I will add the corresponding support in NEML2, a frontend for using torch tensors for constitutive material models

? Now that NEML2 is in the framework I want to get to know it better, and I want to think seriously about how all our libraries that support GPUs can potentially integrate

lindsayad:
> A third option is to migrate NEML2 entirely into MOOSE (not just as a submodule, but embracing the framework as a whole). I'm just not a big fan of the MOOSE material system for solid mechanics.

I think you could both have NEML2 remain a submodule *and* we explore how to make it the backbone of materials in MOOSE. The former would you give you more flexibility if you wanted to retain the ability to have it work with software other than MOOSE

lindsayad:
I can't remember if I've spoken with you directly, but @NamjaeChoi and I have certainly had conversations with @dschwen in which I've talked about exploring making NEML2 our future material system. The initial kokkos implementation that @NamjaeChoi put together doesn't do that in the interest of rolling something out a little bit quicker, but I think we could start looking seriously at this next FY

hugary1995:
> > I will add the corresponding support in NEML2, i.e., as an alternative tensor backend to libtorch
> 
> My understanding is that NEML2 can use torch at the backend, so should this read
> 
> > I will add the corresponding support in NEML2, a frontend for using torch tensors for constitutive material models
> 
> ? Now that NEML2 is in the framework I want to get to know it better, and I want to think seriously about how all our libraries that support GPUs can potentially integrate

Yes, spare my english. NEML2 tensors are just thin wrappers around torch tensors, no magic.

NEML2 really comes with 5 components, functionality-wise:
1. Input file, factory, registry, etc. which MOOSE already has.
2. A tensor backend, as I said, it's just a thin wrapper around torch tensor. we can wrap around another tensor library, e.g., some tensor library written to better support integration with MOOSE.
3. Models, and how models are composed together.
4. The work scheduler and dispatcher.
5. Companion python bindings for training NEML2 models.

hugary1995:
> I think we could start looking seriously at this next FY.

We have an item in the IPL under the SMC technical area to support that. But I'm not sure which FY we put down.

NamjaeChoi:
I will need to see the memory layout of the underlying torch tensors, but if they are regular and contiguous (with the assumption that qp is the outermost), it will be easy to wrap in Kokkos too.

hugary1995:
> I will need to see the memory layout of the underlying torch tensors, but if they are regular and contiguous (with the assumption that qp is the outermost), it will be easy to wrap in Kokkos too.

For the majority of the NEML2 tensors in MOOSE, yes, the layout is contiguous and qp is the outermost tensor dimension. The remaining dimensions are shape of the tensor for each qp.

hugary1995:
@loganharbour and I discussed offline about some build issues. I decided to add `--with-neml2` as a configure option. I'm not terribly familiar with autoconf m4 etc., so I just tried to follow the pattern of --with-libtorch and --with-mfem. I then autoreconf'ed to regenerate several files. Hope I did things correctly. See this commit: https://github.com/idaholab/moose/pull/30799/commits/0949efe7e0dc305b7b3a5ed48e69682957bdd5a8

Maybe a third pair of eyes @lindsayad ?

hugary1995:
@loganharbour I did another submodule update to bring in the `hit::explode` change you made yesterday.

hugary1995:
oh of course the sha isn't trusted...

hugary1995:
Some civet recipes need to change to add a step for "configure neml2"... Breaking next might be inevitable?

loganharbour:
> Some civet recipes need to change to add a step for "configure neml2"... Breaking next might be inevitable?

Wow. First time those "check capabilities" steps actually came in handy! 

Good news is that configure doesn't return non-zero with unused arguments, we just get:

```
configure: WARNING: unrecognized options: --with-neml2
```

I'll make the recipe change.

loganharbour:
Can @roystgnr or @lindsayad take a look at the configure changes in https://github.com/idaholab/moose/commit/0949efe7e0dc305b7b3a5ed48e69682957bdd5a8?

hugary1995:
The phase field threaded timeout seems to be persistent. https://civet.inl.gov/job/2972351/

hugary1995:
> The phase field threaded timeout seems to be persistent. https://civet.inl.gov/job/2972351/

though unrelated to this PR

roystgnr:
I'm not seeing any obvious problems with the `configure` changes.  m4+sh is a great language mishmash in which to hide subtle problems, though.  ;-)

loganharbour:
Failures are unrelated.

loganharbour (approved):

loganharbour:
Still need to figure out next before merging.

lindsayad:
configure changes look good to me as well

PR #30812, opened 2025-06-20
Add projection option to param mesh functions

### LINKED ISSUE #30796: Parameter mesh projection to closest point
### Motivation

For optimization we want the ability to have parameter mesh that may be a dimension lower or not fill the entire simulation domain. In this case, we will project to the closest point on the parameter mesh and use that to determine the parameter value.

### Design

At first limit this to linear edge, quad, and hex parameter meshes for easy projections. 

### Impact

Will allow for better parameter meshes that are only refined where it needs to be.

maxnezdyur:
closes #30796


modules/optimization/src/executioners/AdjointSolve.C
```diff
@@ -89,12 +89,15 @@ AdjointSolve::solve()
...
 
   _problem.execute(OptimizationAppTypes::EXEC_ADJOINT_TIMESTEP_BEGIN);
+
   if (!_problem.execMultiApps(OptimizationAppTypes::EXEC_ADJOINT_TIMESTEP_BEGIN))
   {
     _console << "MultiApps failed to converge on ADJOINT_TIMESTEP_BEGIN!" << std::endl;
     return false;
   }
 
+  _problem.outputStep(OptimizationAppTypes::EXEC_ADJOINT_TIMESTEP_BEGIN);
```
lynnmunday, line 99: Add a code comment why you are doing this here.  I forgot but I don't think its related to the parameterMesh changes.

modules/optimization/include/utils/ParameterMesh.h
```diff
@@ -77,10 +81,35 @@ class ParameterMesh
 protected:
   libMesh::Parallel::Communicator _communicator;
   libMesh::ReplicatedMesh _mesh;
+  /// Find closest projection points
+  bool _find_closest;
```
lynnmunday: const bool _find_closest

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +214,90 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+  {
+
+    Point trial;
+    switch (e->type())
+    {
+      case EDGE2:
+
+        trial = closestPointonEdge(static_cast<const Edge2 *>(e), p);
+        break;
+      case QUAD4:
```
lynnmunday: Would it be hard to add triangles or does it already support triangles?  I thought quad4 usually called through to triangle plane surfaces.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +214,90 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+{
+  Point a = *(elem->node_ptr(0));
+  Point b = *(elem->node_ptr(1));
+  Point c = *(elem->node_ptr(2));
+  Point d = *(elem->node_ptr(3));
+  Plane pl1(a, b, c);
+  Plane pl2(b, c, d);
+  Point trial1 = pl1.closest_point(p);
+  Point trial2 = pl2.closest_point(p);
+  if (!trial1.absolute_fuzzy_equals(trial2, TOLERANCE * TOLERANCE))
```
lynnmunday: Is this a common error?  Is this for a quad4 that is not completely flat.  Similar code might be in the ray tracing kernel.  We could look at using it if it is merged into libmesh.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +214,90 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+{
+
+  LineSegment ls(*(elem->node_ptr(0)), *(elem->node_ptr(1)));
+  return ls.closest_point(p);
+}
+
+Point
+ParameterMesh::closestPointonQuad(const Quad4 * elem, Point p) const
+{
+  Point a = *(elem->node_ptr(0));
```
lynnmunday: this could be broken down into two calls to a triangle surface and then triangle elements would work.  If it already works with triangles, nevermind, but I don't see how it could.  If it doesn't work with triangles, I think we should make an issue and work on that in a seperate PR later.

modules/optimization/test/tests/functions/param_mesh_closest/create_mesh_1d.i
```diff
@@ -0,0 +1,52 @@
...
+    dim = 1
+    nx = 20
+  []
+  [rot]
+    type = TransformGenerator
+    input = gmg
+    transform = ROTATE
+    vector_value = '10 10 10'
+  []
+  parallel_type = REPLICATED
```
lynnmunday, line 13: Do we check in parameterMesh that it is replicated.  What kind of error does it throw for distributed?  I have no idea.

modules/optimization/test/tests/functions/param_mesh_closest/tests
```diff
@@ -0,0 +1,72 @@
...
+      type = CheckFiles
+      input = create_mesh_2d.i
+      check_files = 'create_mesh_2d_out.e create_mesh_2d_out_param_vec_0001.csv'
+      detail = 'creating 2D mesh and'
+    []
+    [evaluate_2d]
+      type = Exodiff
+      input = parameter_mesh_2d.i
+      allow_test_objects = True
+      prereq = parameter_mesh/create_2d
```
lynnmunday, line 29: These could use the gold files and not use prereq.

modules/optimization/test/tests/functions/param_mesh_closest/tests
```diff
@@ -0,0 +1,72 @@
+[Tests]
+  issues = '#30796'
+  design = 'ParameterMeshFunction.md'
+  [parameter_mesh]
+    requirement = 'The system shall be able to evaluate a mesh-based interpolation function by '
+    [create_1d]
+      type = CheckFiles
+      input = create_mesh_1d.i
+      check_files = 'create_mesh_1d_out.e create_mesh_1d_out_param_vec_0001.csv'
```
lynnmunday, line 9: What is the csv file used for?

modules/optimization/test/tests/functions/param_mesh_closest/tests
```diff
@@ -0,0 +1,72 @@
...
+      allow_test_objects = True
+      prereq = check_dbgs/create_second
+      expect_err = "Closet point projection option cannot work with second order elements."
+      detail = 'with closest point projection;'
+      recover = false
+    []
+    [create_tri]
+      type = CheckFiles
+      input = create_mesh_2d.i
+      cli_args = 'Mesh/gmg/elem_type=TRI'
```
lynnmunday: OK!   It doesn't support triangle meshes.  Ignore my previous comments about triangles but they would be easy to add.

modules/optimization/test/tests/functions/param_mesh_closest/tests
```diff
@@ -0,0 +1,72 @@
...
+      check_files = 'create_mesh_2d_out.e create_mesh_2d_out_param_vec_0001.csv'
+      detail = 'using triangle element meshes'
+      recover = false
+    []
+    [evaluate_tri]
+      type = RunException
+      input = parameter_mesh_2d.i
+      allow_test_objects = True
+      prereq = check_dbgs/create_tri
+      expect_err = "Unsupported element type 3 for projection of parameter mesh."
```
lynnmunday: I guess e->type() return an enum that is 3 for triangles, not very helpful error message.  Maybe saying "only edge2 and quad4" are supported element types.

lynnmunday (changes requested):
I thought this worked for a 3D parameter mesh where it projects all of the values from the edge of the 3D parameter mesh to the rest of the FEM volumetric mesh.  What happens if someone tries to project from a 3D parameter mesh to a 2D mesh, would it work?  Does this work for a cube parameter mesh projecting onto a sphere FEM mesh?  Does the sphere need to be totally inside the parameter or can it some of it go outside.   There are lots of corner cases that I think we should document in parameterMeshFunction.md but not implement.  The MD says it works with hex, you should add a test for a volumetric parameter mesh projecting onto a volume.

modules/optimization/test/tests/functions/param_mesh_closest/tests
maxnezdyur: The csv file is used to "parameter" points and values.

modules/optimization/test/tests/functions/param_mesh_closest/tests
maxnezdyur: I prefer this because over time when we add more types like hex, tet, triangle we would have to continue to keep changing some tests.

modules/optimization/src/utils/ParameterMesh.C
maxnezdyur: agreed

modules/optimization/src/utils/ParameterMesh.C
maxnezdyur: While I agree that this isn't common, this is method would fail if we are not coplanar.

modules/optimization/test/tests/functions/param_mesh_closest/create_mesh_1d.i
maxnezdyur: Yes parameterMesh makes sure that the mesh is "libMesh::ReplicatedMesh". We read all the meshes in with exodus so has to be replicated.

modules/optimization/test/tests/functions/param_mesh_closest/tests
maxnezdyur: This is how the other tests do it. I perfer it because it let's me change the element types easily for the errors below.

modules/optimization/test/tests/functions/param_mesh_closest/tests
lynnmunday: You should not print out the enum in the error message if it doesn't have any meaning.

maxnezdyur: fixed this to print out the string. I finally found a utility that did this.

maxnezdyur:
> I thought this worked for a 3D parameter mesh where it projects all of the values from the edge of the 3D parameter mesh to the rest of the FEM volumetric mesh. What happens if someone tries to project from a 3D parameter mesh to a 2D mesh, would it work? Does this work for a cube parameter mesh projecting onto a sphere FEM mesh? Does the sphere need to be totally inside the parameter or can it some of it go outside. There are lots of corner cases that I think we should document in parameterMeshFunction.md but not implement. The MD says it works with hex, you should add a test for a volumetric parameter mesh projecting onto a volume.

This works for 3d parameter mesh with 3d simulation mesh. This works for 3d parameter mesh with 2d simulation mesh. Cube to sphere should work also it's the same as 3d parameter mesh and 3d simulation mesh. Sphere doesn't need to be enclosed.

modules/optimization/src/functions/ParameterMeshFunction.C
```diff
@@ -33,14 +33,19 @@ ParameterMeshFunction::validParams()
   params.addParam<ReporterName>("time_name",
                                 "Name of vector-postprocessor or reporter vector containing time, "
                                 "default assumes time independence.");
+  params.addParam<bool>(
+      "project_points", false, "Wether to find the closest point on parameter mesh.");
```
zachmprince:
```suggestion
      "project_points", false, "Whether to find the closest point on parameter mesh.");
```

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -93,17 +109,25 @@ ParameterMesh::ParameterMesh(const FEType & param_type,
   std::set<dof_id_type> var_indices;
   _sys->local_dof_indices(var_id, var_indices);
   _param_dofs = var_indices.size();
+
+  if (_find_closest)
+  {
+    for (const auto & elem : _mesh.element_ptr_range())
+      if (elem->default_order() != FIRST)
+        mooseError("Closet point projection option cannot work with second order elements.");
```
zachmprince:
```suggestion
        mooseError("Closet point projection currently does not support second order elements.");
```

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,128 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+  Plane pl2(b, c, d);
+  Point trial1 = pl1.closest_point(p);
+  Point trial2 = pl2.closest_point(p);
+  if (!trial1.absolute_fuzzy_equals(trial2, TOLERANCE * TOLERANCE))
+    mooseError("Quad4 element is not coplanar");
+
+  if (elem->contains_point(trial1))
+    return trial1;
+  // Initialize min_distance to infinity
+  double min_distance = std::numeric_limits<double>::max();
```
zachmprince:
```suggestion
  Real min_distance = std::numeric_limits<double>::max();
```

lynnmunday (approved):
I approve.  Thanks for adding the volume element.

lynnmunday:
@zachmprince Will you review and merge.  I reviewed and I approve.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,128 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+  // Initialize min_point to store the point with the smallest distance
+  Point min_point = p;
+
+  for (dof_id_type i : make_range(elem->n_edges()))
+  {
+    std::unique_ptr<const Elem> edge = elem->build_edge_ptr(i);
+    const Edge2 * edge_ptr = static_cast<const Edge2 *>(edge.get());
```
zachmprince:
```suggestion
    const Edge2 * edge_ptr = dynamic_cast<const Edge2 *>(edge.get());
```

zachmprince: Or you can put an assertion if you're worried about overhead:
```suggestion
    mooseAssert(dynamic_cast<const Edge2 *>(edge.get()), "...");
    const Edge2 * edge_ptr = static_cast<const Edge2 *>(edge.get());
```

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,128 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+  // Return the point with the smallest distance
+  return min_point;
+}
+
+Point
+ParameterMesh::closestPointonHex(Hex8 * elem, Point p) const
+{
+
+  // Initialize min_distance to infinity
+  double min_distance = std::numeric_limits<double>::max();
```
zachmprince:
```suggestion
  Real min_distance = std::numeric_limits<double>::max();
```

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,128 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+      }
+    }
+  }
+
+  // Return the point with the smallest distance
+  return min_point;
+}
+
+Point
+ParameterMesh::closestPointonHex(Hex8 * elem, Point p) const
```
zachmprince: Why can't `Hex8` be `const` here?

modules/optimization/include/utils/ParameterMesh.h
```diff
@@ -77,10 +81,37 @@ class ParameterMesh
...
+  Point closestPointonEdge(const Edge2 * elem, Point p) const;
+  /**
+   * Returns the closet point on a Quad4 element from a given point.
+   * @param elem
+   * @param p
+   * @return Point
+   */
+  Point closestPointonQuad(const Quad4 * elem, Point p) const;
+
+  Point closestPointonHex(Hex8 * elem, Point p) const;
```
zachmprince: I think the `elem` arguments in these guys should be a reference.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,128 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+Point
+ParameterMesh::projectToMesh(const Point & p) const
+{
+  // quick path: p already inside an element
+  if ((*_point_locator)(p))
+    return p;
+
+  Real best_d2 = std::numeric_limits<Real>::max();
+  Point best_point = p;
+  for (auto e : _mesh.element_ptr_range())
```
zachmprince: I'm curious on the performance of this. Would it be worth looking into optimizing a bit here? Seems like the number of elements searched could be reduced by skipping internal elements, the points could be cached into a map look up, and/or this loop could be parallelized.

zachmprince (changes requested):
Very cool work! A few requests and general commend/suggestion.

modules/optimization/src/utils/ParameterMesh.C
lynnmunday:
```suggestion
  Real min_distance = std::numeric_limits<Real>::max();
```

modules/optimization/src/utils/ParameterMesh.C
maxnezdyur: So this mesh is the parameter mesh which is generally a lot smaller then the simulation mesh. By order of magnitude usually. In an optimization problem we usually only solve problems in the hundreds of parameters. We can't skip internal elements, if you look at the 2d test, we have a 2d mesh embeddded in a 3d domain. This means that we have have projections that are in the interior of the 2d mesh. Only if the case of a 3d parameter mesh can we gaurentee we will only project onto the surface. Cacheing the location of each projeciton for each quadrature point in the simulation domain would be excessive in my opinion. I am not sure what you mean by this loop can be parallelized? This function would be called within a functionAux which would be threaded already. The main way to speed this up is to have a KDtree. I would prefer to add this functionality first with a suite of tests, since the projectToMesh option is off by default. Then we can add more optimizations when we find that is is a bottle neck. In general, I feel that when there is a good utility in MOOSE to determine the closest point on a mesh to a point not in the mesh then a lot of the code here can be scrapped.

modules/optimization/src/utils/ParameterMesh.C
maxnezdyur: I fixed this with a different method.

Talking to roy this might be a libmesh bug that is getting fixed

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -263,82 +246,119 @@ ParameterMesh::projectToMesh(const Point & p) const
...
+      Point c = *(elem.node_ptr(2));
+      Point d = *(elem.node_ptr(3));
+      Plane pl1(a, b, c);
+      Plane pl2(b, c, d);
+      Point trial1 = pl1.closest_point(p);
+      Point trial2 = pl2.closest_point(p);
+      if (!trial1.absolute_fuzzy_equals(trial2, TOLERANCE * TOLERANCE))
+        mooseError("Quad4 element is not coplanar");
+
+      if (elem.contains_point(trial1))
```
lynnmunday, line 376: You need to check trial2!

Can you set this up to call back on tri for each plane in the quad, then it will return a bool, like found_point if it is inside the plane, false if outside but it will always return a distance.   Then you could have an early return from each tri.  

Also have tri call back on edge for the edge search.

lynnmunday (changes requested):
I like this version better.

modules/optimization/src/utils/ParameterMesh.C
maxnezdyur: No? I have a check just above that confirms that trial1 and trial2 are the same point to a tolerance. If they were not then the two triangles formed from the quad would not be coplanar

lynnmunday: I was wrong.   I didn't understand the code.  I thought it was projecting back to the triangle and if it is outside the triangle it was returning an edge distance.   I was also wrong about calling back on the triangles from quad, that is a waste.

zachmprince (approved):
Great work @maxnezdyur!!

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,148 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+        }
+      }
+      return min_point;
+    }
+
+    default:
+    {
+      if (elem.dim() == 3)
+      {
+        Real min_distance = std::numeric_limits<Real>::max();
```
lynnmunday: add assert if the point is in the volume element.  I know you check for it above but we might change this code later and will miss this case.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,148 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+    mooseError("project_to_mesh failed - no candidate elements.");
+  return best_point;
+}
+
+Point
+ParameterMesh::closestPoint(const Elem & elem, const Point & p) const
+{
+  switch (elem.type())
+  {
+    case EDGE2:
```
lynnmunday, line 344: I think this works for higher order elements so I guess you can add those too. I don't think you need to check that the higher order edge element isn't curved or maybe assert that it isn't curved.  We can assume they aren't.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +217,148 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+    {
+      const Edge2 * edge_ptr = dynamic_cast<const Edge2 *>(&elem);
+      if (!edge_ptr)
+        mooseError("Cannot cast element to Edge2.");
+      LineSegment ls(*(edge_ptr->node_ptr(0)), *(edge_ptr->node_ptr(1)));
+      return ls.closest_point(p);
+    }
+
+    case TRI3:
+    {
```
lynnmunday, line 351: If we add higher order elements, you can also assert the higher order tris are not curved some how or just assume they aren't.  I think the ordering is still corner nodes are the first three.

lynnmunday (approved):
i left some comments if you did make it work for higher order elements.  If you don't then I approve as is.

maxnezdyur:
@zachmprince 
Can you look at this again.
I did a rewrite because I just wanted to get the kdtree in now when I still had the process in my head.

modules/optimization/src/utils/ParameterMesh.C
maxnezdyur: as per our talk, higher order elements won't work because of curves

PR #30833, opened 2025-06-23
Another round of mfem fixups

nmnobre:
## Reason
Tests are failing with `MFEM_DEBUG = ON`. Some documentation is incorrect. I like nice defaults.

## Design
Bug fixes, small amendments to existing code.

## Impact
The tests shall pass, the docs shall be slightly better, and the user shall have to type a bit less when writing inputs.

### DIFF
--- a/framework/doc/content/source/mfem/bcs/MFEMBoundaryIntegratedBC.md
+++ b/framework/doc/content/source/mfem/bcs/MFEMBoundaryIntegratedBC.md
@@ -1,10 +1,10 @@
-# MFEMScalarBoundaryIntegratedBC
+# MFEMBoundaryIntegratedBC
 
 !if! function=hasCapability('mfem')
 
 ## Summary
 
-!syntax description /BCs/MFEMScalarBoundaryIntegratedBC
+!syntax description /BCs/MFEMBoundaryIntegratedBC
 
 ## Overview
 
@@ -16,11 +16,11 @@ Adds the boundary integrator for integrating the linear form
 where the test variable $v \in H^1$ and $f$ is a scalar coefficient. Often used for representing
 Neumann-type boundary conditions.
 
-!syntax parameters /BCs/MFEMScalarBoundaryIntegratedBC
+!syntax parameters /BCs/MFEMBoundaryIntegratedBC
 
-!syntax inputs /BCs/MFEMScalarBoundaryIntegratedBC
+!syntax inputs /BCs/MFEMBoundaryIntegratedBC
 
-!syntax children /BCs/MFEMScalarBoundaryIntegratedBC
+!syntax children /BCs/MFEMBoundaryIntegratedBC
 
 !if-end!
 
--- a/framework/doc/content/source/mfem/bcs/MFEMBoundaryNormalIntegratedBC.md
+++ b/framework/doc/content/source/mfem/bcs/MFEMBoundaryNormalIntegratedBC.md
@@ -1,10 +1,10 @@
-# MFEMVectorNormalIntegratedBC
+# MFEMBoundaryNormalIntegratedBC
 
 !if! function=hasCapability('mfem')
 
 ## Summary
 
-!syntax description /BCs/MFEMVectorNormalIntegratedBC
+!syntax description /BCs/MFEMBoundaryNormalIntegratedBC
 
 ## Overview
 
@@ -16,11 +16,11 @@ Adds the boundary integrator for integrating the linear form
 where $v \in H^1$, $\vec f$ is a vector coefficient, and $\hat n$ is the outward facing unit normal
 vector on the boundary.
 
-!syntax parameters /BCs/MFEMVectorNormalIntegratedBC
+!syntax parameters /BCs/MFEMBoundaryNormalIntegratedBC
 
-!syntax inputs /BCs/MFEMVectorNormalIntegratedBC
+!syntax inputs /BCs/MFEMBoundaryNormalIntegratedBC
 
-!syntax children /BCs/MFEMVectorNormalIntegratedBC
+!syntax children /BCs/MFEMBoundaryNormalIntegratedBC
 
 !if-end!
 
--- a/framework/src/mfem/bcs/MFEMBoundaryIntegratedBC.C
+++ b/framework/src/mfem/bcs/MFEMBoundaryIntegratedBC.C
@@ -20,8 +20,8 @@ MFEMBoundaryIntegratedBC::validParams()
   params.addClassDescription("Adds the boundary integrator to an MFEM problem for the linear form "
                              "$(f, v)_\\Omega$ "
                              "arising from the weak form of the forcing term $f$.");
-  params.addRequiredParam<MFEMScalarCoefficientName>(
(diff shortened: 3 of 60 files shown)


framework/doc/content/source/mfem/bcs/MFEMBoundaryIntegratedBC.md
```diff
@@ -1,10 +1,10 @@
-# MFEMScalarBoundaryIntegratedBC
+# MFEMBoundaryIntegratedBC
```
lindsayad, line 1: I should have commented on this before. Why have `Boundary` in the name with the `BC` suffix?

framework/doc/content/source/mfem/bcs/MFEMBoundaryNormalIntegratedBC.md
```diff
@@ -1,10 +1,10 @@
-# MFEMVectorNormalIntegratedBC
+# MFEMBoundaryNormalIntegratedBC
```
lindsayad, line 1: Same question about having `Boundary` in the name

framework/src/mfem/kernels/MFEMKernel.C
```diff
@@ -41,7 +41,8 @@ MFEMKernel::MFEMKernel(const InputParameters & parameters)
     _subdomain_attributes[i] = std::stoi(_subdomain_names[i]);
   }
   mfem::ParMesh & mesh(getMFEMProblem().mesh().getMFEMParMesh());
-  mfem::common::AttrToMarker(mesh.attributes.Max(), _subdomain_attributes, _subdomain_markers);
+  if (_subdomain_names.size())
```
lindsayad:
```suggestion
  if (_subdomain_attributes.size())
```

unit/src/MFEMMaterialTest.C
```diff
@@ -64,12 +64,8 @@ check_vector(mfem::VectorCoefficient & coef,
   mfem::Vector vec;
   coef.Eval(vec, fe_transform, point);
   int errors = 0;
-  if (vec[0] != expected[0])
-    errors++;
-  if (vec[1] != expected[1])
-    errors++;
-  if (vec[2] != expected[2])
-    errors++;
+  for (const auto i : make_range(vec.Size()))
```
lindsayad, line 67: Sad that we can't use `index_range` because of `Size()` vs `size()` 😦

framework/doc/content/source/mfem/bcs/MFEMBoundaryIntegratedBC.md
nmnobre: Yeah, I hear you, the idea is to be consistent with MFEM integrators names as much as possible.

From @alexanderianblair, "the rule that the root of the name should come from the MFEM integrator should still apply. For kernels, `XXXXIntegrator` becomes `MFEMXXXKernel`, and for BCs, `XXXBoundaryLFIntegrator` should become `MFEMXXXBoundaryIntegratedBC`".

But perhaps we could change the transformation rule such that `XXXBoundaryLFIntegrator` simply becomes `MFEMXXXIntegratedBC`?

unit/src/MFEMMaterialTest.C
nmnobre: ¯\\\_(ツ)\_/¯

framework/src/mfem/kernels/MFEMKernel.C
nmnobre: `_subdomain_attributes` is an `mfem::Array` so I can't directly apply this patch for the same camelCase vs PascalCase reasons as below, but I've nonetheless changed the check to be on `_subdomain_attributes` instead since it does make more sense: `if (!_subdomain_attributes.IsEmpty())`.

framework/doc/content/source/mfem/bcs/MFEMBoundaryIntegratedBC.md
alexanderianblair: The intended scheme is indeed exactly as Nuno's said, which currently includes `Boundary`, to keep things as systematic as possible for mapping from the names of the available `mfem::LinearFormIntegrator` types. 

> But perhaps we could change the transformation rule such that XXXBoundaryLFIntegrator simply becomes MFEMXXXIntegratedBC?

I don't hate this idea (it does make names shorter!) - my main concern is that it still obfuscates the mapping between MFEM integrator names and the corresponding BCs. For example `mfem::BoundaryFlowIntegrator` would become `MFEMFlowIntegratedBC` and `mfem::BoundaryNormalIntegrator` would become `MFEMNormalIntegratedBC`, which I think makes the naming convention less obvious for a new user.

lindsayad (approved):

loganharbour:
Do we need a build with `MFEM_DEBUG`?

alexanderianblair:
> Do we need a build with `MFEM_DEBUG`?

I think this would be a good idea - there are a number of MFEM asserts that this would catch (and not all of them cause failing behaviour when built in opt)

loganharbour:
What are the cmake args?

nmnobre:
> What are the cmake args?

`-DCMAKE_BUILD_TYPE=Debug`

framework/src/mfem/bcs/MFEMBoundaryIntegratedBC.C
```diff
@@ -9,40 +9,39 @@
...
+  params.addParam<MFEMScalarCoefficientName>(
+      "coefficient", "1.", "The coefficient which will be used in the integrated BC");
   return params;
 }
 
-MFEMScalarBoundaryIntegratedBC::MFEMScalarBoundaryIntegratedBC(const InputParameters & parameters)
+MFEMBoundaryIntegratedBC::MFEMBoundaryIntegratedBC(const InputParameters & parameters)
   : MFEMIntegratedBC(parameters),
-    _coef_name(getParam<MFEMScalarCoefficientName>("coefficient")),
-    _coef(getScalarCoefficient(_coef_name))
```
loganharbour, line 31: This is an aside - but is there a reason the `getScalarCoefficient` method doesn't behave like the majority of our interfaces?

Most `getXXX` methods in interfaces act on a parameter name, instead of a proper name. They query the parameter for you. Then we have `getXXXByName` if you want to specify the name directly.

nmnobre: Are you suggesting that `getScalarCoefficient` should be called `getScalarCoefficientByName`?

loganharbour: Yeah, that's what I think. and `getScalarCoefficient("coefficient")` should be the use case here.

nmnobre: I'll add that to my list. :) - This PR includes important bug fixes plus it's blocking another PR, so I'd like it merged soon.

loganharbour: I wouldn't really consider it to be a priority either

PR #30842, opened 2025-06-24
Formatted table speedup

roystgnr:
Refs #30831

## Reason
Makes console output with very large data tables much faster.

## Design
We were spending 90%+ of runtime uselessly in some (ODE with a zillion time step) cases

## Impact
This adds an optional parameter to an existing API, though most of the speedup is just done by reducing the amount of string hashing we do regardless of the parameter's value.

This could probably be made faster still, but with much more intrusive code changes, so I want to see how this benchmarks for @Naktakala first.  If it's fast enough as-is then we can just declare victory and close the issue.

### DIFF
--- a/framework/include/utils/FormattedTable.h
+++ b/framework/include/utils/FormattedTable.h
@@ -279,8 +279,13 @@ private:
   void printRow(std::pair<Real, std::map<std::string, std::shared_ptr<TableValueBase>>> & row_data,
                 bool align);
 
-  /// Fill any values that are not defined (usually when there are mismatched column lengths)
-  void fillEmptyValues();
+  /**
+   *  Fill any values that are not defined (usually when there are mismatched column lengths)
+   *
+   *  If \p last_n_entries is non-zero, only values in that many final
+   *  rows will be examined and filled.
+   */
+  void fillEmptyValues(unsigned int last_n_entries = 0);
 
   /// The optional output file stream
   std::string _output_file_name;
--- a/framework/src/utils/FormattedTable.C
+++ b/framework/src/utils/FormattedTable.C
@@ -310,8 +310,7 @@ FormattedTable::printTablePiece(std::ostream & out,
   printRowDivider(out, col_widths, col_begin, col_end);
   out << "|";
   if (_output_time)
-    out << std::setw(_column_width) << std::left << " time"
-        << " |";
+    out << std::setw(_column_width) << std::left << " time" << " |";
   for (auto header_it = col_begin; header_it != col_end; ++header_it)
     out << " " << std::setw(col_widths[*header_it]) << *header_it << "|";
   out << "\n";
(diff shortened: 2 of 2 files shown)


roystgnr:
This seems to work great on Jan's test case:
```
[Mesh]
  [gmg]
    type = GeneratedMeshGenerator
    dim = 1
    nx = 1
  []
[]

[Variables]
  [n]
    family = LAGRANGE
    order = first
    initial_condition = 1
  []
[]

[Kernels]
  [nada_n] 
    type = NullKernel
    variable = n
  []
[]

[Postprocessors]
  [n01] type = ElementAverageValue variable = n []
  [n02] type = ElementAverageValue variable = n []
  [n03] type = ElementAverageValue variable = n []
  [n04] type = ElementAverageValue variable = n []
  [n05] type = ElementAverageValue variable = n []
  [n06] type = ElementAverageValue variable = n []
  [n07] type = ElementAverageValue variable = n []
  [n08] type = ElementAverageValue variable = n []
  [n09] type = ElementAverageValue variable = n []
  [n10] type = ElementAverageValue variable = n []
  [n11] type = ElementAverageValue variable = n []
  [n12] type = ElementAverageValue variable = n []
  [n13] type = ElementAverageValue variable = n []
  [n14] type = ElementAverageValue variable = n []
  [n15] type = ElementAverageValue variable = n []
  [n16] type = ElementAverageValue variable = n []
  [n17] type = ElementAverageValue variable = n []
  [n18] type = ElementAverageValue variable = n []
  [n19] type = ElementAverageValue variable = n []
  [n20] type = ElementAverageValue variable = n []
[]

[Executioner]
  type = Transient
  dt = 0.001
  end_time = 10.0
[]
```

I can run that 10k time steps in 22 seconds with this branch, or 320s with `devel`.  If I drop down to 5k time steps then it's 51s with `devel` or 11s with the PR.  So I think I've fixed both the nasty constant and the asymptotics.

YaqiWang:
How fast if we turn off console by
```
[Outputs]
  console = false
[]
```
?

roystgnr:
Ooh, that's a good question.  14 seconds.  And if I allow the console output but pipe it to /dev/null, that's 18 seconds.  So it's still not *free* to spit out a million-ish lines of output, but we're efficient enough now that my terminal emulator is half of the overhead.

loganharbour (approved):

loganharbour:
@roystgnr what's the missing line in coverage: https://mooseframework.inl.gov/docs/PRs/30842/coverage_diff/framework/ ?

Naktakala:
@roystgnr This works great for me!

roystgnr:
That missing line is the case we would hit if a row in `_data` has an entry for every column but still has a null for one of those entries.  It might be safe to remove that whole `else` branch (assuming nobody uncomments `FormattedTable::getData()`, and congrats to whoever realized *that* accessor would have been a bad idea), but a priori I didn't want to assume that such a case would never come up nor spend as much time looking through the code as I'd need to be sure it could never come up.

PR #30844, opened 2025-06-24
Maybe this will fix clang thread failures

lindsayad:
Don't use factorization packages which use openmp

I don't trust helgrind with openmp but I also don't trust threading
generally within PETSc solves since PETSc makes little effort at
thread safety. Both superlu_dist and mumps employ openmp so instead
of using those I'm going to default to using the native PETSc LU
factorization (in serial). Only the latter comes out helgrind clean with
respect to not showing any stack traces involving dgemm


lindsayad:
1 clang pass

lindsayad:
2 clang pass

lindsayad:
3 clang pass

loganharbour (approved):
As you see fit

PR #30845, opened 2025-06-24
Add BSpline Utility

### LINKED ISSUE #30820: Create BSpline utility
### Motivation

Will be used in future features and is intended to be a small part of a larger project.

### Design

Creation of a utility which will draw a uniform open B-Spline given control points and interpolation degree.

### Impact

Will give a powerful new capability. This will eventually be added to created features for mesh generation.

lipohlmann:
closes #30820

## Reason
Utility is needed for continued work on a mesh generator to connect objects.

## Design
Implements an arbitrary BSpline recursive formula using a given vector of control points and a desired interpolation degree. Intended to be extremely flexible.

## Impact
Allows for continued work on the mesh generator project described above while also providing a tool found in many engineering applications. Utility will likely find use in others' projects.

### DIFF
--- a/framework/include/utils/BSpline.h
+++ b/framework/include/utils/BSpline.h
@@ -14,18 +14,24 @@
 
 namespace Moose
 {
+/**
+ * Class implementing a B-Spline curve.
+ */
 class BSpline
 {
 public:
   BSpline(const unsigned int degree, const std::vector<libMesh::Point> & control_points);
 
   /**
-   * Get the point on the spline evaluated at \p t
+   * Evaluate the BSpline interpolation at given value of t.
+   *
+   * @param t the parameter value, must lie in [0,1]
+   * @return libMesh::Point type with format (x,y,z)
    */
   libMesh::Point getPoint(const libMesh::Real t) const;
 
   /**
-   * Get the built knot vector associated with the degree and control points
+   * Get the built knot vector associated with the degree and control points.
    */
   const std::vector<libMesh::Real> & getKnotVector() const { return _knot_vector; }
 
--- a/framework/include/utils/BSpline.h
+++ b/framework/include/utils/BSpline.h
@@ -1,0 +1,45 @@
+//* This file is part of the MOOSE framework
+//* https://mooseframework.inl.gov
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include <vector>
+#include "libmesh/point.h"
+
+namespace Moose
+{
+class BSpline
+{
+public:
+  BSpline(const unsigned int degree, const std::vector<libMesh::Point> & control_points);
+
+  /**
+   * Get the point on the spline evaluated at \p t
+   */
+  libMesh::Point getPoint(const double t) const;
+
+  /**
+   * Get the built knot vector associated with the degree and control points
+   */
+  const std::vector<double> & getKnotVector() const { return _knot_vector; }
+
+private:
+  /**
+   * Internal method for building the knot vector given the degree and control points
+   */
+  std::vector<double> buildKnotVector() const;
+
+  /// The polynomial degree
+  const unsigned int _degree;
+  /// The control points
+  const std::vector<libMesh::Point> _control_points;
+  /// The knot vector
+  const std::vector<double> _knot_vector;
+};
+}
--- a/framework/include/utils/BSpline.h
+++ b/framework/include/utils/BSpline.h
@@ -11,6 +11,9 @@
 
 #include <vector>
 #include "libmesh/point.h"
+#include "libMeshReducedNamespace.h"
+
+using namespace libMesh;
 
 namespace Moose
 {
@@ -22,24 +25,33 @@ public:
   /**
    * Get the point on the spline evaluated at \p t
    */
-  libMesh::Point getPoint(const double t) const;
+  libMesh::Point getPoint(const Real t) const;
 
   /**
    * Get the built knot vector associated with the degree and control points
    */
-  const std::vector<double> & getKnotVector() const { return _knot_vector; }
+  const std::vector<Real> & getKnotVector() const { return _knot_vector; }
 
 private:
   /**
-   * Internal method for building the knot vector given the degree and control points
+   * Internal method for building the knot vector given the degree and control points.
    */
-  std::vector<double> buildKnotVector() const;
+  std::vector<Real> buildKnotVector() const;
 
-  /// The polynomial degree
+  /**
+   * Internal method for evaluating the Cox-de-Boor basis functions.
+   * Two additional functions, firstCoeff and secondCoeff, are submethods for CdBBasis
+   * function to evaluate front coefficients in the recursive formula.
+   */
+  Real CdBBasis(const Real & t, const unsigned int i, const unsigned int j) const;
+  Real firstCoeff(const Real & t, const unsigned int i, const unsigned int j) const;
+  Real secondCoeff(const Real & t, const unsigned int i, const unsigned int j) const;
+
+  /// The polynomial degree.
   const unsigned int _degree;
-  /// The control points
+  /// The control points.
   const std::vector<libMesh::Point> _control_points;
-  /// The knot vector
-  const std::vector<double> _knot_vector;
+  /// The knot vector.
+  const std::vector<Real> _knot_vector;
 };
 }
--- a/framework/include/utils/BSpline.h
+++ b/framework/include/utils/BSpline.h
@@ -11,9 +11,6 @@
 
 #include <vector>
 #include "libmesh/point.h"
-#include "libMeshReducedNamespace.h"
-
-using namespace libMesh;
 
 namespace Moose
 {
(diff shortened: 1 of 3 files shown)


framework/src/utils/BSpline.C
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Creates normalized knot vector.
+ */
+std::vector<Real>
+BSpline::buildKnotVector() const
+{
+  const unsigned int num_points_left = _control_points.size() - _degree;
+
+  if (num_points_left <= 0)
+    mooseError("Number of control points must be equal to degree + 1!");
```
roystgnr: This should be "greater than or equal to", no?

roystgnr: Or just "greater than polynomial degree"?

lipohlmann: Yep, you're right.

roystgnr:
This looks good to me.  Since we're not in any hurry to merge before we have the use case ready, though, let's hold off to give others time to look at it.

Honestly, when I recommended that you do the utility PR first separate from the generator PR, I was kind of expecting that the utility PR would have more about it to criticize and fix.

framework/include/utils/BSpline.h
```diff
@@ -0,0 +1,57 @@
...
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include <vector>
+#include "libmesh/point.h"
+#include "libMeshReducedNamespace.h"
+
+using namespace libMesh;
```
GiudGiud: let's not put this in a header, it would get imported accidentally by others, causing them trouble

framework/include/utils/BSpline.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Moose
+{
+class BSpline
```
GiudGiud, line 23:
```suggestion
/**
 * Class implementing a B-Spline curve
 */
class BSpline
```

framework/src/utils/BSpline.C
```diff
@@ -0,0 +1,119 @@
...
+  : _degree(degree), _control_points(control_points), _knot_vector(buildKnotVector())
+{
+}
+
+/**
+ * Evaluate the BSpline interpolation at given value of t.
+ *
+ * @param t the parameter value, must lie in [0,1]
+ * @return libMesh::Point type with format (x,y,z)
+ */
```
GiudGiud:
```suggestion
```
these only go in the header

framework/src/utils/BSpline.C
```diff
@@ -0,0 +1,119 @@
...
+ * recursive formula.
+ *
+ * @param t,i,j parameter t in [0,1], index corresponding to which control point, and the degree of
+ * the basis funtion
+ * @return scalar quantity for the contribution of the basis function at i in the sum.
+ */
+Real
+BSpline::CdBBasis(const Real & t, const unsigned int i, const unsigned int j) const
+{
+  if (j == 0)
```
GiudGiud, line 81: add mooseAssert on i and j not too big

unit/src/BSplineTest.C
```diff
@@ -0,0 +1,114 @@
...
+//* t, the parameter. In addition, a vector of control points is supplied.
+//*
+//* All expected points were independently calculated and verified.
+
+const Real t0 = 0;
+const Real t1 = 0.4;
+const Real t2 = 0.8;
+const Real t3 = 1.0;
+const std::vector<libMesh::Point> control_points = {
+    {1, 2, 0}, {1, 2, 0}, {1.5, 2.5, 0}, {3, 1, 0}, {3, 1, 0}};
```
GiudGiud, line 27: why doubling the two edge control points?

GiudGiud (changes requested):

framework/include/utils/BSpline.h
roystgnr: Oh, man - did I miss it when you got this out of Moose.h last year?  I thought that was going to be stuck there forever.  Good job!

lipohlmann: To be clear, is it the `using namespace libMesh` that is the issue? Or are the other includes a problem?

GiudGiud: other includes are fine, as long as they are justified and cant be replaced by a forward declaration instead

framework/src/utils/BSpline.C
```diff
@@ -0,0 +1,118 @@
...
+                "knot_vector must be normalized to 1 with minimum value 0.");
+  }
+
+  return knot_vector;
+}
+
+libMesh::Real
+BSpline::CdBBasis(const libMesh::Real & t, const unsigned int i, const unsigned int j) const
+{
+  mooseAssert(j >= 0, "j (polynomial degree) must be positive.");
```
GiudGiud:
```suggestion
  mooseAssert(j >= 0, "j (polynomial degree) must be positive.");
  mooseAssert(i < knot_vector.size() - 1, "knot index must stay within bounds");
```

framework/src/utils/BSpline.C
```diff
@@ -0,0 +1,118 @@
...
+    return BSpline::firstCoeff(t, i, j) * BSpline::CdBBasis(t, i, j - 1) +
+           BSpline::secondCoeff(t, i, j) * BSpline::CdBBasis(t, i + 1, j - 1);
+}
+
+/**
+ * Submethod to aid in computing the basis function in CdBBasis.
+ */
+libMesh::Real
+BSpline::firstCoeff(const libMesh::Real & t, const unsigned int i, const unsigned int j) const
+{
```
GiudGiud, line 90:
```suggestion
{
  mooseAssert(i + j < _knot_vector.size(), "Index must stay within bounds");
```

framework/src/utils/BSpline.C
```diff
@@ -0,0 +1,118 @@
...
+  libMesh::Real ti = _knot_vector[i];
+  libMesh::Real tij = _knot_vector[i + j];
```
GiudGiud:
```suggestion
  const auto ti = _knot_vector[i];
  const auto tij = _knot_vector[i + j];
```

framework/src/utils/BSpline.C
```diff
@@ -0,0 +1,118 @@
...
+{
+}
+
+/**
+ * Evaluate B-Spline at given t value using the control points.
+ */
+libMesh::Point
+BSpline::getPoint(const libMesh::Real t) const
+{
+  libMesh::Point returnPoint(0, 0, 0);
```
GiudGiud, line 27:
```suggestion
  mooseAssert((t>=0) && (t <= 1), "t should be in [0, 1]. t=" + std::to_string(t));
  libMesh::Point returnPoint(0, 0, 0);
```

GiudGiud (approved):
@roystgnr if you want to take another look at the first PR in the order

I mostly cleaned up and improved the commit history (and will deal with the conflicts on the later PRs, oh well)

framework/src/utils/BSpline.C
```diff
@@ -47,7 +47,8 @@ BSpline::buildKnotVector() const
     mooseError("Number of control points must be greater than or equal to degree + 1!");
 
   std::vector<libMesh::Real> knot_vector(_degree, 0); // initialize bottom to zeros
-  for (const auto i : make_range(num_points_left))
+
+  for (const auto i : make_range(num_points_left + 1))
```
roystgnr: I see this got changed from plain `make_range(num_points_left)` in a bugfix commit, so I'm assuming the new code really is the correct one, but now it's kind of anti-self-documenting.

How many points are left?  `num_points_left`.

How many points are we going to loop over?  `num_points_left + 1`.

What's actually going on?

unit/src/BSplineTest.C
```diff
@@ -41,10 +41,10 @@ TEST(BSplineTest, Degree2Test)
   const libMesh::Point expected_point_2(2.73, 1.2699999999999996, 0);
   const libMesh::Point expected_point_3(3, 1, 0);
 
-  EXPECT_NEAR((point_0 - expected_point_0).norm(), 0, 1e-9);
-  EXPECT_NEAR((point_1 - expected_point_1).norm(), 0, 1e-9);
-  EXPECT_NEAR((point_2 - expected_point_2).norm(), 0, 1e-9);
-  EXPECT_NEAR((point_3 - expected_point_3).norm(), 0, 1e-9);
+  EXPECT_NEAR((point_0 - expected_point_0).norm(), 0, libMesh::TOLERANCE);
```
roystgnr, line 45: It wouldn't be crazy to use `TOLERANCE*sqrt(TOLERANCE)` (matching the original hard-coded 1e-9 in the standard double-precision case) here.  This is fine too, though.

roystgnr (changes requested):

unit/src/BSplineTest.C
GiudGiud: Yeah this isnt quite up to code anyway. We should be using "saved value" tolerances, which are tighter.
I think it achieves the testing purpose though

framework/src/utils/BSpline.C
GiudGiud: @LP1012 feel free to weigh in here if you see this today. Otherwise I ll look tonight

GiudGiud: ok added some explanations. I agree it was not clear

GiudGiud: Added more. I think this is less general than I thought. We are creating a uniform clamped B-Spline. Both of these adjectives are important here. They are what I want for the project afaik, but less general

GiudGiud: We recover some generality in PR number 2 where the control point creation moves to a util though. I ll amend that one once this one is in

roystgnr (approved):

GiudGiud:
Thanks for the review! 

No need to look at PR 2 and 3 until I rebase them and remove any inconsistency

PR #30853, opened 2025-06-25
Explicit dynamics with zero algebraic ranges

### LINKED ISSUE #30850: neml2 kernels for explicit dynamics
### Motivation

Explicit time integration is a promising alternative to implicit schemes for no-so-stiff systems, as it removes the need for matrix assembly. With the recent introduction of contact constraint enforcement in ExplicitMixedOrder, this becomes even more interesting for contact mechanics problems.

The cost of explicit time integration is dominated by residual evaluation. @maxnezdyur did some profiling and found that a big chunk of time is spent on various `reinit`s and material model evaluation. It would be nice if we can accelerate those bottlenecks.

### Design

The cost of `reinit`s can be significantly reduced if we cache the FE context and zero out the algebraic element/node ranges. The cost of material model evaluation can be reduced utilizing neml2.

### Impact

Faster explicit dynamics calculations. This could be useful in the future for apps like BISON where convergence is sometimes limited by contact enforcement.

hugary1995:
This PR adds a few objects to help cache finite element context and assembly information, so that the residual loop can be made reinit-free.

New objects:

- NEML2Assembly: gather JxW, coord, etc.
- NEML2FEInterpolation: gather dofmap, shape function values/gradiantes for each coupled variable
- NEML2Kernel: define batched operations
- NEML2PreKernel: inherited from NEML2Kernel, for operations _before_ the constitutive update
- NEML2PostKernel: inherited from NEML2Kernel, for operations _after_ the constitutive update

Examples:

- NEML2SmallStrain: inherited from NEML2PreKernel, coupled to displacement gradients, calculates the small strain tensor
- NEML2StressDivergence: inherited from NEML2PostKernel, retrieves stress from the NEML2ModelExecutor and uses it to calculate the residual

Also add corresponding documentation and tests. Retire BatchMaterial.

close #30850


framework/include/neml2/userobjects/NEML2Assembly.h
```diff
@@ -0,0 +1,69 @@
...
+   * is the number of quadrature points per element. The tensor is stored in the device memory.
+   *
+   * @warning The JxWxT value is only available after finalize() is called.
+   */
+  const neml2::Tensor & JxWxT() const { return _neml2_JxWxT; }
+
+  bool upToDate() const { return _up_to_date; }
+
+  /// Invalidate the cached assembly information
+  void invalidate() const;
```
lindsayad: this doesn't seem like a method that should be `const`

framework/include/neml2/userobjects/NEML2Assembly.h
```diff
@@ -0,0 +1,69 @@
...
+  void meshChanged() override { invalidate(); }
+
+  void initialize() override;
+  void execute() override;
+  void finalize() override;
+  void threadJoin(const UserObject &) override;
+
+protected:
+  /// Whether the current assembly cache is up to date
+  mutable bool _up_to_date = false;
```
lindsayad: I have come to like `mutable` less and less over the years

lindsayad (commented):
Just had 2 minutes and glanced at the NEML2Assembly header. Will do a more comprehensive review next week (or maybe maybe this weekend) if no one else beats me to it

framework/include/neml2/userobjects/NEML2Assembly.h
hugary1995: Yeah, I don't like it, too. I'm still trying to think of an alternative approach. The problem is that some other object need to modify this member, and moose only allows me to retrieve a const reference to userobject, IIRC. I guess I can const cast it?

framework/include/neml2/userobjects/NEML2Assembly.h
```diff
@@ -0,0 +1,69 @@
...
+#pragma once
+
+#ifdef NEML2_ENABLED
+
+// MOOSE includes
+#include "ElementUserObject.h"
+
+#include "neml2/tensors/Tensor.h"
+
+/// This user object caches assebmly information from MOOSE.
```
lindsayad:
```suggestion
/// This user object caches assembly information from MOOSE.
```

framework/include/neml2/userobjects/NEML2Assembly.h
lindsayad: Nah I don't think that's the way to go. What about something more dramatic like implementing your concept in regular `Assembly`?

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -162,6 +157,18 @@ ExplicitMixedOrder::solve()
   _nonlinear_implicit_system->nonlinear_solver->converged = converged;
 }
 
+void
+ExplicitMixedOrder::evaluateRHSResidual()
+{
+  // Compute the residual
+  _explicit_residual->zero();
```
maxnezdyur: I don't even think we need this zero? I think we talked about this that computeResidual zeros it out anyway?

modules/solid_mechanics/test/tests/neml2/explicit_dynamics/neml2.i
```diff
@@ -0,0 +1,56 @@
...
+  input = '../elasticity/elasticity_neml2.i'
+  [all]
+    executor_name = 'neml2'
+    model = 'model'
+    verbose = true
+    moose_input_kernels = 'strain'
+    # To run on GPU, pass --libtorch-device=cuda on the command line
+    # and uncomment the following lines:
+    # device = 'cuda'
+    # output_device = 'cuda'
```
maxnezdyur: Will there be any test with gpu?

modules/solid_mechanics/test/tests/neml2/explicit_dynamics/tests
```diff
@@ -0,0 +1,12 @@
+[Tests]
+  issues = '#30850'
+  design = 'NEML2/index.md NEML2Action.md'
+  [explicit_dynamics]
```
maxnezdyur, line 5: Is there some way of testing that the results from neml2 match regular moose? Incase it diverges somehow in the future.

framework/include/neml2/userobjects/NEML2FEInterpolation.h
```diff
@@ -0,0 +1,176 @@
...
+   *
+   * @warning The tensor value is only available after finalize() is called.
+   */
+  const neml2::Tensor & getGradient(const std::string & var_name) const;
+
+  /**
+   * @brief Get the shape function associated with a MOOSE variable.
+   *
+   * The NEML2 tensor will have shape (nelem, ndofe, nqp;), where nelem is the number of elements,
+   * nqp is the number of quadrature points per element, and ndofe is the number of degrees of
```
maxnezdyur, line 69: I'm probably missing this. How does all this work with a mesh that would have tets and hex for example in different subdomains. Do we have an error message to get out of that or does it just work fine?

maxnezdyur (commented):
Really nice capability that I think would mesh well with explicit dynamics in MOOSE. I will give it another go around later.

framework/include/neml2/userobjects/NEML2Assembly.h
dschwen: But this is exactly the use case for `mutable` - a cached state. Let's think this through beyond dislikes for keywords. Is there a danger of thread races? Seems unlikely, as `ElementUserObject`s have per-thread copies. What other risks could this pose? The suggested change in Assembly seems rather intrusive/monolithic, compared to this encapsulated modular approach.

framework/include/neml2/userobjects/NEML2FEInterpolation.h
dschwen: I'd assume we'd use multiple block restricted NEML2Assembly objects etc. for each element type...

framework/include/neml2/userobjects/NEML2Assembly.h
lindsayad: I have thought about it. I have plenty of mutables to my name throughout the framework. I'm serious when I raise the possibility of more comprehensive NEML2 integration in core framework code instead of bolt-ons through user objects. I've been sold by you and Gary on NEML2 and so I want to consider integration opportunities even if we ultimately decide not to pursue it in the moment. You'll note that I didn't request changes but rather left a comment

modules/solid_mechanics/test/tests/neml2/explicit_dynamics/tests
dschwen: the input is already there, I'm adding a test

modules/solid_mechanics/test/tests/neml2/explicit_dynamics/neml2.i
dschwen: added

dschwen (commented):
I like this. I've rebased it and pushed some small changes addressing some of the comments. I'm fine with the UO approach for now. Assembly integration is an option for a follow on PR (good to know that this is a blessed path forward now).

framework/include/neml2/userobjects/NEML2Assembly.h
lindsayad: Can someone point me to a place where this is called such that this needs to be `const`? If it doesn't exist, then we can non-const it, and then does `_up_to_date` have to mutable?

lindsayad: I see it called from `invalidateFEMContext` but that itself only seems to be called from the non-const `meshChanged`

dschwen: Both calls ultimately come from a `meshChanged`. Why does `NEML2FEInterpolation::meshChanged()` invalidate the flag in NEML2Assembly if `NEML2Assembly::meshChanged()` already invalidates it?

dschwen: I'll try making that change and adding a test with refinement

dschwen: Oooooookaaaaay, looks like NEML2 and plain vanilla MOOSE segfault with adaptivity. We may need to update the mass matrix (or its inverse) on mesh changes. I'm investigating this...

framework/src/timeintegrators/ExplicitTimeIntegrator.C
```diff
@@ -45,11 +45,11 @@ ExplicitTimeIntegrator::validParams()
 ExplicitTimeIntegrator::ExplicitTimeIntegrator(const InputParameters & parameters)
   : TimeIntegrator(parameters),
     MeshChangedInterface(parameters),
-
     _solve_type(getParam<MooseEnum>("solve_type")),
     _explicit_residual(addVector("explicit_residual", false, PARALLEL)),
     _solution_update(addVector("solution_update", true, PARALLEL)),
-    _mass_matrix_diag_inverted(addVector("mass_matrix_diag_inverted", true, GHOSTED))
+    _mass_matrix_diag_inverted(addVector("mass_matrix_diag_inverted", true, GHOSTED)),
+    _ones(nullptr)
```
dschwen, line 52: By the way, `_ones` ends up being reset to a nullptr when we enter meshChanged.

dschwen:
Ok, I have figured out how to make ExplicitMixedOrder work with adaptivity. Sorry for working on your PR. If you want I can make it a separate PR, but my next step is to make sure the NEML2 assembly also works with adaptivity.

hugary1995:
no problem, I was occupied with other stuff this week. I'll catch up with your conversations here later next week.

framework/src/timeintegrators/ExplicitTimeIntegrator.C
dschwen: Turns out there is a second time integrator that get's called on meshChanged, which has _ones set to zero (probably because it has `_nl` set to 0). I wonder where that one is coming from?

dschwen:
Referencing `[NEML2 installation](install_neml2.md)` from  `framework/doc/content/source/neml2/actions/NEML2ActionCommon.md` breaks a ton of apps, because `install_neml2.md` is not available in their set of docs.

dschwen:
Hm, adaptivity with NEML2 assembly is still a little off...
output

framework/src/timeintegrators/ExplicitTimeIntegrator.C
dschwen:
```
*** Info ***
TimeIntegrator::ExplicitMixedOrder:aux


*** Info ***
TimeIntegrator::ExplicitMixedOrder:nl0
```

Ok, looks like MOOSE is attaching a separate timeintegrator to both the non-linear _and_ the aux system. That's a bit unexpected. Why would we timeintegrate the aux system?

GiudGiud: If we need a time derivative of an auxvariable we need the time integrator for that

framework/include/neml2/userobjects/NEML2Assembly.h
hugary1995: Alright, I'm finally back to this PR. I sense some heated discussion lol.

In general, I think there's no concern of thread races as long as those mutable methods are called outside the threaded region. I think my main concern is that the current design doesn't provide a mechanism of preventing people from abusing its usage, i.e., doing some heinous operations in a threaded region.

I am not opposed to making those mutable methods non-const and remove the mutable keywords. That way we can prevent people from calling the getter from UserObjectInterface:
```cpp
const T & getUserObject(const std::string & param_name, bool is_dependency = true) const;
```
and force people to use the method from FEProblemBase:
```cpp
T & getUserObject(const std::string & name, unsigned int tid = 0) const
```
That way we still can't prevent thread race entirely, but we can at least force people to think twice about which thread's userobject to retrieve.

lindsayad: Per the above thread I don't understand yet why they were marked `const` to begin with

hugary1995: Got rid of all mutables.

dschwen:
Rebased and fixed the merge conflict.

dschwen:
Failures seem unrelated.

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
```diff
@@ -0,0 +1,26 @@
...
+## Implementation details
+
+Once the shape function caching is done, this object does not actively update the function space information unless the `invalidateFEMContext()` method is called. The `contextUpToDate()` method can be used to check if the current FE context cache is up-to-date. Similarly, the variable interpolations are not actively updated until the `invalidateInterpolations()` method is called.
+
+Several getter methods are provided to access cached data:
+- `getValue` for getting the variable value of a MOOSE nonlinear variable
+- `getGradient` for getting the gradient of a MOOSE nonlinear variable
+- `getPhi` and `getPhiGradient` for getting the value and gradient of a MOOSE variable
+- `getDofMap` and `getGlobalDofMap` for getting the local/global dof map associated with a MOOSE nonlinear variable
+- ...
```
loganharbour:
```suggestion
```

framework/include/neml2/interfaces/NEML2ModelInterface.h
```diff
@@ -86,37 +91,36 @@ InputParameters
 NEML2ModelInterface<T>::validParams()
 {
   InputParameters params = T::validParams();
-  params.addParam<DataFileName>(
-      "input",
-      NEML2Utils::docstring("Path to the NEML2 input file containing the NEML2 model(s)."));
+  params.addParam<DataFileName>("input",
```
loganharbour, line 94: What happened to the removal of `NEML2Utils::docstring`?

framework/include/neml2/userobjects/NEML2Assembly.h
```diff
@@ -0,0 +1,69 @@
...
+   * @brief Get the cached JxWxT for each element, each quadrature point
+   *
+   * The NEML2 tensor will have shape (nelem, nqp;), where nelem is the number of elements, and nqp
+   * is the number of quadrature points per element. The tensor is stored in the device memory.
+   *
+   * @warning The JxWxT value is only available after finalize() is called.
+   */
+  const neml2::Tensor & JxWxT() const { return _neml2_JxWxT; }
+
+  bool upToDate() const { return _up_to_date; }
```
loganharbour, line 44: Doc string for this would be nice

loganharbour (dismissed):
Only partially through.

I have a significant question - do we want another syntax for this instead of just dumping it into userobjects?

dschwen:
> Only partially through.
> 
> I have a significant question - do we want another syntax for this instead of just dumping it into userobjects?

Valid question. My suggestion would be to start with this UO based design and exercise this capability for a while to see if anything needs to be improved and how it works for practical applications.

framework/include/neml2/interfaces/NEML2ModelInterface.h
dschwen: We don't need that anymore. `NEML2Utils::docstring` is a conditionally changing string that switches depending on `NEML2_ENABLED`. This is when we want to always build an object, but with degraded/shim functionality if we don't have `NEML2_ENABLED`. That used to be necessary to satisfy the doc system. But conditionals in moosedocs are a better approach. This object is not built at all now if NEML2 is not available.

framework/include/neml2/userobjects/NEML2PostKernel.h
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#ifdef NEML2_ENABLED
+
+// MOOSE includes
+#include "NEML2Kernel.h"
+#include "NEML2ModelExecutor.h"
+
+/// Base class for NEML2 kernels that operate _after_ the constitutive update
```
GiudGiud, line 18: do you have a doco page that says when the constitutive update is

is on updateMesh(), on mesh displacements, ..?

framework/doc/content/source/neml2/userobjects/NEML2Assembly.md
```diff
@@ -0,0 +1,21 @@
+# NEML2Assembly
+
+!if! function=hasCapability('neml2')
+
+This userobject loops through elements to cache common assembly information such as element transformation Jacobian, quadrature weight, and coordinate transformation factor.
```
GiudGiud: is the list of things always the same? if so let's just include it in a list form

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
```diff
@@ -0,0 +1,58 @@
+# NEML2FEInterpolation
+
+!if! function=hasCapability('neml2')
+
+This userobject provides an interface to NEML2 for finite element interpolation of variables and their gradients. It loops through elements to cache shape functions, shape function gradients, and DOF maps, then provides them as NEML2 tensors for use in NEML2 models.
```
GiudGiud: at quadrature points right

GiudGiud: so in moose-speak, this is where each object's Coupleable would reach into?
sort of a centralized MooseVariableData with all variables

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
```diff
@@ -0,0 +1,58 @@
...
+| `getDofMap` | $(n_e, n_{dofe})$ | | Local DOF indices |
+| `getGlobalDofMap` | | | Global DOF indices (as `std::vector<dof_id_type>`) |
+| `local_ndof` | | | Number of local DOFs including ghost DOFs |
+
+where $n_e$ is the number of elements, $n_q$ is the number of quadrature points per element, and $n_{dofe}$ is the number of degrees of freedom per element.
+
+### Restrictions
+
+- Only variables of type `MooseVariableFE<Real>` are supported
+- Variable scaling factors other than unity are not supported
```
GiudGiud, line 48: this seems like a limitation of something else. the scaling factor is not for variable value/gradients phi etc

GiudGiud: and on the other hand, this FE interpolation has other limitations that seem more obvious

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
```diff
@@ -0,0 +1,58 @@
+# NEML2FEInterpolation
```
GiudGiud, line 1: do you envision this interpolation to do/gather (I ll know when I look at the code) for only ever elementsxqps

or will it do side x qps, lower D x qps, nodes, physical points ...

GiudGiud: same question for NEML2Assembly

I think the name of the classes might be affected, but it could change in later PRs

framework/include/neml2/userobjects/NEML2Assembly.h
```diff
@@ -0,0 +1,70 @@
...
+  /// Number of active elements on this rank
+  int64_t numElem() const { return _nelem; }
+
+  /// Number of quadrature points per element
+  int64_t numQP() const { return _nqp; }
+
+  /**
+   * @brief Get the cached JxWxT for each element, each quadrature point
+   *
+   * The NEML2 tensor will have shape (nelem, nqp;), where nelem is the number of elements, and nqp
```
GiudGiud, line 36:
```suggestion
   * The NEML2 tensor will have shape (nelem, nqp), where nelem is the number of elements, and nqp
```
or did you mean something by ;

framework/include/neml2/userobjects/NEML2Assembly.h
```diff
@@ -0,0 +1,70 @@
...
+   * is the number of quadrature points per element. The tensor is stored in the device memory.
```
GiudGiud, line 37: can we do mixed element types then?

framework/src/neml2/actions/NEML2Action.C
```diff
@@ -208,6 +204,12 @@ NEML2Action::act()
                    input.moose.name);
     }
 
+    // Additional NEML2Kernels that provide input dataAdd commentMore actions
```
GiudGiud: ?

framework/src/neml2/userobjects/NEML2Assembly.C
```diff
@@ -0,0 +1,117 @@
...
+{
+  InputParameters params = ElementUserObject::validParams();
+
+  params.addClassDescription(
+      "This user object gathers the JxWxT values from all elements in the assembly and "
+      "provides them as a neml2 tensor. This is useful for assembling NEML2 models that "
+      "require the JxWxT values for each element.");
+
+  ExecFlagEnum execute_options = MooseUtils::getDefaultExecFlagEnum();
+  execute_options = {EXEC_INITIAL, EXEC_LINEAR};
```
GiudGiud, line 33: we need linear instead of nonlinear/something less often because you are assuming displacements are variables?

framework/src/neml2/userobjects/NEML2FEInterpolation.C
```diff
@@ -0,0 +1,415 @@
...
+    }
+  }
+
+  // shape function values
+  for (const auto & [fetype, phi] : _phis)
+  {
+    auto & moose_phi = _moose_phi[fetype];
+    for (auto i : index_range(*phi))
+      for (auto qp : index_range(_q_point))
+        moose_phi.push_back((*phi)[i][qp]);
```
GiudGiud, line 274: should resize on setup and only set here

or try to assert the starting size if you insist on push_back
or check on the size on finalize()

framework/src/neml2/userobjects/NEML2FEInterpolation.C
```diff
@@ -0,0 +1,415 @@
...
+#include "NEML2FEInterpolation.h"
+
+using namespace libMesh;
+
+registerMooseObject("MooseApp", NEML2FEInterpolation);
+
+InputParameters
+NEML2FEInterpolation::validParams()
+{
+  InputParameters params = ElementUserObject::validParams();
```
GiudGiud, line 32: this is an element UO so it can be block restricted
I think you should check the variables at hand share that block restriction

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
```diff
@@ -64,11 +66,27 @@ class ExplicitMixedOrder : public ExplicitTimeIntegrator
   TimeOrder findVariableTimeOrder(unsigned int var_num) const;
 
 protected:
+  /// compile the dof indices for firs and second order in time variables
```
GiudGiud:
```suggestion
  /// compile the dof indices for first and second order in time variables
```

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
```diff
@@ -64,11 +66,27 @@ class ExplicitMixedOrder : public ExplicitTimeIntegrator
...
 
+  /// Evaluate the RHS residual
+  virtual void evaluateRHSResidual();
+
   /// Whether we are reusing the mass matrix
   const bool & _constant_mass;
 
+  /**
+   * Must be set to true to use adaptivity with a constant mass
+   * matrix.This will recompute the mass matrix on mesh change.The user must make sure that the
```
GiudGiud:
```suggestion
   * matrix. This will recompute the mass matrix when the mesh changes. The user must make sure that the
```

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
```diff
@@ -64,11 +66,27 @@ class ExplicitMixedOrder : public ExplicitTimeIntegrator
...
+   * underlying density stays constant, otherwise simulation results will depend on adaptivity.
```
GiudGiud:
```suggestion
   * underlying mesh density stays constant, otherwise simulation results will depend on adaptivity.
```

modules/solid_mechanics/src/neml2/NEML2SmallStrain.C
```diff
@@ -0,0 +1,53 @@
...
+}
+
+void
+NEML2SmallStrain::forward()
+{
+  const auto & dux = *_grad_disp_x;
+  auto duy = _grad_disp_y ? *_grad_disp_y : neml2::Tensor::zeros_like(dux);
+  auto duz = _grad_disp_z ? *_grad_disp_z : neml2::Tensor::zeros_like(dux);
+  auto du = neml2::R2(neml2::base_stack({dux, duy, duz}, -2));
+  _output = neml2::SR2(du);
```
GiudGiud, line 53: please add comments

modules/solid_mechanics/src/neml2/NEML2StressDivergence.C
```diff
@@ -0,0 +1,90 @@
...
+
+registerMooseObject("SolidMechanicsApp", NEML2StressDivergence);
+
+InputParameters
+NEML2StressDivergence::validParams()
+{
+  InputParameters params = NEML2PostKernel::validParams();
+  params.addClassDescription(
+      "This user object calculates the stress divergence for a given set of displacement "
+      "variables and assemble the contribution into the residual vector.");
```
GiudGiud, line 27: is this PJFNK only then

modules/solid_mechanics/src/neml2/NEML2StressDivergence.C
```diff
@@ -0,0 +1,90 @@
...
+
+void
+NEML2StressDivergence::forward()
+{
+  if (!_constitutive.outputReady())
+    return;
+
+  torch::InferenceMode guard;
+
+  auto dphix = neml2::Tensor(*_grad_test_x, 3);
```
GiudGiud, line 66: const on all these?

GiudGiud (commented):
I think a comparative table of native MOOSE-FE vs NEMLFE-MOOSE would help keep track of what one can do on each side

Would be good to have a framework example to help with coverage (see +/-, "new" column is broken)
Maybe just some auxkernel-like role? Maybe a NEML2ParsedMaterial?

The PR title makes one think this is a mechanics PR, when it's mostly adding objects that are intended to be general

framework/include/neml2/userobjects/NEML2Assembly.h
dschwen: That would be possible by simply using multiple user objects. I suppose.

framework/src/neml2/userobjects/NEML2Assembly.C
dschwen: Primarily because this is at this time only for explicit time integration!

framework/src/neml2/userobjects/NEML2FEInterpolation.C
dschwen: Why wouldn't that be done in a base class, such as `ElementUserObject` or `Coupleable`? Doesn't seem right to delegate that responsibility to a leaf class.

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
dschwen: I don't think so. This is the density that goes into the mass matrix.

modules/solid_mechanics/src/neml2/NEML2StressDivergence.C
dschwen: No. Explicit.

framework/src/neml2/userobjects/NEML2FEInterpolation.C
GiudGiud: if you were using COupleable it would be done that way.
But this isnt I think, it's building and keeping track of its own list of variables

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
GiudGiud: exactly why you need to specify :)

GiudGiud:
```suggestion
   * underlying material density stays constant, otherwise simulation results will depend on adaptivity.
```

WilkAndy:
Hey @hugary1995 , @GiudGiud , just hoping we can get this merged soon, as my simulations seg-fault without it

hugary1995:
> Hey @hugary1995 , @GiudGiud , just hoping we can get this merged soon, as my simulations seg-fault without it

Working on this now!

framework/include/neml2/userobjects/NEML2PostKernel.h
hugary1995: The execution of this object is notionally controlled by NEML2ModelExecutor. NEML2ModelExecutor has a fixed execute_on flag. During its execution, it first calls NEML2PreKernels, then calls the NEML2 constitutive, and finally calls the NEML2PostKernels.

framework/include/neml2/userobjects/NEML2Assembly.h
hugary1995: That's correct -- with multiple block-restricted UOs we can do mixed element types. Though that requires of course each block-restricted UO sees only one type of element.

framework/include/neml2/userobjects/NEML2Assembly.h
hugary1995: Yeah, the semicolon separates batch dimensions from base dimensions. That's a fundamental notation we adopt in neml2. So here it means the batch shape is (nelem, nqp), and the base shape is (). This is documented clearly in neml2, but not so clear in the moose documentation. I am basically assuming any developer coming to this doxygen should at least have a basic understanding of neml2::Tensor...

framework/doc/content/source/neml2/userobjects/NEML2Assembly.md
hugary1995: Sure.

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
hugary1995: That's a good point. They currently only do element loops, not including sides. If later there's a need, we need to add `NEML2SideAssembly` and `NEML2SideFEInterpolation`.

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
hugary1995: Yes, exactly.

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
hugary1995: Can you comment on the other limitations? I can list them here.

framework/src/neml2/actions/NEML2Action.C
hugary1995: LOL, some weird autocompletion at play.

framework/src/neml2/userobjects/NEML2FEInterpolation.C
hugary1995: Good catch.

framework/src/neml2/userobjects/NEML2FEInterpolation.C
hugary1995: Yeah, good catch.

hugary1995: Actually, we can't do the size check here because technically we don't know the number of elements, dofs, and qps yet. We only know that at finalize(). And in finalize() I call updateInterpolations() which has the necessary size checks there.

modules/solid_mechanics/src/neml2/NEML2StressDivergence.C
hugary1995: Hmm, I'm not a big fan of adding const to variables that I am sure
- will destruct when exiting the stack and
- will not be modified for the rest of this stack

I do understand the benefit of adding consts -- it improves readability, clarifies intention, and prevents future changes that violate the constness. However, I was bitten by this before because marking something as const prevents compilers from doing move optimization. I am open to discussion, though.

framework/src/neml2/userobjects/NEML2FEInterpolation.C
hugary1995: I added some domain compatibility check.

framework/include/neml2/userobjects/NEML2PostKernel.h
hugary1995: Oh, and to answer your question, the execution of neml2 constitutive model is documented in NEML2Action/NEML2ActionCommon or the syntax page. The parameter documentation in this object also hints at that.

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
GiudGiud: ok I ll try but I might make mistakes:
single element types per object
no p-adaptivity
no old/older values
no implicit
only element values no face, lowerD, or neighbor
no finite volume
Did you try polyhedras / polygons? In case of of the variable APIs called are not implemented

modules/solid_mechanics/src/neml2/NEML2SmallStrain.C
```diff
@@ -43,10 +43,13 @@ NEML2SmallStrain::NEML2SmallStrain(const InputParameters & parameters) : NEML2Pr
...
 NEML2SmallStrain::forward()
 {
+  // gradient of displacements
   const auto & dux = *_grad_disp_x;
   auto duy = _grad_disp_y ? *_grad_disp_y : neml2::Tensor::zeros_like(dux);
   auto duz = _grad_disp_z ? *_grad_disp_z : neml2::Tensor::zeros_like(dux);
   auto du = neml2::R2(neml2::base_stack({dux, duy, duz}, -2));
+
+  // strain = 0.5 * (grad_u + grad_u^T), neml2::SR2 handles the symmetrization
   _output = neml2::SR2(du);
```
GiudGiud, line 53: is this coded with the radial term at theta-theta (index 22) ?
if not I think we have to mention no RZ support

framework/doc/content/source/neml2/userobjects/NEML2FEInterpolation.md
GiudGiud: see RZ comment

modules/solid_mechanics/src/neml2/NEML2SmallStrain.C
dschwen: I've addressed the limitations in more detail now. I hope we can get this in soonish as we have multiple projects depend on it by now. This PR also cleans up some legacy NEML2 coupling code (removing the batch material system) and fixes issues with changing meshes for explicit time integration.

GiudGiud (approved):
review addressed

PR #30863, opened 2025-06-25
Preliminaries for extension of Convergence system to steady-state detection

joshuahansel:
- Deleted deprecated steady-state check parameters
- Refactored solution diff computation in `TransientBase` to allow refactoring into Convergence

Refs #30858


joshuahansel:
These changes included in https://github.com/idaholab/moose/pull/31010

PR #30865, opened 2025-06-25
Register --key=value command line arguments with petsc

### LINKED ISSUE #30864: Mark `--key=value` command line args as used
### Motivation

The way that PETSc manages unused command line arguments does not work with `--key=value` arguments. That is, setting `--key` as a known command line name does not consider `--key=value` parameters used.

### Design

Manually add `--key=value` arguments as used for PETSc.

### Impact

Better user experience.

loganharbour:
Closes #30864

### DIFF
--- a/framework/src/parser/CommandLine.C
+++ b/framework/src/parser/CommandLine.C
@@ -361,9 +361,9 @@ CommandLine::populateCommandLineParams(InputParameters & params)
       if (param.metadata.global)
         entry.global = true;
 
-      // If the arg is of form "--key=value", PETSc will recognize it as an unused
+      // If the arg is of the form "--key=value", PETSc will recognize it as an unused
       // argument. That is, setting "--key" as a known command line argument is not
-      // sufficient for PETSc consider "--key=value" as known. Thus, we explicitly
+      // sufficient for PETSc to consider "--key=value" as known. Thus, we explicitly
       // add "--key=value" args as known when we come across them.
       if (entry.value_separator && *entry.value_separator == "=")
       {
--- a/framework/src/parser/CommandLine.C
+++ b/framework/src/parser/CommandLine.C
@@ -17,6 +17,7 @@
 
 #include "hit.h"
 
+#include "libmesh/libmesh.h"
 #include "libmesh/simple_range.h"
 #include "libmesh/parallel_algebra.h"
 #include "libmesh/parallel_sync.h"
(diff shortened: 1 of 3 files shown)


framework/src/parser/CommandLine.C
```diff
@@ -358,7 +359,17 @@ CommandLine::populateCommandLineParams(InputParameters & params)
...
+        entry.global = true;
+
+      // If the arg is of form "--key=value", PETSc will recognize it as an unused
+      // argument. That is, setting "--key" as a known command line argument is not
+      // sufficient for PETSc consider "--key=value" as known. Thus, we explicitly
+      // add "--key=value" args as known when we come across them.
+      if (entry.value_separator && *entry.value_separator == "=")
+      {
+        mooseAssert(entry.raw_args.size() == 1, "Should have one value");
+        libMesh::add_command_line_name(entry.raw_args[0]);
```
loganharbour, line 371: @roystgnr: `entry.raw_args[0]` here is `--key=val`

roystgnr (approved):

lindsayad (changes requested):
It would be nice to add a test that shows that something like `--required-capabilities=<foo>` is now absent from the PETSc options left

loganharbour:
I tried to actually create a unit test for this, but there is no way at the moment to query what args are added from libmesh

roystgnr:
Oh, I misunderstood you before - you can call `libMesh::command_line_names()` to get a list of everything that's been queried.

That's not quite the same thing as getting a list of what's been removed from PETSc options (we don't do the removal until the `~LibMeshInit()` destructor), though.

loganharbour:
https://github.com/idaholab/moose/pull/30865/commits/17fa05e90d9b0e6c02bd893cd3f64b8928525546 adds unit tests for checking if they are added to recognized libmesh command line args. @roystgnr sanity check please.

framework/src/parser/CommandLine.C
```diff
@@ -358,7 +359,17 @@ CommandLine::populateCommandLineParams(InputParameters & params)
 
       // If this parameter is global, mark its entry as global
       if (param.metadata.global)
-        entry_it->global = true;
+        entry.global = true;
+
+      // If the arg is of form "--key=value", PETSc will recognize it as an unused
+      // argument. That is, setting "--key" as a known command line argument is not
+      // sufficient for PETSc consider "--key=value" as known. Thus, we explicitly
+      // add "--key=value" args as known when we come across them.
```
nmnobre:
```suggestion
      // If the arg is of the form "--key=value", PETSc will recognize it as an unused
      // argument. That is, setting "--key" as a known command line argument is not
      // sufficient for PETSc to consider "--key=value" as known. Thus, we explicitly
      // add "--key=value" args as known when we come across them.
```

roystgnr:
Unit tests look good to me.

lindsayad (approved):

loganharbour:
Failures are unrelated.

PR #30871, opened 2025-06-25
Remove app input copy-construction, remove old registration

loganharbour:
Refs https://github.com/idaholab/moose/issues/30870

Doesn't quite close it yet. I have a solution in place to slowly deprecate this, but I'll get it in after this.


loganharbour:
Avoided removing the draft status so that I don't notify all of the people, but I believe this is good.

GiudGiud (approved):
good cleanup!

loganharbour:
Failures are unrelated.

PR #30873, opened 2025-06-25
Attempt to see if a new lcov is faster at coverage

loganharbour:


hugary1995:
FYI, there's [fastcov](https://github.com/RPGillespie6/fastcov) which is a parallel wrapper of gcov. It's supposed to be much faster.

PR #30884, opened 2025-06-26
Fix uninitialized memory error introduced in #30653

loganharbour:
refs #30653

In `std::optional` we trust.

### DIFF
--- a/framework/include/problems/FEProblemBase.h
+++ b/framework/include/problems/FEProblemBase.h
@@ -2473,9 +2473,9 @@ protected:
   bool _initialized;
 
   /// Nonlinear system(s) convergence name(s)
-  std::vector<ConvergenceName> _nonlinear_convergence_names;
-  /// Linear system(s) convergence name(s)
-  std::vector<ConvergenceName> _linear_convergence_names;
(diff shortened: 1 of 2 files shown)


lindsayad (approved):
This is probably a good argument for value initialization in the class definition 😄

loganharbour:
> This is probably a good argument for value initialization in the class definition 😄

`+12, -24` or better use of modern standard library features :)

lindsayad:
The changes here rely on default initialization in the class definition ... as I stated in the slack thread, I think value initialization in the class definition jives more consistently with the STL default initialization that frequently occurs in the class definition.

Certainly the change here is better than keeping the booleans around, but in a case where constants are needed ... then it's less easy to make mistakes when you don't have to remember to populate the constructors

loganharbour:
I know. I'm agreeing with you - just joking here. Trying to at least be playful after catching this in < 5 min

loganharbour:
I do however think that it's a good example that we should try to minimize additional, complex state when possible. Expresses intent more, easier to read, so on

loganharbour:
The cancelled valgrind test can be ignored. I tested it on rod - all good.

PR #30888, opened 2025-06-27
Skip some solid mechanics tests in dbg mode

hugary1995:
Many tests in the solid mechanics run significantly slower in dbg than in opt. This pull request identifies tests that run more than 10x slower in dbg and modify their test spec to be `capabilities = 'method!=dbg'`.

ref #27555

### DIFF
--- a/modules/solid_mechanics/examples/coal_mining/tests
+++ b/modules/solid_mechanics/examples/coal_mining/tests
@@ -18,6 +18,7 @@
     type = 'CSVDiff'
     input = 'cosserat_elastic.i'
     csvdiff = 'cosserat_elastic.csv'
+    heavy = true
   [../]
   [./cosserat_wp_only_csv]
     type = 'CSVDiff'
--- a/modules/solid_mechanics/test/tests/2D_geometries/tests
+++ b/modules/solid_mechanics/test/tests/2D_geometries/tests
@@ -63,6 +63,7 @@
     rel_err = 5E-4
     design = 'source/materials/ComputeAxisymmetricRZFiniteStrain.md'
     requirement = 'The ComputeAxisymmetricRZFiniteStrain class shall compute the mechanical response for a pressurized hollow cylinder with a small incremental axisymmetric strain formulation.'
+    capabilities = 'method!=dbg'
   [../]
   [./3D_RZ_finitestrain]
     type = Exodiff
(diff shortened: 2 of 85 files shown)


hugary1995:
Let's see by how much the coverage drops :)

hugary1995:
I'll discuss this with @bwspenc next week and ask for his approval.

bwspenc:
Thanks for doing this, @hugary1995!

I assume these tests have a wide range of dbg run times. You said that running 10x slower than opt is the criterion, but maybe some of them are not actually all that slow. In my mind, the biggest risk in disabling dbg tests is not really in decreasing line coverage (which doesn't seem to be an issue), but in potentially missing asserts when code is changed. 

I also realize that we have limited resources, so it's always a balancing act. Are we disabling tests in dbg across MOOSE?

loganharbour:
> In my mind, the biggest risk in disabling dbg tests is not really in decreasing line coverage (which doesn't seem to be an issue), but in potentially missing asserts when code is changed.

We're in this position because historically tests that were meant to be "unit like" were not actually unit like. Too big, didn't test specific requirements, etc.

bwspenc:
@loganharbour Yeah, I understand. Do we have a target run time for the dbg tests? I know that for the opt tests the guidance was 2 seconds, although maybe that has changed.

bwspenc:
I would say we should focus more on a target run time for the dbg tests rather than on how much slower a given dbg test is than its opt counterpart.

hugary1995:
Yeah, I agree with you. I made a mistake in my original PR description which led to some confusion, so I should correct that:

I said
> This pull request identifies tests that run more than 10x slower in dbg

but what I really did was identifying tests that
1. run more than 10x slower in dbg, AND
2. take more than 5 seconds in dbg (on my personal workstation which is much faster than the civet boxes)

I think that's more or less what you are asking for, right?

bwspenc:
Thanks, @hugary1995 -- that makes sense. @loganharbour is the 5-second cutoff consistent with what we're doing for the rest of the modules?

hugary1995:
See #30876, we have some concensus on reducing the timeout limit to 30 seconds. And 5 seconds on my workstation roughly translate to 30 seconds on civet.

I don't think module developers would object that, but I guess we probably shouldn't push them to make the change immediately. Eventually, I think we will sweep through all modules to enforce the same requirements, eventually.

I think the more important thing is to not abuse regression tests for unit testing. We are actually doing a pretty good job in solid mechanics when I was doing the sweep.

lindsayad (approved):
we can revert this later if someone complains but I'm tired of timeouts. Just had another one: https://civet.inl.gov/job/3074098/

PR #30898, opened 2025-06-28
Add l2 gradient regularizer to parameter mesh optimization

### LINKED ISSUE #30897: Add more regularization methods to the parameter mesh optimization
### Motivation

Add another regularization method that can help when there are not enough data points.

### Design

Have parameterMesh compute the objective and gradient. 
Have parameterMeshOptimization correctly add the values into the gradient.

### Impact

Improves optimization module and gives users more options for harder problems.

maxnezdyur:
closes #30897


maxnezdyur:
Comparison of results
We expect a source term of 100 on the left side and 10 on the right.
Keep in mind of the colorbar extremes in the pictures.
With reg:
image

without:
image

modules/optimization/doc/content/source/optimizationreporters/ParameterMeshOptimization.md
```diff
@@ -29,6 +29,34 @@ This optimization reporter performs the same type of optimization as [Optimizati
...
+\end{equation}
+
+This regularization:
+
+- Enforces spatial smoothness in parameter distributions
+- Prevents oscillatory parameter solutions
+
+To enable L2 gradient regularization, set [!param](/OptimizationReporter/ParameterMeshOptimization/gradient_l2_coeff) to a positive value. The strength of regularization increases with larger coefficient values.
+
 !alert warning
```
lynnmunday: Please move alert above your new section.

modules/optimization/doc/content/source/optimizationreporters/ParameterMeshOptimization.md
```diff
@@ -29,6 +29,34 @@ This optimization reporter performs the same type of optimization as [Optimizati
...
+
+- $J_{data}$ is the data misfit objective
+- $\alpha$ is the regularization coefficient ([!param](/OptimizationReporter/ParameterMeshOptimization/gradient_l2_coeff))
+- $p$ is the parameter field
+- $\Omega$ is the parameter mesh domain
+
+The gradient of this regularization term is:
+
+\begin{equation}
+\frac{\partial}{\partial p} \left( \alpha \int_{\Omega} |\nabla p|^2 \, d\Omega \right) = 2\alpha \nabla p
```
lynnmunday: I think you still need the integral

```suggestion
\frac{\partial}{\partial p} \left( \alpha \int_{\Omega} |\nabla p|^2 \, d\Omega \right) = 2\alpha \left(\int_{\Omega}\nabla p, d\Omega \right) 
```

modules/optimization/src/optimizationreporters/ParameterMeshOptimization.C
```diff
@@ -66,12 +67,24 @@ ParameterMeshOptimization::validParams()
...
+
   return params;
 }
 
 ParameterMeshOptimization::ParameterMeshOptimization(const InputParameters & parameters)
-  : GeneralOptimization(parameters)
+  : GeneralOptimization(parameters), _gradient_l2_coeff(getParam<Real>("gradient_l2_coeff"))
 {
+  // Check that only one regularization method is used
+  if (_gradient_l2_coeff > 0.0 && _tikhonov_coeff > 0.0)
```
lynnmunday: I think you can use these together.  The base class includes the tikhonov regularization so it would be included when you call these.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -108,16 +120,13 @@ ParameterMesh::getIndexAndWeight(const Point & pt,
   // Get the  in the dof_indices for our element
   // variable name is hard coded to _parameter_mesh_var
```
lynnmunday: is this comment still correct?

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -138,17 +147,14 @@ ParameterMesh::getIndexAndWeight(const Point & pt,
...
```
lynnmunday: is this comment still correct?

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -192,3 +198,107 @@ ParameterMesh::getParameterValues(std::string var_name, unsigned int time_step)
...
+      // Add L2 norm squared of gradient for regularization
+      total_variation += param_grad.norm_sq() * JxW[qp];
+    }
+  }
+
+  return total_variation;
+}
+
+std::vector<Real>
+ParameterMesh::computeGradientL2RegularizationGradient(
```
lynnmunday: You could combine these into gradientAndObjective and return both at the same time.  Its less code.  Even though you are just throwing out one of them.

lynnmunday (changes requested):
I think just typos but maybe nothign.

modules/optimization/src/utils/ParameterMesh.C
maxnezdyur: They get called at 2 separate times in parameterMeshOpt. Which means we would need to return a pair of vector and real. In the future we can pull out this loop into another function that makes this cleaner but I wouldn't combine them like that.

lynnmunday:
I approve.  @zachmprince  will you review?

lynnmunday:
@zachmprince Would you review and merge this?  I reviewed it.

modules/optimization/include/utils/ParameterMesh.h
```diff
@@ -22,14 +22,28 @@
...
+ * Enumerations for regularization computations
+ */
+enum class RegularizationType
+{
+  L2_GRADIENT,
+  // Future regularization types can be added here:
+  // L1,
+  // H1,
+  // TV (Total Variation)
+};
```
zachmprince: Put this as a public subclass of ParameterMesh or ParameterMeshOptimization or into a namespace, whichever you think is most appropriate.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -396,3 +400,181 @@ ParameterMesh::closestPoint(const Elem & elem, const Point & p) const
...
+    const std::vector<Real> & JxW = fe->get_JxW();
+    const std::vector<std::vector<Real>> & phi = fe->get_phi();
+    const std::vector<std::vector<RealGradient>> & dphi = fe->get_dphi();
+
+    // Reinitialize for current element
+    fe->reinit(elem);
+
+    // Loop over quadrature points
+    for (const auto qp : make_range(qrule.n_points()))
+      compute_func(phi, dphi, qp, dof_indices, JxW, result);
```
zachmprince: Could this be done instead of passing in a compute_func?

```cpp
for (const auto qp : make_range(qrule.n_points()))
  for (const auto i: make_range(dof_indices))
  {
    if constexpr (std::is_same_v<T, Real>)
      result += computeRegularizationQp(phi[i][qp], dphi[i][qp], JxW[qp]);
    else if constexpr (std::is_same_v<T, std::vector<Real>>)
      for (const auto j: index_range(dof_indices))
        result[dof_indices[i]] += computeRegularizationGradientQp(phi[i][qp], phi[j][qp], dphi[i][qp], dphi[j][qp], JxW[qp]);
  }

```

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -396,3 +400,181 @@ ParameterMesh::closestPoint(const Elem & elem, const Point & p) const
...
+  }
+
+  return result;
+}
+
+Real
+ParameterMesh::computeRegularizationObjective(const std::vector<Real> & parameter_values,
```
zachmprince, line 375: I think with my above comment. You can just make these `typedef`s of templated loop class.

modules/optimization/src/utils/ParameterMesh.C
zachmprince: Of course you would also have to pass in the `reg_type` to this method.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -429,67 +441,127 @@ ParameterMesh::computeGradientL2RegularizationObjective(
...
     const std::vector<Real> & JxW = fe->get_JxW();
+    const std::vector<std::vector<Real>> & phi = fe->get_phi();
     const std::vector<std::vector<RealGradient>> & dphi = fe->get_dphi();
 
     // Reinitialize for current element
     fe->reinit(elem);
 
     // Loop over quadrature points
     for (const auto qp : make_range(qrule.n_points()))
+      compute_func(phi, dphi, qp, dof_indices, JxW, result);
```
lynnmunday: Why not just call the lambda here instead of passing it.  Maybe each regularization should be a private struct in this class with their own compute_func.  I understand what you are doing and this is what we asked for, I just didn't realize it would look like this.  I'm fine with this now, I understand it enough to implement new regularizations.  We can refactor it some day if we have a better design.

modules/optimization/src/utils/ParameterMesh.C
```diff
@@ -401,9 +401,17 @@ ParameterMesh::closestPoint(const Elem & elem, const Point & p) const
...
 
-Real
-ParameterMesh::computeGradientL2RegularizationObjective(
-    const std::vector<Real> & parameter_values) const
+template <typename T>
+T
+ParameterMesh::computeRegularizationLoop(
+    const std::vector<Real> & parameter_values,
+    RegularizationType /*reg_type*/,
+    const std::function<void(const std::vector<std::vector<Real>> &,
```
lynnmunday: Can you call the lambda from this function and not pass it?  Why does it have to be passed?

lynnmunday (commented):
I have some questions but its fine.   When we deprecate one of the optimizationReporters (I forgot which needs deprecated) I think it has Tikhonov regularization in it so that will also go away.   We have been adding tikhonov seperately with the reporterFunction stuff.  Should we add it here too or always leave Tikhonov seperate since it doesn't need a mesh?

lynnmunday:
@maxnezdyur I think this is breaking initial_condition_mesh_variable.   I don't know why.  I am getting a zero residual or nan for the adjoint solve when I use initial_condition_mesh_variable with gradient_l2_coeff.  I'm using your other branch for the PTR simulations so this might not have the same problems but it is something to check.

maxnezdyur:
> @maxnezdyur I think this is breaking initial_condition_mesh_variable. I don't know why. I am getting a zero residual or nan for the adjoint solve when I use initial_condition_mesh_variable with gradient_l2_coeff. I'm using your other branch for the PTR simulations so this might not have the same problems but it is something to check.

Do you have an example with this? Not sure what exactly the problme setup you are describing.

lynnmunday:
> > @maxnezdyur I think this is breaking initial_condition_mesh_variable. I don't know why. I am getting a zero residual or nan for the adjoint solve when I use initial_condition_mesh_variable with gradient_l2_coeff. I'm using your other branch for the PTR simulations so this might not have the same problems but it is something to check.
> 
> Do you have an example with this? Not sure what exactly the problme setup you are describing.

Use this optimization block:
```
[OptimizationReporter]
  type = ParameterMeshOptimization
  objective_name = objective_value
  parameter_names = 'parameter_results'
  parameter_meshes = param_mesh_in.e
  initial_condition_mesh_variable = initialConditions 
  lower_bound_mesh_variable = lowerBounds
  upper_bound_mesh_variable = upperBounds
  gradient_l2_coeff = 10
[]
```
with a parameter mesh containing variables:

```
[Mesh]
  [gmg]
    type = GeneratedMeshGenerator
    dim = 1
    xmin =1
    xmax = 2
    nx = 10 
  []
[]

[Problem]
  solve = false
[]

[Executioner]
  type = Steady
[]

[AuxVariables]
  [lowerBounds]
    family = LAGRANGE
    order = FIRST
  []
  [upperBounds]
    family = LAGRANGE
    order = FIRST
  []
  [initialConditions]
    family = LAGRANGE
    order = FIRST
  []
[]

[AuxKernels]
  [initialConditions_aux]
    type = ParsedAux
    variable = initialConditions
    expression = '50'
  []

  [lowerBounds_aux]
    type = ParsedAux
    variable = lowerBounds
    expression = '.1'
  []
  [upperBounds_aux]
    type = ParsedAux
    variable = upperBounds
    expression = '100'
  []
[]

[Outputs]
  [in]
    type = Exodus
  []
[]

```

lynnmunday (changes requested):
This does not work for initial_condition_mesh_variable being read from the parameter mesh.  It reads the intial conditions in correctly and runs the forwardAdjoint problem fine but it fails right after the main app finishes its first optimization solve.

This can't be merged until that is fixed.  I'm pretty sure it is something in this PR because if I turn regularization off, the optimization will run using initial_condition_mesh_variable.  I can look at this next week.

lynnmunday:
@zachmprince Will you review my changes to the PR.  This fixes the segfault I was hitting in Max's code when initial conditions were being read from a parameter mesh.  I added a test for this too.

modules/optimization/src/optimizationreporters/ParameterMeshOptimization.C
```diff
@@ -66,39 +69,54 @@ ParameterMeshOptimization::validParams()
       "are used for all parameter groups and all meshes, the capability to define different "
       "timesteps for different meshes is not supported.");
 
+  // New parameters for multiple regularization types
+  MultiMooseEnum reg_types("L2_GRADIENT");
```
zachmprince, line 73: Do you not need a default here?
```suggestion
  MultiMooseEnum reg_types("L2_GRADIENT", "L2_GRADIENT");
```

modules/optimization/include/optimizationreporters/ParameterMeshOptimization.h
```diff
@@ -31,8 +34,17 @@ class ParameterMeshOptimization : public GeneralOptimization
...
+  std::vector<Real> parseExodusData(const FEType fetype,
+                                    const FileName mesh_file_name,
+                                    const std::vector<unsigned int> & exodus_timestep,
+                                    const std::string & mesh_var_name) const;
+
+  /// Store parameter meshes for regularization computation
+  std::vector<std::unique_ptr<ParameterMesh>> _parameter_meshes;
+
+  /// Regularization types to apply
+  const MultiMooseEnum & _regularization_types;
```
zachmprince:
```suggestion
  const std::vector<RegularizationType> & _regularization_types;
```

modules/optimization/src/optimizationreporters/ParameterMeshOptimization.C
```diff
@@ -66,39 +69,54 @@ ParameterMeshOptimization::validParams()
...
+  params.addParamNamesToGroup("tikhonov_coeff regularization_types regularization_coeffs",
+                              "Regularization");
+
   return params;
 }
 
 ParameterMeshOptimization::ParameterMeshOptimization(const InputParameters & parameters)
-  : GeneralOptimization(parameters)
+  : GeneralOptimization(parameters),
+    _regularization_types(getParam<MultiMooseEnum>("regularization_types")),
```
zachmprince:
```suggestion
    _regularization_types(getParam<MultiMooseEnum>("regularization_types").getSetValueIDs<RegularizationType>()),
```

modules/optimization/src/optimizationreporters/ParameterMeshOptimization.C
```diff
@@ -180,57 +198,119 @@ ParameterMeshOptimization::setICsandBounds()
...
+
+  // Apply each regularization type with its coefficient
+  for (const auto reg_idx : index_range(_regularization_types))
+  {
+    if (_regularization_coeffs[reg_idx] > 0.0)
+    {
+      Real regularization_value = 0.0;
+
+      // Convert MultiMooseEnum to RegularizationType using get() method
+      auto reg_type = static_cast<RegularizationType>(_regularization_types.get(reg_idx));
```
zachmprince: You won't need this cast with the above two suggestions.

modules/optimization/src/optimizationreporters/ParameterMeshOptimization.C
```diff
@@ -180,57 +198,119 @@ ParameterMeshOptimization::setICsandBounds()
...
+void
+ParameterMeshOptimization::computeGradient(libMesh::PetscVector<Number> & gradient) const
+{
+  // Add regularization gradient contributions to the reporter gradients before base computation
+  for (const auto reg_idx : index_range(_regularization_types))
+  {
+    if (_regularization_coeffs[reg_idx] > 0.0)
+    {
+      // Convert MultiMooseEnum to RegularizationType using get() method
+      auto reg_type = static_cast<RegularizationType>(_regularization_types.get(reg_idx));
```
zachmprince: Same here

zachmprince (changes requested):

modules/optimization/src/optimizationreporters/ParameterMeshOptimization.C
lynnmunday: I want it to be empty if they don't specify a type.

modules/optimization/src/utils/ParameterMesh.C
lynnmunday: I tried to implement something like this.

lynnmunday:
@zachmprince Will you review this again and merge if good?

zachmprince (approved):

PR #30906, opened 2025-06-29
Add `CSVValidationCase` for testing validation with golded CSVs

### LINKED ISSUE #30819: Create reader for reading the test results database
### Motivation

In order to parse data from test results for use in documentation, we need a utility that will read from the database.

### Design

Create a `CIVETTestResultsReader` for reading from test results.

### Impact

New capability.

loganharbour:
Closes https://github.com/idaholab/moose/issues/30819


python/TestHarness/JobDAG.py
```diff
@@ -225,23 +225,29 @@ def _checkOutputCollisions(self):
...
+
+            if race_jobs:
+                all_jobs = [job] + race_jobs
+                job_names = sorted([j.getTestNameShort() for j in all_jobs])
+                output = 'This test spec is set to run in parallel, but a race condition was found\n'
+                output += 'that could lead to multiple tests reading/writing from the same file.\n\n'
+                output += f'  Tests: {", ".join(job_names)}\n'
+                output += f'  File(s): {", ".join(conflicting_files)}\n\n'
+                output += 'You can resolve this issue by setting the approprate prerequisites\n'
+                output += 'between your tests with the "prereq" parameter.'
```
zachmprince: I think this looks nicer, but feel free to ignore
```suggestion
                output = '''This test spec is set to run in parallel, but a race condition was found'
that could lead to multiple tests reading/writing from the same file.

  Tests: {}
  File(s): {}
  
You can resolve this issue by setting the approprate prerequisites
between your tests with the "prereq" parameter.'''.format(', '.join(job_names), ', '.join(conflicting_files))
```

python/TestHarness/schedulers/Job.py
```diff
@@ -348,9 +349,16 @@ def getRunnable(self):
 
     def getOutputFiles(self, options):
         """ Wrapper method to return getOutputFiles (absolute path) """
+        all_files = self.__tester.getOutputFiles(options)
+        for case in self.validation_cases:
+            all_files += case.getOutputFiles()
+        all_files = list(set(all_files))
```
zachmprince:
```suggestion
        all_files = set(self.__tester.getOutputFiles(options))
        for case in self.validation_cases:
            all_files.update(case.getOutputFiles())
        all_files = list(all_files)
```

python/TestHarness/tests/test_CSVValidationCase.py
```diff
@@ -0,0 +1,309 @@
...
+
+                if not skip_write:
+                    df.to_csv(gold_file, index=False)
+
+                if not action:
+                    case.run()
+                else:
+                    action(case)
+        finally:
+            os.chdir(cwd)
```
zachmprince, line 76: Why have a `try` if you are not catching it?

zachmprince: Is it to change back even if there is an exception raised?

python/TestHarness/tests/test_CSVValidationCase.py
```diff
@@ -0,0 +1,309 @@
...
+        self.assertEqual(data_loaded, gold_data)
+
+    def runValidationCase(self, gold_data, data, case_type, action=None,
+                          set_params = {}, skip_write=False) -> CSVValidationCase:
+        gold_file = 'gold.csv'
+
+        params = CSVValidationCase.validParams()
+        params['validation_csv'] = gold_file
+        for k, v in set_params.items():
+            params[k] = v
```
zachmprince, line 55: Does update not work here?
```suggestion
        params.update(set_params)
```

python/TestHarness/validation/cases/csvvalidationcase.py
```diff
@@ -0,0 +1,142 @@
...
+            Remaining arguments passed to addScalarData()
+        """
+        assert isinstance(key, str)
+        assert isinstance(index, int)
+        assert isinstance(store_key, (str, NoneType))
+
+        # Key that we're storing the data as isn't the
+        # same as the column name in the CSV
+        if not store_key:
+            store_key = key
```
zachmprince: 🤪 
```suggestion
        store_key = store_key or key
```

test/tests/test_harness/gold/validation_csv.csv
```diff
@@ -0,0 +1 @@
+validation_ok.csv
```
zachmprince, line 1: I'm a little confused on why this csv file is not the same as validation_ok?

zachmprince (changes requested):
Small suggestions and a couple questions.

python/TestHarness/tests/test_CSVValidationCase.py
loganharbour: It's so that the `finally` is ran. Something else will catch the exception if it happens, but regardless of it works or not I always want to run the `os.chdir()`, which, with this, is required to happen

python/TestHarness/tests/test_CSVValidationCase.py
loganharbour: Nope, not for `InputParameters`

test/tests/test_harness/gold/validation_csv.csv
loganharbour: It is - that's why it's a symlink to it?

zachmprince (approved):

PR #30908, opened 2025-06-30
Store app parameters as const

loganharbour:
refs #30870


permcody (approved):

loganharbour:
SAM patch is in. Pending devel -> master merge on an unrelated issue.

PR #30909, opened 2025-06-30
Add mesh refinement

sean-baccas:
## Reason

1. After updates to the mesh, all associated objects such as gridfunctions, linear forms and FESpaces must all be updated.
2. MFEM has a lot of machinery for performing adaptive mesh refinement, which unlocks a lot of potential uses for platypus-moose

## Design

- Adds a new method to the `EquationSystem` to handle all of the updates to all of the necessary entities after the mesh has updated
- Adds a new `MFEMEstimator` class that can be used to instantiate `mfem::ErrorEstimator` objects. All it needs is a variable name to perform AMR on, a kernel that operates on this variable, and the name of an FESpace from which to derive the FE space order


sean-baccas:
Closed because I need to fork the idaholab repo first.

PR #30911, opened 2025-06-30
Support RealVectorValue in BatchMeshGenerationAction

miaoyinb:
closes #30910


framework/include/actions/BatchMeshGeneratorAction.h
```diff
@@ -101,23 +105,27 @@ class BatchMeshGeneratorAction : public Action,
    * @param params InputParameters object to set the parameters
    * @param param_name Name of the parameter to set
    * @param param_type Type of the parameter to set
+   * @param is_real_vector_value Flag to indicate if the parameter is a RealVectorValue
    * @param param_value Value of the parameter to set in string format
    */
   void setVectorParams(InputParameters & params,
                        const std::string & param_name,
                        const ParameterType & param_type,
+                       const bool & is_real_vector_value,
```
lindsayad:
```suggestion
                       const bool is_real_vector_value,
```

framework/include/actions/BatchMeshGeneratorAction.h
```diff
@@ -138,22 +146,38 @@ class BatchMeshGeneratorAction : public Action,
...
+   * @param is_vector_real_vector_value Vector to store whether the vector parameters are
+   * RealVectorValue
    */
   void checkInputParametersTypes(const InputParameters & params,
                                  const std::string & action_input_param_name,
                                  const std::vector<std::string> & param_names,
                                  const std::vector<ParameterType> & param_types,
-                                 const bool & is_vector = false);
+                                 const bool & is_vector,
+                                 std::vector<bool> & is_vector_real_vector_value);
```
lindsayad:
```suggestion
                                 const std::vector<bool> & is_vector_real_vector_value) const;
```

framework/include/actions/BatchMeshGeneratorAction.h
```diff
@@ -138,22 +146,38 @@ class BatchMeshGeneratorAction : public Action,
...
+   * Check the types of the input parameters are valid, otherwise throw an error
+   * @param params InputParameters object to check the parameters
+   * @param action_input_param_name Name of the action input parameter used for error message
+   * @param param_names Names of the parameters to check
+   * @param param_types Types of the parameters to check
+   */
+  void checkInputParametersTypes(const InputParameters & params,
+                                 const std::string & action_input_param_name,
+                                 const std::vector<std::string> & param_names,
+                                 const std::vector<ParameterType> & param_types);
```
lindsayad:
```suggestion
                                 const std::vector<ParameterType> & param_types) const;
```

framework/src/actions/BatchMeshGeneratorAction.C
```diff
@@ -544,12 +554,13 @@ void
 BatchMeshGeneratorAction::setVectorParams(InputParameters & params,
                                           const std::string & param_name,
                                           const ParameterType & param_type,
+                                          const bool & is_real_vector_value,
```
lindsayad:
```suggestion
                                          const bool is_real_vector_value,
```

framework/src/actions/BatchMeshGeneratorAction.C
```diff
@@ -631,19 +652,32 @@ BatchMeshGeneratorAction::convertAndSetStringLikeVector(
   params.set<std::vector<T>>(param_name) = values;
 }
 
+void
+BatchMeshGeneratorAction::checkInputParametersTypes(const InputParameters & params,
+                                                    const std::string & action_input_param_name,
+                                                    const std::vector<std::string> & param_names,
+                                                    const std::vector<ParameterType> & param_types)
+{
+  std::vector<bool> is_real_vector_value(param_names.size(), false);
```
lindsayad: why do you create this local variable here and not in the overload below?

lindsayad: it seems like this local variable isn't actually used for anything?

framework/src/actions/BatchMeshGeneratorAction.C
```diff
@@ -631,19 +652,32 @@ BatchMeshGeneratorAction::convertAndSetStringLikeVector(
...
+                                                    std::vector<bool> & is_real_vector_value)
 {
   for (const auto i : index_range(param_names))
   {
     switch (param_types[i])
     {
       case (ParameterType::REAL):
-        checkInputParameterType<Real>(params, action_input_param_name, param_names[i], is_vector);
+        is_real_vector_value[i] = checkInputParameterType<Real, RealVectorValue>(
+            params, action_input_param_name, param_names[i], is_vector);
```
lindsayad: Per the comment above, what do we get out of setting the `is_real_vector_value` elements?

framework/include/actions/BatchMeshGeneratorAction.h
```diff
@@ -138,22 +146,38 @@ class BatchMeshGeneratorAction : public Action,
...
    * Check the type of the input parameter is valid, otherwise throw an error
    * @param params InputParameters object to check the parameter
    * @param action_input_param_name Name of the action input parameter used for error message
    * @param param_name Name of the parameter to check
    * @param is_vector Flag to indicate if the parameter is a vector
+   * @return true of the parameter is a RealVectorValue, false otherwise
    */
-  template <typename T>
-  void checkInputParameterType(const InputParameters & params,
+  template <typename T, typename P = std::vector<T>>
```
lindsayad: I prefer to default as few things as possible
```suggestion
  template <typename T, typename P>
```

test/tests/meshgenerators/batch_mesh_generation_action/tests
```diff
@@ -69,6 +69,14 @@
       recover = false
       requirement = 'The system should batch generate meshes of the same type with variation in one single vector parameter'
   []
+  [single_real_vector_value]
+    type = 'Exodiff'
+      input = 'real_vector_value.i'
+      exodiff = 'real_vector_value_in.e'
+      cli_args = '--mesh-only'
+      recover = false
+      requirement = 'The system should batch generate meshes of the same type with variation in one single RealVectorValue parameter, which is treated as a special vector of Real type numbers'
```
lindsayad:
```suggestion
      requirement = 'The system should batch generated meshes of the same type with variation in one single vector value parameter, which is treated as a special vector of real type numbers'
```

lindsayad (changes requested):

framework/src/actions/BatchMeshGeneratorAction.C
miaoyinb: A flag to record whether a vector of `Real` parameter is an ordinary vector<Real> or a `RealVectorValue` so that the parameter can be set properly later.

framework/src/actions/BatchMeshGeneratorAction.C
miaoyinb: Used to record whether a `vector<Real>` is actually a `RealVectorValue`

framework/src/actions/BatchMeshGeneratorAction.C
lindsayad: I searched through the code again and I still don't see where you use the vector version of `is_real_vector_value`

miaoyinb: Each element of this vector is used when I call `setVectorParams()`, which is the fourth argument.

For example,
```
      setVectorParams(params,
                      _fixed_vector_input_param_names[j],
                      _fixed_vector_input_param_types[j],
                      _fixed_vector_input_param_is_real_vector_value[j],
                      _fixed_vector_input_param_values[j]);
```

lindsayad: I'm asking specifically about the variable named `is_real_vector_value`

lindsayad: As the reviewer, I'll resolve this conversation when I'm happy with it

miaoyinb: `is_real_vector_value` is a reference argument of `checkInputParametersTypes()`

`checkInputParametersTypes()` is called twice in this C file with this argument specified.

Line 212, with `_batch_vector_input_param_is_real_vector_value` provided as this argument
Line 221, with `_fixed_vector_input_param_is_real_vector_value` provided as this argument

Both `_batch_vector_input_param_is_real_vector_value` and `_fixed_vector_input_param_is_real_vector_value` are then used as an argument in `setVectorParams()` (Line 474 and Line 485).

framework/src/actions/BatchMeshGeneratorAction.C
lindsayad: This one is just a dummy vector though right? Because this overload is called for the scalar parameters?

https://github.com/idaholab/moose/blob/2bbbe68aee31f3de438f0897047f241f733d8e91/framework/src/actions/BatchMeshGeneratorAction.C#L198-L206

I think this dummy vector is what confused me

framework/src/actions/BatchMeshGeneratorAction.C
lindsayad: That makes sense. I'll resolve this conversation, but I'm going to leave the one above unresolved because to me it's indicative of a non-optimal design

framework/src/actions/BatchMeshGeneratorAction.C
miaoyinb: Yes, when the method is used to check scalar type parameters, this vector is dummy.

miaoyinb: Any suggestion to further optimize this?

lindsayad (changes requested):
I think this should be re-designed a bit. I think that you should add a new enumerator under the `ParameterType` enum for `REALVECTORVALUE`. And then I think you should be able to support mesh generator parameters like `RealVectorValue` under your `scalar` action parameters and (hypothetical) mesh generator parameters like `std::vector<RealVectorValue>` under your `vector` action parameters. Then `RealVectorValue` and e.g. `Real` parameter types are conceptually treated the same. 

Because what happens if in the future someone wants to do a `RealTensorValue` mesh generator parameter? Would you also put support for that in the `vector` action params?

Even though we made `checkInputParameterTypes` semantically `const` they don't feel conceptually `const` because we are now using calls to these methods to modify data members. To me `checkInputParameterTypes` should only be a check. It shouldn't be responsible for actually determining/setting parameter types

miaoyinb:
`REALVECTORVALUE` is a vector consisting of three real elements. It serves as a vector input parameter, rather than a scalar one, and therefore is not syntactically compatible with other scalar-type input parameters. Would it be appropriate to classify `REALVECTORVALUE` under `vector` types instead of `scalar` types?

lindsayad:
> It serves as a vector input parameter, rather than a scalar one, and therefore is not syntactically compatible with other scalar-type input parameters.

I agree that it is a vector in the mathematical sense. But it is not a `std::vector`. Nor is it really considered a vector by the parser

https://github.com/idaholab/moose/blob/2bbbe68aee31f3de438f0897047f241f733d8e91/framework/src/parser/Builder.C#L1080-L1091

miaoyinb:
It would still need some work to make it work with other scalar input in one input parameter. I will think about it.

I think I may be able to support `Point` too once this is figured out.

lindsayad:
I just want to make sure we're thinking about the future. I just think it would be harder to maintain if we need to add a new method parameter to `checkParameterTypes` every time we want to add a new compound type
```c++
  void checkInputParametersTypes(const InputParameters & params,
                                 const std::string & action_input_param_name,
                                 const std::vector<std::string> & param_names,
                                 const std::vector<ParameterType> & param_types,
                                 const bool & is_vector,
                                 std::vector<bool> & is_vector_real_vector_value,
                                 std::vector<bool> & is_vector_point,
                                 std::vector<bool> & is_vector_real_tensor_value,
                                 std::vector<bool> & is_vector_rank_two_tensor,
                                 std::vector<bool> & is_vector_rank_four_tensor) const;
```
so on and so forth

framework/src/actions/BatchMeshGeneratorAction.C
miaoyinb: I am resolving this as the code has been redesigned and needs to be re-reviewed.

miaoyinb:
@lindsayad Just a friendly reminder. This has be redesigned and ready for another round of review. Thanks.

framework/src/actions/BatchMeshGeneratorAction.C
```diff
@@ -636,7 +824,7 @@ BatchMeshGeneratorAction::checkInputParametersTypes(const InputParameters & para
                                                     const std::string & action_input_param_name,
                                                     const std::vector<std::string> & param_names,
                                                     const std::vector<ParameterType> & param_types,
-                                                    const bool & is_vector)
+                                                    const bool & is_vector) const
```
lindsayad, line 836: I love it!

lindsayad (approved):

lindsayad:
@miaoyinb see if you're ok with the last commit I pushed to use `Moose::dim` instead of 3

miaoyinb:
Looks good to me, thanks for the improvements.

PR #30916, opened 2025-07-01
Avoid redundancy between  MooseConfig.h and moose.mk

nmnobre:
## Reason
Avoid redundancy between macros in `MooseConfig.h` and variables in `moose.mk`.

## Design
Include `MooseConfig.h` everywhere so that its macro definitions are visible universally.

## Impact
Hopefully minimal.

### DIFF
--- a/framework/app.mk
+++ b/framework/app.mk
@@ -295,10 +295,10 @@ LIBRARY_SUFFIX :=
 ifeq ($(MOOSE_HEADER_SYMLINKS),true)
 
 # If we are compiling with header symlinks, we don't want to start compiling any
-# object files until all symlinking is completed. The first dependency in the
+# object files until all symlinking is completed. The second dependency in the
 # list below ensures this.
 
-$(all_app_objects) : | $(app_LINKS) $(moose_config_symlink)
+$(all_app_objects) : $(moose_config_symlink) | $(app_LINKS)
 
 else
 
--- a/framework/build.mk
+++ b/framework/build.mk
@@ -121,23 +121,23 @@ unity_files:
 #
 # C++ rules
 #
-pcre%.$(obj-suffix) : pcre%.cc | prebuild
+pcre%.$(obj-suffix) : pcre%.cc | $(prebuild)
 	@echo "Compiling C++ (in "$(METHOD)" mode) "$<"..."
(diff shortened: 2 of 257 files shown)


loganharbour:
Do the old preprocessor variables still work?

There are applications that use things like `LIBTORCH_ENABLED`

nmnobre:
> Do the old preprocessor variables still work?

No, I mean... not at the moment, and I'd rather not allow them back, otherwise there's no point in doing this.

> There are applications that use things like `LIBTORCH_ENABLED`

Such as? Can we not kindly ask them to switch? I'm not introducing any new macros, I'm just increasing the visibility/availability of macros that've always been there... with a bit of luck (i.e. if they happen to already (transitively) include `MooseConfig.h`) those apps can even pre-emptively make the switch.

lindsayad:
I don't see any failures, so it seems like the apps are fine?

lindsayad (approved):
I like this redundancy elimination and I also like MOOSE_ prefixing as is standard macro practice for libraries

lindsayad:
@nmnobre let's wait for @loganharbour to approve this one. Likely libtorch testing in apps is not in MOOSE CI

loganharbour (commented):
Why did all of the prebuild dependencies need to be added?

nmnobre:
> Why did all of the prebuild dependencies need to be added?

I switched it around, such that `prebuild` depends on `$(moose_config)` and not the other way around, so that transitively the configuration header must be present for all rules depending on `prebuild`. I did overdo it a bit; I've changed it now so that only rules whose recipes use `libmesh_CPPFLAGS`, which is where we now inject the header, depend on `prebuild` (in addition to all other rules which already depended on `prebuild` prior to this PR).

loganharbour:
It should just be a few apps that see this, and we can just catch it on their devels

NamjaeChoi:
`-imacros` is troublesome with nvcc. It thinks the following file is an input file. Is there an alternative?

NamjaeChoi:
I will add a filtering rule for now.

nmnobre:
> `-imacros` is troublesome with nvcc. It thinks the following file is an input file. Is there an alternative?

`-include` seems to work I think from some quick testing.

NamjaeChoi:
`-include` looks good. It works with nvcc too. Let me include that change in my PR.

PR #30922, opened 2025-07-01
support damping term in ExplicitMixedOrder

YitaoLiu365:
## Reason
To deal with wave propagation with damping using explicit time integration.

## Design
Modified the ExplicitMixedOrder object, adding a damping_matrix and applying it to re-write the central difference formulation.

## Impact
Making the explicit time integrator capable for one variable with both second and first time derivatives in one equation (mass and damping terms).


hugary1995:
@maxnezdyur can you also review this when you get a chance?

modules/solid_mechanics/doc/content/source/timeintegrators/ExplicitMixedOrder.md
```diff
@@ -33,11 +33,17 @@ then,
...
 ## Example Input File Syntax
 
-An example input file is shown below:
+An example input file is shown below. The file solves the equations:
+\begin{equation}
+    \begin{aligned}
+        \ddot{disp_x} + \dot{disp_x} = 0\\
+        \dot{disp_y}
+    \end{aligned} = 0
+\end{equation}
```
hugary1995, line 46: check this

modules/solid_mechanics/doc/content/source/timeintegrators/ExplicitMixedOrder.md
```diff
@@ -55,6 +61,12 @@ An example input file is shown below:
         matrix_tags = 'system'
         variable = disp_x
     []
+    [dampingmatrix]
+        type = MassMatrix
+        density = 0.8
+        matrix_tags = 'damping'
+        variable = disp_x
+    []
```
hugary1995, line 69: @maxnezdyur @TheGreatCid 

Do you think we should rename `MassMatrix` to be something more general, or just create another object called `DampingMatrix`?

modules/solid_mechanics/include/bcs/ExplicitDirichletBCBase.h
```diff
@@ -38,8 +38,12 @@ class ExplicitDirichletBCBase : public NodalBC
    */
   virtual Real computeQpValue() = 0;
 
-  /// The diagonal of the mass matrix
-  const NumericVector<Number> & _mass_diag;
+  // /// The diagonal of the mass matrix
+  // const NumericVector<Number> & _mass_diag;
```
hugary1995:
```suggestion
```

modules/solid_mechanics/include/bcs/ExplicitDirichletBCBase.h
```diff
@@ -55,5 +59,9 @@ class ExplicitDirichletBCBase : public NodalBC
    * Initialize the lumped mass matrix.
    * @return lumped mass matrix vector
    */
-  const NumericVector<Number> & initLumpedMass();
+  // const NumericVector<Number> & initLumpedMass();
```
hugary1995:
```suggestion
```

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
```diff
@@ -22,34 +24,30 @@ class SparseMatrix;
  * Implements a form of the central difference time integrator that calculates acceleration directly
  * from the residual forces.
  */
-class ExplicitMixedOrder : public ExplicitTimeIntegrator
+class ExplicitMixedOrder : public TimeIntegrator, public MeshChangedInterface
```
maxnezdyur, line 28: Can I get a reason why we changed what EMO derived from? This is will help for the review.

modules/solid_mechanics/src/bcs/ExplicitDirichletBCBase.C
```diff
@@ -24,12 +24,18 @@ ExplicitDirichletBCBase::validParams()
 
 ExplicitDirichletBCBase::ExplicitDirichletBCBase(const InputParameters & parameters)
   : NodalBC(parameters),
-    _mass_diag(initLumpedMass()),
+    // _mass_diag(initLumpedMass()),
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/bcs/ExplicitDirichletBCBase.C
```diff
@@ -24,12 +24,18 @@ ExplicitDirichletBCBase::validParams()
...
+    _damping_lumped(initDampingLumped()),
     _u_old(_var.nodalValueOld()),
     _u_dot_old(_var.nodalValueDotOld()),
     _explicit_integrator(
         dynamic_cast<const ExplicitMixedOrder *>(&_sys.getTimeIntegrator(_var.number())))
 {
+  // if (!_has_lumped_matrix)
+  //   mooseError("Lumped mass matrix is missing. Make sure ExplicitMixedOrder is being used as the
+  //   "
+  //              "time integrator.");
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/bcs/ExplicitDirichletBCBase.C
```diff
@@ -45,13 +51,14 @@ ExplicitDirichletBCBase::computeQpResidual()
...
-      resid /= -_mass_diag(dofnum);
+      resid *= -_mass_lumped(dofnum);
       break;
 
     case ExplicitMixedOrder::SECOND:
       Real avg_dt = (_dt + _dt_old) / 2;
       resid = (computeQpValue() - _u_old) / (avg_dt * _dt) - (_u_dot_old) / avg_dt;
-      resid /= -_mass_diag(dofnum);
+      resid *= -_mass_lumped(dofnum);
+      resid += _damping_lumped(dofnum) * _u_dot_old;
```
hugary1995: What if we don't have a damping term for this variable?

modules/solid_mechanics/src/bcs/ExplicitDirichletBCBase.C
```diff
@@ -64,13 +71,33 @@ ExplicitDirichletBCBase::timestepSetup()
...
+// ExplicitDirichletBCBase::initLumpedMass()
+// {
+//   const auto & nl = _fe_problem.getNonlinearSystemBase(_sys.number());
+//   if (nl.hasVector("mass_matrix_diag_inverted"))
+//     return nl.getVector("mass_matrix_diag_inverted");
+
+//   mooseError("Lumped mass matrix is missing. Make sure ExplicitMixedOrder is being used as
+//   the "
+//              "time integrator.");
+// }
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/bcs/ExplicitDirichletBCBase.C
```diff
@@ -64,13 +71,33 @@ ExplicitDirichletBCBase::timestepSetup()
...
 
   mooseError("Lumped mass matrix is missing. Make sure ExplicitMixedOrder is being used as the "
              "time integrator.");
 }
+
+const NumericVector<Number> &
+ExplicitDirichletBCBase::initDampingLumped()
+{
+  const auto & nl = _fe_problem.getNonlinearSystemBase(_sys.number());
+  // if (nl.hasVector("damping_matrix_lumped"))
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/bcs/ExplicitDirichletBCBase.C
```diff
@@ -64,13 +71,33 @@ ExplicitDirichletBCBase::timestepSetup()
...
+  return nl.getVector("damping_matrix_lumped");
```
hugary1995, line 105: Same here, if we don't have a damping term, do we need to initialize this vector?

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
 
   // Move the residual to the RHS
   *_explicit_residual *= -1.0;
 
   // Perform the linear solve
-  bool converged = performExplicitSolve(mass_matrix);
+  forwardEuler();
+  centralDifference();
+  const bool converged = solutionUpdate();
+  // _console << "u_update: " << *_solution_update << std::endl;
```
maxnezdyur: Get rid of these comments that are old print statements. There are a few, but I'll just call out this one.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -87,6 +92,95 @@ ExplicitMixedOrder::ExplicitMixedOrder(const InputParameters & parameters)
...
+}
+
+void
+ExplicitMixedOrder::meshChanged()
+{
+  *_ones = 1.;
+}
+
+void
+ExplicitMixedOrder::init()
```
hugary1995, line 114: This function is very long and almost identical to `ExplicitTimeIntegrator::init`. Can we somehow reduce code duplication?

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -66,17 +66,22 @@ ExplicitMixedOrder::validParams()
...
+    _has_damping(false),
     _constant_mass(getParam<bool>("use_constant_mass")),
+    _constant_damping(getParam<bool>("use_constant_damping")),
     _mass_matrix(getParam<TagName>("mass_matrix_tag")),
+    _damping_matrix(getParam<TagName>("damping_matrix_tag")),
     _solution_older(_sys.solutionState(2)),
+    _explicit_residual(addVector("explicit_residual", false, PARALLEL)),
+    _solution_update(addVector("solution_update", true, PARALLEL)),
+    _mass_matrix_lumped(addVector("mass_matrix_lumped", true, GHOSTED)),
+    _damping_matrix_lumped(addVector("damping_matrix_lumped", true, GHOSTED)),
```
maxnezdyur, line 85: How common is damping? I wonder if we should have optional to add.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
 
-bool
-ExplicitMixedOrder::performExplicitSolve(SparseMatrix<Number> &)
+void
+ExplicitMixedOrder::centralDifference()
 {
+  // Split lumped mass matrix for first-order variables
+  auto M = NumericVector<Number>::build(_communicator);
+  _mass_matrix_lumped->create_subvector(*M, _local_second_order_indices, false);
+  // _console << "M: " << *M << std::endl;
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
 
-  bool converged = false;
+  const auto r = NumericVector<Number>::build(_communicator);
+  _explicit_residual->create_subvector(*r, _local_second_order_indices, false);
+  // _console << "r: " << *r << std::endl;
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
+  vn->add(*delta_v);
 
-  // Restore acceleration
-  accel->restore_subvector(std::move(accel_second), _local_second_order_indices);
+  // _console << "vel: " << *vn << std::endl;
 
-  vel->restore_subvector(std::move(vel_second), _local_second_order_indices);
+  // close the vectors
+  accel->restore_subvector(std::move(a), _local_second_order_indices);
+  // _console << "acceleration: " << *accel << std::endl;
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
-  // Velocity update for central difference
-  *vel_second += *accel_scaled;
+  // Calculate velocity
+  auto delta_v = a->clone();
+  delta_v->scale((_dt + _dt_old) / 2);
+  vn->add(*delta_v);
 
-  // Restore acceleration
-  accel->restore_subvector(std::move(accel_second), _local_second_order_indices);
+  // _console << "vel: " << *vn << std::endl;
```
hugary1995:
```suggestion
```

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
 
-  vel->restore_subvector(std::move(vel_second), _local_second_order_indices);
+  // close the vectors
+  accel->restore_subvector(std::move(a), _local_second_order_indices);
+  // _console << "acceleration: " << *accel << std::endl;
+  accel->close();
+  vel->restore_subvector(std::move(vn), _local_second_order_indices);
+  // _console << "velocity: " << *vel << std::endl;
```
hugary1995:
```suggestion
```

hugary1995 (changes requested):

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -87,6 +92,95 @@ ExplicitMixedOrder::ExplicitMixedOrder(const InputParameters & parameters)
...
+void
+ExplicitMixedOrder::meshChanged()
+{
+  *_ones = 1.;
+}
+
+void
+ExplicitMixedOrder::init()
+{
+  _has_damping = _sys.subproblem().matrixTagExists(_damping_matrix);
```
maxnezdyur, line 116: Instead see if the user provides a damping matrix named from _damping_matrix paramter.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -87,6 +92,95 @@ ExplicitMixedOrder::ExplicitMixedOrder(const InputParameters & parameters)
...
+
+  meshChanged();
```
maxnezdyur, line 118: Why do we call meshchanged?

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
YitaoLiu365: Previously the inverted mass diagonal matrix was used, but now with the damping term we need some extra operations before inverting the matrix. So directly getting the original matrix from TimeIntegrator is more convenient.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -101,39 +195,69 @@ ExplicitMixedOrder::computeTimeDerivatives()
...
+  }
 
-    // "Invert" the diagonal mass matrix
-    _mass_matrix_diag_inverted->reciprocal();
-    _mass_matrix_diag_inverted->close();
+  // Get the lumped damping matrix
+  // We may only want to compute lumped damping matrix once.
+  if (_has_damping)
+  {
+    auto damping_tag = dampingMatrixTagID();
```
maxnezdyur, line 258: same with mass matrix, maybe we should just save the matrix id once.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
-ExplicitMixedOrder::postResidual(NumericVector<Number> & residual)
+ExplicitMixedOrder::forwardEuler()
 {
-  residual += *_Re_time;
-  residual += *_Re_non_time;
-  residual.close();
+  // Split lumped mass matrix for first-order variables
+  auto Minv = NumericVector<Number>::build(_communicator);
+  _mass_matrix_lumped->create_subvector(*Minv, _local_first_order_indices, false);
+  Minv->reciprocal();
```
maxnezdyur: Why are we doing the reciprocal every iteration? In the previous implementation we did it once if it was a constant_mass.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -142,175 +266,128 @@ ExplicitMixedOrder::solve()
...
-      *_nonlinear_implicit_system->current_local_solution, *_explicit_residual, _nl->number());
+  _fe_problem.computeResidual(sol, *_explicit_residual, _nl->number());
 
   // Move the residual to the RHS
   *_explicit_residual *= -1.0;
 
   // Perform the linear solve
-  bool converged = performExplicitSolve(mass_matrix);
+  forwardEuler();
+  centralDifference();
```
maxnezdyur: I like this. But here is my worry. I would rather have a single function call if the code in the function is more efficicient because this is for explict systems. For example doing this we double the number of "closes".

maxnezdyur (changes requested):
I will give another review later. I have a few questions/concerns that I want to get out so that we can figure out the best path.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
YitaoLiu365: Actually this method is copied from ExplicitTimeIntegrator, and in ExplicitMixedOrder the solve type has been hard coded as LUMPED, which is just the first loop of meshChanged(). Its function is initializing _ones to be 1.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
YitaoLiu365: The matrix needs to be reciprocal in each step for central difference, and so forward Euler has to adapt to this.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
YitaoLiu365: Yes, it's better to combine them. I'll work on it in the future

modules/solid_mechanics/doc/content/source/timeintegrators/ExplicitMixedOrder.md
maxnezdyur: I would keep as is. Using the mass matrix for damping is just one form of damping, so I wouldn't rename it just because the MassMatrix can be used for it.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
maxnezdyur: In the docs can you add the matrix form of the equations that we are solving? I think that would help.

modules/solid_mechanics/src/bcs/ExplicitDirichletBCBase.C
YitaoLiu365: What about setting _damping_lumped to be zero vector if (!nl.hasVector("damping_matrix_lumped"))? I think this can be done along with combining forwardEuler() and centralDifference().

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
hugary1995: Yeah this is my concern as well. If we don't have damping, we could just save the inverted lumped mass. But that's not the case with damping cuz the coefficients now depend on dt. Only if we make sure dt is constant (which might be fine in most cases), we could actually save the inverted coefficients. I think it'll become clear once Yitao adds the math, and I think the constant dt case is worth implementing.

modules/solid_mechanics/doc/content/source/timeintegrators/ExplicitMixedOrder.md
TheGreatCid: Sorry, I'm late to this, but I agree with Max. Makes sense to leave as is.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
TheGreatCid: Definitely should be left as an option.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
hugary1995: Can you elaborate on this? I'm not sure I follow.

framework/src/kernels/ADTimeDerivative.C
```diff
@@ -15,18 +15,21 @@ InputParameters
...
-  : ADTimeKernelValue(parameters)
+  : ADTimeKernelValue(parameters), _coef(getADMaterialProperty<Real>("density"))
 {
 }
 
 ADReal
 ADTimeDerivative::precomputeQpResidual()
 {
-  return _u_dot[_qp];
+  return _coef[_qp] * _u_dot[_qp];
```
dschwen: I'd prefer it if you derived a class instead of adding a required new parameter to a fundamental framework class (which is going to break input files all over the place). We have a kernel like this called SusceptibilityTimeDerivative in phase field.

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
```diff
@@ -87,6 +135,16 @@ class ExplicitMixedOrder : public ExplicitTimeIntegrator
...
 
+  /// Helper functions for phase-field
+  virtual void initPF() {}
+  virtual void upperboundCheck() {}
+  virtual void irreversibilityCheck(NumericVector<Number> * acceleration,
+                                    NumericVector<Number> * velocity)
+  {
+    UNUSED(acceleration);
+    UNUSED(velocity);
+  }
```
dschwen:
```suggestion
  virtual void irreversibilityCheck(NumericVector<Number> * /*acceleration*/,
                                    NumericVector<Number> * /*velocity*/)
  {
  }
```

modules/solid_mechanics/include/timeintegrators/ExplicitMixedOrder.h
```diff
@@ -9,7 +9,11 @@
 
 #pragma once
 
-#include "ExplicitTimeIntegrator.h"
+#include "TimeIntegrator.h"
+#include "MeshChangedInterface.h"
+#include "FEProblemBase.h"
+#include <memory>
+#define UNUSED(x) (void)(x)
```
lynnmunday:
```suggestion
```

framework/include/interfaces/Coupleable.h
```diff
@@ -790,6 +790,13 @@ class Coupleable
   virtual const VariableGradient & coupledGradientDot(const std::string & var_name,
                                                       unsigned int comp = 0) const;
 
+  /**
+   * Returns the time derivatives of the gradient for all of a coupled variable's components
+   * @param var_name Name of coupled variable
+   * @return Vector of VariableValue pointers for each component of \p var_name
+   */
+  std::vector<const VariableGradient *> coupledGradientDots(const std::string & var_name) const;
```
dschwen, line 798: Is this used in this PR at all?

framework/include/interfaces/Coupleable.h
```diff
@@ -1010,6 +1017,13 @@ class Coupleable
   virtual const VariableValue & coupledDotDot(const std::string & var_name,
                                               unsigned int comp = 0) const;
 
+  /**
+   * Returns the second time derivatives for all of a coupled variable's components
+   * @param var_name Name of coupled variable
+   * @return Vector of VariableValue pointers for each component of \p var_name
+   */
+  std::vector<const VariableValue *> coupledDotDots(const std::string & var_name) const;
```
dschwen, line 1025: Is this used in this PR at all?

framework/src/kernels/ADTimeDerivative.C
```diff
@@ -15,8 +15,9 @@ InputParameters
 ADTimeDerivative::validParams()
 {
   InputParameters params = ADTimeKernelValue::validParams();
-  params.addClassDescription("The time derivative operator with the weak form of $(\\psi_i, "
-                             "\\frac{\\partial u_h}{\\partial t})$.");
+  params.addClassDescription(
+      "The time derivative operator with the weak form of $(\\psi_i, "
+      "\\alpha \\dot \\frac{\\partial u_h}{\\partial t})$, where \\alpha is the coefficient.");
```
dschwen: There is no `\alpha` anymore is there? Why is this part of this PR?

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -63,20 +64,31 @@ ExplicitMixedOrder::validParams()
       "A subset of variables that require first-order integration (velocity only) to be applied by "
       "this time integrator.");
 
+  params.addParam<std::vector<VariableName>>(
+      "phase_field",
+      {},
+      "A subset of variables that require phase field to be applied by "
+      "this time integrator.");
```
dschwen: What does that mean?

framework/include/interfaces/Coupleable.h
YitaoLiu365: Not in this PR, but in another repository based on MOOSE.

framework/include/interfaces/Coupleable.h
YitaoLiu365: Same as coupledDotDots, it's applied to another repository.

PR #30925, opened 2025-07-01
Do not allow derivative string classes to have spaces

### LINKED ISSUE #30924: Do not allow derivative strings to contain spaces
### Motivation

For any of the derivative string classes `SubdomainName`, `UserObjectName`, etc, we should not allow values with spaces. For vectors of those types, we split on spaces. So allowing them with spaces makes no sense.

### Design

Add error check in the `Builder` to not allow scalar derivative string classes to contain spaces.

### Impact

Adds additional error checking and consistency.

loganharbour:
Closes #30924

### DIFF
--- a/framework/src/parser/Builder.C
+++ b/framework/src/parser/Builder.C
@@ -1405,9 +1405,9 @@ Builder::setScalarParameter(const std::string & full_name,
   }
 
   // Make sure that derivative string types have no spaces
-  if constexpr (std::is_base_of_v<std::string, T> && !std::is_same_v<std::string, T>)
+  if constexpr (std::is_base_of_v<std::string, T> && std::is_base_of_v<NoSpace, T>)
   {
-    if (param->get().find_first_of("\t\n ") != std::string::npos)
+    if (param->get().find_first_of("\t\n\f\v\r ") != std::string::npos)
       _errmsg += hit::errormsg(root()->find(full_name),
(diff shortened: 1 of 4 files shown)


GiudGiud (approved):

framework/src/parser/Builder.C
```diff
@@ -1404,6 +1404,16 @@ Builder::setScalarParameter(const std::string & full_name,
       throw;
   }
 
+  // Make sure that derivative string types have no spaces
+  if constexpr (std::is_base_of_v<std::string, T> && !std::is_same_v<std::string, T>)
+  {
+    if (param->get().find_first_of("\t\n ") != std::string::npos)
```
hugary1995: How about \v\f\r? I think \v \f are obsolete, but \r is still used from time to time.

hugary1995 (approved):

hugary1995:
`FunctionName` should be an exception.

hugary1995:
Also `ParsedFunctionExpression`, and possibly `CLIArgString`.

hugary1995:
The remaining failures are real.

GiudGiud:
@loganharbour would you like for someone else to finish this?

loganharbour:
There's a significant rebase that needs to be done. I'll do that and see what comes next in terms of passing

PR #30929, opened 2025-07-02
Spline Utilities

### LINKED ISSUE #30928: Add control points utility
### Motivation

For use with recently-submitted BSpline utility, which relies on the input of control points to be interpolated. Builds on the project of creating a pipe junction component

### Design

Utility will auto-create interior control points to be used with BSpline.C. Control points are selected so that the derivatives at supplied points to be connected are honored, which allows the connection to be as smooth as possible. Additional parameter of `sharpness` was included to give the user control over the shape of the curve. 

### Impact

Supplies general utilities for B-Splines (or other spline types).

lipohlmann:
closes #30928

## Reason
For use with recently-submitted BSpline utility, which relies on the input of control points to be interpolated. Builds on the project of creating a pipe junction component.

## Design
Utility will auto-create interior control points to be used with BSpline.C. Control points are selected so that the derivatives at supplied points to be connected are honored, which allows the connection to be as smooth as possible. Additional parameter of `sharpness` was included to give the user control over the shape of the curve. Users can also vary the control point density.

## Impact
Supplies general utility for B-Splines (or other spline types) which can have a wide application base.

### DIFF
--- a/framework/include/utils/BSpline.h
+++ b/framework/include/utils/BSpline.h
@@ -18,8 +18,9 @@ namespace Moose
 /**
  * Class implementing a uniform clamped B-Spline curve.
  * Reference used for implementation: https://mat.fsv.cvut.cz/gcg/sbornik/prochazkova.pdf
- * The B-Spline is uniform because of the spacing in the knot vector
+ * The B-Spline is uniform because of the spacings in the knot vector
  * The B-Spline is clamped because the knot vector starts (and ends) with (degree + 1) 0s (and 1s)
+ * Clamped means that it passes through the specified start and end points with specified slopes
  */
 class BSpline
 {
--- a/framework/include/utils/BSpline.h
+++ b/framework/include/utils/BSpline.h
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include "libMeshReducedNamespace.h"
 #include "libmesh/point.h"
 #include <vector>
 
@@ -23,7 +24,13 @@ namespace Moose
 class BSpline
 {
 public:
-  BSpline(const unsigned int degree, const std::vector<libMesh::Point> & control_points);
+  BSpline(const unsigned int degree,
+          const libMesh::Point & start_point,
+          const libMesh::Point & end_point,
+          const libMesh::RealVectorValue & start_direction,
+          const libMesh::RealVectorValue & end_direction,
+          const unsigned int cps_per_half,
+          const libMesh::Real sharpness);
 
   /**
    * Evaluate the BSpline interpolation at given value of t.
@@ -33,16 +40,16 @@ public:
    */
   libMesh::Point getPoint(const libMesh::Real t) const;
 
-  /**
-   * Get the built knot vector associated with the degree and control points.
-   */
-  const std::vector<libMesh::Real> & getKnotVector() const { return _knot_vector; }
-
 private:
   /**
    * Internal method for building the knot vector given the degree and control points.
    */
   std::vector<libMesh::Real> buildKnotVector() const;
+
+  /**
+   * Creates a vector control points from the SplineUtils set of functions.
+   */
+  std::vector<libMesh::Point> createControlPoints() const;
 
   /**
    * Evaluates the the basis function for a B-Spline according to the Cox-de-Boor
@@ -74,6 +81,18 @@ private:
 
   /// The polynomial degree of the B-Spline.
   const unsigned int _degree;
+  /// starting point of spline
+  const libMesh::Point _start_point;
+  /// ending point of spline
+  const libMesh::Point _end_point;
+  /// starting direction of spline
+  const libMesh::RealVectorValue _start_dir;
+  /// ending direction of spline
+  const libMesh::RealVectorValue _end_dir;
+  /// number of control points per half of the spline (vertex is auto-included)
+  const unsigned int _cps_per_half;
+  /// sharpness of the curve
+  const libMesh::Real _sharpness;
   /// The control points.
   const std::vector<libMesh::Point> _control_points;
   /// The knot vector.
--- a/framework/src/utils/SplineUtils.C
+++ b/framework/src/utils/SplineUtils.C
@@ -1,0 +1,201 @@
+//* This file is part of the MOOSE framework
+//* https://mooseframework.inl.gov
+//*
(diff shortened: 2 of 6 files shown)


framework/src/utils/SplineUtils.C
```diff
@@ -0,0 +1,201 @@
...
+  /// initialize vector of control points
+  std::vector<Point> control_points;
+
+  /// initialize parameter
+  double t;
+
+  /// loop over the number of control points to generate the contorl points
+  for (const auto i : make_range(num_cps))
+  {
+    t = (double)i / (double)(num_cps - 1);
```
roystgnr: Let's get an assertion in here that num_cps > 1?

framework/src/utils/SplineUtils.C
```diff
@@ -0,0 +1,201 @@
...
+circularControlPoints(const libMesh::Point & start_point,
+                      const libMesh::Point & end_point,
+                      const libMesh::RealVectorValue & parallel_direction,
+                      const unsigned int num_cps)
+{
+  /// establish the direction vector between the two points
+  libMesh::RealVectorValue orthogonal_direction = start_point - end_point;
+
+  /// circular control points will only honor the derivatives if the points are colinear along the orthogonal direction
+  if (!(std::abs(parallel_direction * orthogonal_direction) < 1e-12))
```
roystgnr: Should we use `TOLERANCE` here, and do the dot product using the unit vectors to avoid any scaling issues?

framework/src/utils/SplineUtils.C
```diff
@@ -0,0 +1,201 @@
...
+namespace SplineUtils
+{
+std::vector<Point>
+circularControlPoints(const libMesh::Point & start_point,
+                      const libMesh::Point & end_point,
+                      const libMesh::RealVectorValue & parallel_direction,
+                      const unsigned int num_cps)
+{
+  /// establish the direction vector between the two points
+  libMesh::RealVectorValue orthogonal_direction = start_point - end_point;
```
roystgnr: If you're going to use `const` in pass-by-value function parameters (where the C/C++ standard is annoying about making any future removal of it count as a signature change), you might as well use it in local variables where appropriate (where there's really no downside).

framework/src/utils/SplineUtils.C
```diff
@@ -0,0 +1,201 @@
...
+  libMesh::RealVectorValue unit_parallel = parallel_direction / parallel_direction.norm();
+
+  /// calculate circle center
+  libMesh::Point center_point = end_point + radius * unit_normal;
+
+  /// initialize vector of control points
+  std::vector<Point> control_points;
+
+  /// initialize parameter
+  double t;
```
roystgnr: No reason not to have this declared where it's set, inside the loop.  It's a `double`, there's no constructor cost.

framework/include/utils/BSpline.h
```diff
@@ -0,0 +1,70 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+// Reference for B-Spline: https://mat.fsv.cvut.cz/gcg/sbornik/prochazkova.pdf
+
+#pragma once
+
+#include <vector>
+#include "libmesh/point.h"
```
roystgnr: Let's put headers (here and elsewhere) into the "canonical" order?  In Foo.C, you first include Foo.h; after that (or in Foo.h) you include other MOOSE headers first, then dependency headers (like libmesh/whatever.h) second, then standard headers (like vector) third.

That makes it easier to catch bugs where a header doesn't include all it's proper dependencies but "gets away with it" because those dependencies happen to get included before it in its current use case(s).

roystgnr (changes requested):
I have a few nitpicks, but the one thing I think is most worth waiting on is the mesh generator using these utilities; it's possible that you'll want to make API changes here before that's completely finished, and if so then better to make them in this PR than in a new one.

If there's reason to merge earlier though, I wouldn't argue.

framework/include/utils/BSpline.h
GiudGiud: fixing this in the 1st PR and will propagate here

framework/src/utils/SplineUtils.C
GiudGiud: removed

framework/src/utils/SplineUtils.C
GiudGiud: added

framework/src/utils/SplineUtils.C
GiudGiud: added

GiudGiud (approved):
Review should be addressed.
Rebased on top of the first PR

@roystgnr if you want to take a look

GiudGiud:
@roystgnr the additional 7 commits come from PR 3. You have not reviewed those yet

GiudGiud:
Test failures unrelated

framework/src/utils/SplineUtils.C
```diff
@@ -66,38 +70,36 @@ bSplineControlPoints(const libMesh::Point & start_point,
...
-  ///
-  /// start with a couple of checks to be sure the input is good
-  ///
-  /// check that start_point is different from end_point
-  if ((start_point - end_point).norm() < libMesh::TOLERANCE)
-    mooseError("Start and end points must be unique.");
-  /// check that neither direction is the zero vector
-  if (start_direction.norm() < libMesh::TOLERANCE || end_direction.norm() == 0)
+  // check that start_point is different from end_point
+  if ((start_point - end_point).norm_sq() < libMesh::TOLERANCE)
```
roystgnr: For the cases where we're using `norm_sq` we probably also want to use `TOLERANCE * TOLERANCE` or we're getting much looser a tolerance than we'd want.

roystgnr (changes requested):
I think we've *almost* got this.

GiudGiud:
rebased and addressed the comment

GiudGiud:
Test failures are unrelated

roystgnr (approved):
CI failures were unrelated.  I'm kicking the failed recipes because I like to see green, but if they're again thwarted by network/system failures then I'd say let's just merge anyway.

GiudGiud:
passed. thanks for the review!

PR #30937, opened 2025-07-02
Refactor the kernels (IC, aux) that populate q_prime

kyriv-lab:
This PR improves issue refs #30639, by refactoring how the q_prime variable is populated for the Quad and Tri problems. either by ICs or Aux kernels.

The aux kernel is a copy paste version of the IC kernel but has the ability to populate q_prime not just in the initialization of the problem.


modules/subchannel/validation/psbt/psbt_ss/gold/psbt_out.csv
```diff
@@ -1,3 +1,3 @@
 time,PinTemp,T1,T2,T3,T4,T5,T6,total_pressure_drop
 0,0,0,0,0,0,0,0,0
-1,568.16940826925,567.50631420582,566.94159790073,563.67878642056,558.57772522824,553.83624505415,551.39044907974,255877.63333888
+1,612.02990188326,567.99855284885,568.5500625951,564.70705442123,558.19402276689,552.63785457705,549.84218031668,255377.02437708
```
GiudGiud, line 3: this is a huge change in the pin temperature.

what is the total power before and after?

modules/subchannel/validation/ORNL_19_pin/ORNL_19.i
```diff
@@ -62,19 +62,15 @@ P_out = 2.0e5 # Pa
...
-  segregated = true
+  implicit = true
+  segregated = false
   staggered_pressure = false
-  monolithic_thermal = false
+  monolithic_thermal = true
   verbose_multiapps = true
-  verbose_subchannel = false
-  interpolation_scheme = 'upwind'
+  verbose_subchannel = true
```
GiudGiud, line 71: why is this needed? I would rather the test be unchanged with the fix on IC and aux

modules/subchannel/validation/EBR-II/gold/XX09_SCM_SS45R_out.csv
```diff
@@ -1,3 +1,3 @@
 time,TTC-27,TTC-28,TTC-29,TTC-30,TTC-31,TTC-32,TTC-33,TTC-34,TTC-35
 0,0,0,0,0,0,0,0,0,0
-1,730.8246493617,747.87030952825,751.72926821944,753.87560822835,757.33473312294,762.46001148726,766.64909792382,752.99009374971,725.11475484002
+1,734.51559327971,767.69830582236,777.57169040122,778.8349208326,778.83607046477,777.56627203333,767.60662789535,733.80591988158,705.89109115469
```
GiudGiud: these are big changes too. Wasnt the EBR-II mesh aligned with the boundaries???

modules/subchannel/validation/EBR-II/tests
```diff
@@ -29,6 +29,7 @@
     type = CSVDiff
     input = 'XX09_SCM_TR17.i'
     csvdiff  = 'XX09_SCM_TR17_out.csv'
+    cli_args = 'Executioner/num_steps=20'
     heavy = true
     capabilities = 'method!=dbg'
     abs_zero = 1e-5
```
GiudGiud, line 59: since you added more inputs in this folder, I am expecting more tests?

modules/subchannel/test/tests/SCMTriPower/tests
```diff
@@ -0,0 +1,21 @@
+[Tests]
+  design = 'SCMTriPowerAux.md SCMTriPowerIC.md'
+  issues = '#28497 #29194'
+  [SCMTriPower_cells_aligned]
+    type = CSVDiff
+    input = 'test.i'
+    csvdiff = 'test_out_line_check_0001.csv'
+    recover = False
+    abs_zero = 1e-7
+    requirement = 'The system will populate the linear heat rate on subchannels in a triangular lattice using an IC/aux kernel'
```
GiudGiud:
```suggestion
    requirement = 'The system will populate the linear heat rate on subchannels in a triangular lattice using an either an initial condition or an auxiliary kernel'
```

modules/subchannel/test/tests/SCMTriPower/tests
```diff
@@ -0,0 +1,21 @@
...
+  []
+  [SCMTriPower_cells_not_aligned]
+    type = CSVDiff
+    input = 'test.i'
+    csvdiff = 'not_aligned_line_check_0001.csv'
+    cli_args = 'num_cells=11 Outputs/file_base=not_aligned'
+    recover = False
+    abs_zero = 1e-7
+    requirement = 'The system will populate the linear heat rate on subchannels in a triangular lattice using an IC/aux kernel in a not aligned mesh'
```
GiudGiud:
```suggestion
    requirement = 'The system will populate the linear heat rate on subchannels in a triangular lattice using an IC/aux kernel in a mesh that has axial divisions that are not aligned with the heated region entrance and exit heights.'
```

modules/subchannel/test/tests/SCMTriPower/test.i
```diff
@@ -0,0 +1,93 @@
...
+  [line_check]
+    type = LineValueSampler
+    variable = 'q_prime_ic q_prime_aux'
+    execute_on = 'TIMESTEP_END'
+    sort_by = 'z'
+    start_point = '0 0 0'
+    end_point = '0 0 1.5'
+    num_points = ${fparse num_cells + 1}
+  []
+[]
```
GiudGiud, line 192: please add this SCMPowerPostprocessor that can compute the true power

modules/subchannel/test/tests/SCMQuadPower/tests
```diff
@@ -0,0 +1,21 @@
...
+    requirement = 'The system will populate the linear heat rate on subchannels in a quadrilateral lattice using an IC/aux kernel'
+  []
+  [SCMQuadPower_cells_not_aligned]
+    type = CSVDiff
+    input = 'test.i'
+    csvdiff = 'not_aligned_line_check_0001.csv'
+    cli_args = 'num_cells=11 Outputs/file_base=not_aligned'
+    recover = False
+    abs_zero = 1e-7
+    requirement = 'The system will populate the linear heat rate on subchannels in a quadrilateral lattice using an IC/aux kernel in a not aligned mesh'
```
GiudGiud: same changes

modules/subchannel/src/problems/SubChannel1PhaseProblem.C
```diff
@@ -55,7 +55,7 @@ formFunction(SNES, Vec x, Vec f, void * ctx)
 InputParameters
 SubChannel1PhaseProblem::validParams()
 {
-  MooseEnum schemes("upwind downwind central_difference exponential", "central_difference");
+  MooseEnum schemes("upwind downwind central_difference exponential", "exponential");
```
GiudGiud: is this the change causing all those large regolds?

if so please add in the inputs a manual selection of the inteprolation method to prevent regolds.

modules/subchannel/src/ics/SCMTriPowerIC.C
```diff
@@ -96,39 +95,58 @@ SCMTriPowerIC::initialSetup()
...
+    total_power += _estimate_power(i_pin);
     if (_estimate_power(i_pin) == 0.0)
       _estimate_power(i_pin) = 1.0;
   }
   // We need to correct the linear power assigned to the nodes of each pin
   // so that the total power calculated  by the trapezoidal rule agrees with the power assigned by
   // the user.
   _pin_power_correction = _ref_power.cwiseQuotient(_estimate_power);
+  _console << "###########################################" << std::endl;
+  _console << "Total power estimation: " << total_power << " [W] " << std::endl;
```
GiudGiud: is this pre- or post-correction? pre-correction would freak out the user.

GiudGiud (changes requested):
some changes are larger than we expect. please focus the PR on only the IC & aux! one problem at a time, we should only see changes in the meshes-not-aligned case here

modules/subchannel/validation/psbt/psbt_ss/gold/psbt_out.csv
kyriv-lab: This makes sense. I run the test case with the previous version of the ICs and got PinTemp 568.17 at a height of 3.658. Then i changed the height to 3.4 and got 613.5. Which means that the previous IC applies a very low or lower q_prime at the end node. The new results are more consistent to a uniform power profile.

modules/subchannel/validation/ORNL_19_pin/ORNL_19.i
kyriv1980: The exponential scheme doesn't converge for Lead and LBE ... i guess including Prandlt number calculations for these materials makes things harder to converge. The upwind scheme produces reasonable results.

GiudGiud: you are removing `interpolation_scheme = 'upwind'` here ? So you are changing it to exponential?

modules/subchannel/validation/EBR-II/gold/XX09_SCM_SS45R_out.csv
kyriv-lab: I changed the power profile of that test. With the old profile there is no appreciable change.

modules/subchannel/validation/EBR-II/tests
kyriv-lab: I want the local test to run the entirety of the transient. But the civet test to be shorter, thats why i used cli_args.

modules/subchannel/src/ics/SCMTriPowerIC.C
kyriv-lab: pre-correction

modules/subchannel/validation/psbt/psbt_ss/gold/psbt_out.csv
kyriv-lab: For a uniform power profile we expect the temperature at the exit of the pin to be the highest. So now we have corrected that.

modules/subchannel/validation/ORNL_19_pin/ORNL_19.i
kyriv-lab: Sorry my previous comment is wrong. I will revert to upwind.

modules/subchannel/src/problems/SubChannel1PhaseProblem.C
kyriv-lab: no.

GiudGiud: so all these regolds are from misaligned meshes?

kyriv-lab: The change to exponential doesn't affect the distribution of q_prime but it does affect the solution a little bit. All the changes change the solution a little bit. But the most important is due to the changes in q_prime.

kyriv-lab: The change to exponential alone would require the regolding of 13 tests. Though the rel changes are in the range of 1e-3 to 1e-5.

modules/subchannel/src/ics/SCMTriPowerIC.C
GiudGiud: can you add pre-correction then? in the log. Otherwise it looks like we are reporting the wrong power in the IC!

modules/subchannel/src/problems/SubChannel1PhaseProblem.C
GiudGiud: ok let's not do the change to exponential in this PR. So any test that is using central_difference right now should keep using it

modules/subchannel/validation/EBR-II/tests
GiudGiud: you create a new .i file in this folder. But I dont see a test for it?

modules/subchannel/validation/EBR-II/gold/XX09_SCM_SS45R_out.csv
GiudGiud: what motivates the change in power profile? This is a validation case, shouldnt the power profile be the one from the validation study?

modules/subchannel/validation/psbt/psbt_ss/gold/psbt_out.csv
GiudGiud: ok great!

modules/subchannel/test/tests/SCMTriPower/tests
GiudGiud: reminder
"not aligned mesh" is not precise enough

modules/subchannel/examples/MultiApp/gold/fuel_assembly_out_viz0_sample_pin_center_0001.csv
```diff
@@ -1,52 +1,52 @@
 Dpin,Tpin,id,q_prime,x,y,z
```
GiudGiud, line 1: I dont see any change in q_prime in this gold file (everything was aligned) 

yes Tpin is changing. Why is that? Is that the exponential inteprolation change?

modules/subchannel/src/auxkernels/SCMQuadPowerAux.C
```diff
@@ -80,57 +80,78 @@ SCMQuadPowerAux::SCMQuadPowerAux(const InputParameters & parameters)
 void
 SCMQuadPowerAux::initialSetup()
 {
+  auto nx = _quadMesh.getNx();
+  auto ny = _quadMesh.getNy();
+  auto n_pins = (nx - 1) * (ny - 1);
```
GiudGiud, line 85: is that always the case? can't you have instrumented / CR channels?

GiudGiud: For those "pins", the power map should show a 0 right?

maybe it should be called n_cells or n_lattice_pins ?

modules/subchannel/test/tests/SCMQuadPower/gold/test_out.csv
```diff
@@ -0,0 +1,3 @@
+time,Total_power_Aux,Total_power_IC
+0,0,0
+1,4128000,4128000
```
GiudGiud: this is way off... that's why we need a SCM-dedicated post processor

modules/subchannel/validation/EBR-II/XX09_SCM_SS17_corrected.i
```diff
@@ -0,0 +1,400 @@
+# Following Benchmark Specifications and Data Requirements for EBR-II Shutdown Heat Removal Tests SHRT-17 and SHRT-45R
```
GiudGiud, line 1: this input, where is it tested?

modules/subchannel/validation/EBR-II/XX09_SCM_SS45R_corrected.i
```diff
@@ -0,0 +1,400 @@
+# Following Benchmark Specifications and Data Requirements for EBR-II Shutdown Heat Removal Tests SHRT-17 and SHRT-45R
```
GiudGiud, line 1: same question?

GiudGiud (changes requested):

modules/subchannel/validation/EBR-II/gold/XX09_SCM_SS45R_out.csv
kyriv-lab: There are two cases considered. One with a uniform power profile and one with a corrected. In the previous version there was one input file that used the corrected radial profile distribution. I wanted to have two input files and the outputs from these two files to be read by a python script and results printed. So i edited accordingly. The name of the files had to change so i couldn't keep the same gold files.

kyriv-lab: I added two more tests for the two new .i files

modules/subchannel/validation/EBR-II/tests
kyriv-lab: I added two more tests for the two new .i files

modules/subchannel/src/problems/SubChannel1PhaseProblem.C
kyriv-lab: ok, but that means regolding again.

modules/subchannel/examples/MultiApp/gold/fuel_assembly_out_viz0_sample_pin_center_0001.csv
kyriv-lab: This is a test that uses the explicit solver. In that solver values are hardcoded upwinded so the interpolation scheme doesn't affect the results. I figure it is the change in the enthalpy calculation from the heated pins.

modules/subchannel/src/auxkernels/SCMQuadPowerAux.C
kyriv-lab: The pins that have no power get a 0 on the radial power profile map.

modules/subchannel/test/tests/SCMQuadPower/gold/test_out.csv
kyriv-lab: The not aligned case is exact though meaning that the moose way of calculation matches the SCM calculation for not aligned meshes.

modules/subchannel/validation/EBR-II/XX09_SCM_SS17_corrected.i
kyriv-lab: I have added a test for that.

modules/subchannel/validation/EBR-II/XX09_SCM_SS45R_corrected.i
kyriv-lab: I have added a test for that.

modules/subchannel/validation/EBR-II/tests
GiudGiud: ok I see them now!

modules/subchannel/validation/EBR-II/gold/XX09_SCM_SS45R_out.csv
GiudGiud: ok I see the old CSV results are still there, only very slightly different in those _corrected_out files/

This works for me,

modules/subchannel/src/problems/SubChannel1PhaseProblem.C
kyriv-lab: I don't like it, central difference messes up the Blockage validation.

modules/subchannel/validation/ORNL_19_pin/ORNL_19.i
GiudGiud: ok the effect of the other changes is minimal on the test resuls. We can keep them

modules/subchannel/validation/EBR-II/XX09_SCM_SS17_corrected.i
GiudGiud: ok I see it now

modules/subchannel/validation/EBR-II/XX09_SCM_SS17_corrected.i
```diff
@@ -0,0 +1,400 @@
...
+    n_cells = 50
+    flat_to_flat = ${inner_duct_in}
+    unheated_length_exit = ${unheated_length_exit}
+    heated_length = ${heated_length}
+    pin_diameter = ${fuel_pin_diameter}
+    pitch = ${fuel_pin_pitch}
+    dwire = ${wire_diameter}
+    hwire = ${wire_z_spacing}
+    spacer_z = '0.0'
+    spacer_k = '0.0'
```
GiudGiud:
```suggestion
```

modules/subchannel/validation/EBR-II/XX09_SCM_SS17_corrected.i
```diff
@@ -0,0 +1,400 @@
...
+    height = 0.322
+  []
+  [TTC-35]
+    type = SubChannelPointValue
+    variable = T
+    index = 107
+    execute_on = 'TIMESTEP_END'
+    height = 0.322
+  []
+  # [MTC-20]
```
GiudGiud: what s the reason behind not including these in the output?

modules/subchannel/validation/EBR-II/XX09_SCM_SS45R_corrected.i
```diff
@@ -0,0 +1,400 @@
...
+    n_cells = 50
+    flat_to_flat = ${inner_duct_in}
+    unheated_length_exit = ${unheated_length_exit}
+    heated_length = ${heated_length}
+    pin_diameter = ${fuel_pin_diameter}
+    pitch = ${fuel_pin_pitch}
+    dwire = ${wire_diameter}
+    hwire = ${wire_z_spacing}
+    spacer_z = '0.0'
+    spacer_k = '0.0'
```
GiudGiud:
```suggestion
```

modules/subchannel/validation/Blockage/THORS/FFM-5B_high.i
```diff
@@ -30,13 +30,10 @@ P_out = 2.0e5 # Pa
...
     hwire = 0.3048
-    spacer_z = '0.0'
-    spacer_k = '0.0'
     z_blockage = '0.49 0.52'
     index_blockage = '29 31 30 32 34 33 35 15 16 8 17 18 9 19'
-    reduction_blockage = '0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2 0.2'
-    k_blockage = '1.2 1.2 1.2 1.2 1.2 1.2 1.2 1.2 1.2 1.2 1.2 1.2 1.2 1.2 '
-
+    reduction_blockage = '0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1 0.1'
+    k_blockage = '5 5 5 5 5 5 5 5 5 5 5 5 5 5 '
```
GiudGiud, line 36: what s the reason for this?

modules/subchannel/validation/Blockage/THORS/FFM-3A.i
```diff
@@ -77,15 +77,15 @@ P_out = 2.0e5 # Pa
   fp = sodium
   n_blocks = 1
   P_out = 2.0e5
-  CT = 2.2
+  CT = 2
```
GiudGiud, line 80: what motivates this change?

modules/subchannel/test/tests/problems/SFR/sodium-19pin/tests
```diff
@@ -35,7 +36,7 @@
     type = CSVDiff
     input = test19_full_monolithic.i
     csvdiff = test19_full_monolithic_out.csv
-    rel_err = 5e-4
+    rel_err = 2e-3
```
GiudGiud: let's regold? 2e-3 cant be civet vs your mac

kyriv-lab:
- The EBR-II validation tests were regolded because I changed the names of the input files and the radial power profiles.
- The Lead-LBE tests and gold files changed because I changed the interpolation scheme to exponential.
- The Thors Blockage validation changed because I changed the interpolation scheme to exponential and the q_prime IC kernel has changed. Then the validation case was edited (adapted friction coefficients and flow area reduction) to match the experiment.
- The PSBT validation gold file was changed because the q_prime IC kernel corrected the q_prime distribution on the fuel pins.
- The Multiapp example changed due to the change in q_prime IC.
- The ORNL validation case changed because of editing the solver scheme (monolithic = true) and the change in the q_prime IC

modules/subchannel/validation/psbt/psbt_ss/gold/psbt_out.csv
GiudGiud: @tanoret says shouldnt boil

GiudGiud:
For ORNL-19 you should have more data to compare to in the CSVDiff
see bottom of https://mooseframework.inl.gov/virtualtestbed/sfr/subchannel/ornl_19_pin/ornl_19_pin.html

GiudGiud (changes requested):
I think you changed the friction coefficients as well? Should go in the justifications

modules/subchannel/validation/EBR-II/XX09_SCM_SS17_corrected.i
kyriv-lab: I wanted to print one set of measurements at a time.

modules/subchannel/validation/Blockage/THORS/FFM-5B_high.i
kyriv-lab: The changes in the q_prime IC, interpolation scheme, force to redo the validation exercise. Blockage modeling is very sensitive.

modules/subchannel/validation/EBR-II/XX09_SCM_SS17_corrected.i
GiudGiud: I think for validation purposes, 2 sets better than 1

modules/subchannel/validation/Blockage/THORS/FFM-3A.i
kyriv-lab: There was no observable change between the results for these two parameters.

kyriv-lab:
Regarding PSBT boiling concerns: It is not a regression error. I run the input file with the previous version of the code and i get: 
height: 3.658 Tpin: 568.16940826925
height: 3.6 Tpin: 582.90626017904
height: 3.5 Tpin: 608.31462554075
height: 3.4 Tpin: 613.5860154605
height: 3.0 Tpin: 607.74302253826

The expected behavior is to have the highest Pin Temp at the exit of the pin since q_prime is uniform. This leads me to believe that q_prime was not populated correctly in the previous version.

modules/subchannel/src/ics/SCMQuadPowerIC.C
```diff
@@ -93,47 +93,67 @@ SCMQuadPowerIC::initialSetup()
 {
```
GiudGiud, line 93: so this code for computing the power from the function looks exactly the same in SCMQuadPowerAux and SCMQuadPowerIC. 

let's make a  SCMPowerUtils.C file that will house this common code for the 2 objects.

GiudGiud: same for Tri

modules/subchannel/src/problems/QuadSubChannel1PhaseProblem.C
```diff
@@ -426,26 +426,38 @@ QuadSubChannel1PhaseProblem::computeBeta(unsigned int i_gap, unsigned int iz)
 Real
 QuadSubChannel1PhaseProblem::computeAddedHeatPin(unsigned int i_ch, unsigned int iz)
```
GiudGiud, line 427: the SCMPower  postprocessor could just call this routine btw to make the integral over pins

this would make it very consistent imo

modules/subchannel/test/tests/SCMQuadPower/gold/test_out.csv
GiudGiud: yes that's good for that case but it's also sort of a coincidence. The way power is computed in the problem does not match how the PP is doing the integral. Especially since there are two cases: with and without a pin mesh. So in total 4 cases to check.

we also can't have a wrong case (the aligned one)

modules/subchannel/test/tests/SCMTriPower/gold/test_out.csv
```diff
@@ -0,0 +1,3 @@
+time,Total_power_Aux,Total_power_IC
+0,0,0
+1,500000,500000
```
GiudGiud: matches for this case! why? quad does not match

modules/subchannel/validation/psbt/psbt_ss/gold/psbt_out.csv
GiudGiud: resolved in person, 612 is pin not coolant

GiudGiud (commented):
all good after these:

- include all measurements that are part of validation basis for EBR-II/XX09_SCM_... cases in CSVDiff
- include all measurements that are part of validation basis for ORNL-19 (see earlier post) in CSVDiff
These two could be a separate PR but it's best to do it now so we have history for future changes

- add a SCMPowerPostprocessor to verify all conservation cases (tri/quad, pin mesh/no pin mesh, aligned/aligned) directly in the CSVDiff rather than 'coincidentally' through nodal integration of q_prime

modules/subchannel/src/ics/SCMQuadPowerIC.C
GiudGiud: you could delay these changes to a future PR / issue if you want

modules/subchannel/test/tests/SCMQuadPower/gold/test_out.csv
kyriv-lab: I will make an issue to create an SCM-dedicated power calculation post-processor.  Which i think should be based on the solution  (mass-flow rate and enthalpy)???

modules/subchannel/src/ics/SCMQuadPowerIC.C
kyriv-lab: I will make an issue about it.

modules/subchannel/test/tests/SCMTriPower/gold/test_out.csv
kyriv-lab: The power profile in the tri case is sinusoidal and in quad uniform. So for the tri case the moose power post-processor has wings at the edges even for the aligned case.

kyriv-lab:
I will make one Issue/Pr to enrich the ORNL validation base. And one Issue/Pr to refactor the q_prime ICs/Aux kernels .

GiudGiud:
that's a great idea to split into separate PRs. However I would want both to be merged before merging this PR

modules/subchannel/test/tests/SCMTriPower/gold/test_out.csv
GiudGiud: This does not make sense to me? tri + sinusoidal -> exact
quad + uniform -> not exact?

kyriv-lab:
> that's a great idea to split into separate PRs. However I would want both to be merged before merging this PR

The refactoring of ICs/Aux Kernels would include all the changes that are part of this PR so i should have to merge this first or include the refactoring here. Same goes for the new post-processor. 

For the ORNL, i think it should be a new PR that can be merged afterwards.

GiudGiud:
> Same goes for the new post-processor.

The new postprocessor uses a routine on the problem that already exists. You dont need this PR to make it

> For the ORNL, i think it should be a new PR that can be merged afterwards.

No it's independent as well. I'm reluctant to regold things in this PR without having the additional data points

What about the additional EBR measurements?

kyriv-lab:
> > Same goes for the new post-processor.
> 
> The new postprocessor uses a routine on the problem that already exists. You dont need this PR to make it
> 
> > For the ORNL, i think it should be a new PR that can be merged afterwards.
> 
> No it's independent as well. I'm reluctant to regold things in this PR without having the additional data points
> 
> What about the additional EBR measurements?

computeAddedHeatPin(unsigned int i_ch, unsigned int iz) is edited in this PR. So the POWER postprocessor that will be based on this method must be created/merged after or here.

I have added the additional EBR measurements in this PR.

The additional ORNL SCM calculations are not readily available because only one set was included from the start  in both SCM and VTB.

modules/subchannel/test/tests/SCMQuadPower/gold/test_out.csv
kyriv-lab: Post-processor has been created and is included in this PR.

modules/subchannel/test/tests/SCMTriPower/gold/test_out.csv
kyriv-lab: Only if the nodes are aligned the the power profile is uniform does the default q_prime integration is inexact. Because it assumes a linear power distribution at the edge subchannels.

kyriv-lab: for all other cases the moose q_prime integration is exact.

modules/subchannel/src/ics/SCMQuadPowerIC.C
GiudGiud: have you created it yet?

modules/subchannel/src/postprocessors/SCMPowerPostprocessor.C
```diff
@@ -0,0 +1,63 @@
...
+#include "SCMPowerPostprocessor.h"
+#include "SolutionHandle.h"
+#include "FEProblemBase.h"
+#include "Function.h"
+#include "MooseMesh.h"
+#include "MooseVariable.h"
+#include "SubProblem.h"
+#include "libmesh/system.h"
+#include "SCM.h"
+#include "SubChannel1PhaseProblem.h"
```
GiudGiud:
```suggestion
#include "SCMPowerPostprocessor.h"
#include "FEProblemBase.h"
#include "MooseMesh.h"
#include "SCM.h"
#include "SubChannel1PhaseProblem.h"
```

modules/subchannel/src/postprocessors/SCMPowerPostprocessor.C
```diff
@@ -0,0 +1,63 @@
...
+  if (!scm_problem)
+    mooseError("SCMPowerPostprocessor can only be used within a subchannel 1phase problem.");
+  auto Power = 0.0;
+  for (unsigned int iz = 1; iz < nz + 1; iz++)
+  {
+    for (unsigned int i_ch = 0; i_ch < n_channels; i_ch++)
+    {
+      Power += scm_problem->computeAddedHeatPin(i_ch, iz);
+    }
+  }
```
GiudGiud:
```suggestion
  const auto nz = _mesh.getNumOfAxialCells();
  const auto n_channels = _mesh.getNumOfChannels();
  if (!scm_problem)
    mooseError("SCMPowerPostprocessor can only be used within a subchannel 1phase problem.");
  auto power = 0.0;
  for (unsigned int iz = 1; iz < nz + 1; iz++)
    for (unsigned int i_ch = 0; i_ch < n_channels; i_ch++)
      power += scm_problem->computeAddedHeatPin(i_ch, iz);
```

modules/subchannel/test/tests/SCMTriPower/tests
```diff
@@ -0,0 +1,42 @@
...
+    requirement = 'The system will populate a uniform linear heat rate on subchannels in a triangular lattice using an initial condition and auxiliary kernel'
+  []
+  [SCMTriPower_cells_aligned_with_pins]
+    type = CSVDiff
+    input = 'test_with_pins.i'
+    csvdiff = 'test_with_pins_out.csv test_with_pins_out_line_check_0001.csv'
+    recover = False
+    abs_zero = 1e-7
+    max_parallel = 1
+    requirement = 'The system will populate a sunisoidal linear heat rate on pins in a triangular lattice using an initial condition and auxiliary kernel'
```
GiudGiud:
```suggestion
    requirement = 'The system will populate a sinusoidal linear heat rate on pins in a triangular lattice using an initial condition and auxiliary kernel'
```

modules/subchannel/test/tests/SCMTriPower/tests
```diff
@@ -0,0 +1,42 @@
...
+  []
+  [SCMTriPower_cells_not_aligned_with_pins]
+    type = CSVDiff
+    input = 'test_with_pins.i'
+    csvdiff = 'not_aligned_with_pins.csv not_aligned_with_pins_line_check_0001.csv'
+    cli_args = 'num_cells=31 Outputs/file_base=not_aligned_with_pins'
+    recover = False
+    abs_zero = 1e-7
+    max_parallel = 1
+    requirement = 'The system will populate a sunisoidal linear heat rate on pins in a triangular lattice using an initial condition and auxiliary kernel in a mesh that has axial divisions that are not aligned with the heated region entrance and exit heights.'
```
GiudGiud:
```suggestion
    requirement = 'The system will populate a sinusoidal linear heat rate on pins in a triangular lattice using an initial condition and auxiliary kernel in a mesh that has axial divisions that are not aligned with the heated region entrance and exit heights.'
```

GiudGiud (commented):
last things + that issue to create

modules/subchannel/src/ics/SCMQuadPowerIC.C
kyriv-lab: Yes : https://github.com/idaholab/moose/issues/31053

GiudGiud (approved):

PR #30951, opened 2025-07-03
Make optional library check a function and use for both libtorch/mfem

### LINKED ISSUE #30950: Make sure to error for missing libtorch/mfem libraries when make command goals empty
### Motivation

Not so long ago we edited the logic to not error for these optional dependency libs if they were missing but we were doing "non-build" make targets like `make clean`. However, the logic change missed out on addressing the most common make command which is the empty make command goal `make`

### Design

Fix the makefile logic. Make a function that both libtorch/mfem can call

### Impact

Make errors when we should have them

lindsayad:
Closes #30950

### DIFF
--- a/framework/moose.mk
+++ b/framework/moose.mk
@@ -102,10 +102,23 @@ libmesh_LDFLAGS   += $(wasp_LDFLAGS)
 #
 include $(FRAMEWORK_DIR)/contrib/neml2.mk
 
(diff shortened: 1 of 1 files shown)


lindsayad:
failure is unrelated

grmnptr (approved):
Wow, cool generalization!

lindsayad:
thanks!

PR #30956, opened 2025-07-04
Initialise Hypre before setDevice in MFEM problems

### LINKED ISSUE #30955: Hypre::Init not being called in MFEM problems before first usage of Hypre objects
### Bug Description

Hypre is not consistently initialised before the first usage of Hypre in MFEM objects, possibly leading to a segfault if it has not been initialised elsewhere prior. 

Initialisation before creation of Hypre solvers in MFEM problems was addressed in #30660 , but usage of other Hypre objects - including inside internal MFEM methods such as in `mfem::ParFiniteElementSpace::GetEssentialTrueDofs`. Hypre initialisation should instead be performed prior to MFEM device initialisation, since this can depend on Hypre's intialisation state.

### Steps to Reproduce

Segfault reported by @sean-baccas during testing of #30943 with `EnsureNCMesh()` called in the `MFEMMesh` constructor, at the first call of `Moose::MFEM::EquationSystem::ApplyEssentialBCs`.

### Impact

Risk of unexpected segfault due to first usage of Hypre objects, leading to crashes and unclear errors for users.

### [Optional] Diagnostics

_No response_

alexanderianblair:
Closes #30955

### DIFF
--- a/framework/src/mfem/executioners/MFEMExecutioner.C
+++ b/framework/src/mfem/executioners/MFEMExecutioner.C
@@ -32,6 +32,7 @@ MFEMExecutioner::MFEMExecutioner(const InputParameters & parameters)
     _mfem_problem(dynamic_cast<MFEMProblem &>(feProblem())),
     _problem_data(_mfem_problem.getProblemData())
 {
+  mfem::Hypre::Init();
   setDevice();
 }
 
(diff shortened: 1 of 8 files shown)


nmnobre:
Grand! Would it be safe to revert 2cb798a as well then?

alexanderianblair:
> Grand! Would it be safe to revert [2cb798a](https://github.com/idaholab/moose/commit/2cb798a9e407828305dda1f05919811aa112fa9b) as well then?

It should indeed - updated

nmnobre (approved):
Double-checked this still initialises hypre as expected with an example we know fails otherwise: `./moose_test-dbg -i tests/mfem/kernels/diffusion.i Solver/type=MFEMCGSolver`. If I were being picky, I'd say squash and tidy up the commit messages but... it's Friday night :)

PR #30959, opened 2025-07-04
Uniformly refine `bounds` in `LayeredBase`

### LINKED ISSUE #30958: Add uniform splits for the `bounds` parameter in `LayeredBase` derived classes
### Motivation

`LayeredBase` derived classes are incredibly useful for post-processing simulation results, especially if one only cares about an unknown in a single direction (such as axial heating in neutronics). They become rather cumbersome to use when the mesh contains many non-uniform regions in a direction, as it's easily enough to provide `bounds` for subdomains, but if you need many subdivisions to capture rapidly varying fields `bounds` starts to consume a great deal of space in input files.

### Design

A new parameter (called `bound_uniform_splits` or similar) should be added which specifies the number of times the bins in `bounds` should be uniformly split. 

### Impact

New addition to improve solution post-processing with `LayeredBase` derived classes.

nuclearkevin:
Closes #30958


framework/src/userobjects/LayeredBase.C
```diff
@@ -69,7 +73,7 @@ LayeredBase::validParams()
   params.addParam<Real>("direction_max",
                         "Maximum coordinate along 'direction' that bounds the layers");
   params.addParamNamesToGroup("direction num_layers bounds direction_min direction_max",
-                              "Layers extent and definition");
+                              "Layers extent and definition bound_uniform_splits");
```
loganharbour: This got put in the wrong place

framework/src/userobjects/LayeredBase.C
```diff
@@ -117,6 +121,19 @@ LayeredBase::LayeredBase(const InputParameters & parameters)
     // Make sure the bounds are sorted - we're going to depend on this
     std::sort(_layer_bounds.begin(), _layer_bounds.end());
 
+    // If requested, we uniformly split the layers.
+    if (_layered_base_params.isParamValid("bound_uniform_splits"))
```
loganharbour, line 129: I don't see where the actual value of bound_uniform_splits is used?

loganharbour (changes requested):

framework/src/userobjects/LayeredBase.C
nuclearkevin: Shoot, I have no idea how I missed this rookie mistake. Guess this is what happens when one codes at a Starbucks and gets distracted...

nuclearkevin: It's a miracle the test even passes. I checked the output and it split the bins in half uniformly (as intended), but I guess the stars aligned just right to make it "work" (even though the feature does not).

loganharbour (commented):
What happens if `bounds` is set to one value?

nuclearkevin:
> What happens if bounds is set to one value?

The case where only a single value is provided in `bounds` is erroneous and should be handled with an error message. It appears as if `LayeredBase` doesn't do this, and there isn't anything for it to fall back on. This is the resulting error message when setting `bounds = '0.2'` without uniform bound refinement:
```
*** ERROR ***
Layer '0' not found in 'average'.
```
When uniform splits are used, the for loops spins indefinitely. I'll add a check + test to make sure users specify more than a single value in `bounds`.

nuclearkevin:
@loganharbour apologies for being a stick in the mud, but this should be ready for another round of review (when you have the time).

loganharbour (approved):

PR #30995, opened 2025-07-09
[WIP] Hp adaptivity through local smoothness and posteriori error estimate

skvarjun:
## Reason

**Pure h adaptivity may leads to poor convergence in certain scanarios;**

1. Singularities and sharp gradients (e.g., crack tips, re-entrant corners as in 'the L-shaped domains'): require local h-refinement and p-enrichment to accurately capture stress concentrations.
2. Boundary layers and multi-scale phenomena: demand h-refinement and variable polynomial orders to resolve steep gradients without global mesh refinement.
3. Solution smoothnes varying across the domain: hp-adaptivity adapts  higher polynomial degree in smooth regions and mesh size in non-smooth zones for efficiency.

## Design

The key concept used here is to compute the solution smoothness in elements. Here the finite solution is projected onto Legendre polynomial and then the Legendre coeffient decay rate is used to estimate the smoothness; faster the decay, smoother the solution be. 

**Detailed hp-Adaptivity algorithm : logic flow**
1. Projecting local solution in each element to Legendre expansion.
2. Added a function legepoly() to compute tensor-product Legendre basis functions evaluated at a point p.
3. Use the classical recurrence relation to evaluate 1D Legendre polynomials L_n(x), extended to multiple dimensions.
4. Construct the basis up to a specified total degree order of the element.
5. For each element, extract the local dof and compute the finite element solution u(h) at quadrature points.
6. Assemble the projection matrix K and right-hand side  F using   K_ij = integral (psi_i psi_i) dx and F_i = u_h psi_i dx
7. Solve the system K.c=F to obtain the Legendre coefficients c per element.
8. Then index-to-Degree Mapping: Here, a lookup vector mapping each index in the coefficient vector c to its corresponding total polynomial degree.
9. This mapping gives the basis ordering used during legepoly() construction
10. Coefficient Grouping and Filtering: Group coefficients c_i by their associated total degree.
11. From |k_i|, omitting all coefficients where ∣k∣=0 to exclude the constant mode. Then accumulate all coefficients (U_k) of the same degree into vectors keyed by degree.
12. Then for each degree (k >= 1), compute the L2 norm.
13. Finally, log-Log regression for smoothness esimation.
14. Construct vectors log(∣U_k∣) and log(k) using filtered norms.
15. Apply least-squares linear regression as log(∣U_k∣) = −s.log(k) + C
16. Compute the negtaive of slope s (call it regularity), and define the smoothness index as for the given element.
17. Finally calculates p corsen (`p_error`) and h corsen error (`h_error`) (as @roystgnr composed in hp_coarsentest.C) and then if the element's local solution is having highest regularity than mean and `p_error > h_error` then do p refine else h refine.

## Impact

1. Computational cost optimization: combining h- and p-adaptivity reduces degrees of freedom while preserving accuracy, especially in large-scale or real-time simulations.
2. Faster convergency with the optimal DoF in the system. 

PS: Need the updated libmesh


lindsayad:
I'm just going to convert this to draft until your libMesh PR is merged and the libmesh submodule hash is updated

lindsayad:
why are you closing this?

skvarjun:
I think I messed up rebasing

PR #31004, opened 2025-07-09
Mesh generators for 1D B Spline

lipohlmann:
closes #31003 

## Reason
Part of larger scope of work looking to create a pipe connector meshing utility to connect two meshes.

## Design
Uses two points with two directions to draw a B-spline. Additional sharpness parameter created to allow for additional flexibility in the curve's shape. Generator is written such that the user does not need to explicitly specify control points -- the creation of these is handled under the hood based on some user input. Spline is designed to honor the directions specified to create the smoothest curve possible.

## Impact
Creates a new 1D meshed curve. Will eventually be used for mesh extrusion.


GiudGiud (approved):
@roystgnr this is PR number 3.
It has fixes for the utils that were not in PR 1 and 2 but I moved those commits now.

I addressed my review, and fixed the commit history, separating the various changes hence some commits are attributed to me

framework/src/meshgenerators/BSplineCurveGenerator.C
```diff
@@ -0,0 +1,158 @@
...
+                                                    "Direction vector of curve at end point.");
+
+  // Spline shape parameters
+  params.addParam<unsigned int>("degree", 3, "Degree of interpolating polynomial.");
+  params.addRangeCheckedParam<libMesh::Real>(
+      "sharpness", 0.6, "sharpness>0 & sharpness<=1", "Sharpness of curve bend.");
+  params.addParam<unsigned int>(
+      "num_cps",
+      6,
+      "Number of control points used to draw the curve. Miniumum of degree+1 points are required.");
```
GiudGiud:
```suggestion
      "Number of control points used to draw the curve. Minimum of degree+1 points are required.");
```

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
```diff
@@ -0,0 +1,16 @@
+# BSplineCurveGenerator
+
+!syntax description /Mesh/BSplineCurveGenerator
+
+B-Splines are notably desirable because they respect the incoming and outgoing direction of the curve,
```
GiudGiud:
```suggestion
(Clamped) B-Splines are notably desirable because they can respect the incoming and outgoing direction of the curve,
```

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
```diff
@@ -0,0 +1,16 @@
...
+avoiding a discontinuity in the slope.
```
GiudGiud:
```suggestion
avoiding a discontinuity in the slope when connected to another line mesh at either end.
```

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
```diff
@@ -0,0 +1,16 @@
...
+
+The spline shape parameters should be adapted to obtain the desired curve. The sharpness is notably defined here
```
GiudGiud:
```suggestion
The spline shape parameters should be adapted to obtain the desired shape for the curve. The sharpness is notably defined here
```

GiudGiud:
Rebased on top of next.
When you have time for a final check on this one @roystgnr

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
```diff
@@ -0,0 +1,16 @@
...
+as a measure of the proximity of the spline to a curve consisting of three orthogonal segments, two continuing from
```
joshuahansel: They're orthogonal? For instance, couldn't this be used in the simplest case, where the start and end segments are colinear?

framework/src/meshgenerators/BSplineCurveGenerator.C
```diff
@@ -0,0 +1,158 @@
...
+  params.addRequiredParam<libMesh::Point>("end_point", "Ending (x,y,z) point for curve.");
+  params.addRequiredParam<libMesh::RealVectorValue>("start_direction",
+                                                    "Direction vector of curve at start point.");
+  params.addRequiredParam<libMesh::RealVectorValue>("end_direction",
+                                                    "Direction vector of curve at end point.");
+
+  // Spline shape parameters
+  params.addParam<unsigned int>("degree", 3, "Degree of interpolating polynomial.");
+  params.addRangeCheckedParam<libMesh::Real>(
+      "sharpness", 0.6, "sharpness>0 & sharpness<=1", "Sharpness of curve bend.");
```
joshuahansel, line 51: Is "sharpness" some recognized quantity? I guess the bounds show up in documentation in MooseDocs, but maybe it'd be good to say something like "sharpness=0 corresponds to (blah), sharpess=1 corresponds to (blah)"

joshuahansel (changes requested):

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
GiudGiud: when the segments are colinear, we support one case by reverting to a circle iirc, and error on the others

framework/src/meshgenerators/BSplineCurveGenerator.C
GiudGiud: > Is "sharpness" some recognized quantity?

It's not in the mainstream wikipedia page for B spline. I think it might be introduced in this work

> maybe it'd be good to say something like "sharpness=0 corresponds to (blah), sharpess=1 corresponds to (blah)"

I ll run some tests

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
GiudGiud: I verified the behavior

if both direction vectors are equal, we can only do a circle, and we need to bump the control points to get a nice one too

Screenshot 2026-03-31 at 1 42 09 PM

if both direction vectors are opposite, we give the same circle..
I ll add an error for the equal case, we don't support it

joshuahansel: Thanks, I'm just confused by the statement that the line segments are always orthogonal, so I was wondering if that needed to be changed.

framework/src/meshgenerators/BSplineCurveGenerator.C
GiudGiud: sharpness = 1e-8 is a straight line
sharpness = 1 is super smooth but not sure what it really is

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
GiudGiud: yeah that's fair. The circle is kind of an edge case. The aligned is another edge case. if the points are aligned too, we'd just use a line instead of this.

GiudGiud: The case where directions are aligned but points are not aligned (with the directions), and are not along a circle, is not supported either right now but I think we could do it fairly easily by specifying the "closest point" between the two parallel lines to be right in between the two points. 

I'm not necessarily trying to add more to this PR though

GiudGiud: > Thanks, I'm just confused by the statement that the line segments are always orthogonal, 

in that construction, the three lines are orthogonal 2 by 2 in a way. The shortest path from one line to another is to build a 3rd line, orthogonal to both the other lines

joshuahansel: Ah ok, the language made it sound as if they were all orthogonal to one another, but I see it's just that the shortest-path line is orthogonal to both the others.

framework/src/meshgenerators/BSplineCurveGenerator.C
joshuahansel: The doc should show how this sharpness value is used (probably should show up in an equation).

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
roystgnr: I'd change "three orthogonal segments" to "three line segments, each orthogonal to its neighbor(s)".

roystgnr: I'd like to know, though: what happens if we have high sharpness, our end directions (considered as rays starting from their endpoints a distance D apart) are coplanar, and they're, say, 1 degree different in angle.  Those rays intersect at a distance of roughly D/tan(1°), so I'd think that sharpness=1 means we'd get two super-long line segments out to that intersection point.  (plus one length-0 degenerate middle segment, whatever).

But then if we decrease that angle from 1° to 0.0000001° or so, eventually getting below `TOLERANCE` in our comparisons, we'll suddenly snap from a curve that goes D/ε away to a semicircle that goes D/2 away?

I don't actually see anything better we could do there, but that's somewhat unexpected and the "just make a semicircle" fallback should be documented.

roystgnr (changes requested):
We ought to be documenting the semicircle fallback case, and maybe changing "three orthogonal segments" if Josh likes my wording or some other wording better, but I'm not seeing anything else I don't like here.

joshuahansel (changes requested):
Please document how sharpness value is used.

GiudGiud:
documentation should be updated now.

framework/src/meshgenerators/BSplineCurveGenerator.C
GiudGiud: added in last but one commit

framework/doc/content/source/meshgenerators/BSplineCurveGenerator.md
GiudGiud: yeah these end cases are hard. I think users will be thankful for that circle though, better than just erroring out and letting them figure out an alternative generator

joshuahansel (approved):

roystgnr (approved):

GiudGiud:
Thanks for the reviews!

Issue #8770, opened 2017-03-21
Add make target to generate compile_commands.json


dschwen:
### Description of the enhancement or error report
`compile_commands.json` is a JSON database with the compile commands (and options) to build the current project. We can easily generate this using our makefile system.

### Rationale for the enhancement or information for reproducing the error
`compile_commands.json` is useful for clang tooling and tools like rtags. Building the JSON file is a much cleaner way of getting the necessary build info to those tools (compared to screen scraping `make -n`output). (tag @jasondhales)

### Identified impact
Easier to use tooling

PR #31015, opened 2025-07-10
compile_commands.json defaults MOOSE_UNITY and MOOSE_HEADER_SYMLINKS to false

hugary1995:
ref #8770

The compile_commands.json target by default uses MOOSE_UNITY=false and MOOSE_HEADER_SYMLINKS=false.


GiudGiud:
why false on MOOSE_UNITY? just curious

hugary1995:
Because clangd is good at catching missing includes in non-unity builds. I found that to be very useful.

Anyways, @maxnezdyur just pointed out that this should probably be a vscode task, which I agree. I'll think a little more about whether this script has use cases outside vscode, and if not I'll just close this PR.

lindsayad:
Why not just make this the default behavior of `make compile_commands.json`?

hugary1995:
> Why not just make this the default behavior of `make compile_commands.json`?

Uhh I'm not sure how. Can a make rule locally modify other options that might have taken effect earlier? or do you mean making the make compile_commands.json rule spawn a separate shell process? I suppose that could work...

lindsayad:
I was just thinking this kind of thing in `moose.mk`
```
MOOSE_UNITY ?= true
MOOSE_HEADER_SYMLINKS ?= true

ifeq ($(filter compile_commands.json,$(MAKECMDGOALS)),compile_commands.json)
  MOOSE_UNITY := false
  MOOSE_HEADER_SYMLINKS := false
endif
```

loganharbour:
Can we not do

```
ifeq ($(filter compile_commands.json,$(MAKECMDGOALS)),compile_commands.json)
  MOOSE_UNITY ?= false
  MOOSE_HEADER_SYMLINKS ?= false
else
  MOOSE_UNITY ?= true
  MOOSE_HEADER_SYMLINKS ?= true
endif
```

That way it can still be overridden (if one stupidly wants to)

hugary1995:
okay let me do that then

hugary1995:
Should be good to go now.

loganharbour (approved):

loganharbour:
I also build with METHOD=dbg - is it worth putting that in here too?

lindsayad (commented):
I still think it makes sense to auto-remove `compile_commands.json` if it exists when a user requests `make compile_commands.json`. As it is now, I always end up typing `rm compile_commands.json && make compile_commands.json`

hugary1995:
> I still think it makes sense to auto-remove `compile_commands.json` if it exists when a user requests `make compile_commands.json`. As it is now, I always end up typing `rm compile_commands.json && make compile_commands.json`

I marked the target as .PHONY so it should always override existing compile_commands.json now.

hugary1995:
> I also build with METHOD=dbg - is it worth putting that in here too?

I thought a bit about that -- see if you like what I have now. I am not forcefully setting METHOD=dbg because there could be cases when the required dependencies for dbg isn't available, for example a dbg libmesh, and the user might want to control METHOD.

hugary1995:
Default METHOD=dbg if generating compile commands.

loganharbour:
I like this. @zachmprince can you comment on if we're missing anything here?

loganharbour:
> I still think it makes sense to auto-remove `compile_commands.json` if it exists when a user requests `make compile_commands.json`. As it is now, I always end up typing `rm compile_commands.json && make compile_commands.json`

@hugary1995 what do you think about the removal? I think this is a good idea too

hugary1995:
I marked the target as .PHONY so it should always effectively override compile_commands.json now, right? I hope my understanding of .PHONY is correct...

Previously we would just get `target compile_commands.json is up to date.` or something like that.

zachmprince:
I definitely don't understand make as much as you all, but I think all the defaults are perfect and are exactly what I do in my compile_commands.json build.

loganharbour (approved):

lindsayad:
LGTM

PR #31028, opened 2025-07-14
Enable ElementGroupCentroidPositions with distributed mesh

### LINKED ISSUE #31026: Make ElementGroupCentroidPositions support distributed mesh
### Motivation

It is desired to make `ElementGroupCentroidPositions` work with distributed mesh especially in case of pre-split mesh.

### Design

Just add some communications. There is no reason to restrict it with replicated mesh.

### Impact

Minor improvement on an existing position.

YaqiWang:
Closes #31026.

### DIFF
--- a/framework/src/positions/ElementGroupCentroidPositions.C
+++ b/framework/src/positions/ElementGroupCentroidPositions.C
@@ -9,6 +9,8 @@
 
 #include "ElementGroupCentroidPositions.h"
 
+#include "libmesh/parallel_algebra.h"
+
 registerMooseObject("MooseApp", ElementGroupCentroidPositions);
 
 InputParameters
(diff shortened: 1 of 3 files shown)


roystgnr (approved):
Let's wait to merge until the distributed sweeps are done, but this looks like the correct fix to me.

framework/src/positions/ElementGroupCentroidPositions.C
```diff
@@ -9,6 +9,8 @@
 
 #include "ElementGroupCentroidPositions.h"
 
+#include "timpi/communicator.h"
```
YaqiWang: @roystgnr this does not seem fixing the unity build error. Maybe we need a different header like "libmesh/libmesh_call_mpi.h" here?

roystgnr: The non-unity build error, you mean?

Ah, I think I see.  `communicator.h` is sufficient for the basic `Communicator::sum` declaration, but depending on what you're summing you might also need `timpi/parallel_implementation.h` (for all the specializations there) and a header defining a `StandardType` and/or `Packing` specialization for the particular type of data you're summing, in this case probably `libmesh/parallel_algebra.h` for `Point`.

YaqiWang: Trying `libmesh/parallel_algebra.h`. Will see whether it is enough.

roystgnr: You'll probably require both that (for `Point`) and `timpi/parallel_implementation.h`.

GiudGiud (approved):

PR #31029, opened 2025-07-14
Add MooseServer plotting as NEAMS Workbench protocol extension

brandonlangley:
refs #22766

https://github.com/user-attachments/assets/8ec3bf49-dc88-4835-b1ad-2db61b6809b7

### DIFF
--- a/framework/contrib/hit/Makefile
+++ b/framework/contrib/hit/Makefile
@@ -21,8 +21,10 @@ endif
 
 ifeq ($(UNAME), Darwin)
 	DYNAMIC_LOOKUP := -undefined dynamic_lookup
+	hit_pad_LDFLAGS := -Wl,-headerpad_max_install_names
 else
 	DYNAMIC_LOOKUP :=
+	hit_pad_LDFLAGS :=
 endif
 
 $(info Building hit for python with $(pyconfig))
@@ -61,7 +63,7 @@ hit: $(hit_srcdir)/main.cc $(hit_srcfiles) include/hit/braceexpr.h include/hit/l
 bindings: hit.so
 
 hit.so: $(hit_srcfiles)
-	$(hit_CXX) -std=c++17 $(wasp_CXXFLAGS) $(hit_includeflags) -w -fPIC -lstdc++ -shared -L$(PYTHONPREFIX)/lib $(PYTHONCFLAGS) $(DYNAMIC_LOOKUP) $^ $(HITCPP) -o $@ $(wasp_LDFLAGS)
+	$(hit_CXX) -std=c++17 $(wasp_CXXFLAGS) $(hit_includeflags) -w -fPIC -lstdc++ -shared -L$(PYTHONPREFIX)/lib $(PYTHONCFLAGS) $(DYNAMIC_LOOKUP) $^ $(HITCPP) -o $@ $(wasp_LDFLAGS) $(hit_pad_LDFLAGS)
 
 $(HITCPP): hit.pyx chit.pxd
 	$(cython) -o $@ --cplus $<
--- a/conda/moose-dev/conda_build_config.yaml
+++ b/conda/moose-dev/conda_build_config.yaml
@@ -7,7 +7,7 @@ moose_libmesh:
   - moose-libmesh 2025.08.06 openmpi_0
 
 moose_wasp:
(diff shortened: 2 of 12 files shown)


brandonlangley:
@loganharbour @dschwen -

This adds plotting capabilities to the MooseServer as an LSP extension for the NEAMS Workbench.

The MooseServer has been updated to include this plotting LSP extension in its server capabilities.

For language servers that support this plotting extension, Workbench allows plots to be requested.

If the query is from a `Piecewise` function, the MooseServer builds a `wasp::CustomPlot` response.

The `wasp::CustomPlot` objects could be bar charts, color maps, or line graphs allowing error bars.

The `Piecewise` functions create line graph objects now with data defined in the input or CSV files.

This has no effect on the MooseServer in VS Code since it does not request this custom extension.

It is all being unit tested, could one of you review this if you have time and merge it if you approve?

An example of this plotting extension being used by the NEAMS Workbench is shown in this video:

https://github.com/user-attachments/assets/d608ba83-ba02-46cb-8a8c-993292fdd550

brandonlangley:
@loganharbour -

It looks like some conflicts crept in with these files since my conda `wasp` update also updated the `moose-dev` version.

 - `conda/moose-dev/meta.yaml`
 - `conda/moose/conda_build_config.yaml`
 - `scripts/tests/versioner_hashes.yaml`
 - `scripts/versioner.yaml`

Should I continually regenerate my conda config files and bump my versioner hashes after other updates get merged?

Or would it be okay to just make those changes once more after you review and approve this right before it is merged?

loganharbour:
WASP update was added in https://github.com/idaholab/moose/pull/31141

brandonlangley:
@loganharbour -

I [managed](https://github.com/idaholab/moose/pull/31029/commits/43230645daa9f9637524dc632536307a6ea6db7f) to fix the [Conda moose-dev intel mac](https://civet.inl.gov/job/3167258/) job using `-headerpad_max_install_names` for the HIT shared library.

If that update looks okay, this should be good to merge and plots all functions derived from the `PiecewiseBase` class.

The `PiecewiseBase` class contains the plottable 2d data, so `PiecewiseLinear` and `PiecewiseConstant` are included.

I feel like this could be merged and easily updated in the future if an interface is added to handle more function types.

framework/contrib/hit/Makefile
```diff
@@ -21,8 +21,10 @@ endif
 
 ifeq ($(UNAME), Darwin)
 	DYNAMIC_LOOKUP := -undefined dynamic_lookup
+	hit_pad_LDFLAGS := -Wl,-headerpad_max_install_names
```
loganharbour, line 24: Isn't this only an Intel Mac issue? Arm mac is okay? If so, we should only apply it there. Darwin is both arches here

loganharbour (changes requested):

framework/contrib/hit/Makefile
brandonlangley: @loganharbour -

No this is not an Intel Mac only issue. It is Mac specific but may happen with any Mac (Intel or Apple Silicon).

It happens if the linker leaves too little room and a later tool (e.g., install_name_tool) has to grow the RPATH.

We happen to see it on x86_64 now, but this may appear on arm64 later if paths grow or conditions change.

My theory is it finally crossed the limit on x86_64 from the new -lwaspplot library, and it will on arm64 soon.

So even though it hit Intel Mac first, it should remain for all Darwin to avoid it with Apple Silicon in the future.

brandonlangley: @loganharbour -

What do you think? I could make it so it only adds the `-headerpad_max_install_names` flag in this condition:

```
ifeq ($(shell uname -s) $(shell uname -m),Darwin x86_64)
```

But it seems like making this Intel only is setting the `arm64` Mac up for future failure since it applies to both.

loganharbour (changes requested):
Bump the date for moose-dev to today (2025.09.04) and redo the versioner hashes file. Otherwise, good to go.

brandonlangley:
@loganharbour -

I bumped the wasp and moose-dev conda dates to 2025.09.04 and redid the versioner hashes.

This should be good to merge now.

loganharbour (approved):

Issue #30655, opened 2025-06-03
GPU Acceleration of MOOSE with Kokkos


NamjaeChoi:
### Motivation

GPU computing is a leading technology in the high-performance computing field and is becoming essential for scientific computing. The improvements in the GPU computing technology is keep accelerating, which is driven by the accelerated development of AI. The DOE's leadership-class exascale supercomputers all employ GPUs as the primary processors: Doudna (2026, LNBL, NVIDIA), El Capitan (2024, LLNL, AMD), Aurora (2023, ANL, Intel), Frontier (2022, ORNL, AMD). MOOSE must incorporate these new technologies or else risk alienating customers interested in the benefits these technologies can offer.

### Design

Considering the diversity of the GPU architectures, the GPU acceleration of MOOSE will leverage Kokkos, a performance portability library developed as part of the Exascale Computing Project. Kokkos provides a unified programming model that allows developers to write code that can efficiently run on diverse hardware architectures. The user interface of MOOSE for GPUs will remain as close as possible to the current interface, which will minimize the users' burden of porting from CPUs to GPUs.

### Impact

A new technology that will bring a substantial performance improvement and allow MOOSE to run on the state-of-the-art HPC systems.

lindsayad:
@NamjaeChoi at what point do you want to close this issue? After your UO PR is in?

NamjaeChoi:
When Kokkos-MOOSE becomes a masterpiece

lindsayad:
I'm updating the title to be focused on the `framework` as I want to create separate issue(s) regarding plans for the modules

lindsayad:
> When Kokkos-MOOSE becomes a masterpiece

Hahaha. You've already done masterful work. Perhaps some more tangible criteria could be good 😄

NamjaeChoi:
Okay let's close the first phase when we have
- [ ] Kokkos-MOOSE UO system
- [ ] MetaPhysicL 2.0 integration
- [ ] RDC enabled
- [ ] HIP working

NamjaeChoi:
Second phase will have to involve libMesh.

lindsayad:
Do you want programming help with any items in your list like MetaPhysicL 2 integration?

NamjaeChoi:
Sure, I might bother you a lot once I start working on it.

lindsayad:
[...]

PR #31047, opened 2025-07-15
Add common device command and option in test harness

NamjaeChoi:
Refs #30655. Refs #30915.

### DIFF
--- a/python/TestHarness/testers/RunApp.py
+++ b/python/TestHarness/testers/RunApp.py
@@ -120,15 +120,14 @@ class RunApp(Tester):
                 self.setStatus(self.skip)
                 return False
 
-        if self.specs['libtorch'] or self.specs['mfem']:
-            if self.specs.isValid('libtorch_devices'):
-                devices_lower = [x.lower() for x in self.specs['libtorch_devices']]
-            else:
-                devices_lower = [x.lower() for x in self.specs['devices']]
-            if options.device not in devices_lower:
-                self.addCaveats(f'{options.device} not in devices')
-                self.setStatus(self.skip)
-                return False
+        if self.specs.isValid('libtorch_devices'):
+            devices_lower = [x.lower() for x in self.specs['libtorch_devices']]
+        else:
+            devices_lower = [x.lower() for x in self.specs['devices']]
+        if options.device not in devices_lower:
+            self.addCaveats(f'{options.device} not in devices')
+            self.setStatus(self.skip)
+            return False
 
         if options.hpc and self.specs.isValid('command_proxy') and os.environ.get('APPTAINER_CONTAINER') is not None:
             self.addCaveats('hpc unsupported')
--- a/python/TestHarness/testers/RunApp.py
+++ b/python/TestHarness/testers/RunApp.py
@@ -53,7 +53,8 @@ class RunApp(Tester):
         # Valgrind
         params.addParam('valgrind', 'NORMAL', "Set to (NONE, NORMAL, HEAVY) to determine which configurations where valgrind will run.")
 
-        params.addParam('libtorch_devices', ['CPU'], "The devices to use for this libtorch test ('CPU', 'CUDA', 'MPS'); default ('CPU')")
+        params.addParam('libtorch_devices', "The devices to use for this libtorch test ('CPU', 'CUDA', 'MPS')")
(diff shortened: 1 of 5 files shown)


python/TestHarness/testers/RunApp.py
```diff
@@ -114,10 +120,13 @@ def checkRunnable(self, options):
                 self.setStatus(self.skip)
                 return False
 
-        if self.specs['libtorch']:
-            devices_lower = [x.lower() for x in self.specs['libtorch_devices']]
-            if options.libtorch_device not in devices_lower:
-                self.addCaveats(f'{options.libtorch_device} not in libtorch_devices')
+        if self.specs['libtorch'] or self.specs['mfem']:
```
NamjaeChoi: I don't see the point of these. They always seem to return `ALL`

loganharbour (commented):
I don't think that our previous design was the best, we just needed to get something going. And with that I'm not sure extending it is a good idea either

NamjaeChoi:
The other option I can think of is to use capabilities to determine the device to run, but that will be complicated.

python/TestHarness/testers/Tester.py
```diff
@@ -83,7 +83,7 @@ def validParams():
         params.addParam('libpng',        ['ALL'], "A test that runs only if libpng is available ('ALL', 'TRUE', 'FALSE')")
         params.addParam('libtorch',      ['ALL'], "A test that runs only if libtorch is available ('ALL', 'TRUE', 'FALSE')")
         params.addParam('libtorch_version', ['ALL'], "A list of libtorch versions for which this test will run on, supports normal comparison operators ('<', '>', etc...)")
-        params.addParam('mfem', ['ALL'], "A test that runs only if mfem is available ('ALL', 'TRUE', 'FALSE')")
+        params.addParam('mfem',          ['ALL'], "A test that runs only if MFEM is available ('ALL', 'TRUE', 'FALSE')")
```
lindsayad: we should remove this parameter as we can just use `capabilities`

lindsayad:
I definitely think we want a `--device` CLI arg for both MOOSE applications and `run_tests`. As to the test specs themselves, I think an opt-in approach is currently the right way to go, so having a default of `CPU` is right. However, once we have https://gitlab.com/petsc/petsc/-/merge_requests/8505 merged into MOOSE, then I think it makes more sense to do opt-out instead of opt-in, so the default `devices` would be `ALL`

NamjaeChoi:
I want to eventually remove `--libtorch-device` tester CLI argument and `libtorch_devices` test spec parameter. The former would be able to be removed after merging this and updating the recipe. The latter will require going through all apps that can potentially be still using it. How do you think

lindsayad:
I agree that once we have `--device` we definitely don't want `--libtorch-device`

lindsayad (approved):

NamjaeChoi:
Why does this affect coverage?

lindsayad:
you're talking about that single `break` line? I think that's something spurious. Don't worry about it

NamjaeChoi:
> spurious

Exactly

lindsayad:
@loganharbour you good with this?

loganharbour (approved):
It's not ideal, but it's good for now

NamjaeChoi:
Thank you folks

python/TestHarness/testers/RunApp.py
```diff
@@ -274,8 +282,11 @@ def getCommand(self, options):
         if options.scaling and specs['scale_refine'] > 0:
             cli_args.insert(0, ' -r ' + str(specs['scale_refine']))
 
-        if specs['libtorch']:
-            cli_args.append(f'--libtorch-device {options.libtorch_device}')
+        if specs.isValid('capabilities'):
+            if 'libtorch' in specs['capabilities']:
+                cli_args.append(f'--libtorch-device {options.device}')
+            if 'mfem' in specs['capabilities']:
+                cli_args.append(f'Executioner/device={options.device}')
```
dschwen, line 289: This is not ok. Just because the string `libtorch` or `mfem` is in the capabilities expression does not mean that the tester is using either of those. The tester could for example contain `!mfem` (_not_ mfem). The right thing to do is to make the `Executioner/device` parameter in `MFEMExecutioner` optional and use the command line parameter currently named `--libtorch-device` (better name would be `--compute-device`)!

PR #31050, opened 2025-07-16
Add get functions to FunctionElementAverage postprocesssor

### LINKED ISSUE #31049: Volume and integral getter functions for FunctionElementAverage postprocessor
### Motivation

I would like to calculate the sum and mean of a function in an app, and think the best way to do so is via the FunctionElementAverage postprocesser. FunctionElementAverage does not have methods to return the integral value or volume, despite calculating the two independently. So as to not need both FunctionElementIntegral and FunctionElementAverage to calculate the sum and mean (doing the integral two times), the getter functions could be used to manually calculate the mean.

### Design

This would require two getter functions to be added to FunctionElementAverage, one to return the volume and one to return the integral value.

### Impact

This will remove the need to use both FunctionElementIntegral and FunctionElementAverage to calculate the sum and mean.

nglaser3:
Added constant get functions to FunctionElementAverage to return the volume and the integral value.

Closes #31049.

### DIFF
--- a/framework/include/postprocessors/FunctionElementAverage.h
+++ b/framework/include/postprocessors/FunctionElementAverage.h
@@ -27,6 +27,9 @@ public:
   virtual void finalize() override;
   virtual void threadJoin(const UserObject & y) override;
 
+  const Real & getVolume() const { return _volume; }
+  const Real & getIntegralValue() const { return _integral_value; }
+
 protected:
   Real _volume;
 };


pbehne (approved):
Looks good to me. Thanks for adding this flexibility!

pbehne (changes requested):
Changing approval, will update review with suggestions.

pbehne:
The purpose of a postprocessor is to compute a single scalar value. This value is provided by the `getValue` method. Aside from this API, we shouldn't be adding other getters to access the internals of the postprocessor. We want all postprocessors to have/use the same APIs as much as possible, so we definitely don't want to add getters for some postprocessors and not others.

For the reasons above, I will approve this PR.

However, I do recognize that it is undesirable from a code optimization standpoint to use two separate postprocessors to compute two closely related values (total and mean). I would use the `FunctionElementIntegral` postprocessors to compute the total, and then use the following code to compute the mesh volume:

```
Real volume = 0.;
for (const auto * elem : mesh.active_local_element_ptr_range())
  volume += elem->volume();
mesh.comm().sum(volume);
```

Once you have the volume, use to get the average from the total.

lindsayad:
> then use the following code to compute the mesh volume:

we also already have a `VolumePostprocessor`. To save computations as @pbehne you could use a `ParsedPostprocessor` to compute the average from the result of `FunctionElementIntegral` and `VolumePostprocessor` if you want access to all 3 values: integral, volume, and average

PR #31058, opened 2025-07-16
Using a functor in HeatSourceVolumetric1Phase

### LINKED ISSUE #31057: Allow HeatSourceVolumetric1Phase to take a functor to define the heat source
### Motivation

Using a functor instead of a function for the input parameter q would give more flexibilty in using this component

### Design

Define q as a functor instead of a function

### Impact

More flexibility

licharlot:
Closes #31057


modules/thermal_hydraulics/src/components/HeatSourceVolumetric1Phase.C
```diff
@@ -43,11 +43,11 @@ HeatSourceVolumetric1Phase::addMooseObjects()
   const FlowChannelBase & fch = getComponent<FlowChannel1Phase>("flow_channel");
 
   {
-    std::string class_name = "OneD3EqnEnergyHeatSource";
```
joshuahansel, line 46: Any reason to keep the old one?

joshuahansel (changes requested):
Just the one comment.

modules/thermal_hydraulics/src/components/HeatSourceVolumetric1Phase.C
licharlot: No. We kept a bunch of other non-ADKernels, so I didn't remove that one. My guess is that at the time, they may have been used by other applications. Maybe it is time to clean that up?

joshuahansel: I think we might have kept the non-AD ones for before RELAP-7 was made AD. We could probably start removing those. I won't ask you to do all of the cleanup here, just try this kernel maybe. And I think it's pretty unlikely that the kernel is used directly in inputs, so I'd just try actually deleting it directly instead of deprecating. If that doesn't work, we can deprecate -> patch -> delete instead.

joshuahansel:
I guess you see that the class deletion broke RELAP-7 docs. Looks like some 7-equation class are calling 3-equation syntax commands 🤦 So you should be able to do a RELAP-7 fix beforehand to prevent this breakage.

licharlot:
> I guess you see that the class deletion broke RELAP-7 docs. Looks like some 7-equation class are calling 3-equation syntax commands 🤦 So you should be able to do a RELAP-7 fix beforehand to prevent this breakage.

Yep, I saw that. Working on fixing RELAP-7

licharlot:
Turns out Casey fixed it in RELAP-7. 
HPC failure is unrelated.

joshuahansel (approved):

PR #31064, opened 2025-07-17
Try turning on MFEM CUDA tests

NamjaeChoi:
Refs #30915.

### DIFF
--- a/framework/include/mfem/outputs/MFEMDataCollection.h
+++ b/framework/include/mfem/outputs/MFEMDataCollection.h
@@ -26,6 +26,11 @@ public:
 protected:
   void registerFields();
   void output() override;
+  /**
+   * Whether the grid functions need to be explicitly read on the host before output. This seems to
+   * be required for Conduit data output whereas other formats handle the host read themselves
+   */
+  virtual bool gridFunctionsNeedHostRead() const { return false; }
 
   /// Reference to the MFEMProblemData struct storing the output variables.
   MFEMProblemData & _problem_data;
--- a/framework/src/mfem/outputs/MFEMDataCollection.C
+++ b/framework/src/mfem/outputs/MFEMDataCollection.C
@@ -53,11 +53,8 @@ MFEMDataCollection::output()
   dc.SetCycle(getFileNumber());
   dc.SetTime(time());
   if (gridFunctionsNeedHostRead())
-    for (auto const & [gf_name, gf_ptr] : _problem_data.gridfunctions)
-    {
-      libmesh_ignore(gf_name);
+    for (auto & [_, gf_ptr] : _problem_data.gridfunctions)
       gf_ptr->HostRead();
-    }
   dc.Save();
   _file_num++;
 }
--- a/framework/src/mfem/outputs/MFEMDataCollection.C
+++ b/framework/src/mfem/outputs/MFEMDataCollection.C
@@ -52,6 +52,12 @@ MFEMDataCollection::output()
   // Write fields to disk
   dc.SetCycle(getFileNumber());
   dc.SetTime(time());
+  if (gridFunctionsNeedHostRead())
+    for (auto const & [gf_name, gf_ptr] : _problem_data.gridfunctions)
+    {
+      libmesh_ignore(gf_name);
+      gf_ptr->HostRead();
+    }
   dc.Save();
   _file_num++;
 }
--- a/framework/src/mfem/problem_operators/ProblemOperatorInterface.C
+++ b/framework/src/mfem/problem_operators/ProblemOperatorInterface.C
@@ -39,7 +39,7 @@ ProblemOperatorInterface::Init(mfem::BlockVector & X)
   for (size_t i = 0; i < _test_variables.size(); ++i)
   {
     auto * const test_var = _test_variables.at(i);
-    X.GetBlock(i) = (*test_var);
+    X.GetBlock(i) = test_var->GetTrueVector();
     // Sync the flags from sub-block vectors to global vector
     X.SyncFromBlocks();
 
--- a/framework/src/mfem/problem_operators/ProblemOperatorInterface.C
+++ b/framework/src/mfem/problem_operators/ProblemOperatorInterface.C
@@ -36,23 +36,21 @@ void
 ProblemOperatorInterface::Init(mfem::BlockVector & X)
 {
   X.Update(_block_true_offsets);
-  for (size_t i = 0; i < _test_variables.size(); ++i)
-  {
-    auto * const test_var = _test_variables.at(i);
-    X.GetBlock(i) = test_var->GetTrueVector();
-    // Sync the flags from sub-block vectors to global vector
-    X.SyncFromBlocks();
+  for (const auto i : index_range(_test_variables))
+    X.GetBlock(i) = _test_variables[i]->GetTrueVector();
+  // Sync the flags from sub-block vectors to global vector
+  X.SyncFromBlocks();
 
-    // After initial assignment of X from the grid function, which may contain initial conditions,
-    // we alias the grid function to X
-    test_var->MakeTRef(_test_variables.at(i)->ParFESpace(), X, _block_true_offsets[i]);
-    _test_true_vector = &X;
-    // This might seem silly but after making the tref the memory flags of the grid function and its
-    // true vector are in an empty state other than the aliasing. This operation syncs the flags and
-    // should be a no-op in terms of actual data transfer
-    test_var->GetTrueVector().SyncMemory(*_test_true_vector);
-    test_var->SetFromTrueVector();
-  }
+  // After initial assignment of X from the grid function, which may contain initial conditions,
+  // we alias the grid function to X
+  for (const auto i : index_range(_test_variables))
+    _test_variables[i]->MakeTRef(_test_variables[i]->ParFESpace(), X, _block_true_offsets[i]);
+  _test_true_vector = &X;
+
+  // This might seem silly but after making the tref the memory flags of the grid function and its
+  // true vector are in an empty state other than the aliasing. This operation syncs the flags and
+  // should be a no-op in terms of actual data transfer
+  SetTestVariablesFromTrueVectors();
 }
 
 void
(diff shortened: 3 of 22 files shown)


framework/src/mfem/problem_operators/ProblemOperatorInterface.C
```diff
@@ -48,7 +51,13 @@ ProblemOperatorInterface::SetTestVariablesFromTrueVectors()
 {
   for (unsigned int ind = 0; ind < _test_variables.size(); ++ind)
   {
-    _test_variables.at(ind)->SetFromTrueVector();
+    auto * const test_var = _test_variables.at(ind);
+
+    // We must sync the memory flags from the true true vector to the grid function copy of the true
+    // vector
+    mooseAssert(_test_true_vector, "The true vector should already have been set");
+    test_var->GetTrueVector().SyncMemory(*_test_true_vector);
```
alexanderianblair, line 66: I'm not sure I understand this - `X` is the vector storing all (true) grid function DoFs. If `_test_variables.size()` is greater than 1, `_test_true_vector` will be longer than the last `test_var` true DoF vector

lindsayad: It will be longer for sure, but this call is not syncing the data arrays. It's just syncing the memory flags, such as validity of data on host/device

lindsayad: It is a little silly/rendudant to do this for all `_test_variables`. We should just do this for the zeroth test variable

alexanderianblair: Ah, I see - yes, doing it for just the first variable in `_test_variables` makes sense, and should avoid triggering the assert in `mfem::Memory<T>::Sync` that checks the two pointers are equal (which I don't think will be true for all `test_var`, but should be for the first).

lindsayad: I take back what I said. Since each grid function has its own copy of a "true vector" we must make sure to sync the flags for all the grid functions

lindsayad: I'm not sure what the `Registered` flag is for in MFEM, but I do know that `flags & Registered` is `true` so we never do the check in the `MFEM_ASSERT`

lindsayad: Sorry, rather it's that the block vector `_problem_data.f` is not registered. That's the reason we don't execute the assertion

lindsayad: Sorry for the sorry ... the former post is accurate when `Executioner/device=cuda`. The latter post is accurate when `Executioner/device=cpu`. So `Registered` is used to keep track of whether device memory has been registered with the global MFEM memory manager

lindsayad:
The `time_steppers/time_adaptive.test` failure on previous clang testing is documented as an intermittent failure in https://github.com/idaholab/moose/issues/27555

NamjaeChoi:
Looks like our CUDA machine is too busy.

NamjaeChoi:
Tests passed!

lindsayad:
Since I made important changes to this PR I'd like another set of eyes to review

test/tests/mfem/kernels/tests
```diff
@@ -199,6 +210,8 @@
     requirement = 'The system shall have the ability to solve for an irrotational vortex in two dimensions using MFEM.'
     abs_zero = 1e-6
     capabilities = 'mfem'
+    devices = 'cpu cuda'
     recover = false
+    max_parallel = 1 # mfem-hypre doesn't appear to work with multiple ranks with cuda
```
nmnobre: Let's rebase so we don't miss any tests. Given pretty much all tests (including beyond this file) require `max_parallel = 1`, I wonder if we should move this comment up to the beginning of the file or somewhere else.

nmnobre (changes requested):
Haven't reviewed/tested the new syncing changes/logic yet.

test/tests/mfem/kernels/tests
lindsayad: We have a bit of interweaving issues. The `XMLDiffs` innately can't be used with more than one rank based on conversations with @alexanderianblair. And then we have this MOOSE-MFEM-HypreGMRES issue

lindsayad: It would be really nice to convert those XMLDiff tests to something we can run in parallel. Because it prevents from identifying other parallel issues. For instance I'm seeing now that the transient scalar IC test just crashes in parallel

alexanderianblair: > We have a bit of interweaving issues. The `XMLDiffs` innately can't be used with more than one rank based on conversations with @alexanderianblair. And then we have this MOOSE-MFEM-HypreGMRES issue

This is true, since the partitioning may differ between runs, and the number of output files in a `ParaViewDataCollection` will depend on the number of processors used.

We have a couple of `VectorPostProcessors` inbound for sampling MFEM FE variables at arbitrary points using  GSLib from @k-collie - which should enable us to convert many of the existing XMLDiff tests over to CSVDiff tests (also significantly reducing output sizes...) - WIP branch is [here](https://github.com/aurora-multiphysics/platypus-moose/tree/k-collie/mfem-point-and-line-sampler)

framework/src/mfem/outputs/MFEMDataCollection.C
```diff
@@ -52,6 +52,12 @@ MFEMDataCollection::output()
   // Write fields to disk
   dc.SetCycle(getFileNumber());
   dc.SetTime(time());
+  if (gridFunctionsNeedHostRead())
+    for (auto const & [gf_name, gf_ptr] : _problem_data.gridfunctions)
+    {
+      libmesh_ignore(gf_name);
```
alexanderianblair: `libmesh_ignore` usage here seems clunky in an MFEM object -  preference for use of `[[maybe_unused]]` (or the upcoming `_` dummy variable in the structured binding? We can confirm the latter works to suppress warnings due to its usage in `CoefficientMap.h`.

framework/src/mfem/problem_operators/ProblemOperatorInterface.C
```diff
@@ -38,8 +38,20 @@ ProblemOperatorInterface::Init(mfem::BlockVector & X)
   X.Update(_block_true_offsets);
   for (size_t i = 0; i < _test_variables.size(); ++i)
   {
-    X.GetBlock(i) = _test_variables.at(i)->GetTrueVector();
-    _test_variables.at(i)->MakeTRef(_test_variables.at(i)->ParFESpace(), X, _block_true_offsets[i]);
+    auto * const test_var = _test_variables.at(i);
+    X.GetBlock(i) = test_var->GetTrueVector();
+    // Sync the flags from sub-block vectors to global vector
+    X.SyncFromBlocks();
```
alexanderianblair: This is iterating over all blocks at every iteration of the loop (even when some blocks have been yet to be assigned to their respective variable true dofs).

Might be more readable to split this loop into two, where the data in `X` is updated in the first, then `X.SyncFromBlocks()` is called, before a second loop to update the test variables to reference the true DOF data stored in `X`.

framework/src/mfem/problem_operators/ProblemOperatorInterface.C
```diff
@@ -38,8 +38,20 @@ ProblemOperatorInterface::Init(mfem::BlockVector & X)
...
+
+    // After initial assignment of X from the grid function, which may contain initial conditions,
+    // we alias the grid function to X
+    test_var->MakeTRef(_test_variables.at(i)->ParFESpace(), X, _block_true_offsets[i]);
+    _test_true_vector = &X;
+    // This might seem silly but after making the tref the memory flags of the grid function and its
+    // true vector are in an empty state other than the aliasing. This operation syncs the flags and
+    // should be a no-op in terms of actual data transfer
+    test_var->GetTrueVector().SyncMemory(*_test_true_vector);
+    test_var->SetFromTrueVector();
```
alexanderianblair: This is `SetTestVariablesFromTrueVectors()` - using the method would help reduce duplication if we ever need to modify how test vars are updated from their true vectors at a later date

framework/src/mfem/problem_operators/ProblemOperatorInterface.C
alexanderianblair: Understood - this makes sense now with the usage of `_test_true_vector` formed in `Init`. I agree each grid function needs its own flags synced with X.

lindsayad:
@nmnobre can you re-review when you have a chance?

alexanderianblair (approved):

framework/include/mfem/outputs/MFEMDataCollection.h
```diff
@@ -26,6 +26,11 @@ class MFEMDataCollection : public FileOutput
 protected:
   void registerFields();
   void output() override;
+  /**
+   * Whether the grid functions need to be explicitly read on the host before output. This seems to
+   * be required for Conduit data output whereas other formats handle the host read themselves
+   */
+  virtual bool gridFunctionsNeedHostRead() const { return false; }
```
nmnobre, line 33: Okay for now due to all the incoming complications with updating MFEM, but once that inevitably happens, I'd vote to remove this, you okay with that?

lindsayad (approved):
I approve the changes that aren't mine

lindsayad:
I like the changes @nmnobre. Is this gtg?

nmnobre (approved):

nmnobre:
Likewise, I approve the changes that aren't mine.

framework/include/mfem/outputs/MFEMDataCollection.h
lindsayad: certainly

loganharbour (approved):

lindsayad:
@loganharbour have you looked at https://github.com/idaholab/civet_recipes/pull/2416 ?

lindsayad:
Let's wait to merge this until we have the cuda gpu recipe figured out. We can't have this failing 2 out of every 3 times everywhere

lindsayad:
failure is unrelated due to multiple base sha's

PR #31099, opened 2025-07-22
Avoid redundant add/remove of GhostingFunctors

roystgnr:
Refs #31069

## Reason
This should avoid soon-to-be-deprecated calls to the libMesh add/remove methods

## Design
Fixing the tracking of `_attached_relationship_managers` fixes one to-be-deprecated redundant add.  Avoiding a manually-reattached `default_coupling()` object fixes the second.  Manually checking for attempted removals of non-attached functors fixes that problem.

## Impact
If any apps are directly removing GhostingFunctors independent of the MOOSE RelationshipManager system, then that `default_coupling()` add might have been a necessary re-add.  Otherwise we should see no impact other than forwards compatibility.

### DIFF
--- a/framework/src/base/MooseApp.C
+++ b/framework/src/base/MooseApp.C
@@ -3147,7 +3147,10 @@ MooseApp::attachRelationshipManagers(MeshBase & mesh, MooseMesh & moose_mesh)
     if (rm->isType(Moose::RelationshipManagerType::GEOMETRIC))
     {
       if (rm->attachGeometricEarly())
+      {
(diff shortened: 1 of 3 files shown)


loganharbour (approved):

lindsayad:
It seems like that time stepping failure is getting more frequent. It's documented as an intermittent failure in #27555

lindsayad:
For the big roar you made on slack, I was anticipating the required changes to be bigger 😄

PR #31104, opened 2025-07-22
Add kokkos+libceed to all petsc builds, update container mpi, update containers, fix conda zsh on linux

### LINKED ISSUE #31041: Linux ZSH Conda initializations not working
### Bug Description

The following option is not supported in ZSH
https://github.com/idaholab/moose/blob/b57d1f86f40635e2248990e9de09151eaf643b4a/conda/mpi/base_build.sh#L53

### Steps to Reproduce

Create a Conda environment using ZSH on Linux. An orphaned `-Wl,` is present, causing issues during builds/configs (especially with CMake).

### Impact

This prevents ZSH users on Linux from using our packages.

### [Optional] Diagnostics

relevant info from `./diagnostics.sh`:

```pre
LDFLAGS=-Wl, -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,/home/someone/micromamba/envs/moosemoose/lib -Wl,-
        ^^^^ problematic
```

### LINKED ISSUE #31089: Add libCEED support for MFEM backend
### Motivation

MFEM's cmake options include the possibility of making it against a [libCEED](https://github.com/CEED/libCEED) build, which extends its capabilities. Notably, it allows the user to perform matrix-free solves and partial assembly on tet meshes, both of which can be particularly performant on GPUs.

### Design

To include libCEED in the MFEM build, all that is needed is to add a script that clones and makes a libCEED submodule (which is small and builds very quickly), and add libCEED as a cmake option in the MFEM build script.

Eventually, it would also be necessary to include libCEED in the CIVET CI workflow, so that we can add static and transient matrix-free tests. However, this more likely will need to be a separate PR.

### Impact

The enhancement will add new capabilities, particularly improving GPU performance by allowing partial assembly on tet meshes and matrix-free solves.

loganharbour:
- Adds `--download-kokkos` and `--download-kokkos-kernels` to all PETSc builds (including CPU)
- Updates openmpi in all containers to openmpi 5.0.8
- Updates mpich in all containers to mpich 4.3.1
- Updates base container (system packages)
- Adds `--download-libceed` to all PETSc builds and uses it in MFEM (closes https://github.com/idaholab/moose/issues/31089)
- Fixes conda zsh prompt on linux (closes https://github.com/idaholab/moose/issues/31041)

### DIFF
--- a/conda/mpi/base_build.sh
+++ b/conda/mpi/base_build.sh
@@ -12,6 +12,7 @@ function baked_flags()
     # support ZSH initializations
     if setopt &>/dev/null; then
         setopt local_options BASH_REMATCH
+        setopt rematchpcre
     fi
 
     # flags that were set during the build process
@@ -60,6 +61,14 @@ function baked_flags()
         b_FFLAGS=\${b_FFLAGS//\${strip_flag}/}
         b_LDFLAGS=\${b_LDFLAGS//\${strip_flag}/}
     done
+
+    # Remove orphaned -Wl, flags
+    # Note we cannot use variable replacement here as it will not work with spaces in zsh
+    b_CXXFLAGS=$(echo "\$b_CXXFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_CPPFLAGS=$(echo "\$b_CPPFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_CFLAGS=$(echo "\$b_CFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_FFLAGS=$(echo "\$b_FFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_LDFLAGS=$(echo "\$b_LDFLAGS" | sed 's/-Wl,[[:space:]]//g')
 
     # append necessary std c library
     export CXXFLAGS="\${b_CXXFLAGS} -std=c++17"
--- a/apptainer/mpi.def
+++ b/apptainer/mpi.def
@@ -31,18 +31,14 @@
 
 {#- The files directory for this definition                                                   -#}
 {%- set FILES_DIR = MOOSE_DIR + '/apptainer/files/mpi' -%}
-{#- Whether or not to enable openmpi                                                          -#}
-{%- set ENABLE_OPENMPI = ALTERNATE_FROM is not defined or ALTERNATE_FROM == "cuda" -%}
-{#- Whether or not to enable mpich                                                            -#}
-{%- set ENABLE_MPICH = ALTERNATE_FROM is not defined or ALTERNATE_FROM != "cuda" -%}
 
 Bootstrap: oras
 {%- if ALTERNATE_FROM == "gcc_min" or ALTERNATE_FROM == "clang" or ALTERNATE_FROM == "clang_min" %}
-From: mooseharbor.hpc.inl.gov/base/rocky-x86_64:8.10-3
+From: mooseharbor.hpc.inl.gov/base/rocky-x86_64:8.10-5
 {%- elif ALTERNATE_FROM == "cuda" %}
-From: mooseharbor.hpc.inl.gov/moose-hpcbase/rocky-cuda-x86_64:8.10-cuda12.4-1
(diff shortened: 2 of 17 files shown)


NamjaeChoi:
I tested this on my own and works for me. Thanks for the update

milljm (approved):

NamjaeChoi:
If this is merged, how long do the new containers take to be deployed to CIVET?

loganharbour:
> If this is merged, how long do the new containers take to be deployed to CIVET?

They'll be in tomorrow morning.

loganharbour:
Failure is unrelated - timeout.

conda/mpi/base_build.sh
```diff
@@ -61,6 +62,14 @@ function baked_flags()
         b_LDFLAGS=\${b_LDFLAGS//\${strip_flag}/}
     done
 
+    # Remove orphaned -Wl, flags
+    # Note we cannot use variable replacement here as it will not work with spaces in zsh
+    b_CXXFLAGS=$(echo "\$b_CXXFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_CPPFLAGS=$(echo "\$b_CPPFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_CFLAGS=$(echo "\$b_CFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_FFLAGS=$(echo "\$b_FFLAGS" | sed 's/-Wl,[[:space:]]//g')
+    b_LDFLAGS=$(echo "\$b_LDFLAGS" | sed 's/-Wl,[[:space:]]//g')
```
hugary1995, line 71: I don't believe this part is working. What I see in `$CONDA_PREFIX/etc/conda/activate.d/activate_zzz_moose-mpi-base.sh` is

```bash
    # Remove orphaned -Wl, flags
    # Note we cannot use variable replacement here as it will not work with spaces in zsh
    b_CXXFLAGS=$b_CXXFLAGS
    b_CPPFLAGS=$b_CPPFLAGS
    b_CFLAGS=$b_CFLAGS
    b_FFLAGS=$b_FFLAGS
    b_LDFLAGS=$b_LDFLAGS
```
It almost seems like conda evaluated these commands when generating the activate.d scripts. We probably need to escape the entire command, maybe add double quotes.

@loganharbour could you confirm this?

loganharbour: Do we still need this? Can add to https://github.com/idaholab/moose/pull/31935

hugary1995: Yes, please!

loganharbour: You can actually edit the activate script in your installed environment. can you edit it and see how you can get it working with what your use case is?

hugary1995: If I have this in activate.d/activate_zzz_moose-mpi-base.sh, I can get sane LDFLAGS

```bash
    b_CXXFLAGS=$(echo "$b_CXXFLAGS" | sed 's/-Wl,[[:space:]]//g')
    b_CPPFLAGS=$(echo "$b_CPPFLAGS" | sed 's/-Wl,[[:space:]]//g')
    b_CFLAGS=$(echo "$b_CFLAGS" | sed 's/-Wl,[[:space:]]//g')
    b_FFLAGS=$(echo "$b_FFLAGS" | sed 's/-Wl,[[:space:]]//g')
    b_LDFLAGS=$(echo "$b_LDFLAGS" | sed 's/-Wl,[[:space:]]//g')
```

loganharbour: Does this work in bash as expected?

hugary1995: Yes it does, I just tried both zsh and bash. This makes sense because sed is not a shell feature. The only thing we need to worry about is whether the sed utility is available on the platforms we support. Apparantly sed is very widely supported, so we should be safe.

We just need to modify the base_moose conda script with the correct escaping so that the activate.d script coming out has the correct commands.

hugary1995: ^ edited as I misspoke about echo...

PR #31113, opened 2025-07-23
Advanced Extruder Generator Extension -- Extrude Along Curves

### LINKED ISSUE #31112: Add Extrude Along Curve to AdvancedExtruderGenerator
### Motivation

Extrusion is currently limited to only along a single direction. To create items such as pipe bends, an extrusion needs to take place along a 1D curve. This will aid in applications for the Navier-Stokes module.

### Design

Feature should take a 1D meshed curve that is fixed to a face or side and extrude the mesh along.

### Impact

Will allow users to easily create meshed curves and bends in their geometry without using an external mesher like Cubit.

lipohlmann:
Closes #31112

## Reason
Feature was created in pursuit to the creation of an action component that allows users to easily connect two meshes. Also extends capabilities of existing code (i.e. AdvancedExtruderGenerator) for users to be able to extrude their meshes along 1D meshed curves.

## Design
The user is able to supply a 1D meshed curve that is touching the edge or face they want to extrude. The tool then extrudes each layer along the 1D curve. Additional feature was added to allow users to control a radius change (implemented for arbitrary shapes, not just for circles) and the rate of that change (currently, CUBIC and LINEAR are the only available options, thought CUBIC also allows the user to specify the start and end rate of change).

## Impact
Expands the current capabilities of the AdvancedExtruderGenerator for an increase in flexibility. Meshing along a curve allows users to create elements such as pipe bends within MOOSE.


GiudGiud:
ugh this is going to be a nasty rebase. lessgo

GiudGiud:
Rebased, cleaned the commit history and it seems to pass the relevant tests locally. 

I think I'll extract the spline curve generator changes from this PR though, to make it all about extrusion. 
I'll also rework the advanced extrusion generator per the group discussion a couple weeks ago

GiudGiud:
Going to reduce the footprint in AEG.C a bit, will undraft after

GiudGiud:
@roystgnr this should be ready for review

framework/include/meshgenerators/AdvancedExtruderGenerator.h
```diff
@@ -63,6 +63,12 @@ class AdvancedExtruderGenerator : public MeshGenerator
   /// The direction of the extrusion
   Point _direction;
 
+  /// Mesh to extrude along
+  std::unique_ptr<MeshBase> & _extrusion_curve;
+  const libMesh::Point _start_extrusion_direction;
+  const libMesh::Point _end_extrusion_direction;
+  bool _extrude_along_curve;
```
GiudGiud:
```suggestion
  // Attributes for extruding along a curve mesh of edges
  /// Curve mesh to extrude along
  std::unique_ptr<MeshBase> & _extrusion_curve;
  /// Extrusion direction to follow at the start (first layer) of the extrusion
  const libMesh::Point _start_extrusion_direction;
  /// Extrusion direction for the final layer of the extrusion
  const libMesh::Point _end_extrusion_direction;
  /// Whether we are extruding along a curve
  bool _extrude_along_curve;
```

framework/include/meshgenerators/AdvancedExtruderGenerator.h
```diff
@@ -83,4 +89,18 @@ class AdvancedExtruderGenerator : public MeshGenerator
...
+  /// Radial extent of the extruded geometry at the end  of the extrusion
+  const Real _end_radial_extent;
+  /// Function type for modifying the radial extent of the extruded shape
+  const MooseEnum _radial_expansion_method;
+  /// Derivative of the radial expansion function at the beginning of the extrusion
+  const Real _start_radial_growth_rate;
+  /// Derivative of the radial expansion function at the end of the extrusion
+  const Real _end_radial_growth_rate;
+
+  /// Calculate ratio to apply to the expansion
```
GiudGiud:
```suggestion
  /// Calculate the share of the radial expansion to apply at the local node
  /// This share goes from 0 at the beginning of the extrusion to 1 at the end
```

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -409,44 +542,190 @@ AdvancedExtruderGenerator::generate()
...
 
       // k is the element layer ordering within each elevation layer
-      for (unsigned int k = 0; k < order * num_layers + (e == 0 ? 1 : 0); ++k)
+      for (const auto k : make_range(num_heights_at_elevation))
       {
+        // Compute current_step, the update vector
         // For the first layer we don't need to move
         if (e == 0 && k == 0)
-          current_distance.zero();
+          current_step.zero();
```
GiudGiud: this is still confusing imo

maybe cumulative_step ?

framework/doc/content/source/meshgenerators/AdvancedExtruderGenerator.md
```diff
@@ -28,6 +28,25 @@ Boundary ID remapping also works similarly to subdomain ID remapping. During ext
...
+- the [!param](/Mesh/AdvancedExtruderGenerator/start_extrusion_direction) if specified at the first curve node, else the direction from the first to the second node
+- the direction from the previous node to the next node (ignoring the current node) for every other curve node
+- the [!param](/Mesh/AdvancedExtruderGenerator/end_extrusion_direction) if specified at the last curve node, else the direction from the last but one to the last node
+
+!alert note
+When extruding along an line mesh curve, specifying the biases, heights, number of layers and directions is not currently supported.
+
+## Radial growth during extrusion
+
+By setting the [!param](/Mesh/AdvancedExtruderGenerator/end_radial_extent), the radial extent of the extruded mesh can be expanded or
```
roystgnr, line 45: It's not clear to me from these docs what "radial extent" would mean if we're ever extruding anything that's not a circle (or annulus, or at least *something* with circular symmetry); we should give a definition here.

test/tests/meshgenerators/advanced_extruder_generator/tests
```diff
@@ -130,14 +121,22 @@
...
+  [radial_growth]
+    type = 'Exodiff'
+    input = 'gen_extrude.i'
+    cli_args = "Mesh/extrude/end_radial_extent='${fparse 2 * sqrt(2)}' --mesh-only straight_radial_growth_in.e"
+    exodiff = 'straight_radial_growth_in.e'
+    recover = false
 
-  [errors]
-    requirement = 'The system shall error'
+    requirement = 'The system shall be able to extrude a mesh and proportionally increase its radial extent as it is being extruded.'
```
roystgnr: Is 'proportionally' what we want here?  I think it's fine, but to my ears 'linearly' sounds more intuitive.

roystgnr: Maybe just 'with linear radial expansion' to match the 'cubic' text below?

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -241,43 +323,76 @@ AdvancedExtruderGenerator::AdvancedExtruderGenerator(const InputParameters & par
...
+       _upward_boundary_ids.size() != num_elevations))
+  {
+    if (_extrude_along_curve)
+      paramError(
+          "upward_boundary_ids",
+          "This parameter must have the same length (" +
+              std::to_string(_upward_boundary_ids.size()) + ") as upward_boundary_source_blocks (" +
+              std::to_string(_upward_boundary_source_blocks.size()) + ") and num_elevations (" +
+              std::to_string(num_elevations) +
+              "). Note that the number of heights is set to 1 when extrudingn along a curve.");
```
roystgnr:
```suggestion
              "). Note that the number of heights is set to 1 when extruding along a curve.");
```

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -241,43 +323,76 @@ AdvancedExtruderGenerator::AdvancedExtruderGenerator(const InputParameters & par
...
+  {
+    if (_extrude_along_curve)
+      paramError(
+          "downward_boundary_ids",
+          "This parameter must have the same length (" +
+              std::to_string(_downward_boundary_ids.size()) +
+              ") as downward_boundary_source_blocks (" +
+              std::to_string(_downward_boundary_source_blocks.size()) + ") and (" +
+              std::to_string(num_elevations) +
+              "). Note that the number of heights is set to 1 when extrudingn along a curve.");
```
roystgnr:
```suggestion
              "). Note that the number of heights is set to 1 when extruding along a curve.");
```

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -351,14 +466,32 @@ AdvancedExtruderGenerator::generate()
 
   std::unique_ptr<MeshBase> input = std::move(_input);
 
+  std::unique_ptr<MeshBase> extrusion_curve;
+  if (_extrude_along_curve)
+  {
+    extrusion_curve = std::move(_extrusion_curve);
+
+    // We must serialize the curve to know how to extrude across all ranks
+    libMesh::MeshSerializer serial(*extrusion_curve);
```
roystgnr: This won't work.  The `MeshSerializer` will serialize the mesh, sure, and then will happily let it be distributed again when `serial` is destructed at the end of the `if` block!

The idiom we want is, after/outside the `if` block, `libMesh::MeshSerializer serial(*extrusion_curve, _extrude_along_curve)`; that second parameter can be false and then the serializer is a no-op.

Also, why are we doing the `move` here? Why not just keep operating on `_extrusion_curve` if we're using it?

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -409,44 +542,190 @@ AdvancedExtruderGenerator::generate()
...
+
+          // Keep track of the cumulative step size as an extrusion axis coordinate
+          cumulative_step_size += step_size;
 
-          current_distance = old_distance + _direction * step_size;
+          // Handle radial expansion. No need to perform expansion at the centerline
+          if (_end_radial_extent && start_node_radius > libMesh::TOLERANCE)
+          {
+            // How far along we are in the extrusion
+            // TODO: this assumes equi-distance of the nodes!!
```
roystgnr: Are we not totally breaking this assumption?

roystgnr: I guess we don't actually need to support expansion that's linear/cubic in extrusion distance rather than in layer count, we just need to do one of two things:

1) scream and die if someone asks for radial expansion but doesn't give an extrusion mesh with equispaced nodes

or

2) Document that we're defining "linear" and "cubic" to be in terms of layer count rather than of extrusion distance.

My vote would be (1) but either would be fine.

roystgnr (changes requested):

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: the move needs to be done somewhere, it s a quirk of the MG system to require all input meshes to have been moved

GiudGiud: I ll fix the rest

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: we are very much breaking this assumption. 

I'll probably go for (2) for now

framework/doc/content/source/meshgenerators/AdvancedExtruderGenerator.md
GiudGiud: added a sentence with the definition I use

test/tests/meshgenerators/advanced_extruder_generator/tests
GiudGiud: updated to linearly

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: nevermind I did (3) which I think will be nice for improving the quality of our outputs

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: > libMesh::MeshSerializer serial(*extrusion_curve, _extrude_along_curve); that second parameter can be false and then the serializer is a no-op.

right but I can't dereference extrusion_curve if it a null unique_ptr  right

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: changed to something explicit

framework/src/meshgenerators/AdvancedExtruderGenerator.C
roystgnr: Yeah, that's definitely undefined behavior.  Putting the serializer before the move is probably the cleanest solution.

framework/doc/content/source/meshgenerators/AdvancedExtruderGenerator.md
roystgnr: That's the natural definition IMHO, but it's still better to have it spelled out.

framework/src/meshgenerators/AdvancedExtruderGenerator.C
roystgnr: Ah:

3. Compute the arc lengths and just do the expansion right.

Much much better than my suggestions.

framework/src/meshgenerators/AdvancedExtruderGenerator.C
roystgnr: Ah, but if we don't have an extrusion curve mesh, we can't actually dereference that mesh safely whether we've moved it or not.

Damn.  I think the current code is probably an "invalid C++ but most compilers will let it slide" time-bomb; IIRC there was a similar idiom that led to a security exploit in the Linux kernel years ago, after an optimizer decided "I see they're dereferencing that pointer, so I can assume it's not null!"

The proper fix is something we've had to use a couple places in libMesh, stl_io.C and unstructured_mesh.C: instead of putting the `MeshSerializer` on the stack, we put a `unique_ptr<MeshSerializer>` there, and then the `make_unique<MeshSerializer>` goes in the `if` block so it never gets called unless the mesh pointer being dereferenced really isn't null.

GiudGiud: > Damn. I think the current code is probably an "invalid C++ but most compilers will let it slide" time-bomb; 

my workaround was not *extrusion_curve, I created a dummy mesh unique ptr to dereference if extrusion_curve = nullptr. 

anyway, I switched to your solution

framework/src/utils/MooseMeshUtils.C
```diff
@@ -273,13 +273,23 @@ changeSubdomainId(MeshBase & mesh, const subdomain_id_type old_id, const subdoma
   mesh.unset_is_prepared();
 }
 
+Real
+meshVolumeCalculator(const MeshBase & mesh)
+{
+  mooseAssert(mesh.is_prepared(), "Ranges need to be known");
```
GiudGiud: is there something finer grained for the ranges?
I didnt know we needed to prepare for getting element ranges but i hit it pretty hard

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -464,16 +464,16 @@ AdvancedExtruderGenerator::generate()
...
-  {
     extrusion_curve = std::move(_extrusion_curve);
 
-    // We must serialize the curve to know how to extrude across all ranks
-    libMesh::MeshSerializer serial(*extrusion_curve);
-  }
+  // We must serialize the curve to know how to extrude across all ranks
+  std::unique_ptr<libMesh::MeshSerializer> serializer;
+  if (_extrude_along_curve)
+    serializer = std::make_unique<libMesh::MeshSerializer>(*extrusion_curve);
```
roystgnr, line 476: We could move the `serializer` declaration up a few lines and then merge these two `if` statements into one block.

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -349,16 +464,34 @@ AdvancedExtruderGenerator::generate()
...
+  unsigned int total_num_layers;
+  unsigned int total_num_elevations;
+  if (!_extrude_along_curve)
+  {
+    total_num_layers = std::accumulate(_num_layers.begin(), _num_layers.end(), 0);
+    total_num_elevations = _heights.size();
+  }
+  else
+  {
+    total_num_layers = extrusion_curve->n_elem();
```
roystgnr, line 492: Huh.  I was going to say it's possible that this is why we need to prepare (since n_elem() can get out of sync when a DistributedMesh is built or refined in a distributed fashion), but you've got the prepare going on afterward.  (and also in practice it's pretty hard for n_elem() to get out of sync; I'm not sure what would do it *other* than mesh refinement that isn't followed by a prepare_for_use())

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -409,44 +542,213 @@ AdvancedExtruderGenerator::generate()
...
+    for (const auto e : make_range(total_num_elevations))
     {
-      auto num_layers = _num_layers[e];
-
-      auto height = _heights[e];
+      unsigned int num_layers = libMesh::invalid_uint;
+      Real height = std::numeric_limits<Real>::max(), bias = std::numeric_limits<Real>::max();
+      if (_extrude_along_curve)
+      {
+        num_layers = extrusion_curve->n_elem();
```
roystgnr, line 603: Maybe the `n_elem()` call here is what's getting us, or the `n_nodes()` call later?

I guess we could try backing off to testing just `extrusion_curve->preparation().has_synched_id_counts` and calling `update_parallel_id_counts()` rather than a full `prepare_for_use()`.

framework/src/utils/MooseMeshUtils.C
```diff
@@ -273,13 +273,23 @@ changeSubdomainId(MeshBase & mesh, const subdomain_id_type old_id, const subdoma
   mesh.unset_is_prepared();
 }
 
+Real
+meshVolumeCalculator(const MeshBase & mesh)
```
roystgnr: So ... `libMesh::MeshTools::volume(mesh)`?

And actually, if this is what's failing on us, I think I see another possibility for the preparedness problem: if you pass in a mesh that hasn't been entirely partitioned, your code is only summing over the elements which do have a `processor_id()` assigned; the ones that are still `invalid_processor_id` aren't getting counted.  `MeshTools::volume()` counts those too.

I don't want to get too smug here, though; it looks like `MeshTools::volume()` relies on `_elem_dimensions` being a prepared cache, but neither it nor the `mesh_dimension()` call itself assert that.  I'll fix that on the libMesh side; on this side, if we do weaken the `is_prepared()`/`prepare_for_use()` idiom, we'll need to check those caches.

roystgnr (changes requested):
A full `prepare_for_use()` should be fine on a tiny 1D mesh, so don't treat my microoptimization speculations as suggestions, much less requirements... but could we just call `MeshTools::volume()` rather than reinventing it?

framework/src/utils/MooseMeshUtils.C
GiudGiud: sounds good I can try to see if that fixes it without preparing

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: the explanation of what we are doing looks nicer this way though so I kept it that way

framework/src/utils/MooseMeshUtils.C
GiudGiud: ok prepareing is still needed

GiudGiud: (well let me check `extrusion_curve->preparation().has_synched_id_counts` real quick)

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: I tried. needs prepare for use to pass

framework/src/utils/MooseMeshUtils.C
GiudGiud: see other thread.

added a commewnt to look at mesh_tools. I honestly always forget

roystgnr (approved):
Disable-HDF5 test failure is unrelated

PR #31121, opened 2025-07-24
Added hs_boundary in HeatTransferFromHeatStructure1Phase

### LINKED ISSUE #31120: Allow flow channel to be coupled with sub-section of heat structure boundary
### Motivation

Sometimes we want to perform heat transfer between a flow channel and only a sub-section of a heat structure's boundary, rather than a whole boundary.

### Design

In `HeatTransferFromHeatStructure1Phase`:

- Make `hs_side` optional.
- Add `hs_boundary` as an alternative.

These changes are expected to require some light refactoring.

### Impact

Backwards compatible input. Derived applications may require code patch.

joshuahansel:
Closes #31120

Other changes:

- Made all the parameter checking methods in Component public instead of protected so that they can be called from interface classes, for example. Note the git diff is misleading.


joshuahansel:
This will break at least RELAP-7. I will look into if it can be avoided without too much hassle.

modules/thermal_hydraulics/include/interfaces/HSBoundaryInterface.h
```diff
@@ -21,27 +21,37 @@ class Component;
...
+   * @param[in] component   Component pointer
+   */
+  Component2D::ExternalBoundaryType
+  getHSExternalBoundaryType(const Component * const component) const;
 
   /// Heat structure name
   const std::string & _hs_name;
-  /// Heat structure side enum
-  const MooseEnum & _hs_side_enum;
+  // TODO: make the following parameters private after transition to use above APIs
```
licharlot, line 52: To do?

licharlot (changes requested):

modules/thermal_hydraulics/include/interfaces/HSBoundaryInterface.h
joshuahansel: That's there intentionally. And it's conventionally "TODO" instead of "TO DO" if that's what you mean.

licharlot: I thought this was something you intended to do in the PR. If this is intentional, this is fine.

licharlot (approved):

PR #31124, opened 2025-07-25
Fix a logic for ADConvectiveHeatFluxBC with distributed mesh

YaqiWang:
Closes #31123.

### DIFF
--- a/modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
+++ b/modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
@@ -115,6 +115,11 @@ ADConvectiveHeatFluxBC::initialSetup()
         htc_can_use_neighbor && neighbor && _htc_functor->hasBlocks(neighbor->subdomain_id());
   }
 
+  _communicator.max(_T_infinity_use_neighbor);
+  _communicator.max(_htc_use_neighbor);
+  _communicator.min(T_inf_can_use_neighbor);
+  _communicator.min(htc_can_use_neighbor);
+
   const std::string error_msg =
       "Functor must either be defined on all of the primary side of the boundary or on all "
       "of the secondary side.";
--- a/modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
+++ b/modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
@@ -95,7 +95,7 @@ ADConvectiveHeatFluxBC::initialSetup()
   {
     const auto & [elem, side, bid] = *bnd_elem;
     // Skip if this boundary is not part of the restriction
-    if (!hasBoundary(bid))
+    if (!hasBoundary(bid) || (elem->processor_id() != this->processor_id()))
       continue;
 
     // Use neighbors if the functor is not defined on all primary blocks
@@ -105,13 +105,14 @@ ADConvectiveHeatFluxBC::initialSetup()
 
     // Determine if neighbor can be used, just in case
     const auto neighbor = elem->neighbor_ptr(side);
-    if (neighbor && !neighbor->is_remote())
-    {
-      T_inf_can_use_neighbor =
-          T_inf_can_use_neighbor && _T_infinity_functor->hasBlocks(neighbor->subdomain_id());
-      htc_can_use_neighbor =
-          htc_can_use_neighbor && _htc_functor->hasBlocks(neighbor->subdomain_id());
-    }
+    if (neighbor)
+      mooseAssert(
+          !neighbor->is_remote(),
+          "The neighbor best not be remote because we request at least one layer of ghosting");
+    T_inf_can_use_neighbor = T_inf_can_use_neighbor && neighbor &&
+                             _T_infinity_functor->hasBlocks(neighbor->subdomain_id());
+    htc_can_use_neighbor =
+        htc_can_use_neighbor && neighbor && _htc_functor->hasBlocks(neighbor->subdomain_id());
   }
 
   const std::string error_msg =
--- a/modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
+++ b/modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
@@ -105,10 +105,13 @@ ADConvectiveHeatFluxBC::initialSetup()
 
     // Determine if neighbor can be used, just in case
     const auto neighbor = elem->neighbor_ptr(side);
-    T_inf_can_use_neighbor = T_inf_can_use_neighbor && neighbor &&
-                             _T_infinity_functor->hasBlocks(neighbor->subdomain_id());
-    htc_can_use_neighbor =
-        htc_can_use_neighbor && neighbor && _htc_functor->hasBlocks(neighbor->subdomain_id());
+    if (neighbor && !neighbor->is_remote())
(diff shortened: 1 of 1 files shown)


roystgnr (approved):
Good diagnosis!

YaqiWang:
Can we get this merged? My thermal conduction model cannot run without this with a pre-split mesh.

modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
```diff
@@ -105,10 +105,13 @@ ADConvectiveHeatFluxBC::initialSetup()
...
-                             _T_infinity_functor->hasBlocks(neighbor->subdomain_id());
-    htc_can_use_neighbor =
-        htc_can_use_neighbor && neighbor && _htc_functor->hasBlocks(neighbor->subdomain_id());
+    if (neighbor && !neighbor->is_remote())
+    {
+      T_inf_can_use_neighbor =
+          T_inf_can_use_neighbor && _T_infinity_functor->hasBlocks(neighbor->subdomain_id());
+      htc_can_use_neighbor =
+          htc_can_use_neighbor && _htc_functor->hasBlocks(neighbor->subdomain_id());
+    }
```
lindsayad: This changes the logic. If initially `T_inf_can_use_neighbor` and `!neighbor` your PR will change from a final result of `T_inf_can_use_neighbor == false` to `T_inf_can_use_neighbor == true`. This indicates we don't have test coverage for this if no tests are failing

lindsayad (changes requested):

modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
roystgnr: I'm trying to figure out what's going on with the test coverage for this.  The only test that sets `T_infinity_functor` and `heat_transfer_coefficient_functor` is `fe_fv_coupled.i` ... but that test *isn't* restricted to ReplicatedMesh, so you'd have expected it to already have revealed any distributed-mesh bugs.

And in hindsight, I'm not sure how Yaqi is hitting a `neighbor->is_remote()` case.  This BC should be getting evaluated only on local elements, right?  Those should only have ghosted neighbors, not remote neighbors.

YaqiWang: I doubt this is only visiting local elements with `*_mesh.getBoundaryElementRange()`. This makes me wondering that the other fix might be skipping element not on local.

YaqiWang: @lindsayad yes, this does change the logic. When neighbor is null, `_T_infinity_use_neighbor` need to be true to trigger the error, which means the functor is not defined on a boundary element. We probably do not have a test for that.

YaqiWang: Note that `neighbor->is_remote()` does not mean we cannot use neighbor values. So I think skipping the ghost elements in this loop is possibly a better fix.

lindsayad: I get a failure with current moose with 4 procs and distributed mesh. Some day we'll have distributed mesh sweeps on modules

roystgnr: getBoundaryElementRange() should be visiting non-local elements too in general, yeah.  But now I'm even more baffled by how that regression test isn't screaming when run distributed.

Wait.  Is it just not ever getting run on enough processors?  `--re functor/fv --distributed-mesh -p 3` dies with an error for me.

lindsayad: Yea @roystgnr see my post above yours

lindsayad: > Note that `neighbor->is_remote()` does not mean we cannot use neighbor values. So I think skipping the ghost elements in this loop is possibly a better fix.

What are you talking about? You're going to try to evaluate a functor with a remote element? That in general will not work. The only things that could potentially work for are postprocessors/Functions, but it will certainly fail for variable functors

lindsayad: I am fine with only checking local `elem`. But `foo_can_use_neighbor` absolutely requires that `neighbor` is not null and that `neighbor` is not remote

roystgnr: So long as `computeQpResidual` is only getting called on local elems, there shouldn't be a need to check neighbor pointers of non-local elems.

YaqiWang: We basically check neighbor of neighbors of local elements here. So if that is remote, it is fine still for computeResidual. I can revert my change and add a if-continue for checking whether the element is local. Do you think that is okay?

lindsayad: I pushed what I think the code should be

YaqiWang: Okay, skipping the non-local element there causes another test under the same folder `functor.error` failure because expected error was not thrown. I suspect that different processors are not having the same flag. Do you think we need an allreduce operation for these flags?

modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
```diff
@@ -105,13 +105,14 @@ ADConvectiveHeatFluxBC::initialSetup()
...
-      htc_can_use_neighbor =
-          htc_can_use_neighbor && _htc_functor->hasBlocks(neighbor->subdomain_id());
-    }
+    if (neighbor)
+      mooseAssert(
+          !neighbor->is_remote(),
+          "The neighbor best not be remote because we request at least one layer of ghosting");
+    T_inf_can_use_neighbor = T_inf_can_use_neighbor && neighbor &&
+                             _T_infinity_functor->hasBlocks(neighbor->subdomain_id());
+    htc_can_use_neighbor =
```
YaqiWang, line 114: We will need a gather for `_T_infinity_use_neighbor` and `_htc_use_neighbor`, I think.

modules/heat_transfer/src/bcs/ADConvectiveHeatFluxBC.C
YaqiWang: I pushed another commit, which fixes the `functor.error` test.

lindsayad: Thanks. I force pushed a commit that's a superset of yours

YaqiWang: Actually, I do not think we need the other two. I intentionally left them for micro-op.

lindsayad: that's smart. Will remove

lindsayad (approved):

YaqiWang:
I guess we can merge this? The failing test is a timeout and irrelevant.

lindsayad:
failure is unrelated

PR #31126, opened 2025-07-25
Add matrix free ScaleIntegrator methods

### LINKED ISSUE #31125: Matrix-free assembly for MFEM transient problems
### Motivation

With the recent introduction of libCEED capabilities in #31104, we can now perform matrix-free assembly in static problems. However, because matrix-free methods have not yet been introduced into the `ScaleIntegrator` object, which is necessary for scaling the time derivative grid functions in transient problems, we cannot yet perform matrix-free assembly in these cases.

### Design

Fixing this issue is straightforward: add matrix-free methods in the `ScaleIntegrator` which multiply the vectors passed to it by a constant value, in the same way that is already done for element and partial assembly.

### Impact

This will allow us to perform matrix-free assembly in transient problems, which is particularly beneficial to GPU-accelerated problems.

Heinrich-BR:
Closes #31125

### DIFF
--- a/framework/include/mfem/integrators/ScaleIntegrator.h
+++ b/framework/include/mfem/integrators/ScaleIntegrator.h
@@ -92,6 +92,12 @@ public:
   virtual void
   AssembleEA(const mfem::FiniteElementSpace & fes, mfem::Vector & emat, const bool add);
 
+  virtual void AssembleMF(const mfem::FiniteElementSpace & fes);
+
+  virtual void AddMultMF(const mfem::Vector & x, mfem::Vector & y) const;
+
+  virtual void AssembleDiagonalMF(mfem::Vector & diag);
+
   virtual ~ScaleIntegrator();
 };
 } // namespace Moose::MFEM
--- a/framework/src/mfem/integrators/ScaleIntegrator.C
+++ b/framework/src/mfem/integrators/ScaleIntegrator.C
@@ -121,6 +121,31 @@ ScaleIntegrator::AssembleEA(const mfem::FiniteElementSpace & fes,
   }
 }
 
+void
+ScaleIntegrator::AssembleMF(const mfem::FiniteElementSpace & fes)
+{
+  CheckIntegrator();
+  _integrator->AssembleMF(fes);
+}
+
+void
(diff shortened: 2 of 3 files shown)


Heinrich-BR:
@lindsayad The test is failing because in order to use libceed to run a matrix-free assembly script, it is necessary to set `Executioner/device=ceed-cpu`. instead of just `cpu`. The issue is that I can see that `TestHarness.py` sets the device for every MFEM test to be `cpu`. So for the matrix-free tests we end up with a command line that has  `Executioner/device=ceed-cpu` and also `Executioner/device=cpu`. Do you think there's a way to override this default device only for the matrix-free tests? Alternatively we could forgo matrix-free tests, but I would prefer having them in now that we've added libceed.

lindsayad:
We definitely need to fix what we're doing. I definitely don't want to forego matrix-free testing

lindsayad:
I'm desperate to get more testing coverage for MFEM, which is why we put in https://github.com/idaholab/moose/pull/31047 pretty rapidly. @dschwen has also pointed out issues with what we're doing there. We'll try to get this cleaned up quickly so that PRs like this one can go ahead as they should

lindsayad:
@Heinrich-BR I added some code to our test harness to support setting `devices = 'ceed-cpu'` and pushed it here. Hopefully we see the tests running correctly now or rather we will likely see them not run at all...

Heinrich-BR:
@lindsayad most of the tests are passing, but as you can see the CI is complaining that the SHA for `TestHarness` is not trusted. How does one go about "adding trust", as the CI message suggests?

lindsayad:
@loganharbour has to add trust for it. But we might as well just wait on getting https://github.com/idaholab/moose/pull/31151 in because that has just the commits that I pushed here

lindsayad:
Once that's in we can add a recipe that includes ceed-cpu device testing so that you can get coverage of your code here

lindsayad:
Good to see we got coverage for your new stuff working! Ignore cancelled recipes. Those aren't your fault. The documentation failure is legit though

lindsayad (approved):

PR #31142, opened 2025-07-28
Generalise makefiles for better compiler support including nvcc

nmnobre:
## Reason
In general, better support for more compilers and compilation flows.
In particular, we'll soon need to start compiling MOOSE with nvcc, the changes here are enough to enable that workflow on my machine with the following GNU make invocation (which should be straightforward to adapt for a CIVET recipe):
`make -jXX libmesh_CXX=/usr/local/cuda/bin/nvcc CXXFLAGS="--x=cu --forward-unknown-to-host-compiler --extended-lambda -DEIGEN_NO_CUDA" LDFLAGS="--forward-unknown-to-host-compiler"`

## Design
Changed from `-imacros` to the more widely available `-include` to inject `MooseConfig.h`, from the GNU-extended `gnu++11` to the standard `c++14` for exodiff, and from `CXXFLAGS` to a more flexible `LDFLAGS` for linking rules.

## Impact
Hopefully none for existing builds, but that's why we have CIVET.
Better support for more compilers going forward, including nvcc.

### DIFF
--- a/framework/app.mk
+++ b/framework/app.mk
@@ -375,7 +375,7 @@ $(app_LIB): curr_additional_libs := $(ADDITIONAL_LIBS)
 $(app_LIB): $(app_HEADER) $(app_plugin_deps) $(depend_libs) $(app_objects) $(ADDITIONAL_DEPEND_LIBS)
 	@echo "Linking Library "$@"..."
 	@$(libmesh_LIBTOOL) --tag=CXX $(LIBTOOLFLAGS) --mode=link --quiet \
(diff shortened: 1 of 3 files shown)


lindsayad:
Will this be necessary even if https://github.com/mfem/mfem/issues/4959 is addressed?

nmnobre:
> Will this be necessary even if https://github.com/mfem/mfem/issues/4959 is addressed?

Maybe not. But that's why I mentioned nvcc only as a particular use case, nothing here is specific to that compiler. Indeed, the point is to open up possibilities and everything here _adds_ to our capability, nothing detracts. Besides, at least one of the changes will still be needed for the Kokkos functionality. Is there anything in particular you dislike?

lindsayad (approved):
No. I just got scared looking at the compilation line in your OP and I hadn't looked at the actual changes yet 😆

loganharbour:
Looks like this broke the conda install of moose for arm mac: https://civet.inl.gov/job/3095344/. looking into it

Issue #25958, opened 2023-11-03
Make improvements to volume junction model


joshuahansel:
## Reason
Volume junctions are a critical component in systems analysis, since it is the only way to connect pipes with different orientations. Numerous users have reported deficiencies in the current formulation, such as https://github.com/idaholab/moose/discussions/25947.

## Design
This would make model improvements to the existing volume junction model (or make a new one if necessary). For a start, see https://www.sciencedirect.com/science/article/pii/S0045782519303524 for a volume junction model. This paper also includes some test problems that can be used for the junction.

## Impact
Improved systems modeling capability.

joshuahansel:
I found a definite issue that needs to be fixed: the 3D state from the volume junction just takes the normal component of velocity and passes it into the 1D Riemann solver to compute the flux for a 1D connection, but the total energy was not adjusted to exclude the tangential components of velocity, so there's a different junction state (v,e) for each connection. What needs to happen is an actual 3D Riemann solve for the normal flux. This flux should be applied to the junction volume and then a projection of this flux should be applied to the flow channel. Note that due to this projection, the tangential components of the momentum flux are dropped, leading to non-conservation of tangential momentum. This is thought to be acceptable - energy is still conserved, so basically the tangential kinetic energy is converted into internal energy.

joshuahansel:
Another Discussion about volume junction issues: https://github.com/idaholab/moose/discussions/28116

PR #31148, opened 2025-07-29
Lightened a THM heavy test

joshuahansel:
Refs #25958


joshuahansel:
Locally takes about 25 seconds to run now instead of 250 seconds, so I figure the default of 300 seconds is good.

lindsayad (approved):
thanks!

PR #31179, opened 2025-07-31
Add neml2 crystal plasticity model using exact kinematics

### LINKED ISSUE #31178: neml2 crystal plasticity benchmark
### Motivation

The neml2 crystal plasticity model current uses a formulation using approximate kinematics. This leads to a slight difference when compared to moose crystal plasticity models. We should add an example for using neml2 crystal plasticity with the exact kinematics (i.e. with FeFp multiplicative decomposition).

### Design

Add example input files under `modules/solid_mechanics/test/tests/neml2/crystal_plasticity`

### Impact

This would make it easier for users who wish to benchmark neml2 results against moose.

hugary1995:
This PR adds an example moose input file and the corresponding neml2 input file for using neml2 crystal plasticity model with the FeFp multiplicative decomposition.

close #31178


jiangwen84 (approved):
Looks good!

modules/solid_mechanics/src/materials/lagrangian/ComputeLagrangianStressCustomPK2.C
```diff
@@ -0,0 +1,41 @@
...
+
+registerMooseObject("SolidMechanicsApp", ComputeLagrangianStressCustomPK2);
+
+InputParameters
+ComputeLagrangianStressCustomPK2::validParams()
+{
+  InputParameters params = ComputeLagrangianStressPK1::validParams();
+  params.addRequiredParam<MaterialPropertyName>("custom_pk2_stress", "The name of the PK2 stress.");
+  params.addRequiredParam<MaterialPropertyName>(
+      "custom_pk2_jacobian", "The name of the PK2 Jacobian (w.r.t. the deformation gradient).");
```
jiangwen84, line 20: Very minor thing: either "custom_pk2 and custom_pk2_jacobian" or "custom_pk2_stress and custom_pk2_stress_jacobian"

modules/solid_mechanics/src/materials/lagrangian/ComputeLagrangianStressCustomPK2.C
```diff
@@ -0,0 +1,41 @@
...
+}
+
+ComputeLagrangianStressCustomPK2::ComputeLagrangianStressCustomPK2(
+    const InputParameters & parameters)
+  : ComputeLagrangianStressPK1(parameters),
+    _pk2(getMaterialProperty<RankTwoTensor>("custom_pk2_stress")),
+    _dpk2_dF(getMaterialProperty<RankFourTensor>("custom_pk2_jacobian"))
+{
+  if (!_large_kinematics)
+    paramError("large_kinematics", "This material requires large kinematics to be enabled.");
```
jiangwen84, line 31: Can you enforce this parameter to be false?

modules/solid_mechanics/src/materials/lagrangian/ComputeLagrangianStressCustomPK2.C
hugary1995: uhh this is just to be consistent with `ComputeLagrangianCauchyCustomStress` where we use `custom_cauchy_stress` and `custom_cauchy_jacobian`. Also all of the lagrangian mechanics kernels and materials use `pk2_jacobian` or `pk1_jacobian`, so for the sake of consistency let's keep it like this...

modules/solid_mechanics/src/materials/lagrangian/ComputeLagrangianStressCustomPK2.C
hugary1995: I could. But I did this on purpose in case user sets `GlobalParams/large_kinematics=false` and this error check would warn user about the inconsistency.

PR #31184, opened 2025-07-31
Added PP to compute a discrete, block-restricted variable norm

joshuahansel:
Closes #31177


framework/src/postprocessors/DiscreteVariableResidualNorm.C
```diff
@@ -0,0 +1,76 @@
...
+  _norm = 0;
+  _local_dof_indices.clear();
+}
+
+void
+DiscreteVariableResidualNorm::execute()
+{
+  std::vector<dof_id_type> dof_indices;
+  _var.getDofIndices(_current_elem, dof_indices);
+  for (const auto dof_index : dof_indices)
```
lindsayad:
```suggestion
  for (const auto dof_index : _var.dofIndices())
```

lindsayad (changes requested):

joshuahansel:
I'm also planning to add a parameter "correct_mesh_bias", which divides by N for l-1 and sqrt(N) for l-2, where N is the number of dofs in the norm.

joshuahansel:
Added the other norms and the mesh bias option, so this is ready now.

framework/doc/content/source/postprocessors/DiscreteVariableResidualNorm.md
```diff
@@ -0,0 +1,50 @@
+# DiscreteVariableResidualNorm
+
+This postprocessor computes one of the following discrete norms of the nonlinear residual
```
GiudGiud, line 3: is it only nonlinear? solve_type = linear could have a residual vector too

framework/doc/content/source/postprocessors/DiscreteVariableResidualNorm.md
```diff
@@ -0,0 +1,50 @@
...
+
+- $\ell_\infty$:
+
+  !equation
+  \|r\|_\infty = \max\limits_{i\in \mathcal{I}_{u,\Omega}} |r_i|
+
+where $\mathcal{I}_{u,\Omega}$ is the set of residual vector indices corresponding to the
+solution variable $u$ and the domain $\Omega$.
+
+Note that some of these norms have a mesh-size bias. For example, if all vector entries
```
GiudGiud:
```suggestion
Note that some of these algebraic vector norms have a mesh-size bias. For example, if all vector entries
```

GiudGiud (approved):

framework/doc/content/source/postprocessors/DiscreteVariableResidualNorm.md
joshuahansel: It's `_fe_problem.getNonlinearSystemBase(_sys.number()).RHS()` - is that one used for linear solves even though it's from the "nonlinear system"?

framework/doc/content/source/postprocessors/DiscreteVariableResidualNorm.md
joshuahansel: I probably wouldn't say this since the continuous norms have a different mesh-related bias - they're biased on the domain size rather than the number of elements.

framework/doc/content/source/postprocessors/DiscreteVariableResidualNorm.md
GiudGiud: that 's pretty nonlinear yeah.

framework/doc/content/source/postprocessors/DiscreteVariableResidualNorm.md
GiudGiud: ok edited. I just want to get across it s not a spatial integral. There are no perfect norms

framework/doc/content/source/postprocessors/DiscreteVariableResidualNorm.md
joshuahansel: We could probably extend it to linear systems later.

joshuahansel:
Hmm, I think the parallel failure might be real:
```
postprocessors/discrete_variable_residual_norm.test: ERROR: In file test_out.csv: The values in column "fe_A_l1" don't match. 
postprocessors/discrete_variable_residual_norm.test: 	relative diff:   1.322e+04 ~ 8.937e+03 = 3.239e-01 (3.239e-01)
postprocessors/discrete_variable_residual_norm.test: In file test_out.csv: The values in column "fe_A_l1_nobias" don't match. 
postprocessors/discrete_variable_residual_norm.test: 	relative diff:   7.343e+02 ~ 5.958e+02 = 1.886e-01 (1.886e-01)
postprocessors/discrete_variable_residual_norm.test: In file test_out.csv: The values in column "fe_A_l2" don't match. 
postprocessors/discrete_variable_residual_norm.test: 	relative diff:   4.134e+03 ~ 3.240e+03 = 2.163e-01 (2.163e-01)
postprocessors/discrete_variable_residual_norm.test: In file test_out.csv: The values in column "fe_A_l2_nobias" don't match. 
postprocessors/discrete_variable_residual_norm.test: 	relative diff:   9.745e+02 ~ 8.366e+02 = 1.415e-01 (1.415e-01)
```
The l-infinity norm matches but not the l-1 or l-2 norms (at least for subdomain A and FE). At first I thought this was due to taking a slightly different nonlinear iteration, but I tried using `-pc_type lu` and got the same all-dof l-2 norm for both serial and 2 proc (both the console output and in the PP). It also matches with `VariableResidual` 🤔

joshuahansel:
I think I see what the problem is. For the case where a proc owns an element on block A but not one of the dofs on it, that dof's index does not get added into the owning proc's set because that elem never executes on the owning proc. Let me think about the solution here.

joshuahansel:
I guess what I need is something like `_var.dofMap().send_to_owning_proc(dof_index, _local_dof_indices)` if such a thing exists...

framework/src/postprocessors/DiscreteVariableResidualNorm.C
```diff
@@ -0,0 +1,127 @@
...
+    if (_var.dofMap().local_index(dof_index))
+      _local_dof_indices.insert(dof_index);
+    else
+    {
+      // if a Dof is non-local add it to a map, to be communicated to the owner in finalize()
+      const auto dof_owner = _var.dofMap().dof_owner(dof_index);
+      if (_nonlocal_dof_indices_map.count(dof_owner))
+        _nonlocal_dof_indices_map[dof_owner].push_back(dof_index);
+      else
+        _nonlocal_dof_indices_map[dof_owner] = {dof_index};
```
lindsayad:
```suggestion
      _nonlocal_dof_indices_map[dof_owner].push_back(dof_index);
```

lindsayad (changes requested):

framework/src/postprocessors/DiscreteVariableResidualNorm.C
```diff
@@ -0,0 +1,124 @@
...
+void
+DiscreteVariableResidualNorm::finalize()
+{
+  // communicate the non-local Dofs to their processors
+  auto receive_functor = [&](processor_id_type /*pid*/, const std::vector<dof_id_type> & indices)
+  { _local_dof_indices.insert(indices.begin(), indices.end()); };
+  Parallel::push_parallel_vector_data(_communicator, _nonlocal_dof_indices_map, receive_functor);
+
+  // compute the total number of Dofs for the variable on the subdomain
+  unsigned int n_dofs = _local_dof_indices.size();
```
lindsayad: woops I should have caught this earlier. `unsigned int` might not be big enough
```suggestion
  auto n_dofs = _local_dof_indices.size();
```

lindsayad (changes requested):

lindsayad (approved):

PR #31199, opened 2025-08-02
Add a new parameter in parsed node set generator to exclude subdomains

### LINKED ISSUE #31198: Improve ParsedGenerateNodeset slightly to include a parameter excluded_subdomains
### Motivation

Currently, the mesh generator has a parameter `included_subdomains` to add a node set whose nodes are a node of an element within the included subdomains. It is desired to have a generator to remove certain nodes in a list of subdomains, i.e. to have another parameter `excluded_subdomains`. It is noted that the mesh generator has another two parameters already `included_nodesets` and `excluded_nodesets` similarly as the pair of the two parameters considering subdomains.

### Design

It is straight-forward to add this new parameter. One caveat is that a node may belong to both `included_subdomains` and `excluded_subdomains` when the node is on the interface of the two parts of subdomains. In such a case, we will make `excluded_subdomains` win.

### Impact

A minor improvement of an existing mesh generator for generating node sets.

YaqiWang:
To close #31198.


GiudGiud (approved):
looks good

YaqiWang:
Thanks @GiudGiud for your quick review!

PR #31217, opened 2025-08-05
Fix for LDFLAGS during conda moose application build on mac

loganharbour:
Annoying conda is setting a flag (which is probably good...) that is not compatible with our builds on mac. This removes it.

refs #31142

### DIFF
--- a/conda/moose/build.sh
+++ b/conda/moose/build.sh
@@ -8,21 +8,32 @@ function do_build(){
     fi
     # shellcheck disable=SC2086  # we want word spliting when dealing with passing arguments
     ./configure --prefix="${PREFIX:?}/moose" ${MOOSE_OPTIONS:-''} || return 1
-    CORES="${MOOSE_JOBS:-2}"
+    local CORES="${MOOSE_JOBS:-2}"
 
-    # moose_test-opt
+    # moose_test-opt; docs will come from combined so explicitly
+    # don't build them here
     cd test
-    make -j "${CORES:?}"
-    make install -j "${CORES:?}"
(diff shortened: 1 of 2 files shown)


loganharbour:
Welp @nmnobre what was originally a temp fix is an actual fix. Hopefully the comments in the build script explain things.

nmnobre (approved):
I'll try to avoid getting rid of that flag (which as you said is probably useful) in another PR, best to get this merged asap.

conda/moose/build.sh
```diff
@@ -8,22 +8,33 @@ function do_build(){
...
 # shellcheck disable=SC1091  # made available through meta.yaml src path
 source "${SRC_DIR:?}/retry_build.sh"
 
+# Conda on mac sets "-Wl,-dead_strip_dylibs" in LDFLAGS, which is picked up
+# during the application executable linking. Unfortunately, we end up with
+# intermediate dead dylibs during our build/install (which we do fix later!).
+# So... remove that flag on macs? Good, try conda.
+if [[ "$(uname)" == "Darwin" ]]; then
+  export LDFLAGS="${LDFLAGS//-Wl,-dead_strip_dylibs/}"
+fi
```
lindsayad, line 36: We had already done this in 38dc8b9f77263af9db766eed00838a18ad2a7eda. Is that not being used?

loganharbour: Not within the conda build environment it looks like, just the user environment?

lindsayad: damn

Issue #31171, opened 2025-07-30
Enable WebServerControl of stochastic tools apps


zachmprince:
### Motivation

The WebServerControl enables the control of moose simulations via python. Python contains many MANY advanced capabilities involving stochastic simulations and optimization. However, the utilizing these capabilities while driving moose simulations is somewhat clunky, especially when it comes to parallelism. As such, coupling the WebServerControl with the STM can provide a extremely efficient, and flexible interface to prototype optimization and stochastic routines in python while executing real moose-based simulations.

### Design

The Sampler system in the STM is very robust and flexible, so making that system controllable with the WebServerControl would probably be the easiest way to do this. The easiest way to do this would leverage the `InputMatrix` sampler and make it's `matrix` parameter controllable and ensure that it detects changes in the size of the sampling matrix to adjust partitioning. The WebServerControl will need to be enable control of this type of parameter and provide an interface for gathering more general output data provided by the STM.

### Impact

Running python efficiently with the STM.

PR #31230, opened 2025-08-06
Enhancements to web server control

zachmprince:
- Enable control of RealEigenMatrix parameters
- Enable retrieval of reporter values by converting nlohmann JSON to miniJSON
- Provide "terminate" hook to abort simulation without going through every timestep requested

Refs #31171

### DIFF
--- a/framework/include/controls/WebServerControl.h
+++ b/framework/include/controls/WebServerControl.h
@@ -47,6 +47,13 @@ public:
   template <typename T, miniJson::JsonType json_type>
   static T getScalarJSONValue(const miniJson::Json & json_value);
 
+  /**
+   * Convert values to a miniJson::Json node.
+   *
+   * @tparam T Data type
+   * @param value The value to be converted.
+   * @return miniJson::Json The miniJson::Json node of the converted value.
+   */
   template <typename T>
   static miniJson::Json toMiniJson(const T & value);
 
--- a/framework/include/controls/WebServerControl.h
+++ b/framework/include/controls/WebServerControl.h
@@ -13,6 +13,8 @@
 
 #include "WebServerControlTypeRegistry.h"
 
+#include "libmesh/parallel_eigen.h"
+
 #include "tinyhttp/http.h"
 
 #include <atomic>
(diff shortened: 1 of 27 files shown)


framework/include/controls/WebServerControl.h
```diff
@@ -45,6 +47,9 @@ class WebServerControl : public Control
   template <typename T, miniJson::JsonType json_type>
   static T getScalarJSONValue(const miniJson::Json & json_value);
 
+  template <typename T>
```
loganharbour, line 57: Docstring

framework/include/reporters/ReporterInterface.h
```diff
@@ -81,6 +81,13 @@ class ReporterInterface
 
   ///@}
 
+  /**
+   * Get the reporter context to allow non-typed operations with the data
+   * @param reporter_name A ReporterName object that for the desired Reporter context.
+   * @returns ReporterContextBase of the reporter value
+   */
+  const ReporterContextBase & getReporterContextBaseByName(const ReporterName & reporter_name);
```
loganharbour: Should be able to be a const method, no?

framework/include/reporters/ReporterName.h
```diff
@@ -35,6 +35,8 @@ class ReporterName
   ReporterName(const char * combined_name);
   ReporterName(){}; // empty constructor for InputParameters
 
+  static bool isValidName(const std::string & object_and_value_name);
```
loganharbour, line 45: Docstring

framework/src/controls/WebServerControl.C
```diff
@@ -315,6 +380,8 @@ WebServerControl::execute()
 
     for (const auto & value_ptr : _controlled_values)
       name_and_types.emplace_back(value_ptr->name(), value_ptr->type());
+
+    terminate_solve = _terminate_requested.load();
```
loganharbour, line 384: Should we set this back to false now that it's been used?

framework/src/controls/WebServerControl.C
```diff
@@ -364,3 +436,62 @@ WebServerControl::stringifyJSONType(const miniJson::JsonType & json_type)
...
+  const auto nrows = array_of_array_value.size();
+  if (nrows == 0)
+    return RealEigenMatrix::Zero(0, 0);
+
+  RealEigenMatrix matrix;
+  for (const auto i : make_range(nrows))
+  {
+    if (array_of_array_value[i].getType() != miniJson::JsonType::kArray)
+      throw ValueBase::Exception(
+          "The value '" + array_of_array_value[i].serialize() + "' of type " +
```
loganharbour: Is it more useful to have the index of these errors instead of the value itself?

loganharbour (changes requested):

framework/src/controls/WebServerControl.C
zachmprince: Does this make sense?
```cpp
      throw ValueBase::Exception(
          "Element " + std::to_string(i) + " of '" + json_value.serialize() + "' of type " +
          stringifyJSONType(array_of_array_value[i].getType()) + " is not an array");
```

The second error I think makes sense to show the whole thing to show its jagged-ness.

loganharbour (approved):

PR #31232, opened 2025-08-06
Add better name collision checks. 

### LINKED ISSUE #31231: Clarify responsible object for name collisions
### Motivation

When adding a userobject, postprocessor, vectorpostprocessor, or reporter (possibly other systems too) and there is a name collision, MOOSE throws an error like

```
A UserObject with the name 'my_name' already exists. You may not add a Postprocessor by the same name.
```

This can be confusing to users because the actual name collision might not be from something in a `[UserObjects]` block, but rather in `[Postprocessors]`, `[VectorPostprocessors]`, ...

### Design

Update the name collision checks to return information on the actual collision object type and name.

### Impact

Improved user experience.

aprilnovak:
Closes #31231

I also added coverage for a reporter name collision with a userobject, which was not already covered (as far as I could tell). Much of the tests for this also had extraneous parts not necessary for testing.

### DIFF
--- a/framework/src/problems/FEProblemBase.C
+++ b/framework/src/problems/FEProblemBase.C
@@ -4239,7 +4239,8 @@ FEProblemBase::addPostprocessor(const std::string & pp_name,
 {
   // Check for name collision
   if (hasUserObject(name))
-    mooseError("A ", getUserObjectBase(name).typeAndName(),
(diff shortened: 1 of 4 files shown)


loganharbour (approved):
Beautiful 👍🏼 and thanks for simplifying the tests too

PR #31271, opened 2025-08-12
Added default value to RunApp's  parameter to catch unused petsc opti…

pbehne:
…on warnings in the moose test suite.


loganharbour:
This should probably controlled by the allow unused option, and should be an additional check next to absent out

PR #31281, opened 2025-08-13
Time sequence stepper fix

### LINKED ISSUE #31279: TimeSequenceStepper can segfault when restarting
### Bug Description

TimeSequenceStepper will segfault in any situation when it is restarted from a simulation that did not also have a TimeSequenceStepper. This is because of the code below where it is assume a sequence of times (`_time_sequence`) has been restarted and the vector, which could be empty, is access without any checking if it is in fact empty:
```c++
    // also we need to do something different when restarting
    if (!_app.isRestarting())
    {
      // sync _executioner.startTime and endTime with _time_sequence
      Real start_time = _executioner.getStartTime();
      Real end_time = _executioner.endTime();

      // make sure time sequence is in strictly ascending order
      if (!std::is_sorted(times.begin(), times.end(), std::less_equal<Real>()))
        paramError("time_sequence", "Time points must be in strictly ascending order.");

      _time_sequence.push_back(start_time);
      for (unsigned int j = 0; j < times.size(); ++j)
      {
        if (times[j] > start_time && times[j] < end_time)
          _time_sequence.push_back(times[j]);
      }
      _time_sequence.push_back(end_time);
    }
    else
    {
      // in case of restart it should be allowed to modify _time_sequence if it follows the
      // following rule:
      // all times up to _current_step are identical
      // 1. start time cannot be modified
      // 2. the entries in _time_sequence and times must be equal up to entry with index
      // _current_step

      if (!MooseUtils::absoluteFuzzyEqual(_executioner.getStartTime(), _time_sequence[0]))
        mooseError("Timesequencestepper does not allow the start time to be modified.");
```

### Steps to Reproduce

Use TimeSequenceStepper in a restarted simulation where the simulation being restarted from did not have a TimeSequenceStepper

### Impact

Segfault

### [Optional] Diagnostics

_No response_

jthano:
closes #31279

### DIFF
--- a/framework/src/timesteppers/TimeSequenceStepperBase.C
+++ b/framework/src/timesteppers/TimeSequenceStepperBase.C
@@ -46,7 +46,7 @@ TimeSequenceStepperBase::setupSequence(const std::vector<Real> & times)
   if (!_app.isRecovering())
   {
     // also we need to do something different when restarting
-    if (!_app.isRestarting())
+    if (!_app.isRestarting() || _time_sequence.empty())
     {
       // sync _executioner.startTime and endTime with _time_sequence
       Real start_time = _executioner.getStartTime();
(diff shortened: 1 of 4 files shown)


MengnanLi91 (approved):
Looks good to me. I run it with my changes in [#31156](https://github.com/idaholab/moose/pull/31156). All works fine.

Issue #31298, opened 2025-08-13
Segfault when using ElementSubdomainModifier with mechanical contact


bwspenc:
### Bug Description

I have a solid mechanics model that uses an ElementSubdomainModifer to "delete" elements, and also uses mechanical contact. The first time an element gets moved to the other subdomain, it segfaults.

### Steps to Reproduce

I have a model that shows this, but it is unwieldy and probably isn't the best test case for this. I believe this would happen with any model that uses contact with ElementSubdomainModifier, though, so I'll make a smaller test case.

I ran it in a debugger, and it segfaults inside `NearestNodeLocator::findNodes()` because one of the boundary nodes has an invalid pointer to a node. I believe the problem is that the displaced mesh hasn't been updated before the equation systems are reinitialized. I have a code change that I'll submit for review that fixes this problem for me.

### Impact

This prevents us from running a problem that we need to work for a project.

bwspenc:
I confirmed that this was fixed by #30630

PR #30630, opened 2025-05-30
Support more strategies for variable (re)initialization in `ElementSubdomainModifier`

ChengHauYang:
close #30627
close #29054

Add more strategies for variable (re)initialization in `ElementSubdomainModifier`

The `ElementSubdomainModifier` dynamically changes subdomains. When elements change subdomain, we should allow user to specify if/how variables should be (re)initialized.

Currently, we only support initial condition projection as the initialization strategy. This PR adds three strategies for variable (re)initialization:
- `POLYNOMIAL_NEIGHBOR`
- `POLYNOMIAL_NEARBY`
- `POLYNOMIAL_WHOLE`

All three strategies employ the same algorithm:
1. Identify the updated subdomain on which a variable should be (re)initialized.
2. Construct a patch of elements.
3. Fit a polynomial over the element patch.
4. Project the extrapolation of the fitted polynomial onto the updated subdomain.

At step 3, the polynomial is fitted in the least-squares sense against variable value interpolated onto quadrature points. User has control over the polynomial degree. The current implementation hard codes the polynomial basis to be monomial -- this could be relaxed in the future.

The three strategies differ only in step 2. `POLYNOMIAL_NEIGHBOR` constructs the patch using the direct element neighbors of the updated subdomain. `POLYNOMIAL_NEARBY` performs a geometric search to include elements within a given distance. `POLYNOMIAL_WHOLE` uses the entire mesh as the patch.

PR #31299, opened 2025-08-13
Update displaced mesh in meshChanged() 

bwspenc:
ref #31298


bwspenc:
I'll add a test case to this tomorrow that exercises the bug that this fixes.

lindsayad (changes requested):
In `DisplacedProblem::meshChanged` we also call `_mesh.clearQuadratureNodes()` before calling `_mesh.meshChanged()`. I would be a proponent for absorbing the logic from `MooseMesh::clearQuadratureNodes()` into `MooseMesh::meshChanged()` and deleting the `clearQuadratureNodes()` API.

Also it would be great to have some kind of test somewhere to ensure we don't regress on this in the future

bwspenc:
Agreed about setting up the test case, @lindsayad . I've got one set up now, and unfortunately it looks like my test segfaults with this change, so maybe my fix doesn't completely take care of the problems. I'm looking into that now.

bwspenc:
@lindsayad I moved the call to `clearQuadratureNodes()` into `MooseMesh::meshChanged`. It doesn't seem like a bad idea to me to keep that function around as it is still since it encapsulates some operations and since it's also called from the destructor. I'm happy to further change it, though.

framework/src/problems/FEProblemBase.C
```diff
@@ -8110,6 +8112,8 @@ FEProblemBase::meshChanged(const bool intermediate_change,
   // "active" and "local" may have been *changed* by refinement and
   // repartitioning done in EquationSystems::reinit().
   _mesh.meshChanged();
+  if (_displaced_problem)
+    _displaced_mesh->meshChanged();
```
lindsayad, line 8116: I don't think this one should be necessary. `es().reinit()` should not touch the displaced mesh

lindsayad (changes requested):

framework/src/problems/FEProblemBase.C
bwspenc: I didn't have that in there initially, but then I added it because the `meshChanged()` call is apparently needed for the undisplaced mesh because of adaptivity, so I thought it might also be needed to keep the displaced mesh in sync. If it's not needed I'll get rid of it.

Moving the call to `clearQuadratureNodes()` into `MooseMesh::meshChanged()` broke a test (geomsearch/quadrature_locator_plus_constraint.quadrature_locator_plus_constraint). Moving it there made it so it gets called on the undisplaced mesh, which wasn't happening before. Maybe the test doesn't re-create those quadrature nodes, which would be necessary to do if they are deleted? It seems odd to me that they were only being deleted on the displaced mesh.

bwspenc: I'm calling it a day on this. That test that I added doesn't actually trigger the segfault without the changes. I've been working on it for quite a while trying to figure out what it's missing that's in the actual problem that got me started on this. It's really not that different, so I'm kind of stumped. Maybe it's not worth spending too much more time on that though -- at least the test case verifies that contact and subdomain modifiers can coexist.

hugary1995:
Given the test case I believe this might be related to #30630 and #31306

framework/src/problems/FEProblemBase.C
lindsayad: > so I thought it might also be needed to keep the displaced mesh in sync

There is a `meshChanged` call on the displaced mesh in `DisplacedProblem::meshChanged` that should cover what your second add here is doing

bwspenc:
I confirmed that the issue this was attempting to address was fixed by #30630, so I'm closing this PR.

PR #31303, opened 2025-08-14
Stm web server control

### LINKED ISSUE #31171: Enable WebServerControl of stochastic tools apps

zachmprince:
Closes #31171

Some additional features enabled from this effort:

- batch-restore no longer resets between timesteps
- InputMatrix is now controllable
- StochasticReporters can be executed on `EXEC_TRANSFER` to make sure it has the proper size when the transfer is inserting values and gather on ROOT earlier.
- The python in stochastic tools is reorganized to look like an actual python module.
- Previous STM python was actually never tested, now it is.


modules/stochastic_tools/doc/content/modules/stochastic_tools/python/StochasticControl.md
```diff
@@ -0,0 +1,119 @@
...
+  ```python
+  with StochasticControl(...) as runner:
+      runner.configCache()
+
+      y1 = runner([0, 1, 2])              # Runs STM input for the first time
+      y2 = runner([0, 1, 2])              # Does not run input and returns y1
+      y3 = runner([[0, 1, 2], [3, 4, 5]]) # Only runs input for [3, 4, 5]
+  ```
+
+- Use `min_procs_per_sample` if each sample itself needs multiple processes (e.g., heavy physics solves requiring a lot of memory).
```
loganharbour:
```suggestion
- Set `min_procs_per_sample` in `StochasticRunOptions` if each sample itself needs multiple processes (e.g., heavy physics solves requiring a lot of memory).
```

modules/stochastic_tools/doc/content/modules/stochastic_tools/python/StochasticControl.md
```diff
@@ -0,0 +1,119 @@
...
+
+## API Reference
+
+!alert! construction
+
+The following requires MooseDocs to add `stochastic_tools` python directory into the system path:
```
loganharbour, line 109: Should we be systematically adding all of the python paths for each application at moosedocs runtime?

modules/stochastic_tools/doc/content/modules/stochastic_tools/index.md
```diff
@@ -29,6 +29,10 @@ as well. However, to enable it one needs to compile MOOSE with the C++ APIs of
 
 - [enable_pytorch.md]
 
+## Python Utilities
+
+- [StochasticControl.md]
```
loganharbour, line 34: Probably should add a few words as to what it is?

modules/stochastic_tools/examples/optimization/annulus_shape/test_optimize_annulus.py
```diff
@@ -0,0 +1,35 @@
...
+
+    args = cli_args()
+
+    cmd = os.environ.get("RUNAPP_COMMAND")
+    if not cmd is None:
+        cmd = cmd.split()
+        # Gather MPI options if running with MPI
+        if "-n" in cmd:
+            mpi_index = cmd.index("-n") - 1
+            args.num_procs = int(cmd[mpi_index + 2])
```
loganharbour: Where is num procs used outside of the command?

modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
```diff
@@ -0,0 +1,625 @@
...
+
+class StochasticControl(MooseControl):
+    """
+    A MOOSE control interface for executing stochastic sampling runs
+    using the StochasticTools module.
+
+    This class automates the setup of input files, execution, and retrieval
+    of results (quantities of interest, QoIs) based on sampled parameters.
+
+    Example Usage::
```
loganharbour, line 91:
```suggestion
    Example Usage:
```

modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
```diff
@@ -0,0 +1,625 @@
...
+            try:
+                waiting_flag = self.getWaitingFlag()
+            except:
+                waiting_flag = None
+            if not waiting_flag is None:
+                self.setTerminate()
+
+            # Don't finalize if we never started
+            if self._initialized:
+                self.finalize()
```
loganharbour, line 216: Can this logic not be somewhat recycled from the MooseControl?

modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
```diff
@@ -0,0 +1,625 @@
...
+        return num_params
+
+    @staticmethod
+    def _matrixToParameter(matrix: np.ndarray) -> str:
+        """Convert a numpy array to a MOOSE-input matrix."""
+        assert len(matrix.shape) == 2
+        return "; ".join([" ".join([f"{elem:.16g}" for elem in row]) for row in matrix])
+
+
+class _ResultCache:
```
loganharbour, line 467: This probably deserves some docstring love too

modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
```diff
@@ -0,0 +1,625 @@
...
+
+    def set(self, x: np.ndarray, y: np.ndarray):
+        k = self._key(x)
+        self._cache[k] = (x.copy(), y.copy())
+        self._cache.move_to_end(k)
+        if len(self._cache) > self.maxsize:
+            self._cache.popitem(False)
+
+
+class StochasticRunner:
```
loganharbour, line 578: Also missing some docstrings

modules/stochastic_tools/test/src/controls/InputMatrixControl.C
```diff
@@ -0,0 +1,54 @@
...
+{
+  const dof_id_type nrow = _nrow_function.value(_t);
+  const dof_id_type ncol = _ncol_function.value(_t);
+
+  RealEigenMatrix matrix(nrow, ncol);
+  for (const auto i : make_range(nrow))
+    for (const auto j : make_range(ncol))
+      matrix(i, j) = _sample_function.value(_t, Point(i, j));
+
+  _console << matrix << std::endl;
```
loganharbour:
```suggestion
```

modules/stochastic_tools/test/src/postprocessors/OptimizationTestFunction.C
```diff
@@ -0,0 +1,66 @@
...
+OptimizationTestFunction::getValue() const
+{
+  if (_function == "rosen")
+    return rosen(_x);
+  else if (_function == "eggholder")
+    return eggholder(_x);
+  else
+    paramError("function", "Unknown function: ", _function);
+
+  return 0.0;
```
loganharbour: `paramError` has the `[[ noreturn ]]` modifier so you shouldn't need to return something here

modules/stochastic_tools/test/src/postprocessors/OptimizationTestFunction.C
```diff
@@ -0,0 +1,66 @@
...
+
+PostprocessorValue
+OptimizationTestFunction::getValue() const
+{
+  if (_function == "rosen")
+    return rosen(_x);
+  else if (_function == "eggholder")
+    return eggholder(_x);
+  else
+    paramError("function", "Unknown function: ", _function);
```
loganharbour: Should have this check in the constructor, no?

modules/stochastic_tools/test/src/postprocessors/OptimizationTestFunction.C
```diff
@@ -0,0 +1,66 @@
...
+
+Real
+OptimizationTestFunction::rosen(const std::vector<Real> & x)
+{
+  if (x.size() < 2)
+    return 0.0;
+
+  PostprocessorValue sum = 0.0;
+  for (const auto i : make_range(x.size() - 1))
+    sum += 100.0 * Utility::pow<2>(x[i + 1] - x[i] * x[i]) + Utility::pow<2>(1 - x[i]);
```
loganharbour: No `std::accumulate`? :D 

Also:

```
for (const auto i : index_range(x))
```

instead

loganharbour (commented):
Have only looked through the control components

modules/stochastic_tools/doc/content/modules/stochastic_tools/index.md
zachmprince: Did you look at the rendered docs? What words would you recommend?
image

modules/stochastic_tools/examples/optimization/annulus_shape/test_optimize_annulus.py
zachmprince: So for integration testing of the control, I use the `RUNAPP_COMMAND` to gather how many processors are being tested with. This is set in the `StochasticRunOptions`, which `StochasticControl` uses to build the `_moose_command` sent to `MooseControl`. So, in essence, the `num_procs` is only used for building the command that `MooseControl` uses to spawn the process.

modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
zachmprince: This is actually a python docstring trick. The double colon indicates that the next set of indented lines should be rendered as code.

modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
zachmprince: MooseControl will throw if it tries to finalize when the process is waiting. For this application, we want to terminate if the stochastic simulation is finished (the process will still be waiting) and then finalize.

modules/stochastic_tools/test/src/postprocessors/OptimizationTestFunction.C
zachmprince: Is there a way to `accumulate` when the function considers two neighboring elements? Also, the loop is from 1 to N-1, so I don't think `index_range` can be used unless I break on the last element.

modules/stochastic_tools/test/src/postprocessors/OptimizationTestFunction.C
zachmprince: It's a MooseEnum, show it should error during parsing. I really will never be hit, anyway.

modules/stochastic_tools/doc/content/modules/stochastic_tools/examples/annulus_shape.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Problem Description
+
+In this problem, we focus on an annular (ring-shaped) domain subjected to thermal conditions. The domain is defined by two concentric circles with an inner radius and thickness. The physical problem is heat conduction governed by the heat equation with convective flux type boundary conditions at the inner radius and an insulated outer radius. The convective heat transfer coefficient is dependent on the inner radius. A constant source term is distributed throughout the domain.
+
+The optimization goal is to find the inner radius and thickness that minimize the maximum temperature within the domain while maintaining a fixed volume of the annulus. In other words, the design variables are the inner radius and thickness, and the objective function is the maximum temperature within the annulus. This is a constrained optimization problem because the volume of the annulus needs to remain constant during the optimization process.
```
lynnmunday, line 9: Can you reference the optimization example write-up that solves this same problem:
https://mooseframework.inl.gov/modules/optimization/examples/shapeOpt_Annulus.html
And add a link for the optimization module example to this.  
Or if you can't reference, can you mention there is also a test in the optimization module for constrained optimization problem and vice versa.

modules/stochastic_tools/doc/content/modules/stochastic_tools/examples/annulus_shape.md
```diff
@@ -0,0 +1,134 @@
+# Annulus Shape Optimization — SciPy + `StochasticControl`
+
+This example demonstrates how to couple [SciPy optimization](https://docs.scipy.org/doc/scipy/reference/optimize.html) routines with [StochasticControl](StochasticControl.md) to minimize the maximum temperature in a 2D annular geometry while enforcing a constant volume constraint. It is a practical pattern for black‑box optimization where each function evaluation runs a MOOSE simulation and returns quantities of interest (QoIs).
+
+## Problem Description
+
+In this problem, we focus on an annular (ring-shaped) domain subjected to thermal conditions. The domain is defined by two concentric circles with an inner radius and thickness. The physical problem is heat conduction governed by the heat equation with convective flux type boundary conditions at the inner radius and an insulated outer radius. The convective heat transfer coefficient is dependent on the inner radius. A constant source term is distributed throughout the domain.
```
grmnptr:
```suggestion
In this problem, we focus on an annular (ring-shaped) domain subjected to thermal conditions. The domain is defined by two concentric circles with an inner radius and thickness. The physical problem is heat conduction governed by the heat equation with convective flux type boundary condition at the inner radius and an insulated outer radius. The convective heat transfer coefficient is dependent on the inner radius. A constant source term is distributed throughout the domain.
```

modules/stochastic_tools/doc/content/modules/stochastic_tools/examples/annulus_shape.md
```diff
@@ -0,0 +1,134 @@
...
+
+The optimization goal is to find the inner radius and thickness that minimize the maximum temperature within the domain while maintaining a fixed volume of the annulus. In other words, the design variables are the inner radius and thickness, and the objective function is the maximum temperature within the annulus. This is a constrained optimization problem because the volume of the annulus needs to remain constant during the optimization process.
+
+This example problem is identical to the [shape optimization example](optimization/examples/shapeOpt_Annulus.md optional=True) in the [Optimization module](modules/optimization/index.md optional=True). The optimization module version utilizes gradients for its minimization solve, which makes it a faster method, but is more involved to formulate.
+
+## Physics Inputs
+
+There are two MOOSE inputs representing the physics. The first is utilizes input variables for the inner radius and thickness to directly change the mesh that is generated. The other uses mesh displacement to change the domain of the ring.
```
grmnptr:
```suggestion
There are two MOOSE inputs representing the physics. The first utilizes input variables for the inner radius and thickness to directly change the mesh that is generated. The other uses mesh displacement to change the domain of the ring.
```

modules/stochastic_tools/doc/content/modules/stochastic_tools/examples/annulus_shape.md
```diff
@@ -0,0 +1,134 @@
...
+
+!listing annulus_shape/optimize_annulus.py start=def optimize_annulus end=#
+
+### Configuring the StochasticControl
+
+The parameters to control are dependent on which input file is being utilized:
+Next, the postprocessors defining the objective and constraints of the optimization are specified as QoIs of the simulation:
+A `StochasticRunOptions` is then built, to be given as options for the `StochasticControl`.
+The `StochasticControl` is created as a context manager which returns a `StochasticRunner` object.
+Since the optimizers often run objective and constraint evaluation separately, but with the same sample, it is recommended to cache the input-output pairs using `configCache`. Finally, somewhat arbitrary bounds for the parameters is defined.
```
grmnptr:
```suggestion
Since the optimizers often run objective and constraint evaluation separately, but with the same sample, it is recommended to cache the input-output pairs using `configCache`. Finally, somewhat arbitrary bounds for the parameters are defined.
```

modules/stochastic_tools/doc/content/modules/stochastic_tools/python/StochasticControl.md
```diff
@@ -0,0 +1,119 @@
...
+parameters = ["Materials/K/value",
+              "Kernels/convection/alpha",
+              "beta"]                     # Parameters to sample (names as seen by the app)
+qois = ["pp/value"]                       # Reporter(s) to collect as QoIs
+
+# Configure how the stochastic run is executed
+opts = StochasticRunOptions(
+    num_procs=4,                 # mpiexec -n 4 ...
+    mpi_command="mpiexec",
+    input_name="stochastic_run.i",
```
grmnptr: Might be beneficial adding a `#` comment here as well to make sure everyone understands that this is the generated input name. You discuss it in the following sections but the reader hasn't gotten there yet.

modules/stochastic_tools/src/samplers/InputMatrixSampler.C
```diff
@@ -27,6 +28,20 @@ InputMatrixSampler::InputMatrixSampler(const InputParameters & parameters)
   setNumberOfCols(_data.cols());
 }
 
+void
+InputMatrixSampler::executeSetUp()
+{
+  if ((dof_id_type)_data.cols() != getNumberOfCols())
+    paramError("matrix",
+               "Inputted matrix as changed the number of columns from ",
```
grmnptr:
```suggestion
               "Input matrix changed the number of columns from ",
```

modules/stochastic_tools/test/include/postprocessors/OptimizationTestFunction.h
```diff
@@ -0,0 +1,33 @@
...
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+#pragma once
+
+#include "GeneralPostprocessor.h"
+
+class OptimizationTestFunction : public GeneralPostprocessor
```
grmnptr, line 13: I know a lot of people are interested in these functions (Dewen was, Som as well, etc). 
I think Som actually tried to add the Rosenbrock function at some point. 
Do you think we could split this function into two different functions and move them to STM from test?
Actually I am fine having it in the test but I still think we should split this in 2 different functions instead of doing the if-else in here.

grmnptr (changes requested):
Very nice! Almost optimal!

modules/stochastic_tools/test/include/postprocessors/OptimizationTestFunction.h
zachmprince: Made two different function classes with this class being the base.

grmnptr (approved):
Looks good to me! Very cool work.

Issue #31234, opened 2025-08-06
Add materials for deviatoric and volumetric strain increments due to vacancy concentration change


sourabhkadambi:
### Motivation

The existing [FluxBasedStrainIncrement](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/FluxBasedStrainIncrement.C#L4) material object describes strain increment as the symmetric part of vacancy flux gradient tensor. However, to simulate shape changes (stretch and shear), only the deviatoric part of the tensor needs to be considered. Volume change due to the hydrostatic part needs to be considered separately and scaled by a lattice relaxation/ expansion coefficient. Additional volume change due to creation/ removal of lattice sites due to sources/ sinks needs to be described.

### Design

- Create a material inheriting from [FluxBasedStrainIncrement](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/FluxBasedStrainIncrement.C#L4) that subtracts the hydrostatic part of tensor to yield the deviatoric part.
- Create a new material that accounts for the volumetric strain increments under both lattice site conserved and non-conserved conditions.
- The deviatoric and volumetric parts can be added together using the existing [SumTensorIncrements](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/SumTensorIncrements.C#L4) object.

### Impact

The new capability will help simulate bulk deformation of material under diffusion creep and volume swelling under irradiation. It will be backward compatible and will not break existing tests but will add new options for users to specify the treatment of the deviatoric and hydrostatic parts of the tensor separately.

PR #31307, opened 2025-08-15
Added materials for deviatoric and volumetric strain increments due to vacancy concentration change

YKumar1999:
refs #31234 

## Reason
MOOSE currently implements [FluxBasedStrainIncrement](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/FluxBasedStrainIncrement.C#L4) to compute strain rate as a function of the symmetric part of the vacancy flux gradient tensor. However, to simulate stretch/shear phenomena in materials, the hydrostatic (volumetric part) needs to be subtracted from the total strain to obtain the deviatoric component. The volumetric part is considered separately in a different material that accounts for volume changes due to lattice site creation/removal and vacancy redistribution (diffusion-based processes). 

## Design
1. Firstly, a new material "**DeviatoricStrainIncrement**" is inherited from [FluxBasedStrainIncrement](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/FluxBasedStrainIncrement.C#L4) where the hydrostatic strain (correction) is subtracted from the total strain as described by [FluxBasedStrainIncrement](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/FluxBasedStrainIncrement.C#L4). 
2. Another, material "**VolumeStrainIncrement**" has been created which is inherited again from [FluxBasedStrainIncrement](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/FluxBasedStrainIncrement.C#L4), where **only the volumetric strain is considered (combination of divergence of fluxes and production rate of vacancies, both scaled by relaxation factors provided as a material property, in this material, deviatoric part of strain rate is not added)**. 
3. Both of these materials "**DeviatoricStrainIncrement**" and "**VolumeStrainIncrement**" can then be added together using the existing material [SumTensorIncrements](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/SumTensorIncrements.C#L4) in MOOSE.
4. Additionally, a small change has been made in [FluxBasedStrainIncrement](https://github.com/idaholab/moose/blob/50dfe9fc4d0a6f4c6f43f7cffa8fa053699fd962/modules/solid_mechanics/src/materials/FluxBasedStrainIncrement.C#L4), where a function for computing **Identity Tensor** has been added, but it will not affect the existing tests. This has been added to support the two new materials **DeviatoricStrainIncrement** and **VolumeStrainIncrement** which uses IdentityTensor for isotropic case. 

## Impact
These new materials added for deviatoric and volumetric strain increments will be beneficial for simulating bulk deformation in materials under diffusion creep. Additionally, local volume changes caused by dense dislocations (acting as sinks/sources) can also be studied explicitly. The current materials will not break the existing tests; new options are provided to users for accounting for volume changes due to lattice site creation/removal or lattice relaxation (local changes in concentration) separately.

@sourabhkadambi @dschwen


hugary1995:
Does this fit inside the solid_mechanics module? I feel this fits better in some more specialized moose-based application.

YKumar1999:
> Does this fit inside the solid_mechanics module? I feel this fits better in some more specialized moose-based application.

I think it should be in solid-mechanics module itself because here we have inherited both of our classes from FluxBasedStrainIncrement and these new materials are just providing a newer way to study diffusion creep, they are not entirely a different class. 

But yes if there is a better way, i am open for it. thanks for your suggestion

hugary1995:
> I think it should be in solid-mechanics module itself because here we have inherited both of our classes from FluxBasedStrainIncrement and these new materials are just providing a newer way to study diffusion creep, they are not entirely a different class.

Maybe those other classes that inherit from `FluxBasedStrainIncrement` should be moved to a separete app.

modules/combined/test/tests/DiffuseCreep/Test_VolumeChange.i
```diff
@@ -0,0 +1,260 @@
...
+  scheme = bdf2
+  solve_type = NEWTON
+  petsc_options_iname = '-pc_type'
+  petsc_options_value = 'lu'
+  # petsc_options_iname = '-pc_type -pc_factor_mat_solver_package'
+  # petsc_options_value = 'lu superlu_dist'
+  # petsc_options_iname = '-pc_type -pc_hypre_type -ksp_gmres_restart -pc_hypre_boomeramg_strong_threshold'
+  # petsc_options_value = 'hypre    boomeramg      31                 0.7'
+  # petsc_options_iname = '-pc_type -ksp_grmres_restart -sub_ksp_type -sub_pc_type -pc_asm_overlap'
+  # petsc_options_value = 'asm      31                  preonly       lu           2'
```
sourabhkadambi: remove lines you're not using in the executioner

modules/combined/test/tests/DiffuseCreep/Test_VolumeChange.i
```diff
@@ -0,0 +1,260 @@
...
+
+  line_search = 'none' # SK
+  automatic_scaling = false # SK
+  nl_forced_its = 2 # SK
+
+  # [Adaptivity]
+  #   max_h_level = 2
+  #   refine_fraction = 0.3
+  #   coarsen_fraction = 0.2
+  # []
```
sourabhkadambi: cleanup similar unused lines in input files

hugary1995 (changes requested):
placeholder. I'm concerned about further inflating solid mechanics.

modules/combined/test/tests/DiffuseCreep/Test_Deviatoric.i
```diff
@@ -0,0 +1,261 @@
+#This test demonstrates the volume change due to gradient in vacancy flux
+#For this test, deviatoric part of strain rate is considered by substracting out the hydrostatic part from total creep strain tensor
+#Equal fluxes of vacancy are considered for the test, with diffusivity value of 1 (imposed by the factor option in GenericGradientComponent kernel)
+
+[Mesh]
+  type = GeneratedMesh
+  dim = 2
+  nx = 100
+  ny = 100
```
sourabhkadambi: reduce mesh elements

modules/solid_mechanics/src/materials/DeviatoricStrainIncrement.C
sourabhkadambi, line 1: is it better to call it DeviatoricFluxBasedStrainIncrement?

YKumar1999: isn't it too long name ?... although reading it makes more sense.

YKumar1999: Name changed from DeviatoricStrainIncrement to DeviatoricFluxBasedStrainIncrement

bwspenc:
I have similar concerns -- this seems a little to application-specific for solid_mechanics. The same thing could probably be said for FluxBasedStrainIncrement as well!

YKumar1999:
@moosebuild could you elaborate the coverage error I have to address and also for the previous documentation failed check ?

hugary1995:
moosebuild is a bot. If you click on the failed check, you can see the complete log with error messages. In your case, the new models you added are missing markdown documentation.

@bwspenc and I have both left comments regarding the appropriate ownership of your material models. They are high level comments that should be addressed first. I think keep working on the code without resolving the concern isn't very respectful.

YKumar1999:
Yes that makes sense... actually I got a gist of what you @hugary1995  and @bwspenc commented about the solid mechanics module inflating thing that this Deviatoric and Volumetric based material properties are bit specific to studying volume changes due to diffusion creep, and it has to go in a separate application or something ?....do you have any suggestions or more elaborative comment exactly what application we can target to add these materials into ?

hugary1995:
I'm sure @sourabhkadambi and @dschwen have suggestions for you regarding which app this belongs to. If there isn't already an app suitable for this, you can always [create one](https://mooseframework.inl.gov/getting_started/new_users.html).

YKumar1999:
@hugary1995 Yeah we think as of now it should go well into MARMOT.

PR #31311, opened 2025-08-15
Add `--only-tests-that-require` to TestHarness

### LINKED ISSUE #31310: Enable filtering of tests by if they depend on a capability
### Motivation

Take the "libtorch" capability as an example. In certain cases, we want to test _just_ the libtorch tests. Because those tests explicitly have `libtorch` in their capabilities string, we should be able to filter and find the tests that have a dependency on libtorch. This should be a test harness option.

### Design

Add a `--require-capability` option that filters tests whether they depend on a capability.

### Impact

New feature.

loganharbour:
Closes https://github.com/idaholab/moose/issues/31310

This lets you pass `--required-capability foo` to only run tests that explicitly depend on the capability `foo` given the current configuration. You can also use multiple, i.e., `--only-tests-that-require foo --only-tests-that-require !bar`.

Thus, you can do something like:

```
./run_tests --only-tests-that-require cuda
```

to only run the tests that depend on cuda.

### DIFF
--- a/modules/doc/content/newsletter/2025/2025_09.md
+++ b/modules/doc/content/newsletter/2025/2025_09.md
@@ -7,6 +7,27 @@ for a complete description of all MOOSE changes.
 
 ## MOOSE Improvements
 
+### TestHarness `--only-tests-that-require` option
+
+The `capability` parameter for test specifications allows for run-time checking of if a test can or cannot run based on capabilities defined by the application executable. The `--only-tests-that-require` command line option has been added to the [TestHarness.md], which allows filtering of tests based on whether or not they depend on a capability.
+
+For example, take a situation in which `./run_tests --only-tests-that-require cuda` is ran along with the following test specification:
+
+```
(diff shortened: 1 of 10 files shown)


dschwen (approved):
I was yet again confused by the parameter name, but don't have a much better suggestion other than maybe `--only-tests-that-require`

loganharbour:
@lindsayad what do you think? See the PR header for info.

lindsayad:
tbh it's long but I kind of like @dschwen's suggestion

loganharbour:
Fine with me. I don't have a strong feeling either way.

loganharbour:
Renamed `--require-capability` to `--only-tests-that-require` in https://github.com/idaholab/moose/pull/31311/commits/fead0bcdea8091cb179b69a592d9ece7ff7c1a8f and added a newsletter entry.

dschwen (approved):

PR #31319, opened 2025-08-16
Add `sfcurves` capability

dschwen:
## Reason
Need to be able to guard parallel tests that need a space filling curve partitioner

## Design
Add `sfcurves` capability.

## Impact
Improved test trequirements.


loganharbour (approved):
Can you figure out what recipes that we need to add this to the capabilities checks after the build

PR #31343, opened 2025-08-21
changing type of PorousFlowPorosity.solid_bulk to Functor

### LINKED ISSUE #31342: Making PorousFlowPorosity.solid_bulk a functor
### Motivation

PorousFlowPorosity.solid_bulk` is currently of type `Real`. I have a model where the stiffness of my soil skeleton softens significantly. Due to this I'd like to reduce `PorousFlowPorosity.solid_bulk` over time.

### Design

Motivated by #31336 I'd make a PR changing the type of `PorousFlowPorosity.solid_bulk` to `MooseFunctorName`.

### Impact

Current functionality is maintained. All existing input files should not be affected.

jmeier:
closes #31342

### Motivation

`PorousFlowPorosity.solid_bulk` is currently of type `Real`. I have a model where the stiffness of my soil skeleton softens significantly. Due to this I'd like to reduce `PorousFlowPorosity.solid_bulk` over time.

### Design

Motivated by #31336 I'd make a PR changing the type of `PorousFlowPorosity.solid_bulk` to `MooseFunctorName`.

### Impact

Current functionality is maintained. All existing input files should not be affected.


jmeier:
Dear Moose Community,

There is one point I'd like to request special attention during your review: I call the functor using `elem_arg = makeElemArg(_current_elem)` - ignoring the current `_qp`. Consequently, the effective solid_bulk in `decayQp` and `ddecayQp_dvar` will be a constant for one and the same element (spatially; but not in time). It is not clear to me, if this is a problem.

jmeier:
For the parameter "solid_bulk" `Unit:(no unit assumed)` is shown in the docs. Should we change this to "Pressure"? If yes, how?
image

cpgr:
> Dear Moose Community,
> 
> There is one point I'd like to request special attention during your review: I call the functor using `elem_arg = makeElemArg(_current_elem)` - ignoring the current `_qp`. Consequently, the effective solid_bulk in `decayQp` and `ddecayQp_dvar` will be a constant for one and the same element (spatially; but not in time). It is not clear to me, if this is a problem.

Any reason in particular for this?

jmeier:
> > ... ignoring the current `_qp` ...
> 
> Any reason in particular for this?

Please advice how to do it for `_qp`.

Would be the following the correct argument?
`const Moose::ElemQpArg qp_arg = {_current_elem, _qp, _qrule, _q_point[_qp]};`

cpgr:
> > > ... ignoring the current `_qp` ...
> > 
> > 
> > Any reason in particular for this?
> 
> Please advice how to do it for `_qp`.
> 
> Would be the following the correct argument? `const Moose::ElemQpArg qp_arg = {_current_elem, _qp, _qrule, _q_point[_qp]};`

I think that is correct. The simplest example is probably in FunctorKernel
```
  const Moose::ElemQpArg space_arg = {_current_elem, _qp, _qrule, _q_point[_qp]};
  return _sign * _functor(space_arg, Moose::currentState());
```

jmeier:
> I think that is correct. The simplest example is probably in FunctorKernel
> 
> ```
>   const Moose::ElemQpArg space_arg = {_current_elem, _qp, _qrule, _q_point[_qp]};
>   return _sign * _functor(space_arg, Moose::currentState());
> ```

Thanks! I've updated the code of this PR accordingly.

jmeier:
@cpgr, may I ask if the functor call is now correct?

cpgr:
I think that is the right way, but it seems that there is a real failure in the modules debug tests. Not sure what the bug is?

GiudGiud:
@jmeier I can take a look if needed

jmeier:
> @jmeier I can take a look if needed

@GiudGiud, this would be great. For example, `test:poro_elasticity.terzaghi` is "OK" on my (local) machine - but fails in "Civet: Modules debug".

GiudGiud:
Ok I see it locally with a Debug executable. 

The first thing that comes to mind is that you use an ElemQpArg but a boundary material or an interface material would use ElemSideQpArg but it might not be that. 

I ll keep digging

jmeier:
> Ok I see it locally with a Debug executable.

Oh - a dbg executable also gives an error: `Assertion 'i < _size' failed`. 

btw: `ElemQpArg` comes from [here](https://github.com/idaholab/moose/pull/31343#issuecomment-3215887014)

I hope your digging points me to the right direction.

GiudGiud:
> btw: ElemQpArg comes from https://github.com/idaholab/moose/pull/31343#issuecomment-3215887014

yup was correct in the context of a kernel but materials can be computed on sides

GiudGiud:
so while we don't need boundary materials here (I think? BCs are not using materials), we are still re-initing them for stateful properties. I don't see why so I'll try to change that

ok this is why that happens. that s not really the complete picture
```
bool
NonlinearSystemBase::needBoundaryMaterialOnSide(BoundaryID bnd_id, THREAD_ID tid) const
{
  return _integrated_bcs.hasActiveBoundaryObjects(bnd_id, tid);
}
```
not our solution but something to improve

cpgr:
Thanks for looking into this @GiudGiud

modules/porous_flow/src/materials/PorousFlowMaterial.C
```diff
@@ -161,7 +161,7 @@ PorousFlowMaterial::nearestQP(unsigned nodenum) const
 {
   unsigned nearest_qp = 0;
   Real smallest_dist = std::numeric_limits<Real>::max();
-  for (unsigned qp = 1; qp < _qrule->n_points(); ++qp)
+  for (const auto qp : make_range(_qrule->n_points()))
```
GiudGiud, line 164: this should be fine righg @cpgr

modules/porous_flow/src/materials/PorousFlowPorosity.C
```diff
@@ -286,7 +307,32 @@ PorousFlowPorosity::ddecayQp_dvar(unsigned pvar) const
...
+    {
+      const std::set<SubdomainID> subdomain_set = {_current_elem->subdomain_id()};
+      const Moose::NodeArg space_arg = {_current_elem->node_ptr(qp_used), &subdomain_set};
+      solid_bulk = (*_solid_bulk)(space_arg, Moose::currentState());
+    }
+    else if (_bnd)
+    {
+      const Moose::ElemSideQpArg space_arg = {
+          _current_elem, _current_side, qp_used, _qrule, _q_point[qp_used]};
+      solid_bulk = (*_solid_bulk)(space_arg, Moose::currentState());
```
GiudGiud, line 325: just a note that if dsollid_bulk/dvar !=0 you are not capturing that here
since it s not an AD kernel you could not use an AD functor either

modules/porous_flow/src/materials/PorousFlowMaterial.C
cpgr: I'm actually not sure why the old one was starting at 1? At first look what you have seems correct

GiudGiud: yeah me neither. looks better to me now.

I think the intent of the old one was to have the `  unsigned nearest_qp = 0;` handle the qp=0 case but it was not finished

modules/porous_flow/src/materials/PorousFlowPorosity.C
jmeier: I'm not sure what I should do here.

modules/porous_flow/src/materials/PorousFlowMaterial.C
jmeier: I'm not sure what I should do here. Click "resolve"?

GiudGiud: I can resolve before merging

modules/porous_flow/src/materials/PorousFlowPorosity.C
GiudGiud: well in your use case, is it going to depend on pp or any other variable?
if so you should consider having the object request the d_solidbulk_dvar functor and use it to form the contribution the Jacobian
or try to live with an inexact Jacobian but keep it in mind if you ever hit convergence issues

jmeier: In my intended use case, solid-bulk only depends on time and varies in space. No dependence on other variables. So, is this use case all right then?

However, I think it would be a good idea to add a note to the documentation. Ok?

GiudGiud: note in the documentation works for me.

GiudGiud (approved):

Issue #31344, opened 2025-08-21
Perf graph live can create data races


lindsayad:
### Bug Description

We have a test failing occasionally in CI and running with thread sanitizer I see
```
WARNING: ThreadSanitizer: data race (pid=1223)
  Write of size 8 at 0x729000008000 by main thread:
    #0 memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115 (libtsan.so.2+0x8bd30) (BuildId: 38097064631f7912bd33117a9c83d08b42e15571)
    #1 memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:107 (libtsan.so.2+0x8bd30)
    #2 std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) <null> (libstdc++.so.6+0x166e5f) (BuildId: ca77dae775ec87540acd7218fa990c40d1c94ab1)
    #3 void moose::internal::mooseDeprecatedStream<libMesh::BasicOStreamProxy<char, std::char_traits<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(libMesh::BasicOStreamProxy<char, std::char_traits<char> >&, bool, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&) /data/lindad/projects/thread-debugging/framework/build/header_symlinks/MooseError.h:286 (libmoose-dbg.so.0+0x69aa57f) (BuildId: bba36506d1cc6776f3db154c3c8091d2d5812e6c)
    #4 void mooseDeprecated<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&) /data/lindad/projects/thread-debugging/framework/build/header_symlinks/MooseError.h:386 (libmoose-dbg.so.0+0x696ac85) (BuildId: bba36506d1cc6776f3db154c3c8091d2d5812e6c)
[...]

PR #31345, opened 2025-08-21
Disable perf graph live in the test harness by default

lindsayad:
Refs #31344


YaqiWang:
Maybe also disable screen coloring?

roystgnr (approved):

lindsayad:
I thought about that but I do think coloring can be a way to help people quickly find test failures when they're scrolling in their browser. Whereas perf graph live has zero value in terms of improving interaction with the test harness

lindsayad:
But I suppose since we can obviously have multiple threads writing to output even when perf graph is disabled, turning off color by default is the right thing to do at least for CIVET

lindsayad:
I opened up https://github.com/idaholab/civet_recipes/pull/2450 to disable coloring any time we're running the test harness with multiple threads

lindsayad:
failures are unrelated

loganharbour:
> Whereas perf graph live has zero value in terms of improving interaction with the test harness

I don't know if I agree with that actually. We don't have any other way of capturing memory usage when someone foolishly adds in a test that causes an OOM event

lindsayad:
that's fair. I still think it's generally more harmful than helpful

lindsayad:
failures are unrelated

PR #31350, opened 2025-08-21
SP3 radiation

ahnsininl:
Closes #31346.


ahnsininl:
@tanoret Made Draft PR. Docs WIP.

modules/heat_transfer/include/base/HeatConductionNames.h
```diff
@@ -19,5 +19,8 @@ static const std::string T_ambient = "T_ambient";
 namespace Constants
 {
 static const Real sigma = 5.670374419e-8;
+static const Real hp = 6.62607015e-34; // J.s
+static const Real c = 2.99792458e8;    // m/s
+static const Real kb = 1.380649e-23;   // J/K
```
hugary1995: Change them to `constexpr`s.

hugary1995: I'm generally concerned about adding constants with hard-coded units into an upstream location like this. I glanced at the places where you are using them in the source code, in some places the units cancel out and become dimensionless, which is fine, but in some places the units are not cancelled out. The simplest example is `kb * T` and so you are assuming temperature is in the unit of Kelvin -- I know that's pretty common, but there might be places where people want to use Celsius. Alternatively, you can just add these constants with hard-coded units into specific objects where you are using them. Then you can explicitly mention this in the class description, something like "this object only works with temperature in the unit of Kelvin".

hugary1995 (dismissed):

YaqiWang:
Why do you choose FV instead of FEM? The input of the test is very long. It should be a good idea to add custom actions for adding the system. Griffin has a thermal radiation transport capability, we could either separate it from Griffin or merge this into Griffin.

modules/heat_transfer/include/base/HeatConductionNames.h
ahnsininl: 1. Changed into `constexpr`.
2. Regarding your concern, I understand that user could be confused about units of constants. For the 3 constant I added (Plank constant, Speed of light, Boltzmann constant),  they are used in several kernels, and MOOSE already provides similar constants such as Stefan-Boltzmann constant `HeatConductionName::Constants::sigma` `PhysicalConstants::avogadro_number` and `PhysicalConstants::avogadro_number`.
3. To avoid confusion, I added unit descriptions of these constants in the Markdown file, and also included comments with the unit names directly in the code. Updated documents (`FVSP3TemperatureBC`) also include specific units for the constants.

hugary1995: Thanks for changing them to constexpr!

I'm glad you brought up the other already existing constants. We should come up with a fix for them as well. I think however it's unfair to ask you to fix the ones that you didn't introduce. But we shouldn't continue the bad practice.

I'll help you think about this...

hugary1995: How about something like this?

```cpp
#include "Units.h"

namespace HeatConduction::Constants
{
/**
 * @brief Get the Stefan-Boltzmann constant.
 *
 * @param unit The unit to return the constant in (optional). Default to "W/m^2/K^4"
 * @return Real The Stefan-Boltzmann constant in the specified unit.
 */
Real
StefanBoltzmann(const std::optional<std::string> & unit)
{
  constexpr Real v = 5.670374419e-8;

  if (!unit)
    return v; // W/m^2/K^4

  return MooseUnits(unit.value()).convert(v, MooseUnits("W/m^2/K^4"));
}
}
```

You can then replace `HeatConduction::Constants::sigma` with `HeatConduction::Constants::StefanBoltzmann()` (for default SI units) or `HeatConduction::Constants::StefanBoltzmann("kW/mm^2/K^4")`. You can do similar things for other constants that you introduce.

hugary1995: @dschwen As a related note, it would be nice to make MooseUnits `constexpr`s. I think it's technically possible with `string_view`s.

hugary1995: @ahnsininl You can even allow user to specify the units for the constants, i.e.
```cpp
params.addParam<std::string>("stefan_boltzmann_units", "W/m^2/K^4", "Units for the Stefan-Boltzmann constant");
```
I think this would make your radiation objects much more flexible. But it's up to you to decide whether to adopt this.

ahnsininl: Thank you for your suggestion
1. I've updated the constants in `HeatConduction::Constants` and added optinal parameters for unit conversion in my kernels. 
2. While reviewing MOOSE, I noticed that constants are defined in various ways across the framework. Unifying these would require a broader effort beyond the scope of this PR. For now, I've focused on addressing the current issue.

hugary1995: Thanks for the change. This looks great to me. I'm marking this as resolved and unsubscribing now.

ahnsininl:
> Why do you choose FV instead of FEM? The input of the test is very long. It should be a good idea to add custom actions for adding the system. Griffin has a thermal radiation transport capability, we could either separate it from Griffin or merge this into Griffin.

Thank you for the thoughtful suggestions!
We implemented the radiation model with FVM because it is planned to be coupled with fluid dynamics. I agree that adding an Action would be useful for this module, but I considered it as future work rather than part of the current implementation.
Regarding Griffin, unfortunately I don't have access to it, so I couldn't explore integration options at this time.

GiudGiud:
Griffin does not have SPN so there is no overlap. 
This capability makes sense to me in the heat transfer module.

modules/heat_transfer/doc/content/source/fvbcs/FVSP3TemperatureBC.md
```diff
@@ -0,0 +1,44 @@
+# FVSP3TemperatureBC
+
+This boundary condition implements the Robin boundary conditions for the Temperature in Semi-Transparent boundary for the SP3 method.
+
+## Overview
+
+The Robin boudnary condition implemented reads as follows [!citep](larsen2002):
```
GiudGiud:
```suggestion
The Robin boundary condition implemented reads as follows [!citep](larsen2002):
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3ThermalRadiationBC.md
```diff
@@ -0,0 +1,31 @@
+# FVSP3ThermalRadiationBC
+
+This boundary condition implements the Robin boundary conditions for the radiative heat flux moments for the SP3 method.
+
+## Overview
+
+The Robin boudnary condition implemented reads as follows [!citep](larsen2002):
```
GiudGiud:
```suggestion
The Robin boundary condition implemented reads as follows [!citep](larsen2002):
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3ThermalRadiationBC.md
```diff
@@ -0,0 +1,31 @@
...
+
+where:
+
+- $\kappa$ is the absorpsivity of medium (SI unit ($m^{-1}$))
+- $\varepsilon$ is the optical thickenss of medium (SI unit (-))
+- $\alpha_i$ is the first term coefficient with idential order of flux moments (SI unit (-))
+- $\beta_j$ is the second term coefficient with coupled order of flux moments (SI unit (-))
+- $\eta_i$ is the third term coefficient with idential order of flux moments (SI unit (-))
+- $\nu_{i}$ is the lower boundary of radiation frequency band (SI units ($s^{-1}$))
+- $\nu_{i+1}$ is the upper boudnary of radiation frequency band (SI unit ($s^{-1}$))
```
GiudGiud:
```suggestion
- $\nu_{i+1}$ is the upper boundary of radiation frequency band (SI unit ($s^{-1}$))
```

modules/heat_transfer/doc/content/source/fvkernels/FVSP3ThermalRadiationSourceSink.md
```diff
@@ -0,0 +1,26 @@
...
+\begin{equation}
+{\kappa}({\psi_i}-{\int_{\nu_i}^{\nu_{i+1}}}4{\pi}Bd{\nu})
+\end{equation}
+
+where:
+
+- $\kappa$ is the optical thickenss of medium (SI unit ($m^{-1}$))
+- $\psi_i$ is the radiative heat flux moment of targeting order integrated in frequency band \[$\nu_i$, $\nu_{i+1}$\] (SI unit ($W m^{-2}$))
+- $\nu_{i}$ is the lower boundary of radiation frequency band (SI units ($s^{-1}$))
+- $\nu_{i+1}$ is the upper boudnary of radiation frequency band (SI unit ($s^{-1}$))
```
GiudGiud:
```suggestion
- $\nu_{i+1}$ is the upper boundary of radiation frequency band (SI unit ($s^{-1}$))
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3TemperatureBC.md
```diff
@@ -0,0 +1,44 @@
...
+\left[
+h\,(T_b - T) +
+\alpha \pi \left( \frac{n_2}{n_1} \right)^2
+\int_{0}^{\nu_1} \left( B(\nu, T_b) - B(\nu, T) \right)\, d\nu
+\right]
+\end{equation}
+
+where:
+
+- $\varepsilon$ is the optical thickenss of medium (SI unit (-))
```
GiudGiud:
```suggestion
- $\varepsilon$ is the optical thickness of medium (SI unit (-))
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3ThermalRadiationBC.md
```diff
@@ -0,0 +1,31 @@
...
+
+The Robin boudnary condition implemented reads as follows [!citep](larsen2002):
+\begin{equation}
+\frac{\kappa}{\varepsilon}(-\alpha_i\psi_i -\beta_j\psi_j + \int_{\nu_i}^{\nu_{i+1}}\eta_i{B_b}d\nu)
+\end{equation}
+
+where:
+
+- $\kappa$ is the absorpsivity of medium (SI unit ($m^{-1}$))
+- $\varepsilon$ is the optical thickenss of medium (SI unit (-))
```
GiudGiud:
```suggestion
- $\varepsilon$ is the optical thickness of medium (SI unit (-))
```

modules/heat_transfer/doc/content/source/fvkernels/FVSP3ThermalRadiationDiffusion.md
```diff
@@ -0,0 +1,23 @@
...
+## Overview
+
+The diffusion term implemented reads as follows [!citep](larsen2002):
+\begin{equation}
+-{\nabla}{\cdot}{\frac{{\varepsilon}^2{\mu_i}^2}{{\kappa}}}{\nabla}{\psi_i}
+\end{equation}
+
+where:
+
+- $\varepsilon$ is the optical thickenss of medium (SI unit (-))
```
GiudGiud:
```suggestion
- $\varepsilon$ is the optical thickness of medium (SI unit (-))
```

modules/heat_transfer/doc/content/source/fvkernels/FVSP3ThermalRadiationSourceSink.md
```diff
@@ -0,0 +1,26 @@
...
+## Overview
+
+The source and sink term implemented reads as follows [!citep](larsen2002):
+\begin{equation}
+{\kappa}({\psi_i}-{\int_{\nu_i}^{\nu_{i+1}}}4{\pi}Bd{\nu})
+\end{equation}
+
+where:
+
+- $\kappa$ is the optical thickenss of medium (SI unit ($m^{-1}$))
```
GiudGiud:
```suggestion
- $\kappa$ is the optical thickness of medium (SI unit ($m^{-1}$))
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3TemperatureBC.md
```diff
@@ -0,0 +1,44 @@
...
+- $T_b$ is the ambient temperature (SI unit (K))
+- $n_1$ is the refraction coefficient of medium 1 (SI unit (-))
+- $n_2$ is the refraction coefficient of medium 2 (SI unit (-))
+- $\alpha$ is the hemispheric emissivity of the medium (SI unit (-))
+- $\nu_{1}$ is the maximum opaque frequency of the medium (SI units ($s^{-1}$))
+- $B$ is the monochromatic black body radiation intensity from under Planck function (SI unit ($W sr^{-1} m^{-2}$))
+
+Plank function:
+\begin{equation}
+{B_m}({T},{\nu}) = {\frac{n_m^2}{c_0^2}}{\frac{2{h_p}{\nu^3}}{e^{{{h_p}{\nu}}/{({k_B}{T})}}-1}}
```
GiudGiud: what is the index m here? The medium? If so should it appear in the equation above?

modules/heat_transfer/doc/content/source/fvbcs/FVSP3TemperatureBC.md
```diff
@@ -0,0 +1,44 @@
...
+\int_{0}^{\nu_1} \left( B(\nu, T_b) - B(\nu, T) \right)\, d\nu
+\right]
+\end{equation}
+
+where:
+
+- $\varepsilon$ is the optical thickenss of medium (SI unit (-))
+- $k$ is the thermal conductivity of medium (SI unit ($W m^{-1} K^{-1}$))
+- $h$ is the convective heat transfer coefficient of medium (SI unit ($W m^{-2} K^{-1}$))
+- $T_b$ is the ambient temperature (SI unit (K))
```
GiudGiud:
```suggestion
- $T_b$ is the ambient (boundary) temperature (SI unit (K))
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3ThermalRadiationBC.md
```diff
@@ -0,0 +1,31 @@
...
+- $\alpha_i$ is the first term coefficient with idential order of flux moments (SI unit (-))
```
GiudGiud:
```suggestion
- $\alpha_i$ is the first term coefficient with identical order of flux moments (SI unit (-))
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3ThermalRadiationBC.md
```diff
@@ -0,0 +1,31 @@
...
+- $\beta_j$ is the second term coefficient with coupled order of flux moments (SI unit (-))
+- $\eta_i$ is the third term coefficient with idential order of flux moments (SI unit (-))
```
GiudGiud:
```suggestion
- $\eta_i$ is the third term coefficient with identical order of flux moments (SI unit (-))
```

modules/heat_transfer/doc/content/source/fvkernels/FVSP3ThermalRadiationDiffusion.md
```diff
@@ -0,0 +1,23 @@
...
+- $\mu_i$ is the dimensionless eigenvalue parameter of targeting order (SI unit (-))
+- $\kappa$ is the absorpsivity of medium (SI unit ($m^{-1}$))
```
GiudGiud:
```suggestion
- $\kappa$ is the absorptivity of medium (SI unit ($m^{-1}$))
```

modules/heat_transfer/include/base/HeatConductionNames.h
```diff
@@ -19,5 +21,38 @@ static const std::string T_ambient = "T_ambient";
 namespace Constants
 {
 static const Real sigma = 5.670374419e-8;
+
+static constexpr Real
+PlanckConstant(const std::optional<std::string> & unit)
```
GiudGiud:
```suggestion
planckConstant(const std::optional<std::string> & unit)
```

modules/heat_transfer/include/base/HeatConductionNames.h
```diff
@@ -19,5 +21,38 @@ static const std::string T_ambient = "T_ambient";
...
+  constexpr Real v = 6.62607015e-34;
+
+  if (!unit)
+    return v; // Js
+
+  return MooseUnits(unit.value()).convert(v, MooseUnits("J*s"));
+}
+
+static constexpr Real
+SpeedOfLight(const std::optional<std::string> & unit)
```
GiudGiud:
```suggestion
speedOfLight(const std::optional<std::string> & unit)
```

modules/heat_transfer/include/base/HeatConductionNames.h
```diff
@@ -19,5 +21,38 @@ static const std::string T_ambient = "T_ambient";
...
+  constexpr Real v = 2.99792458e8;
+
+  if (!unit)
+    return v; // m/s
+
+  return MooseUnits(unit.value()).convert(v, MooseUnits("m/s"));
+}
+
+static constexpr Real
+BoltzmannConstant(const std::optional<std::string> & unit)
```
GiudGiud:
```suggestion
boltzmannConstant(const std::optional<std::string> & unit)
```

modules/heat_transfer/doc/content/source/fvbcs/FVSP3TemperatureBC.md
```diff
@@ -0,0 +1,44 @@
...
+- $n_1$ is the refraction coefficient of medium 1 (SI unit (-))
+- $n_2$ is the refraction coefficient of medium 2 (SI unit (-))
```
GiudGiud: need to distinguish medium 1 and 2
which one is the cavity and which one is the boundary

modules/heat_transfer/include/fvbcs/FVSP3TemperatureBC.h
```diff
@@ -0,0 +1,60 @@
...
+  const Moose::Functor<ADReal> & _epsilon;
+
+  /// hemispheric emissivity
+  const Moose::Functor<ADReal> & _alpha;
+
+  /// Opaque frequency of medium
+  const Real & _nu1;
+
+  /// Minimum frequency
+  const Real & _nu_min;
```
GiudGiud:
```suggestion
  /// Opaque frequency of medium
  const Real _nu1;

  /// Minimum frequency
  const Real _nu_min;
```
storing them locally on the class has better memory locality

modules/heat_transfer/include/fvbcs/FVSP3ThermalRadiationBC.h
```diff
@@ -0,0 +1,67 @@
...
+  const Moose::Functor<ADReal> & _n1;
+
+  /// Optical Thickness of medium
+  const Moose::Functor<ADReal> & _optical_thickness;
+
+  /// Radiation function from the conjugated other
+  const Moose::Functor<ADReal> & _psi;
+
+  /// Order or radiation moment
+  const MooseEnum & _order;
```
GiudGiud:
```suggestion
  const MooseEnum _order;
```

modules/heat_transfer/test/tests/sp3_radiation/tests
```diff
@@ -0,0 +1,25 @@
...
+  [glass_1D_thin]
+    type = CSVDiff
+    input = 'adaptive1D_e10.i'
+    csvdiff = 'adaptive1D_e10_csv_y0_0005.csv'
+    requirement = 'The system shell run a short simulation with SP3 kernels and BCs in 1D optically thin cooling'
+  []
+
+  [single_frequency]
+    type = CSVDiff
+    input = 'single_frequency.i'
```
GiudGiud, line 21: how do you handle two frequency bands? same full set of kernels & BCs?

modules/heat_transfer/include/fvkernels/FVSP3TemperatureSourceSink.h
```diff
@@ -0,0 +1,46 @@
...
+
+#pragma once
+
+#include "FVFluxKernel.h"
+
+/// FVSP3TemperatureSourceSink implements irradiation-consistent temperature source
+///
+///     - strong form: \int dV \int d\nu \nabla \cdot (1/\kappa) \nabla (a1 * psi_1 + a2 * psi_2)
+///
+///     - weak form: \int_{A} dA \int d\nu (1/\kappa)_f \nabla (a1 * psi_1 + a2 * psi_2)_f
```
GiudGiud:
```suggestion
///     - Integrated by parts : \int_{A} dA \int d\nu (1/\kappa)_f \nabla (a1 * psi_1 + a2 * psi_2)_f
```

GiudGiud: so which BC does this go with? SP3 temperature?

modules/heat_transfer/include/fvkernels/FVSP3TemperatureSourceSink.h
```diff
@@ -0,0 +1,46 @@
...
+class FVSP3TemperatureSourceSink : public FVFluxKernel
+{
+public:
+  static InputParameters validParams();
+  FVSP3TemperatureSourceSink(const InputParameters & params);
+
+protected:
+  virtual ADReal computeQpResidual() override;
+
+  /// kappa absorptivities per bank
```
GiudGiud: bank ? introduce this terminology in the doco please

modules/heat_transfer/include/fvkernels/FVSP3TemperatureSourceSink.h
```diff
@@ -0,0 +1,46 @@
...
+  const std::vector<MooseFunctorName> & _psi2_vec;
+
+  /// Vector to store the functors
+  std::vector<const Moose::Functor<ADReal> *> _absorptivity_vec_functors;
+  std::vector<const Moose::Functor<ADReal> *> _psi1_vec_functors;
+  std::vector<const Moose::Functor<ADReal> *> _psi2_vec_functors;
+
+  /// Closure parameters
+  const Real _a1 = 1. / 30. * (5. - 3. * std::sqrt(5. / 6.)); // * 5.05e-7;
+  const Real _a2 = 1. / 30. * (5. + 3. * std::sqrt(5. / 6.)); // * 5.05e-7;
```
GiudGiud:
```suggestion
  const Real _a1 = 1. / 30. * (5. - 3. * std::sqrt(5. / 6.));
  const Real _a2 = 1. / 30. * (5. + 3. * std::sqrt(5. / 6.));
```

modules/heat_transfer/include/fvkernels/FVSP3ThermalRadiationDiffusion.h
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+
+#include "FVFluxKernel.h"
+#include "FVDiffusionInterpolationInterface.h"
+
+/// FVSP3ThermalRadiationDiffusion implements a diffusion term for the SP3 thermal radiation diffusion/
+///
+///     - strong form: \nabla \cdot (\epsilon^2 \mu_n^2) / \kappa) \nabla u / coef
+///
+///     - weak form: \int_{A} (\epsilon^2 \mu_n^2) \nabla u / coef \cdot \vec{n} dA
```
GiudGiud:
```suggestion
///     - Integration by parts: \int_{A} (\epsilon^2 \mu_n^2) \nabla u / coef \cdot \vec{n} dA
```

modules/heat_transfer/include/fvkernels/FVSP3ThermalRadiationDiffusion.h
```diff
@@ -0,0 +1,49 @@
...
+  virtual ADReal computeQpResidual() override;
+
+  /// Optical Thickness
+  const Moose::Functor<ADReal> & _optical_thickness;
+
+  /// kappa Thickness
+  const Moose::Functor<ADReal> & _absorptivity;
+
+  /// Order
+  const MooseEnum & _order;
```
GiudGiud:
```suggestion
  const MooseEnum _order;
```

modules/heat_transfer/include/fvkernels/FVSP3TemperatureSourceSink.h
GiudGiud: I think you mean band

modules/heat_transfer/include/fvkernels/FVSP3ThermalRadiationSourceSink.h
```diff
@@ -0,0 +1,48 @@
...
+  FVSP3ThermalRadiationSourceSink(const InputParameters & parameters);
+
+protected:
+  /// Overriding residual function
+  ADReal computeQpResidual() override;
+
+  /// Temperature
+  const Moose::Functor<ADReal> & _T;
+
+  /// Frquency
```
GiudGiud:
```suggestion
  /// Frequency
```

modules/heat_transfer/include/fvkernels/FVSP3ThermalRadiationSourceSink.h
```diff
@@ -0,0 +1,48 @@
...
+  const Moose::Functor<ADReal> & _nu;
+
+  /// Frequency bounds for numerical integration
+  const Moose::Functor<ADReal> * _nu_low;
+  const Moose::Functor<ADReal> * _nu_high;
```
GiudGiud: why are these functors? do they frequency bounds typically vary?

modules/heat_transfer/include/fvkernels/FVSP3ThermalRadiationSourceSink.h
```diff
@@ -0,0 +1,48 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "FVElementalKernel.h"
+
+/**
+ * Computes source the sink terms for the turbulent kinetic energy.
```
GiudGiud:
```suggestion
 * Computes the source and sink terms for the thermal radiation
```

modules/heat_transfer/include/fvkernels/FVSP3ThermalRadiationSourceSink.h
```diff
@@ -0,0 +1,48 @@
...
+ */
+class FVSP3ThermalRadiationSourceSink : public FVElementalKernel
+{
+public:
+  static InputParameters validParams();
+
+  FVSP3ThermalRadiationSourceSink(const InputParameters & parameters);
+
+protected:
+  /// Overriding residual function
```
GiudGiud:
```suggestion
```
this would override the docstring for this routine

modules/heat_transfer/include/utils/HeatTransferModels.h
```diff
@@ -87,4 +87,81 @@ cylindricalGapRadiationHeatFlux(const T1 & r_inner,
   return sigma * (std::pow(T_inner, 4) - std::pow(T_outer, 4)) / rad_resistance;
 }
 
+/* ------------------------------------------------------------------ */
+/*       Generic, error-controlled adaptive-Simpson integrator        */
+/* ------------------------------------------------------------------ */
```
GiudGiud: add docstring in regular format

modules/heat_transfer/include/utils/HeatTransferModels.h
```diff
@@ -87,4 +87,81 @@ cylindricalGapRadiationHeatFlux(const T1 & r_inner,
...
+
+  const Scalar S0 = simpson(a, b);
+  const Scalar tol = std::max(abs_tol, rel_tol * std::fabs(S0));
+
+  return recurse(a, b, S0, tol, max_depth);
+}
+
+/* ------------------------------------------------------------------ */
+/*                   Band-integrated Plank emission                   */
+/* ------------------------------------------------------------------ */
```
GiudGiud: same

modules/heat_transfer/src/fvbcs/FVSP3TemperatureBC.C
```diff
@@ -0,0 +1,89 @@
...
+  InputParameters params = FVFluxBC::validParams();
+  params.addClassDescription("Semi-Transparent Boundary Condition for SP3 Radiation Temperature");
+
+  params.addRequiredParam<MooseFunctorName>("Tb", "The temperature of the boundary");
+  params.addParam<MooseFunctorName>("n1", 1.0, "The refraction coefficient of medium 1.");
+  params.addParam<MooseFunctorName>("n2", 1.0, "The refraction coefficient of medium 2.");
+  params.addRequiredParam<MooseFunctorName>("h",
+                                            "The convective heat transfer coefficient of medium.");
+  params.addRequiredParam<MooseFunctorName>("k",
+                                            "The conductive heat transfer coefficient of medium.");
```
GiudGiud:
```suggestion
                                            "The thermal conductivity of the medium.");
```

modules/heat_transfer/src/fvbcs/FVSP3TemperatureBC.C
```diff
@@ -0,0 +1,89 @@
...
+FVSP3TemperatureBC::validParams()
+{
+  InputParameters params = FVFluxBC::validParams();
+  params.addClassDescription("Semi-Transparent Boundary Condition for SP3 Radiation Temperature");
+
+  params.addRequiredParam<MooseFunctorName>("Tb", "The temperature of the boundary");
+  params.addParam<MooseFunctorName>("n1", 1.0, "The refraction coefficient of medium 1.");
+  params.addParam<MooseFunctorName>("n2", 1.0, "The refraction coefficient of medium 2.");
+  params.addRequiredParam<MooseFunctorName>("h",
+                                            "The convective heat transfer coefficient of medium.");
```
GiudGiud: needs to be clear what is medium, medium 1, medium 2

modules/heat_transfer/src/fvbcs/FVSP3TemperatureBC.C
```diff
@@ -0,0 +1,89 @@
...
+
+  // Build the convective source at the boundary
+  const auto T = _var(face, state);
+  const auto Tb = _Tb(face, state);
+  const auto epsilon = _epsilon(face, state);
+  const auto thermal_conv_source =
+      _h(face, determineState()) * (_Tb(face, determineState()) - _var(face, determineState()));
+
+  // Build the radiative source at the boundary
+  const auto n1 = _n1(face, state);
```
GiudGiud:
```suggestion
  const auto & T = _var(face, state);
  const auto & Tb = _Tb(face, state);
  const auto & epsilon = _epsilon(face, state);
  const auto & thermal_conv_source =
      _h(face, determineState()) * (_Tb(face, determineState()) - _var(face, determineState()));

  // Build the radiative source at the boundary
  const auto & n1 = _n1(face, state);
```
avoids copies since they are AD numbers it could matter

modules/heat_transfer/doc/content/source/fvkernels/FVSP3TemperatureSourceSink.md
```diff
@@ -0,0 +1,24 @@
...
+## Overview
+
+The temperature source implemented reads as follows [!citep](larsen2002):
+\begin{equation}
+{\sum}_{i}{\nabla}{\cdot}{\frac{1}{\kappa}}{\nabla}{(a_1\psi_1+a_2\psi_2)}
+\end{equation}
+
+where:
+
+- $\kappa$ is the absorpsivity of medium (SI unit ($m^{-1}$))
```
GiudGiud:
```suggestion
- $\kappa$ is the absorptivity of medium (SI unit ($m^{-1}$))
```

modules/heat_transfer/src/fvbcs/FVSP3TemperatureBC.C
```diff
@@ -0,0 +1,89 @@
...
+  const Real abs_tol = 1E-8;
+  const Real rel_tol = 1E-6;
+
+  const auto boundary_source = HeatTransferModels::integratedPlanckBand<ADReal>(
+      n1, 1.0, Tb, _nu_min, _nu1, abs_tol, rel_tol, _planck_units, _sol_units, _boltzmann_units);
+  const auto cell_source = HeatTransferModels::integratedPlanckBand<ADReal>(
+      n1, 1.0, T, _nu_min, _nu1, abs_tol, rel_tol, _planck_units, _sol_units, _boltzmann_units);
```
GiudGiud, line 83: why is kappa hard coded to 1?

modules/heat_transfer/src/fvbcs/FVSP3ThermalRadiationBC.C
```diff
@@ -0,0 +1,148 @@
...
+{
+  InputParameters params = FVFluxBC::validParams();
+  params.addClassDescription("Marshak Boundary Condition for SP3 Radiation Transport");
+
+  params.addRequiredRangeCheckedParam<MooseFunctorName>(
+      "Tb", "Tb>0", "The temperature of the boundary.");
+  params.addRangeCheckedParam<MooseFunctorName>(
+      "nu", "nu>0", "The mean frequency of the thermal radiation band.");
+  params.addRangeCheckedParam<MooseFunctorName>(
+      "nu_low", "nu_low>0", "The lower frequency for integration of the thermal radiation band.");
```
GiudGiud: did you check rhat rangeCheckParam works with functors? I would think it only works with Real

modules/heat_transfer/src/fvbcs/FVSP3ThermalRadiationBC.C
```diff
@@ -0,0 +1,148 @@
...
+                                                                             nu_high,
+                                                                             abs_tol,
+                                                                             rel_tol,
+                                                                             _planck_units,
+                                                                             _sol_units,
+                                                                             _boltzmann_units);
+    thermal_rad_source = -1 * _eta(face, state) * rad_source / (4.0 * libMesh::pi);
+  }
+  else
+  { // equatino remains in single frequency
```
GiudGiud:
```suggestion
  { // equation for a single frequency
```

modules/heat_transfer/test/tests/sp3_radiation/single_frequency.i
```diff
@@ -0,0 +1,201 @@
...
+      nx = 51
+      xmin = -0.5
+      xmax = 0.5
+  []
+[]
+
+[Problem]
+  nl_sys_names = 't
+                  psi1
+                  psi2'
```
GiudGiud: so this is solving them in a segregated manner
could you compare to a fully coupled solve to make sure we converge in both cases

modules/heat_transfer/test/tests/sp3_radiation/tests
```diff
@@ -0,0 +1,25 @@
+[Tests]
+  design = 'FVSP3ThermalRadiationBC.md FVSP3TemperatureBC.md FVSP3ThermalRadiationDiffusion.md FVSP3ThermalRadiationSourceSink.md FVSP3TemperatureSourceSink.md'
+  issues = '#31346'
+
+  [glass_1D_thick]
+    type = CSVDiff
+    input = 'adaptive1D_e1.i'
+    csvdiff = 'adaptive1D_e1_csv_y0_0005.csv'
+    requirement = 'The system shell run a short simulation with SP3 kernels and BCs in 1D optically thick cooling'
```
GiudGiud:
```suggestion
    requirement = 'The system shall run a short simulation with SP3 kernels and BCs in 1D optically thick cooling'
```

modules/heat_transfer/test/tests/sp3_radiation/tests
```diff
@@ -0,0 +1,25 @@
...
+  []
+
+  [glass_1D_thin]
+    type = CSVDiff
+    input = 'adaptive1D_e10.i'
+    csvdiff = 'adaptive1D_e10_csv_y0_0005.csv'
+    requirement = 'The system shell run a short simulation with SP3 kernels and BCs in 1D optically thin cooling'
```
GiudGiud:
```suggestion
    requirement = 'The system shall run a short simulation with SP3 kernels and BCs in 1D optically thin cooling'
```

modules/heat_transfer/test/tests/sp3_radiation/tests
```diff
@@ -0,0 +1,25 @@
...
+  []
+
+  [single_frequency]
+    type = CSVDiff
+    input = 'single_frequency.i'
+    csvdiff = 'single_frequency_csv_y0_0005.csv'
+    requirement = 'The system shell run a short simulation with SP3 kernels and BCs with single frequency band'
```
GiudGiud:
```suggestion
    requirement = 'The system shall run a short simulation with SP3 kernels and BCs with single frequency band'
```

modules/heat_transfer/src/fvkernels/FVSP3ThermalRadiationDiffusion.C
```diff
@@ -0,0 +1,104 @@
...
+  // If we are on internal faces, we interpolate the diffusivity as usual
+  if (_var.isInternalFace(*_face_info))
+  {
+    // Compute the diffusion coefficients at the element and the neighbor
+    const auto optical_thickness_elem_squared =
+        Utility::pow<2>(_optical_thickness(elemArg(), state));
+    const auto optical_thickness_neighbor_squared =
+        Utility::pow<2>(_optical_thickness(neighborArg(), state));
+    const auto absorptivity_elem =
+        (_absorptivity(elemArg(), state) > 1e-12 ? _absorptivity(elemArg(), state) : 1e-12);
```
GiudGiud: this is evaluating the functor twice. let s avoid that

modules/heat_transfer/src/fvkernels/FVSP3ThermalRadiationDiffusion.C
```diff
@@ -0,0 +1,104 @@
...
+  }
+  // Else we just use the boundary values (which depend on how the diffusion
+  // coefficient is constructed)
+  else
+  {
+    const auto face = singleSidedFaceArg();
+
+    const auto optical_thickness_face_squared = Utility::pow<2>(_optical_thickness(face, state));
+    const auto absorptivity_face =
+        (_absorptivity(face, state) > 1e-12 ? _absorptivity(face, state) : 1e-12);
```
GiudGiud, line 96: same

modules/heat_transfer/src/fvkernels/FVSP3ThermalRadiationDiffusion.C
```diff
@@ -0,0 +1,104 @@
...
+                optical_thickness_elem_squared / absorptivity_elem,
+                optical_thickness_neighbor_squared / absorptivity_neighbor,
+                *_face_info,
+                true);
+  }
+  // Else we just use the boundary values (which depend on how the diffusion
+  // coefficient is constructed)
+  else
+  {
+    const auto face = singleSidedFaceArg();
```
GiudGiud, line 92: to get coverage on this part of the kernel you ll want a case with a dirichletBC

GiudGiud (commented):
you should be able to group all those suggestions from the files tab

modules/heat_transfer/test/tests/sp3_radiation/tests
ahnsininl: Two frequency bands are handled in the same way as the multi-band (e.g., 8-band) tests. The difference is that the number of psi variables is reduced to 4, two (psi1, psi2) for each band.

modules/heat_transfer/include/fvkernels/FVSP3ThermalRadiationSourceSink.h
ahnsininl: Frequency bounds are determined by the radiating species and remain fixed in my problems, so I changed them to `Real`. Since users can already model space-dependent opacity with `kappa` (absorptivity), I think constant frequency bounds are sufficient. Even if a user later wants to represent space-dependent radiating species, the effect of each frequency band can still be captured through `kappa`, provided the spectrum is split into enough bands.

modules/heat_transfer/src/fvbcs/FVSP3TemperatureBC.C
ahnsininl: Since `HeatTransferModels::integratedPlanckBand` calculate $4\pi \kappa \int_{\nu_a}^{\nu_b} B d \nu$, $\kappa$ is hard coded to 1 and devided by 4 to get integrated $\pi B(\nu)$ in line 84-86.

modules/heat_transfer/src/fvbcs/FVSP3ThermalRadiationBC.C
ahnsininl: As you mentioned, rangecheck was not available for functors. Based on your other comment, I converted the frequency band inputs to Real, so rangecheck now works.

modules/heat_transfer/test/tests/sp3_radiation/single_frequency.i
ahnsininl: I added a test for the segregated solve.

modules/heat_transfer/src/fvkernels/FVSP3ThermalRadiationDiffusion.C
ahnsininl: Thank you for your suggestion, I added a test with Dirichlet BC.

modules/heat_transfer/test/tests/sp3_radiation/adaptive1D_e1_coupled.i
```diff
@@ -0,0 +1,778 @@
...
+      xmin = -0.5
+      xmax = 0.5
+  []
+[]
+
+[Variables]
+  [T]
+    type = MooseVariableFVReal
+    initial_condition = ${T0}
+  []
```
GiudGiud, line 54: since all variables are actually in the same nonlinear system, this is the fully coupled approach not the segregated.

this is the test I wanted to see, just change the name / description please

GiudGiud (commented):
Do you have a spatial MMS test for single band?
There are a couple examples in the repo. The python utility will compute the forcing functions for each equations

modules/heat_transfer/test/tests/sp3_radiation/adaptive1D_e1_coupled.i
ahnsininl: changed

ahnsininl:
> Do you have a spatial MMS test for single band? There are a couple examples in the repo. The python utility will compute the forcing functions for each equations

Thank you for suggestion. I've added an MMS test using a frequency band.
Also updated the test set to exclude temperature calculation in the single frequency value case to prevent confusion, since that is non-physical. (Single frequency value input is meant to calculate the radiative heat flux moment before frequency integration, if the user wants to inspect that value. If temperature calculation is needed, a single frequency band should be used instead.)

modules/heat_transfer/test/tests/sp3_radiation/mms_simple_band_spatial.py
```diff
@@ -0,0 +1,24 @@
...
+        fig = mms.ConvergencePlot(xlabel='Average Element Size ($h$)', ylabel='$L_2$ Error')
+        fig._axes.set_xscale('log')
+        fig._axes.set_yscale('log')
+
+        fig.plot(df, label='CONST', marker='o', markersize=8)
+        fig.save('simple_band_mms_spatial.png')
+
+        for key, value in fig.label_to_slope.items():
+            print("%s slope, %f" % (key, value))
+            self.assertTrue(fuzzyAbsoluteEqual(value, 2., .1))
```
GiudGiud: ok this is great this is the order we expect

modules/heat_transfer/test/tests/sp3_radiation/simple_band_mms_temporal.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+
+import unittest
+import mms
+from mooseutils import fuzzyEqual, fuzzyAbsoluteEqual
+
+class TestSimpleBand(unittest.TestCase):
+    def test_convergence(self):
+        df = mms.run_temporal('simple_band_temporal.i', 4, console=False, dt=1e-5, executable='../../../heat_transfer-opt')
```
GiudGiud: are you doing anything special for time derivatives that require this new MMS test?

ahnsininl: This MMS test was created because the equation includes a time-derivative term of temperature. I didn’t implement any special handling for the time derivative, and the test verifies that the temporal term is correctly solved using the default implicit Euler scheme.

PR #31352, opened 2025-08-22
Relax regexp so can pass with and without color output

lindsayad:
Refs failures after idaholab/civet_recipes#2450

### DIFF
--- a/test/tests/problems/no_solve/tests
+++ b/test/tests/problems/no_solve/tests
@@ -22,9 +22,8 @@
     type = 'RunApp'
     prereq = eigen_problem_solve_fail
     requirement = 'The system shall have the ability to disable the nonlinear-eigen system solve in a simulation.'
-    # regex meant to make sure no messaged about solve failing is printed last
-    # \S+ is to consume the ansi color characters
-    expect_out = 'Solve Skipped!\S+\s+(?!Aborting)'
+    # regex meant to make sure no message about solve failing is printed
+    expect_out = 'Solve Skipped!(?![\s\S]*Aborting)'
     absent_out = 'WARNING! There are options you set that were not used!'
     input = 'ne_fail.i'
     cli_args = Problem/solve=false


loganharbour (approved):

test/tests/problems/no_solve/tests
```diff
@@ -24,7 +24,7 @@
     requirement = 'The system shall have the ability to disable the nonlinear-eigen system solve in a simulation.'
     # regex meant to make sure no messaged about solve failing is printed last
     # \S+ is to consume the ansi color characters
-    expect_out = 'Solve Skipped!\S+\s+(?!Aborting)'
+    expect_out = 'Solve Skipped!\S*\s*(?!Aborting)'
```
nmnobre: Either:
```suggestion
    expect_out = 'Solve Skipped!'
```
Or, if we want to keep the _intended_ behavior:
```suggestion
    expect_out = 'Solve Skipped!((?!Aborting).)*$'
```
The current suggested change is way too loose, https://regex101.com/r/SHDq1R/1, it's effectively equivalent to my first suggestion above.

nmnobre (changes requested):

test/tests/problems/no_solve/tests
nmnobre: Perhaps I should've said greedy, not loose, for proper regex lingo...

loganharbour:
```suggestion
    expect_out = 'Solve Skipped!'
    absent_out = 'Aborting'
```

loganharbour (changes requested):

test/tests/problems/no_solve/tests
loganharbour: Opted for more simple... https://github.com/idaholab/moose/pull/31352#discussion_r2295030981

lindsayad: thoughtless work by me on an off day

test/tests/problems/no_solve/tests
```diff
@@ -24,8 +24,8 @@
     requirement = 'The system shall have the ability to disable the nonlinear-eigen system solve in a simulation.'
     # regex meant to make sure no messaged about solve failing is printed last
     # \S+ is to consume the ansi color characters
```
nmnobre:
```suggestion
```

test/tests/problems/no_solve/tests
nmnobre: > Opted for more simple...

Yes, I thought of that too, but you had to sacrifice part of the checks.
If the idea is we're already testing that warning is absent in another test (the first in this same file), that's fine, but please remove the outdated comments as suggested below.

lindsayad: We pass `re.MULTILINE | re.DOTALL`, so the latter regexp would not work if `Aborting` is on a later line

lindsayad:
Hopefully the commit I just pushed satisfies everyone including CIVET

test/tests/problems/no_solve/tests
```diff
@@ -22,9 +22,8 @@
     type = 'RunApp'
     prereq = eigen_problem_solve_fail
     requirement = 'The system shall have the ability to disable the nonlinear-eigen system solve in a simulation.'
-    # regex meant to make sure no messaged about solve failing is printed last
-    # \S+ is to consume the ansi color characters
-    expect_out = 'Solve Skipped!\S+\s+(?!Aborting)'
+    # regex meant to make sure no message about solve failing is printed
+    expect_out = 'Solve Skipped!(?![\s\S]*Aborting)'
```
nmnobre, line 26:
```suggestion
    expect_out = 'Solve Skipped!(?!.*Aborting)'
```

nmnobre (approved):
I'm happy with this :), my suggestion is just a slight simplification leveraging `re.DOTALL`.

test/tests/problems/no_solve/tests
lindsayad: I think there's some value in trying to make the regex work independently of the options we pass to re.search in case someone wants to try and change them in the future

loganharbour (approved):

PR #31356, opened 2025-08-22
Make homogenization constraints in solid_mechanics work with NEML2

### LINKED ISSUE #31355: solid_mechanics homogenization system does not work with NEML2
### Bug Description

The current homogenization constraint system does not work:
1. With NEML2 material models
2. When `residual_and_jacobian_together = true`, which we typically use when running NEML2 models.

### Steps to Reproduce

Take any test (e.g. `modules/solid_mechanics/test/tests/lagrangian/cartesian/total/homogenization/action/action_3d.i` and replace the constitutive model with NEML2).

### Impact

Will prevent us from running multiscale NEML2 calculations, e.g. crystal plasticity.

### [Optional] Diagnostics

_No response_

reverendbedford:
The solid_mechanics homogenization system does not work with NEML2 models, nor does it work with the `residual_and_jacobian_together` option which we typically use with NEML2 models.

This PR fixes this by:
1. Swapping the user object based system with a new system using `KernelScalar`, already [merged](https://github.com/idaholab/moose/commits/db004415602203a8d05b7a82a114b3fccfed69fe/modules/tensor_mechanics/test/src/kernels/HomogenizedTotalLagrangianStressDivergenceS.C?browsing_rename_history=true&new_path=modules/solid_mechanics/test/src/kernels/HomogenizedTotalLagrangianStressDivergenceS.C&original_branch=next) into the test app.
2. Fixes a bug in `KernelScalar` that prevents the scalar jacobian contributions being entered when using a `KernelScalar` with `residual_and_jacobian_together`
3. Updates the existing homogenization system tests to use the `KernelScalar` implementation.  No regolding is necessary -- the `KernelScalar` approach gives exactly the same answers as the `UserObject` method.  I also kept the syntax of the action/physics the same, so it's only the tests that "manually" add the objects that need to be updated.
4. Added new tests to cover the neml2 and `residual_and_jacobian_together` issues.

For those curious why the user object system doesn't work with NEML2 it has to do with the ordering of how we call NEML2 and then dump the results back into the materials.  We call NEML2 in a (general) user object then the actual MOOSE materials get updated in the element residual loop.  Unfortunately the current homogenization constraint system uses another (element) user object, which means even if we push it to execute after the NEML2 user object we still don't have the new stresses/strains/whatever available.

Edit: should add that @ttruster originally wrote the `KernelScalar` system.

closes #31355


hugary1995 (approved):

hugary1995 (approved):
I approve the changes in modules/solid_mechanics.

hugary1995:
@lindsayad could you please review the small changes in the framework? They look right to me, but I think I better have someone from the framework team to confirm.

lindsayad (approved):

Issue #20482, opened 2022-03-07
Unscale reaction forces


dschwen:
## Reason
When using nodal sums of `TagVectorAux` on a variable vector tag to obtain the reaction forces, the results are actually scaled with the variable scale (this is particularly confusing when using autoscaling). 

## Design
Provide an option to reverse the scaling in `TagVectorAux`.

## Impact
Added capability.

recuero:
Should the reverse scaling be default?

dschwen:
> Should the reverse scaling be default?

I don't know. It comes down what the user should expect from this object.

- The raw plumbing, the residual contributions as PETSc sees them (with scaling applied)
- The physically meaningful quantities (without scaling applied)

@lindsayad , any opinions on this?

lindsayad:
I think the default should be the scaled values corresponding to what is being fed into PETSc and the nonlinear solve

dschwen:
@lindsayad for matrix tags the scaling relationship is not as clear. I suppose boundary conditions break it. So I'll just do it for vector tags.

lindsayad:
Scaling factors have to be uniformly applied (or not applied) to the residual and to the Jacobian in order for the latter to be an accurate representation of the derivatives of the former. I believe we apply scaling factors to anything that is a "physics" residual/Jacobian. "Strong" constraints such as Dirichlet boundary conditions typically do not have scaling factors applied to their residual/Jacobian.

dschwen:
![image](https://user-images.githubusercontent.com/202302/157352643-3971df31-4162-4eab-af37-30d4de70157a.png)

Yeah, I was honestly quite confused when the unscaling didn't work for the Jacobian. he image above has the unscaled run on the left and the 10x scaled run on the right (plotting TagMatrixAux).

lindsayad:
Do you have Dirichlet BCs in this problem?

dschwen:
Yeah, sorry, I could have sworn I wrote this :eyes:

lindsayad:
I would expect that you would have seen similarly confusing behavior for a `TagVectorAux`. Do you not?

dschwen:
I don't

lindsayad:
Do your ICs match your BCs? Also, I'm guessing you aren't putting `preset = false` so that would make sense to me ... your DirichletBC residuals are 0 since your dofs are getting preset to the dirichlet value

dschwen:
[...]

PR #31357, opened 2025-08-22
Add ReactionForceAux

hugary1995:
ref #20482
    
Deprecate the "scaled" parameter in TagAuxBase. The TagVectorAux never removes variable scaling from the dof value. A new object ReactionForceAux is added which always removes variable scaling from the dof value.
    
This also affects other aux kernels that inherit from TagAuxBase.


framework/include/auxkernels/TagAuxBase.h
```diff
@@ -50,10 +50,13 @@ TagAuxBase<T>::validParams()
...
-  params.addParam<bool>("scaled",
-                        true,
-                        "Return value depending on the variable scaling/autoscaling. Set this to "
-                        "false to obtain unscaled physical reaction forces.");
+  params.addDeprecatedParam<bool>(
+      "scaled",
+      true,
+      "Return value depending on the variable scaling/autoscaling. Set this to false to obtain "
+      "unscaled values with units consistent with the rest of the simulation.",
+      "This parameter will be removed in the future. If you are currently setting scaled = true, "
```
dschwen: Mention `ReactionForceAux` here

hugary1995: It's a little tricky because this is a decorator used by many objects. I added a deprecation message in TagVectorAux specifically mentioning ReactionForceAux.

framework/src/auxkernels/ReactionForceAux.C
```diff
@@ -0,0 +1,31 @@
...
+
+registerMooseObject("MooseApp", ReactionForceAux);
+
+InputParameters
+ReactionForceAux::validParams()
+{
+  InputParameters params = TagVectorAux::validParams();
+  params.addClassDescription("Couple a tag vector, and return its dof value. Variable scaling is "
+                             "removed from the dof value.");
+  params.addRequiredParam<TagName>("vector_tag", "Tag Name this AuxKernel works on");
```
GiudGiud:
```suggestion
  params.addRequiredParam<TagName>("vector_tag", "Name of the vector tag this AuxKernel works on");
```

GiudGiud (commented):
do you need this in the framework? Seems like it's only a force in solid mechanics. 

In the mass conservation equation, or a simple diffusion equation, for example it's not a force

framework/doc/content/source/auxkernels/ReactionForceAux.md
```diff
@@ -0,0 +1,14 @@
+# ReactionForceAux
+
+The value of a tagged vector for a given variable is coupled to
+the current AuxVariable. ReactionForceAux return the unscaled coupled DOF value with units consistent with the simulation.
+AuxVariable then is written out in an exodus file.
+
+!warning
+This AuxKernel assumes that the coupled tagged vector is affected by variable scaling. This assumption holds for the residual vector, but does not for the solution vector.
```
GiudGiud: right so since we know what the solution vector tag is, we could warn the user in the code

bwspenc:
> do you need this in the framework? Seems like it's only a force in solid mechanics.
> 
> In the mass conservation equation, or a simple diffusion equation, for example it's not a force

I agree with this concern regarding naming of this. This is generally useful across physics models, but it has different names depending on the physics. It's a force in mechanics, but it's an integrated flux for diffusion models. I'm not sure what a good generic name would be for it. TagVectorAux? Just kidding, kinda.

GiudGiud:
> but it's an integrated flux for diffusion models

as a volumetric term I m not sure how to understand it at all in diffusion

lindsayad:
We call the Laplacian `Diffusion` and I haven't really heard anyone complain about that

lindsayad:
Even though some people might think of diffusion as mass transport whereas the PDE term that's implemented is just as applicable for heat conduction/mass transport and viscous momentum transport

lindsayad:
We also have `Reaction` for the incredibly generic PDE term $u$

bwspenc:
I didn't think about the use of "Reaction" in the chemical reaction sense, but that's another reason why generically using ReactionForce in this name could be a problem.

bwspenc:
My brainstorming on names:
UnscaledTagVectorAux: Precisely describes what this is without misleading anyone
IntegratedFluxAux: I think this might kind of work as a generic physics-independent description of it, assuming you have set your volumetric kernels to accumulate to this tag vector and not done that on the BCs.

As much as UnscaledTagVectorAux is a horribly non-user-friendly name for someone trying to get the reactions in a mechanics model, it's currently my favorite.

bwspenc:
> > but it's an integrated flux for diffusion models
> 
> as a volumetric term I m not sure how to understand it at all in diffusion

If you have Dirichlet BCs in a diffusion problem and have the volumetric kernels assemble to this tag vector, this gives you the integrated flux across the boundary they are applied to. It's integrated in the sense that it's integrated over the elements tributary to the node that it's reported on. You could sum it up over the boundary to get it integrated over the boundary.

GiudGiud:
You are doing this for user friendliness. There is already a "scaled" parameter which isnt friendly enough. 

So 
```
As much as UnscaledTagVectorAux is a horribly non-user-friendly name for someone trying to get the reactions in a mechanics model, it's currently my favorite.
```
that's not an option. 

IMO ReactionForce-Something in SolidMechanics is the path forward. And dont deprecate the "scaled" parameter, the other Physics can use it

Any "Flux" name is wrong because the definition of the flux depends both on the equation and on the numerical discretization and you cannot hope to catch them all in one object. 

I renamed the "SideFlux" objects for that reason.

GiudGiud:
> We call the Laplacian Diffusion and I haven't really heard anyone complain about that

I like to think Laplacian is the math name for the operator and Diffusion is the name of the physical phenomenon, and in that thought, what we do totally works. 
You think it does not?

I could see Diffusion become "DiffusionGalerkin" or "VolumetricDiffusionGalerkin" to have a more descriptive name, but it's not what I m thinking is the better path forward, which is using Physics actions to describe equations instead of kernels

bwspenc:
Just curious, what was the SideFlux rename (i.e., what were they named before and what are they named now)? This is a similar idea.

bwspenc:
I still think the best solution to this problem is making "scaled" a required parameter in TagVectorAux. Then the user's intent would be clear.

GiudGiud:
> I still think the best solution to this problem is making "scaled" a required parameter in TagVectorAux. Then the user's intent would be clear.

That also works for me. Though that will be a lot of inputs

bwspenc:
I think the 3 people who have actually used TagVectorAux in practical models all made the wrong assumption about what the default value of "scaled" was, and if more people use it, they're going to make the same mistake.

bwspenc:
> That also works for me. Though that will be a lot of inputs

I agree that it's more inputs, but this is a pretty small one, and it forces the user to think about what they actually want.

lindsayad:
The ironic thing is that what you call Unscaled is actually scaled

dschwen:
If you're objected to anything with "Force" I'd propose `PhysicalResidualAux` or something along those lines.

GiudGiud:
Force is fine in a solid mechanics context

lindsayad:
> I like to think Laplacian is the math name for the operator and Diffusion is the name of the physical phenomenon, and in that thought, what we do totally works.
> You think it does not?

I have no qualms with our `Diffusion` object's name. I was saying that we have precedent for one name being applicable to multiple types of physics, e.g. what some may think of as a mass transport name can also be used in heat/momentum transport. But it's perhaps not the best example because some people may still feel that "diffusion" is a fine name to describe heat conduction/viscous momentum transport, and I wouldn't disagree with them

lindsayad:
I like @dschwen's `PhysicalResidualAux`

lindsayad:
> I still think the best solution to this problem is making "scaled" a required parameter in TagVectorAux. Then the user's intent would be clear.

I'm going to re-paste my response from the issue:

> TagVectorAux is a very general class and allows providing both residual and solution vector tags. Right now there is no check in that class for whether the provided vector tag is a solution or residual one (nor do I think there should be for a class with this general a name). And so actually changing the default to always apply _v_var.scalingFactor() would be completely erroneous for a whole class of vector tag types.

It doesn't make sense to me to make a parameter required which makes no sense for a solution tag type

lindsayad:
I think even scaled/unscaled language is confusing. In my opinion the raw vector is unscaled, even if was composed through "scaling" local element residuals. In others' opinions and as it's written in the doc string, the raw vector is *scaled*. So to me that's another argument for creating two distinct objects where we no longer even need a `scaled` parameter to get confused about.

lindsayad:
If you were to take the complete residual vector, even in the presence of automatic/manual scaling you will have some vector elements that are not scaled by the variable's scaling factor. Dirichlet residuals are unscaled for example. I doubt that anyone is using `TagVectorAux` with the full residual vector though. But these kind of nuances are why I simply view the raw vector as unscaled. (Again a solution vector will never be scaled in any kind of way, so it's nice to be able to view all the raw vectors through one common lens... and my lens is as unscaled)

GiudGiud:
> It doesn't make sense to me to make a parameter required which makes no sense for a solution tag type

could do non-required, but error if using a non-solution tag without specifying it

lindsayad:
I feel pretty strongly that we should have a framework object for "physical" vector quantities. If we can't agree on a name, I'd even prefer @bwspenc's suggested change of making `scaled` required in `TagVectorAux` to removing the ability to compute "physical" vector quantities in the framework. This should be something we support for all kinds of physics

lindsayad:
> could do non-required, but error if using a non-solution tag without specifying it

So require specifying it for residual vector tag types and require not specifying it for solution vector tag types 😄 You like that better than `PhysicalResidualAux`

lindsayad:
To me this is a perfect opportunity for sub-classing ... making a general object (supports residual and solution tags) more specific (supports only residual types/always has variable scaling factors undone to become physical, `return TagVectorAux::computeValue() / _v_var.scalingFactor();`)

GiudGiud:
You are saying it's physical because it's not scaled, it's not intuitive

lindsayad:
I agree it's not intuitive. That's why you have a general base object that just returns elements of a vector. And then you have a subclass that returns physical residuals. No need for any user parameters

GiudGiud:
A "physical residual" is not a thing though.
If we name an object with a name that won't be what people look for, we will always have to point people to it (well maybe not with AI...)

I like ReactionForceAux because that's actually a good name in solid mechanics. I dont want it in the framework.
If you come up with any name that is misleading (Flux... etc), I think it's doing more harm than good
If you come up with any name that is less intuitive than the TagVectorAux, new people will just use that old object, which they will have found first from search for "Tag"

hugary1995:
I admit I'm not terribly familiar with the tagging system -- I have a few questions:
1. Can we tell if a tagged vector is a residual vector or a solution vector just from the tag name?
2. Is a tagged vector either residual or solution?
3. What happens to user declared extra vector tags?
4. What if I am creating a petsc vector myself in the code that's neither solution nor residual, and I'm doing some of my own magical operations with it. Can I still use the TagVectorAux (or some new objects with a better name)?

lindsayad:
> 1. Can we tell if a tagged vector is a residual vector or a solution vector just from the tag name?

Yes, and I pushed a commit to this PR that errors in your new class if someone uses a solution vector tag

> 2\. Is a tagged vector either residual or solution?

Yes, at the moment

> 3\. What happens to user declared extra vector tags?

`Problem/extra_tag_vectors` are residual vector tags. `Problem/extra_tag_solutions` are solution vector tags

> 4\. What if I am creating a petsc vector myself in the code that's neither solution nor residual, and I'm doing some of my own magical operations with it. Can I still use the TagVectorAux (or some new objects with a better name)?

You would need to call `SubProblem::addVectorTag` and then `SystemBase::associateVectorToTag`, and then as long as your vector lives in the libMesh system that your coupled variable belongs to then you could use `TagVectorAux`

lindsayad:
> A "physical residual" is not a thing though.

I would disagree with that

lindsayad:
It's definitely a thing or else we wouldn't have @dschwen, @bwspenc, @hugary1995 etc. trying to use residual vectors to visualize physical quantities

hugary1995:
Yeah, in the context of solid mechanics, we tend to think of this as the reaction force (actual force). For dofs constrained with dirichlet bcs, i.e. a fixed amount of displacement is applied, the reaction force would be the equivalent amount of force to hold the body in place with the same amount of displacement.

As far as I can tell, the residual values associated with the dirichlet dofs are affected by variable scaling. That's why we've been wanting to remove the effect of variable scaling for postprocessing/visualization purposes.

hugary1995:
If the tags are either associated with solution or residual, why don't we simply call them "residual tags" and "solution tags"? Similarly, we could rename `extra_tag_vectors` to `extra_tag_residuals`. We can then have
- TagSolutionAux
- TagResidualAux
- PhysicalTagResidualAux (though this name is debatable)

lindsayad:
> As far as I can tell, the residual values associated with the dirichlet dofs are affected by variable scaling

I just went through the call sequence starting from `NodalBC::computeResidual` and I don't see any place where the variable scaling factor is applied. This would be big news to me

hugary1995:
> If you were to take the complete residual vector, even in the presence of automatic/manual scaling you will have some vector elements that are not scaled by the variable's scaling factor. Dirichlet residuals are unscaled for example.

This seems to contradict my observation though. As I mentioned above, I think the dirichlet residuals (components of the residual vector associated with dirichlet dofs) are in fact affected by variable scaling. That's the whole reason Ben and I were tricked. Could I be missing something?

lindsayad:
```
(moose-python) lindad@rod:/data/lindad/projects/moose/test/tests/kernels/simple_diffusion(tagvectoraux_scaled)$ ../../../moose_test-dbg -i simple_diffusion.i Mesh/dim=1 Mesh/nx=2 Variables/u/scaling=1e3 GlobalParams/preset=false --color off -snes_monitor_residual

 0 Nonlinear |R| = 1.000000e+00
Vec Object: 1 MPI process
  type: seq
0.
0.
-1.
```

hugary1995:
Here's a minimal reproducing example. I must be confusing myself somehow.

The input file is
```python
[Problem]
  extra_tag_vectors = 'dummy'
[]

[Mesh]
  [fmg]
    type = GeneratedMeshGenerator
    dim = 1
    nx = 5
  []
[]

[Variables]
  [u]
  []
[]

[Kernels]
  [diffusion]
    type = MatDiffusion
    variable = u
    diffusivity = 1
    extra_vector_tags = 'dummy'
  []
[]

[BCs]
  [fixed_left]
    type = DirichletBC
    variable = u
    boundary = left
    value = 0
  []
  [fixed_right]
    type = DirichletBC
    variable = u
    boundary = right
    value = 1
  []
[]

[AuxVariables]
  [r]
    [AuxKernel]
      type = TagVectorAux
      v = u
      vector_tag = 'dummy'
    []
  []
[]

[Postprocessors]
  [R]
    type = NodalSum
    variable = r
    boundary = right
  []
[]

[Executioner]
  type = Steady
[]

```

Below I summarize the execution commands and the corresponding postprocessor values:
- `./solid_mechanics-opt -i simple.i Kernels/diffusion/diffusivity=1 Executioner/automatic_scaling=true`, R = 0.1
- `./solid_mechanics-opt -i simple.i Kernels/diffusion/diffusivity=2 Executioner/automatic_scaling=true`, R = 0.1
- `./solid_mechanics-opt -i simple.i Kernels/diffusion/diffusivity=3 Executioner/automatic_scaling=true`, R = 0.1
- `./solid_mechanics-opt -i simple.i Kernels/diffusion/diffusivity=1 Executioner/automatic_scaling=false`, R = 1
- `./solid_mechanics-opt -i simple.i Kernels/diffusion/diffusivity=2 Executioner/automatic_scaling=false`, R = 2
- `./solid_mechanics-opt -i simple.i Kernels/diffusion/diffusivity=3 Executioner/automatic_scaling=false`, R = 3

hugary1995:
Similarly,
`./solid_mechanics-opt -i simple.i Kernels/diffusion/diffusivity=1 Variables/u/scaling=1e3` gives R = 1e3

lindsayad:
Oh yea that's totally expected. For your extra vector tag, you're only feeding in from `MatDiffusion`. `MatDiffusion` element residuals/Jacobians are affected by the variable scaling

hugary1995:
> Oh yea that's totally expected. For your extra vector tag, you're only feeding in from `MatDiffusion`. `MatDiffusion` element residuals/Jacobians are affected by the variable scaling

Ahh I see. Even though it might be obvious to you, I find it to be somewhat confusing, i.e., extra vector tag which accumulates residual contributions behaves differently from the whole residual, etc.. And I think if I find it to be confusing, it's definitely confusing to our users. I'd love to see some improvement in the framework to avoid/reduce the confusion.

lindsayad:
The `Variables/u/scaling` parameter is specifically meant for "physical" residuals. If young's modulus is 1e12 then you might want to specify `scaling=1e-12`. But since for constrained dofs there is nothing physical (generally) in the constraint Jacobian row, it would not make sense to apply a scaling factor there. If we did, then our Jacobian would look something like this and we would have accomplished nothing in terms of trying to improve the conditioning of our system

1
0 1
0 0 1
0 0 0 1e-12

lindsayad:
> Ahh I see. Even though it might be obvious to you, I find it to be somewhat confusing, i.e., extra vector tag which accumulates residual contributions behaves differently from the whole residual

It doesn't behave differently. If you include your extra vector tag in the Dirichlet BC, then it's vector elements will be ~zeroed in advance of applying~ overwritten by the Dirichlet BC just as for the full residual.

But I agree that the nuances of why is scaling applied to Kernels/IntegrateBCs/DGKernels but not constraint objects should definitely be documented

hugary1995:
To add more motivation for this, calculating reaction forces from a solid mechanics simulation is almost certainly the first thing a new user tries to do for benchmarking, when they migrate from another software package and start using moose for solid mechanics.

hugary1995:
Okay, I have a much better understanding now. Circling back to the original issue now, when Ben and I were talking about residual vector, we were actually referring to the element residual contributions from a specific kernel that get accumulated into an extra tagged vector.

hugary1995:
And with the nuances between kernel residual and dirichlet bc residual, I think "to scale or not to scale" is a real tricky choice in the TagVectorAux.

hugary1995:
Now all of your earlier comments start to make a lot of sense to me :)

I'd add to the discussion that if there's a way to hide all of the setup within an Action or Physics, that would also address the user friendliness issue.

lindsayad:
> If the tags are either associated with solution or residual, why don't we simply call them "residual tags" and "solution tags"? Similarly, we could rename `extra_tag_vectors` to `extra_tag_residuals`.

I see no downside to this suggested parameter name change. It only adds clarity

> * TagSolutionAux
> * TagResidualAux
> * PhysicalTagResidualAux (though this name is debatable)

I don't see a need for all three of these. The existence of a `Physical(Tag)ResidualAux` would imply to me that `TagSolutionAux` and `TagResidualAux` do the same thing in terms of constructor, data members, and `computeValue()`.

> To add more motivation for this, calculating reaction forces from a solid mechanics simulation is almost certainly the first thing a new user tries to do for benchmarking, when they migrate from another software package and start using moose for solid mechanics.

For sure I want to support you guys. I understand the issue. Despite what @bwspenc implied here

> I think the 3 people who have actually used TagVectorAux in practical models all made the wrong assumption

I have used `TagVectorAux` to visualize balances between inertial/viscous/convection/gravitational/Boussinesq momentum terms in 3D additive manufacturing simulations (I believe a somewhat practical model) as well as when debugging user convergence issues (in which case I truly do want the raw residual vector).  I don't *want* people to get bitten by this issue of scaling and get frustrated.

> And with the nuances between kernel residual and dirichlet bc residual, I think "to scale or not to scale" is a real tricky choice in the TagVectorAux.

And this is why I vote to have separate objects that totally removes the `scale` parameter

lindsayad:
> But I agree that the nuances of why is scaling applied to Kernels/IntegrateBCs/DGKernels but not constraint objects should definitely be documented

Yea I'm missing documentation of this in [NonlinearSystemBase.md](https://mooseframework.inl.gov/source/systems/NonlinearSystemBase.html)

GiudGiud:
> > A "physical residual" is not a thing though.

> I would disagree with that

We can accumulate in a tagged vector a physical quantity, I do not contest that. 
But is the residual of the equation the physical quantity? Certainly is not before convergence. And after, I would think once we solve the equation at convergence, that residual is just 0?

So what i think is more likely happening, is that we obtain this physical quantity from a subset of kernels / BCs. For example, at a dirichlet BC, we tag everything but the BC and we see the force that compensates for the setting of the BC.
That's my understanding of this at this time. 

A tagged vector accumulated from kernels is just not always physical. And if it's physical, it's likely not just the residual. "PhysicalResidualAux" is misleading.

modules/solid_mechanics/test/tests/free_expansion_abs_ref/free_expansion_abs_ref.i
```diff
@@ -47,13 +47,13 @@
...
   [ref_x]
-    type = TagVectorAux
+    type = ReactionForceAux
     variable = 'ref_x'
     vector_tag = 'ref'
     v = 'disp_x'
   []
   [ref_y]
-    type = TagVectorAux
+    type = ReactionForceAux
```
GiudGiud: I dont think it is a force here? This is an absolute value vector tag for 'ref'

lindsayad:
> But is the residual of the equation the physical quantity? Certainly is not before convergence. And after, I would think once we solve the equation at convergence, that residual is just 0?

I believe it might have even been you who made it so `TagVectorAux` could only be executed on `TIMESTEP_END`, so before convergence is not a concern. A zero residual could be interpreted physically as a balance of your physical forces/stresses/fluxes etc.

hugary1995:
This might be deviating from the original discussion, but I think it's helpful to describe the problem formally.

We are solving for a system of equations (P1) in the form
$r_i(u) = 0$, $\forall i \in \mathcal{I}$
subject to
$c_i(u) = 0$, $\forall i \in \mathcal{A}$
where $i$ is the dof index, $\mathcal{I}$ is the set of inactive constraints, and $\mathcal{A}$ is the set of active constraints. $\mathcal{I}$ and $\mathcal{A}$ disjointly partition the dofs.

The solution $u^\*$ to P1 must also be a solution to the following problem (P2)
$r_i(u) = f_i $, where $f_i = 0$, $\forall i \in \mathcal{I}$ and $f_i = r_i(u^*)$, $\forall i \in \mathcal{A}$

Note P1-->P2 holds, but the converse doesn't necessarily hold.

Anyways, $f_i, \forall i \in \mathcal{A}$ is what we are ultimately looking for. Looking at this from a framework point of view, we could evaluate $f_i$ by accumulating the associated kernel contributions into the same tagged vector.

This approach has been working fine in solid mechanics, because, in the quasi static case, we usually only have N kernels (with N = number of displacements), and those kernels are being set up by the action. It's not a huge amount of effort to set this up.

For a general multiphysics application, one would have to laboriously go through all kernels corresponding to the same equation and add `extra_vector_tags` for each of them. I think it would be much nicer if one could simply set the vector tag in the corresponding strong constraint object (e.g., Dirichlet BCs) and let the framework calculate $f_i$ corresponding to $c_i$.

modules/solid_mechanics/test/tests/free_expansion_abs_ref/free_expansion_abs_ref.i
hugary1995: This might have to be a `AbsoluteReactionForceAux` or something like that... @bwspenc can you comment?

lindsayad:
There's not always strong constraints. Additionally users may want to split their tagging arbitrarily, as I wanted to when visualizing advection/diffusion/external-source contributions. But it could be nice to have a short-cut for the specific case you are describing

bwspenc:
> For sure I want to support you guys. I understand the issue. Despite what @bwspenc implied here

@lindsayad I apologize if something I said came across wrong. I definitely didn't mean to imply that.

bwspenc:
I think @hugary1995 and I are on the same page about what we want, which is the unscaled (hopefully I'm using the right term!) value of the residual vector before the Dirichlet BC zeroes it out. TagVectorAux works for that purpose if we use the right options for scaling there, and also if we request all kernels associated with that variable to assemble to the same tag vector that is referenced in TagVectorAux. That can be a lot to orchestrate correctly, and it's counterintuitive that the kernels, not the BCs are the objects responsible for assembling to the tag.

Users coming from other codes are accustomed to having an easy way of requesting reaction forces in mechanics, or maybe equivalent quantities for other physics. If there is some way of getting this in one step using an object with a name that reflects doing that in a somewhat generic fashion, that would be amazing. TagVectorAux is great for the use cases @lindsayad described, but it's probably too low-level of a thing to expect users who want the reaction forces to wrap their heads around, and its name doesn't at all imply that it could be used for getting reaction forces. Even a power user who understands tags and the roles of the kernels and BCs can also easily get tripped up on the scaled/non-scaled issue there.

Are we wrong in thinking that the tagging system is the best thing to use to get what @hugary1995 and I want? 
Would it make any sense, for example, to allow DirichletBC to directly save off that value into an AuxVariable (with scaling applied)? I know that's got its downsides, but maybe worth considering.

GiudGiud:
> Are we wrong in thinking that the tagging system is the best thing to use to get what @hugary1995 and I want?
Would it make any sense, for example, to allow DirichletBC to directly save off that value into an AuxVariable (with scaling applied)? I know that's got its downsides, but maybe worth considering.

you are thinking of the old save_in and the idea was to deprecate that

> Users coming from other codes are accustomed to having an easy way of requesting reaction forces in mechanics, or maybe equivalent quantities for other physics.

Then alias TagVectorAux to ReactionForcesAux but only in the SolidMechanics module. And set the "scaled" parameter to the expected value

hugary1995:
> > Are we wrong in thinking that the tagging system is the best thing to use to get what @hugary1995 and I want?
> Would it make any sense, for example, to allow DirichletBC to directly save off that value into an AuxVariable (with scaling applied)? I know that's got its downsides, but maybe worth considering.
> 
> you are thinking of the old save_in and the idea was to deprecate that

I think that's very different from save_in. In my understanding, save_in is similar to accumulating to tagged vectors in terms of functionality (not in terms of implementation). What we are picturing above is "saving the residual before zeroing out the constraint row" -- note that would account for all residual objects that write to the constrained dofs -- which is different from setting extra vector tags for each and every residual object (excluding the constraint object itself!).

hugary1995:
So, let me attempt to summarize and see if we are converging here. Does the following sound right to everyone?

framework:
- TagSolutionAux, this object can only be coupled to the solution vector, and no scaling is applied whatsoever.
- TagResidualAux, this object can only be coupled to the residual vector, and "scaled" is an optional (?) parameter default to true (?).
- TagVectorAux will be deprecated.

solid mechanics:
- ReactionForceAux, this object inherits from TagResidualAux and forces scaled = false.

GiudGiud:
yeah this works for me. Though I can keep arguing :D

hugary1995:
> yeah this works for me. Though I can keep arguing :D

lol, which one(s) are you not satisfied with? Perhaps document your thoughts here for future reference...

@lindsayad in response to your concern regarding ambiguity in the "scaled" parameter, would it be better to rename the parameter to something like "remove_variable_scaling" which default to false?

grmnptr:
I'm on vacation so I didn't have too much time to look into this
(practically 0), just thought about the fact that we are also using
tagvector/matrix aux for linear systems where we dont compute residuals and
jacobians. Could you make sure the new system looks fine for that too?
Apologies if this comment is irrelevant and just creates noise here.

bwspenc:
> > > Are we wrong in thinking that the tagging system is the best thing to use to get what @hugary1995 and I want?
> > > Would it make any sense, for example, to allow DirichletBC to directly save off that value into an AuxVariable (with scaling applied)? I know that's got its downsides, but maybe worth considering.
> > 
> > 
> > you are thinking of the old save_in and the idea was to deprecate that
> 
> I think that's very different from save_in. In my understanding, save_in is similar to accumulating to tagged vectors in terms of functionality (not in terms of implementation). What we are picturing above is "saving the residual before zeroing out the constraint row" -- note that would account for all residual objects that write to the constrained dofs -- which is different from setting extra vector tags for each and every residual object (excluding the constraint object itself!).

Exactly what I was thinking! That's an important distinction from the behavior of save_in. We could probably do something like this with the tagging system too, but that would lead us back to the question of how to extract the value of the tag into an AuxVariable and whether it should be scaled.

bwspenc:
> So, let me attempt to summarize and see if we are converging here. Does the following sound right to everyone?
> 
> framework:
> 
> * TagSolutionAux, this object can only be coupled to the solution vector, and no scaling is applied whatsoever.
> * TagResidualAux, this object can only be coupled to the residual vector, and "scaled" is an optional (?) parameter default to true (?).
> * TagVectorAux will be deprecated.
> 
> solid mechanics:
> 
> * ReactionForceAux, this object inherits from TagResidualAux and forces scaled = false

I'm OK with this (although I still think just making `scaled` a required parameter and documenting what it does better would be an easier solution). It raises some questions for me that users would also have:
* If a user creates an extra tag vector (which they have to do to use this for our intended purpose) is it a residual tag or a solution tag? I actually didn't know that there were two different types of tags until yesterday. I assume ours are always residual tags, but this raises questions.
* What should be the default for "scaled" or (whatever better name we have for it) on the object that gets the residual tags? People will come at that with different mindsets, so I think the parameter that selects this should be required.
* I kind of like the idea of ReactionForceAux in SolidMechanics, but it might cause some confusion about what exactly it does that TagResidualAux doesn't do. I guess one big benefit to it is that we could explain in the documentation for it how you need to set up assembly to the tags in the kernels to actually get the reaction forces from this object, which wouldn't be a good fit in the framework documentation.

lindsayad:
I think it would fit in the framework documentation. I still think two classes makes the most sense with both in the framework. But it's not going to be the hill that I die on

I think renaming the parameter to something like Gary suggested including language about undoing variable scaling would help eliminate confusion.

GiudGiud:
>  just thought about the fact that we are also using
tagvector/matrix aux for linear systems where we dont compute residuals and
jacobians. 

Yes we dont cover this well here. We need a TagAssemblyVectorAux essentially, instead of TagResidualAux maybe

dschwen:
> Yes we dont cover this well here. We need a TagAssemblyVectorAux essentially, instead of TagResidualAux maybe

While technically more correct it might start getting incomprehensible for most users.

GiudGiud:
We could just use a LinearFVAssemblyTagAux that is inside the NavierStokes module if preferred

dschwen:
Yeah, just register a bunch of aliases in the modules

hugary1995:
I would disagree with that. There's nothing different in a linear system. The concept of residual, solution, and Jacobian isn't specific to a nonlinear system. In a linear system you still have residual, solution, and Jacobian.

GiudGiud:
yes but that is not how we solve them in the linear FV system. We solve Ax = b not Ax - b = 0
And that (residual, solution) is not what we tally using the Tagging system.

hugary1995:
I just renamed all `vector_tag` to `residual_tag`. You can still tag it to names like "rhs" -- no one stopping you from doing that. Well, you were also correct in that you can always alias something and/or rename parameters.

framework/include/auxkernels/TagAuxBase.h
```diff
@@ -50,16 +50,23 @@ TagAuxBase<T>::validParams()
...
-                        "Return value depending on the variable scaling/autoscaling. Set this to "
-                        "false to obtain unscaled physical reaction forces.");
+  params.addDeprecatedParam<bool>(
+      "scaled",
+      true,
+      "Return value depending on the variable scaling/autoscaling. Set this to false to obtain "
+      "unscaled values with units consistent with the rest of the simulation.",
+      "This parameter has been replaced by remove_variable_scaling.");
+  params.addParam<bool>(
+      "remove_variable_scaling", false, "If true, the variable scaling will be removed.");
```
lindsayad, line 60: We suppress this in the `TagSolutionAux` class and we do not yet have an implementation for this parameter in the tagged matrix aux kernel. I think this should be moved to `TagResidualAux`

framework/src/auxkernels/TagVectorAux.C
```diff
@@ -1,35 +0,0 @@
...
-
-#include "TagVectorAux.h"
-
-registerMooseObject("MooseApp", TagVectorAux);
-
-InputParameters
-TagVectorAux::validParams()
-{
-  InputParameters params = TagAuxBase<AuxKernel>::validParams();
-  params.addRequiredParam<TagName>("vector_tag", "Tag Name this Aux works on");
```
lindsayad, line 18: Given the comments about the linear system vectors, I wonder if it's worth keeping this class around and having `TagResidualAux` inherit from it and `params.renameParam("vector_tag", "residual_tag", "new_docstring");`, similarly for `TagSolutionAux`. The new scaling parameter would still go into `TagResidualAux`

hugary1995:
I was shocked when I was doing the code sweep for the renaming. We only have three vector tag types currently
```cpp
enum VectorTagType
{
  VECTOR_TAG_RESIDUAL = 0,
  VECTOR_TAG_SOLUTION = 1,
  VECTOR_TAG_ANY = 2
};
```
I think it's a mistake not to add tag types for absolute value residual, "not zeroed" residual, and rhs, etc. They are truly different types of vectors that can't be mixed up. I feel like the tagging system has been operating based on the assumption that the users know all the differences between these tag types.

I think this is becoming too much of a change for me to do. I guess I'll just revert all of the changes and only keep the bits and pieces about TagVectorAux, TagResidualAux, TagSolutionAux, and ReactionForceAux.

framework/src/auxkernels/TagVectorAux.C
hugary1995: Sure, I'll keep this one.

framework/include/auxkernels/TagAuxBase.h
hugary1995: That's true.

hugary1995:
My suggestion for a future refactor would be to
1. Add more vector tag types
```cpp
enum VectorTagType
{
  RESIDUAL,
  ABS_RESIDUAL,
  NOT_ZEROED_RESIDUAL,
  SOLUTION,
  RHS,
  ANY 
};
```
2. Add the corresponding APIs like `hasVectorTag`, `hasResidualTag`, `hasAbsResidualTag`, `hasRHSTag`, etc.
3. Add more tag parameters in FEProblemBase such as `extra_tag_residuals`, `tag_absolute_value_residuals`, `tag_not_zeroed_residuals`, `extra_tag_rhs`, etc.
4. Similarly, add more tag parameters in TaggingInterface.

This is too much for me to do unless someone from the framework team would like to take this over. I'll keep this branch and close this PR. I'll start a new PR with a different branch.

hugary1995:
I also find "extra_tag_xxx" vs "extra_xxx_tag" very confusing. Some standardization would be nice, probably following the convention of "extra_xxx_tag".

lindsayad:
Whether signed residuals or the absolute value of the residual is loaded into a tagged vector is a decision made on a per-residual object basis. I think there's a fundamental difference between the tag type of a vector vs. the operations made on it

lindsayad:
The `not_zeroed_tag_vectors` stuff should be fixed-up. I don't think it should introduce new vectors but instead signal what vectors added in `extra_tag_vectors` (or whatever we rename that parameter to) should not be zeroed

lindsayad:
> I also find "extra_tag_xxx" vs "extra_xxx_tag" very confusing. Some standardization would be nice, probably following the convention of "extra_xxx_tag".

I totally agree with this

hugary1995:
> Whether signed residuals or the absolute value of the residual is loaded into a tagged vector is a decision made on a per-residual object basis. I think there's a fundamental difference between the tag type of a vector vs. the operations made on it

I see. I can buy that. Though mathematically this makes little to no sense to me. It's very hard for me to think of a use case for absolute value contribution to the residual. I wonder what the motivation was for even supporting it.

There could be a funny case where a residual object writes absolute values into a residual vector, and others do not, i.e.
r = r_1 + | r_2 | + r_3
what does that even mean?

This is also bad in the sense that the absolute value operation depends on the grouping of terms in the weak form, for example
r = | r_1 + r_2 | + | r_3 |
is different from
r = | r_1 | + | r_2 + r_3 |
so different combinations of kernels that resulting in the same PDE weak form will give different absolute value residual...

lindsayad:
> It's very hard for me to think of a use case for absolute value contribution to the residual. I wonder what the motivation was for even supporting it.

I believe @tophmatthews drafted the initial work. I vaguely remember that @bwspenc supported it. I equally vaguely recall that this was to help build a more robust reference residual(?)

bwspenc:
> > It's very hard for me to think of a use case for absolute value contribution to the residual. I wonder what the motivation was for even supporting it.
> 
> I believe @tophmatthews drafted the initial work. I vaguely remember that @bwspenc supported it. I equally vaguely recall that this was to help build a more robust reference residual(?)

That's right. We're using it regularly for the reference residual (i.e. the denominator in the relative convergence check), and it's working very well for that purpose. Please don't remove it!

hugary1995:
Why can't you take the absolute value of the reference residual vector directly inside ReferenceResidualConvergence? In fact, you already have a parameter called "normalization_type" with a bunch of options. Why bother taking the absolute value inside the residual object?

lindsayad:
Presumably because you get some cancellation from your various residual object contributions

bwspenc:
That's right -- we want to save off the values before they get cancelled out by the contributions from other things that assemble to them.

hugary1995:
That is still... hacky. Whatever, this is derailing the original discussion. But to be honest, something "working very well" for your cases doesn't necessarily mean it's correct. I still can't make sense of it. The fact that this hackiness only exists in solid mechanics and contact makes me somewhat sad.

hugary1995:
From a framework point of view though, if we keep this, when one residual object accumulates absolute value into a tagged vector, shouldn't we prevent other residual objects from accumulating non-absolute-value into the same tagged vector? @lindsayad 

Does that warrant a separate vector type?

hugary1995:
If we let this develop, in some future day, we may have something like

r = r_1 + |r_2| + exp(r_3) + sin(r_4) + cos(r_5)

and that would be total madness.

hugary1995:
Anyhow, this is replaced by #31448 which has a much reduced scope.

I think we have reached consensus on almost all the relevant issues in this thread. I do hope at some point we can address all of them. The tagging system requires some clean up.

tophmatthews:
> The fact that this hackiness only exists in solid mechanics and contact makes me somewhat sad.

Coming in late here, but not sure that is a fair assessment, I think rather it hasn't been widely adopted. I use it for all input files I touch now, from thermo-chemical-mechanical simulations to cluster dynamics/rate theory. I can't argue that there isn't a better implementation out there, you guys are the pros on that, but I will certainly defend the usefulness of the absolute value contributions to the residual as a convergence check.

lindsayad:
> From a framework point of view though, if we keep this, when one residual object accumulates absolute value into a tagged vector, shouldn't we prevent other residual objects from accumulating non-absolute-value into the same tagged vector? @lindsayad
> 
> Does that warrant a separate vector type?

To be honest I don't care what random things people do with their extra residual vectors. They're not touching the core residual vector so they can do

> r = r_1 + |r_2| + exp(r_3) + sin(r_4) + cos(r_5)

as far as I'm concerned. The thing I will stand firm on is not introducing tagged vector types that correspond to a given kind of operation. Only having solution and residual (and whatever we decide for the linear system) tag types feels right to me. You are hypothesizing "mad" scenarios that have not come to pass. There are lots of things that keep me up at night but this isn't one of them

hugary1995:
I'm fine with all that. I am against further spread of the use of absolute vector tags in the solid mechanics module.

tophmatthews:
> I'm fine with all that. I am against further spread of the use of absolute vector tags in the solid mechanics module.

Enlighten me on a better way to compute convergence then? Vanilla relative is not connected to physical reality, vanilla absolute doesn't work if you include other variables that have different residuals, and vanilla reference residual doesn't work if you move beyond simple basic cases?

hugary1995:
Uhh please don't take me wrong -- I am fully aware of the importance of a robust and practical convergence criterion, and I'm more than happy to discuss with you about a potentially better convergence check for your cases.

For the reasons I mentioned above, the absolute value residual is prone to regrouping of residual terms. If we keep adding more tests using that, it would be a huge maintenance burden if we ever decide to change the residual group in our kernels. For that reason, I say
> I am against further spread of the use of absolute vector tags in the solid mechanics module.

which is purely driven by maintenance considerations. I think before we come up with a better, more robust criterion, we should try to refrain from adding more regression tests relying on that particular capability.

joshuahansel:
A while back, I had independently thought of the idea of using the absolute value of residual contributions as the denominator in relative criteria, so I can see some reasoning to get there, but of course I also see that it's not perfect.

We've implemented the Convergence system, so now you can use whatever criteria you can dream up. I'm thinking maybe we should all come up with some general advice on "good" criteria, and we can implement whatever is missing to facilitate it. My thoughts so far:

- Step convergence criteria for some set of variables (either the solution variables or any mixture of solution+aux variables)
- Loose residual criteria to guard against convergence of step criteria when we've just artificially stagnated in the step criteria.

Of course, residual criteria is still a component here, but it's at least not the main criteria. For the residual criteria:

- Remove as much mesh-size and domain-size bias as possible
- For the denominator, use some reference value that depends on the physics. This part is the art. It might be some constant times a corresponding source term.

This might require some experimentation for each application. You can take a large number of iterations and plot out the error of some QoIs (taking the final iterate as the "exact" value) as a function of the residual error and test out different normalizations.

PR #31389, opened 2025-08-28
switch from absoluteFuzzyEqual to relative in the fillGeneralOrthotro…

### LINKED ISSUE #31270: Change absoluteFuzzyEqual to relativeFuzzyEqual in fillGeneralOrthotropicFromInputVector
### Bug Description

When using the absoluteFuzzyEqual without an ability to specify the tolerance it is incredibly difficult to meet the symmetry checks when using values on the order of 1.00e12 which is normal for Pa/m. Most of the examples I have seen seem to move towards reducing the precision of the input values.

I am trying to create a functional version of the ComputeElasticityTensor which calculates the minor PRs from the symmetry condition. But its unable to pass the abs equal check, even when manually comparing the values is 0.0

### Steps to Reproduce

https://github.com/idaholab/moose/discussions/25325

The way around here was to reduce the precision.

### Impact

This causes values that for the most part are valid entries to be flagged as errors. This gets more difficult to battle when creating a functional version of the ComputeElasticityTensor. 

A relative tolerance should be ok here since the values are expected to be large enough...though could have some sort of logic to switch based on expected magnitude. 

### [Optional] Diagnostics

_No response_

vtnick:
Switching from an absolute check to relative given the scales of the numbers here it makes more sense for the relative check. Also brings in line with the symmetric case.

Closes #31270

### DIFF
--- a/framework/include/utils/RankFourTensorImplementation.h
+++ b/framework/include/utils/RankFourTensorImplementation.h
@@ -806,9 +806,9 @@ RankFourTensorTempl<T>::fillGeneralOrthotropicFromInputVector(const std::vector<
   const T & nubc = input[11];
 
   // Input must satisfy constraints.
-  bool preserve_symmetry = MooseUtils::absoluteFuzzyEqual(nuab * Eb, nuba * Ea) &&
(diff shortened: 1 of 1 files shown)


GiudGiud (approved):
note that RankTwoTensor are still using absolute tolerances

PR #31396, opened 2025-08-28
Only print trace and abort from a single thread

lindsayad:
`print_trace` relies on static command line information and this is destroyed by exit handlers triggered by `MPI_Abort`

This is part of the continuing saga to address failures like those on libmesh/libmesh#4236 and libmesh/libmesh#4237

### DIFF
--- a/framework/src/base/MooseError.C
+++ b/framework/src/base/MooseError.C
@@ -57,43 +57,55 @@ mooseErrorRaw(std::string msg,
   if (Moose::_throw_on_error)
     throw MooseRuntimeError(msg, node);
 
-  // If we have a node available, add in the hit context (file location)
-  if (node)
-    msg = Moose::hitMessagePrefix(*node) + msg;
-
-  msg = mooseMsgFmt(msg, "*** ERROR ***", COLOR_RED);
-
-  std::ostringstream oss;
-  oss << msg << "\n";
-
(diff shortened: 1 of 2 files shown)


framework/src/base/MooseError.C
```diff
@@ -57,43 +57,55 @@ mooseErrorRaw(std::string msg,
...
-
+  static std::atomic_flag aborting = ATOMIC_FLAG_INIT;
+  if (aborting.test_and_set(std::memory_order_acq_rel))
+    // Another thread has gone before us. MPI_Abort, despite its name, does not behave like
+    // std::abort but instead calls exit handlers and destroys statics. So we don't want to touch
+    // anything static at this point. We'll just wait until the winning thread has incurred program
+    // exit
+    for (;;)
+      // Don't burn cpu
+      sched_yield();
```
roystgnr: IIRC `sched_yield()` *does* burn some CPU; it re-triggers the kernel scheduler with each call.  Can we just `sleep(1e9)`?

lindsayad: How about `pause()`?

roystgnr (approved):

lindsayad:
failure due to multiple base shas. This is my best hope to fixing the PR failures on libmesh

PR #31405, opened 2025-08-29
Globally constrained current flow on closed toroidal domain

### LINKED ISSUE #31404: Add MFEM example enforcing a global constraint on a domain of non-trivial topology
### Motivation

The need to enforce global constraints on flows around embedded domains of non-trivial topology (such as toroids) is a common requirement in electromagnetic problems, in order to specify non-zero current flows around internal conductors. Such domains are commonly used to approximate coil geometries for magnet systems, and the ability to strongly enforce these constraints is important to enable the future simulation of the response of highly nonlinear conductors in the domain.

### Design

Following the method described [here](https://orbi.uliege.be/bitstream/2268/191358/1/J.00.4-ICS-PDular-EdgNodCoupl-PDular.pdf), this could be achieved by introduce a cut plane into our closed domain and defining a new `mfem::SubMesh` object on it, consisting of all elements that lie on one side of the cut plane and have at least one vertex touching it. 

A source field can then be specified on this 'thick cut' region that can drive a net flow - this could be done with an initial condition imposed on the cut plane. This object will need to modify parent mesh attribute IDs in order to define the new subdomain.

### Impact

Existing APIs should not require modification - this proposed design adds a new SubMesh object to allow specification of the domain of the new source term.

alexanderianblair:
Adds an MFEM example enforcing a global (topological) constraint on the loop voltage of a closed toroidal conductor to drive a non-trivial current flow. Closes #31404.

Solution originally based on the method detailed [here](https://orbi.uliege.be/bitstream/2268/191358/1/J.00.4-ICS-PDular-EdgNodCoupl-PDular.pdf), as implemented in the MFEM-based library [hephaestus](https://github.com/aurora-multiphysics/hephaestus/blob/master/src/sources/closed_coil.cpp), but with a number of fixes to the hephaestus implementation for adding transition regions in parallel and from multiple domains.

Adds a new SubMesh class to specify the set of elements adjacent to a planar cut surface, used to define the (constrained) source field.

Also adds a number of auxiliary AuxKernels and ICs as required to impose the constraint and for testing:

- `MFEMSumAux` for storing the sum of two `MFEMVariables`, used to combine the external and internal fields here
- `MFEMScalarBoundaryIC`, for specifying an initial condition on a scalar variable on a boundary, used here set the constrained values of an auxiliary source potential on the cut plane
- `MFEMMixedGradGradKernel` and `MFEMMixedVectorMassKernel`, to add `MixedBilinearFormIntegrators` similar to the diffusion and vector mass integrators, but for cases where the test and trial variable may differ - used here for adding source terms to the weak form
-  `MFEMVectorFEInnerProductIntegralPostprocessor`, to report the integral of the dot product of two vector MFEM variables, used here to report the power dissipated by the coil as a test metric.

Additional documentation has also been added to the MFEM examples index detailing the method implemented in the new test.

image


framework/doc/content/source/mfem/kernels/MFEMMixedGradGradKernel.md
```diff
@@ -0,0 +1,35 @@
+# MFEMMixedGradGradKernel
+
+!if! function=hasCapability('mfem')
+
+## Summary
+
+!syntax description /Kernels/MFEMMixedGradGradKernel
+
```
lindsayad:
```suggestion
```

lindsayad: redundant with the overview

framework/doc/content/source/mfem/kernels/MFEMMixedGradGradKernel.md
```diff
@@ -0,0 +1,35 @@
...
+!equation
+(k\vec\nabla u, \vec\nabla v)_\Omega \,\,\, \forall v \in V
+
+where $u$ and $v$ are both $\in H^1$, and
+$k$ is a scalar coefficient.
+
+This term arises from the weak form of the Laplacian operator
+
+!equation
+k \vec\nabla u
```
lindsayad:
```suggestion
\nabla\cdot k \vec\nabla u
```

framework/doc/content/source/mfem/kernels/MFEMMixedGradGradKernel.md
lindsayad: consider removing these lines for all the new objects in this PR if they are similarly redundant

framework/doc/content/source/mfem/submeshes/MFEMCutTransitionSubMesh.md
```diff
@@ -0,0 +1,64 @@
...
+!syntax description /SubMeshes/MFEMCutTransitionSubMesh
+
+## Overview
+
+An `MFEMCutTransitionSubMesh` specifies and builds an `mfem::ParSubMesh` object from a
+user-specified (interior or exterior) planar cut boundary in a specified closed volumetric
+subdomain, consisting of all elements that:
+
+- Have at least one vertex that lies on the boundary,
+- Lie on one side of the boundary, and
```
lindsayad, line 16: Are there any fuzzy comparisons here?

framework/doc/content/syntax/MFEM/ClosedCoilMagnetostatic.md
```diff
@@ -0,0 +1,174 @@
...
+We must now find a suitable choice of basis ${\vec c_i \in H(\mathrm{curl})(\Omega_c)}$ for the
+external electric field, such that
+
+\begin{equation}
+    \vec E_{ext} = - \sum_{i \in \Omega_c} V_i \vec c_i
+\end{equation}
+
+for some unique set of degrees of freedom $\{V_i\}$. We shall also wish to restrict the support of $\{\vec c_i\}$ to the smallest number of elements possible in $\Omega_c$ to minimise the number of degrees of freedom required to represent $\vec E_{ext}$.
+
+To do this, we first identify a 'cut' surface $\Gamma_c$ of the toroidal the conductor that is required to make it simply connected.
```
lindsayad: This sentence doesn't seem quite right

framework/doc/content/syntax/MFEM/ClosedCoilMagnetostatic.md
```diff
@@ -0,0 +1,174 @@
...
+Next, we define a one element wide 'transition' region $\Omega_t \subset \Omega_c$
+on one side of the cut surface, using an `MFEMCutTransitionSubMesh` object, comprised of all
+elements with at least one vertex lying on the cut surface that lie on one side of the cut.
+
+Next, to enforce $\vec \nabla \times \vec E_{ext} =0$ in this region, we define a scalar variable $v$ defined on a (nodal) $H^1$ conforming finite element space on this submesh such that $\vec E_{ext} = -\vec \nabla v$:
```
lindsayad:
```suggestion
Next, to enforce $\vec \nabla \times \vec E_{ext} =0$ in this region, we define a scalar field variable $v$ defined on a (nodal) $H^1$ conforming finite element space on this submesh such that $\vec E_{ext} = -\vec \nabla v$:
```
This helps clarify that this is not a "SCALAR" (generally single-dof) variable (libMesh/classic-MOOSE terminology)

framework/include/mfem/auxkernels/MFEMSumAux.h
```diff
@@ -0,0 +1,45 @@
...
+#ifdef MOOSE_MFEM_ENABLED
+
+#pragma once
+#include "libmesh/ignore_warnings.h"
+#include "mfem/miniapps/common/pfem_extras.hpp"
+#include "libmesh/restore_warnings.h"
+#include "MFEMAuxKernel.h"
+
+/*
+Class to scale and sum two MFEM variables, and store the result in a third variable.
```
lindsayad:
```suggestion
 * Class to scale and sum two MFEM variables, and store the result in a third variable.
```

framework/include/mfem/auxkernels/MFEMSumAux.h
```diff
@@ -0,0 +1,45 @@
...
+
+  virtual ~MFEMSumAux() = default;
+
+  // Computes the auxvariable.
+  virtual void execute() override;
+
+protected:
+  // Names of input MFEMVariables to sum.
+  const VariableName _v1_var_name;
+  const VariableName _v2_var_name;
```
lindsayad:
```suggestion
  const VariableName & _v1_var_name;
  const VariableName & _v2_var_name;
```

framework/include/mfem/kernels/MFEMTimeDerivativeVectorFEMassKernel.h
```diff
@@ -0,0 +1,36 @@
...
+class MFEMTimeDerivativeVectorFEMassKernel : public MFEMVectorFEMassKernel
+{
+public:
+  static InputParameters validParams();
+
+  MFEMTimeDerivativeVectorFEMassKernel(const InputParameters & parameters);
+
+  /// Get name of the trial variable (gridfunction) the kernel acts on.
+  /// Defaults to the name of the test variable labelling the weak form.
+  virtual const VariableName & getTrialVariableName() const override { return _var_dot_name; };
```
lindsayad:
```suggestion
  virtual const VariableName & getTrialVariableName() const override { return _var_dot_name; }
```

framework/include/mfem/submeshes/MFEMCutTransitionSubMesh.h
```diff
@@ -0,0 +1,54 @@
...
+
+#ifdef MOOSE_MFEM_ENABLED
+
+#pragma once
+#include "MFEMSubMesh.h"
+#include "MFEMBlockRestrictable.h"
+
+/**
+ * Modifies the MFEM Mesh to label a subdomain consisting of elements adjacent to an
+ * interior surface on one side, and constructs an stores an mfem::ParSubMesh object
```
lindsayad:
```suggestion
 * interior surface on one side, and constructs and stores an mfem::ParSubMesh object
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+      "transition_subdomain_boundary",
+      "Name to assign boundary of transition subdomain not shared with cut surface.");
+  params.addRequiredParam<SubdomainName>(
+      "transition_subdomain",
+      "The name of the subdomain to be created on the mesh comprised of the set of elements "
+      "adjacent to the cut surface on one side.");
+  params.addRequiredParam<SubdomainName>(
+      "closed_subdomain",
+      "The name of the subdomain attribute to be created comprised of the set of all elements "
+      "of the closed goemetry, including the new transition region.");
```
lindsayad:
```suggestion
      "of the closed geometry, including the new transition region.");
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+  _submesh->bdr_attribute_sets.attr_sets = getMesh().bdr_attribute_sets.attr_sets;
+}
+
+void
+MFEMCutTransitionSubMesh::labelMesh(mfem::ParMesh & parent_mesh)
+{
+  int mpi_comm_rank = 0;
+  int mpi_comm_size = 1;
+  MPI_Comm_rank(getMFEMProblem().mesh().getMFEMParMesh().GetComm(), &mpi_comm_rank);
+  MPI_Comm_size(getMFEMProblem().mesh().getMFEMParMesh().GetComm(), &mpi_comm_size);
```
lindsayad, line 72: It doesn't necessarily have to be in this PR, but it would be nice to convert these raw MPI calls to TIMPI calls which have a much more C++ feel. If you understandably don't want to do it here, let me know and I'll make an issue

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+  MPI_Bcast(_cut_normal.GetData(),
+            _cut_normal.Size(),
+            MPI_DOUBLE,
+            rank_with_submesh,
+            getMFEMProblem().mesh().getMFEMParMesh().GetComm());
+  /// Iterate over all vertices on cut, find elements with those vertices,
+  /// and declare them transition elements if they are on the +ve side of the cut
+  mfem::Array<int> transition_els;
+  std::vector<int> global_cut_vert_ids;
+  mfem::Array<HYPRE_BigInt> gi;
```
lindsayad, line 99: why not `int` here? You assign to elements in `global_cut_vert_ids` from this container so I think we should be consistent

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+                 &n_vert_offset[0],
+                 MPI_INT,
+                 getMFEMProblem().mesh().getMFEMParMesh().GetComm());
+
+  /// Detect shared vertices and add corresponding elements
+  for (int g = 1, sv = 0; g < parent_mesh.GetNGroups(); g++)
+  {
+    for (int gv = 0; gv < parent_mesh.GroupNVertices(g); gv++, sv++)
+    {
+      // all els touching this shared vertex plvtx should be updated
```
lindsayad: "plvtx"?

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+      int cut_vert = parent_mesh.GroupVertex(g, gv);
+      for (size_t i = 0; i < all_cut_verts.size(); i += 1)
```
lindsayad:
```suggestion
      for (std::size_t i = 0; i < all_cut_verts.size(); i += 1)
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+            MPI_DOUBLE,
+            rank_with_submesh,
+            getMFEMProblem().mesh().getMFEMParMesh().GetComm());
+  /// Iterate over all vertices on cut, find elements with those vertices,
+  /// and declare them transition elements if they are on the +ve side of the cut
+  mfem::Array<int> transition_els;
+  std::vector<int> global_cut_vert_ids;
+  mfem::Array<HYPRE_BigInt> gi;
+  parent_mesh.GetGlobalVertexIndices(gi);
+  mfem::Table * vert_to_elem = parent_mesh.GetVertexToElementTable();
```
lindsayad:
```suggestion
  std::unique_ptr<mfem::Table> vert_to_elem = parent_mesh.GetVertexToElementTable();
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+            if (isInDomain(el_adj_to_cut, getSubdomainAttributes(), parent_mesh) &&
+                sideOfCut(el_adj_to_cut, cut_vert, parent_mesh) == 1)
+              transition_els.Append(el_adj_to_cut);
+          }
+        }
+      }
+    }
+  }
+
+  delete vert_to_elem;
```
lindsayad:
```suggestion
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+{
+  const int sdim = parent_mesh.SpaceDimension();
+  mfem::Vector el_center(3);
+  parent_mesh.GetElementCenter(el, el_center);
+  mfem::Vector vertex_coords(parent_mesh.GetVertex(el_vertex_on_cut), sdim);
+  mfem::Vector relative_center(sdim);
+  for (int j = 0; j < sdim; j++)
+  {
+    relative_center[j] = el_center[j] - vertex_coords[j];
+  }
```
lindsayad:
```suggestion
    relative_center[j] = el_center[j] - vertex_coords[j];
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+    new_attrs[i] = i + 1;
+  for (int i = 0; i < transition_els.Size(); ++i)
+  {
+    const auto & transition_el = transition_els[i];
+    const int old_attr = parent_mesh.GetAttribute(transition_el);
+    new_attrs[old_attr - 1] = old_max_attr + old_attr % old_max_attr + 1;
+  }
+
+  /// Distribute local attribute IDs to other MPI ranks to create set of new
+  /// global attribute IDs for the transition region.
```
lindsayad: should use `//` for in-line comments

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+
+  mfem::AttributeSets & attr_sets = parent_mesh.attribute_sets;
+  mfem::AttributeSets & bdr_attr_sets = parent_mesh.bdr_attribute_sets;
+  /// Create attribute set labelling the newly created transition region
+  /// on one side of the cut
+  attr_sets.CreateAttributeSet(_transition_subdomain);
+  /// Create attribute set labelling the entire closed geometry
+  attr_sets.SetAttributeSet(_closed_subdomain, getSubdomainAttributes());
+  /// Add the new domain attributes to new attribute sets
+  std::set<std::string> attr_set_names = attr_sets.GetAttributeSetNames();
```
lindsayad:
```suggestion
  const std::set<std::string> attr_set_names = attr_sets.GetAttributeSetNames();
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+
+  parent_mesh.SetAttributes();
+}
+
+bool
+MFEMCutTransitionSubMesh::isInDomain(const int & element,
+                                     const mfem::Array<int> & subdomains,
+                                     const mfem::ParMesh & mesh)
+{
+  bool is_in_domain = false;
```
lindsayad:
```suggestion
```

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -0,0 +1,317 @@
...
+  /// element<0 for ghost elements
+  if (element < 0)
+    return true;
+
+  for (const auto & subdomain : subdomains)
+  {
+    if (mesh.GetAttribute(element) == subdomain)
+      is_in_domain = true;
+  }
+  return is_in_domain;
```
lindsayad:
```suggestion
    if (mesh.GetAttribute(element) == subdomain)
      return true;
  return false;
```

test/tests/mfem/submeshes/av_magnetostatic.i
```diff
@@ -0,0 +1,163 @@
...
+[]
+
+[AuxKernels]
+  [curl]
+    type = MFEMCurlAux
+    variable = b_field
+    source = a_field
+    scale_factor = 1.0
+    execute_on = TIMESTEP_END
+    execution_order_group = 3
```
lindsayad, line 51: I forget whether we have an issue open for investigating how we can eliminate the need for `execution_order_group` in MFEM inputs? If we don't, we should. This is the kind of thing that a framework should provide. If users need to perform direct dependency resolution intervention like this, I don't think we're doing the best job we can wrapping MFEM

test/tests/mfem/submeshes/tests
```diff
@@ -49,4 +49,38 @@
...
+    design = 'syntax/MFEM/ClosedCoilMagnetostatic.md'
+    issues = '#31404'
+    type = CSVDiff
+    input = hphi_magnetodynamic.i
+    csvdiff = 'OutputData/HPhiMagnetodynamicClosedCoilCSV.csv'    
+    requirement = 'MOOSE shall have the ability to apply boundary conditions on boundaries of embedded volumetric subdomains.'
+    capabilities = 'mfem'
+    compute_devices = 'cpu cuda'
+    recover = false
+    restep = false
```
lindsayad, line 84: This test spec needs this but the others don't?

framework/doc/content/source/mfem/submeshes/MFEMCutTransitionSubMesh.md
alexanderianblair: No, there shouldn't be - the set of vertices is generated from the vertices of the boundary elements themselves, and the orientation check uses the element centre relative to the element vertex on the boundary (which should never be coincident AFAICT)

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
alexanderianblair: Happy to do this, but I would indeed like to separate it out from this PR if possible!

test/tests/mfem/submeshes/tests
alexanderianblair: Unlike the other two, this one is a transient test, which is why `restep = false` is needed here - mainly because the magnetodynamic example is a nice user-relevant test for applying a DirichletBC on an internal DomainSubMesh boundary, but also because it helps improve coverage for the other postprocessors/auxkernels used in this set of tests to include transient problems

test/tests/mfem/submeshes/av_magnetostatic.i
alexanderianblair: I agree - I've added one up at #31434. Need to dig in to how the dependency resolver should be set up for user objects - the affected classes all have `UserObject` (and thus `DependencyResolverInterface`) already in their inheritance tree

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
alexanderianblair: Removed (came from an outdated name within MFEM for a similar loop)

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
alexanderianblair: Used `HYPRE_BigInt` here since the `GetGlobalVertexIndices` method acts on an array of type `mfem::Array<HYPRE_BigInt`. Agree this should be consistent with `global_cut_vert_ids` though - I've updated the latter to use `HYPRE_BigInt` as well, to be consistent using `HYPRE_BigInt` for global vertex IDs here.

framework/doc/content/syntax/MFEM/ClosedCoilMagnetostatic.md
alexanderianblair: Indeed - removed extra 'the'

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
lindsayad: I opened https://github.com/idaholab/moose/issues/31466

lindsayad (approved):

PR #31417, opened 2025-09-01
Xfem mark

donggiang:
## Reason
Crack tip locations in XFEM code are not consistently updated through the object of **CrackFrontDefinition**. Additionally, to allow crack propagation, additional block to define the crack tip location is required in the input file, while the tip location is already handled by the object **MeshCut2DFractureUserObject**.

## Design
Crack tip locations handled by CrackFrontDefinition  will be automatically updated through a block in the input file defining **MeshCut2DFractureUserObject**. Classes that have CrackFrontDefinition as their member need to initialize it by calling initialSetup() and getting CrackFrontDefinition pointer from **_fe_problem**.
## Impact
This improvement allows consistency of crack front points as the tips move to new elements, avoiding convergence issues in inelastic problems with crack tip enriched solutions.


lynnmunday:
@donggiang We need to split this up into two separate PRs if we can.  One for the XFEM mark flag.  And a seperate PR for the enrichment element changes.  I want to see what changes are needed for only the XFEM mark flag.  I thought we could limit the changes to the solid_mechanics and xfem modules. I want to isolate those changes from the changes we need for the enrichment functions.  

I want to know if we can update the xfem crack tip without making changes to transient.C and fixedPointSolve.C. I can help you with this on Monday.

donggiang:
> @donggiang We need to split this up into two separate PRs if we can. One for the XFEM mark flag. And a seperate PR for the enrichment element changes. I want to see what changes are needed for only the XFEM mark flag. I thought we could limit the changes to the solid_mechanics and xfem modules. I want to isolate those changes from the changes we need for the enrichment functions.
> 
> I want to know if we can update the xfem crack tip without making changes to transient.C and fixedPointSolve.C. I can help you with this on Monday.

@lynnmunday I separeted the original branch to two. Now, this branch only has changes to the XFEM_mark flag.

lindsayad (dismissed):
This doesn't feel right to me. I thought the `ExecFlagRegistry` was for the purpose of not having to pollute execution flags known by the framework with flags from modules

lindsayad:
@loganharbour ?

framework/src/base/Moose.C
```diff
@@ -31,6 +31,7 @@ const ExecFlagType EXEC_INITIAL = registerDefaultExecFlag("INITIAL");
 const ExecFlagType EXEC_LINEAR = registerDefaultExecFlag("LINEAR");
 const ExecFlagType EXEC_LINEAR_CONVERGENCE = registerDefaultExecFlag("LINEAR_CONVERGENCE");
 const ExecFlagType EXEC_NONLINEAR = registerDefaultExecFlag("NONLINEAR");
+const ExecFlagType EXEC_XFEM_MARK = registerDefaultExecFlag("XFEM_MARK");
```
lynnmunday: revert this

framework/include/base/Moose.h
```diff
@@ -122,6 +122,7 @@ extern const ExecFlagType EXEC_LINEAR_CONVERGENCE;
 extern const ExecFlagType EXEC_LINEAR;
 extern const ExecFlagType EXEC_NONLINEAR_CONVERGENCE;
 extern const ExecFlagType EXEC_NONLINEAR;
+extern const ExecFlagType EXEC_XFEM_MARK;
```
lynnmunday: revert this

modules/xfem/src/base/XFEMAppTypes.C
```diff
@@ -8,6 +8,4 @@
 //* https://www.gnu.org/licenses/lgpl-2.1.html
 
 #include "XFEMAppTypes.h"
-#include "ExecFlagRegistry.h"
-
-const ExecFlagType EXEC_XFEM_MARK = registerDefaultExecFlag("XFEM_MARK");
+#include "ExecFlagRegistry.h"
```
lynnmunday: mv this file to solid_mechanics/src/base and rename something like:  solid_mechanicsAppTypes.C and h

Only include the execution flag part of this file. 

In the original XFEMAppTypes.h file move 

typedef unsigned int CutSubdomainID 

to XFEMApp.h right after the includes.

lynnmunday (changes requested):
First run the xfem and solidmechanics tests with the changes I requested below.  If they fail, try to modify the xfem geometric cut user object executioner order flag to -2.

loganharbour:
Unfortunately there's xfem stuff littered all over the framework as its insertion points weren't available in a pluggable way at the time of development.

We should, however, be able to do this dynamically. Register the static variable in XFEM. And then query the registry for that flag (if it exists) in the framework. We'll still have xfem in framework, but it'll be something like `ExecFlagRegistry::getFlag("SOMETHING_XFEM")`.

That's probably the best we can do. @lynnmunday do you need help with this?

lynnmunday:
@loganharbour I think we can get this to work without making any changes to framework.  We only need that flag in the solid_mechanics and xfem modules.

donggiang:
Thanks @lindsayad @loganharbour and @lynnmunday for suggestions. 
However, before addressing them, it seems there are two tests that failed in the XFEM modules on upstream/devel. Has any of you checked this?

Screenshot 2025-09-10 at 09 41 59

lindsayad:
Those tests are passing in CIVET

donggiang:
> Those tests are passing in CIVET

@lindsayad I checkouted upstream/devel, run tests in  module XFEM and got the message above.

lindsayad:
It's possible those tests are brittle and can fail on different architectures/environments

bwspenc (changes requested):
This looks right to me as far as the ordering. As you have this now, we've got the following execution order groups set:
CrackFrontDefinition: -1 (already set in DomainIntegralAction in solid_mechanics)
fracture integral VPPs: 0 (default setting, these are created in DomainIntegralAction)
GeometricCutUserObject: 1 (set in those classes' params, added in this PR)

A few things to do:
1) This is failing the precheck, complaining about trailing whitespace in SolidMechanicsAppTypes.C. It doesn't have trailing whitespace, but it's missing a newline on the last line, which might be triggering it. Try re-saving the file or something to fix that. I was able to do that in my editor. You need to fix that to get the tests to run.
2) Your last commit message is wrong: it says that you set the order to -1, but it was 1. Fix that.
3) I would squash these commits into a single commit since the earlier commits in the history touched framework files that weren't necessary.
4) I think we need to do some more work in the DomainIntegralAction where we're setting the execute flags up. The current behavior there depends on the execute flag being defined in xfem, but I think it isn't right for having it in solid_mechanics. Now I'm wondering whether we even needed to move the execute flag definition at all, actually!
5) It sounds like this does break some tests, so we'll need to investigate those.

modules/xfem/src/userobjects/MeshCut2DFractureUserObject.C
```diff
@@ -94,6 +94,7 @@ MeshCut2DFractureUserObject::initialize()
 {
   _is_mesh_modified = false;
   findActiveBoundaryGrowth();
+
```
bwspenc: This isn't a huge deal, but I would revert the changes to this file that just add whitespace. Otherwise, I think the algorithm is correct now for critical crack growth.

bwspenc (approved):

bwspenc:
It looks like we have a test that fails in parallel: test:solid_mechanics_basic.mesh_cutter_3d/mhs

lynnmunday:
I'll look at the failing test. I'm having trouble with the 3D mesh object for some other stuff.

lynnmunday:
@donggiang    Will you change the commit message for 153c8da8e4c5313dff78da387d792010a23762da
This is wrong:    
set execution_order_group=-1 for geometricCutUserObject.

it should be:
set execution_order_group=**1** for geometricCutUserObject!

I saw this and had to find my notes to figure out the ordering again.

donggiang:
> @donggiang Will you change the commit message for [153c8da](https://github.com/idaholab/moose/commit/153c8da8e4c5313dff78da387d792010a23762da) This is wrong: set execution_order_group=-1 for geometricCutUserObject.
> 
> it should be: set execution_order_group=**1** for geometricCutUserObject!
> 
> I saw this and had to find my notes to figure out the ordering again.

donggiang:
> @donggiang Will you change the commit message for [153c8da](https://github.com/idaholab/moose/commit/153c8da8e4c5313dff78da387d792010a23762da) This is wrong: set execution_order_group=-1 for geometricCutUserObject.
> 
> it should be: set execution_order_group=**1** for geometricCutUserObject!
> 
> I saw this and had to find my notes to figure out the ordering again.

updated

bwspenc:
It was unintentionally closed

bwspenc:
To fix that diffing test, add this line to the 'mhs' test definition in moose/modules/xfem/test/tests/solid_mechanics_basic:
partial=false

modules/xfem/test/tests/solid_mechanics_basic/tests
```diff
@@ -71,6 +71,7 @@
       exodiff = 'edge_crack_3d_mhs_out.e edge_crack_3d_mhs_out.e-s002 edge_crack_3d_mhs_out.e-s003 edge_crack_3d_mhs_out_XFEMCutMeshOutput.e-s0004'
       abs_zero = 1e-4 #crack face is normal to z so this gets rid of exodiffs caused by z-direction variables
       map = false
+      partial=false
```
bwspenc:
```suggestion
      partial = true
```

bwspenc: Sorry, I told you the wrong thing. false is the default, and we need to set it to true for the check to pass on the cutting mesh. It will still fail the test as it is currently.

bwspenc (changes requested):

bwspenc (approved):

bwspenc:
@lindsayad I'm going to dismiss your review when the tests pass (which they should this time around) so we can merge this. This no longer touches framework code, which should address your concern.

bwspenc:
Test failure is unrelated to this PR. Merging.

PR #31441, opened 2025-09-04
Fix variable in app apptainer definition

loganharbour:
Closes #31440


loganharbour:
@dschwen this fixes the issue with setting `WITH_LIBTORCH` for the swift apptainer build

dschwen (approved):

PR #31478, opened 2025-09-10
Resolve mooseError issues introduced in #31396

loganharbour:
refs #31396


lindsayad (approved):
Looks good to me. When the libmesh submodule is updated we can remove some (all?) of the stream locks since `Moose::out` and `Moose:err` will then have thread sync'd buffering

PR #31495, opened 2025-09-11
Adding Metal Fuel Pin Temperature Solver to Subchannel Module

### LINKED ISSUE #30571: 1D Fuel Pin Heat Transfer Model in SCM
### Motivation

The SCM subchannel analysis model currently lacks a dedicated fuel pin heat transfer model. To enable this functionality, coupling with other MOOSE modules is required. Incorporating both steady-state and transient fuel pin heat transfer, along with the associated material properties, is essential for comprehensive subchannel analysis. Providing a user-friendly solution within SCM is therefore critical. Moreover, the upcoming Clad Damage Propagation (CDAP) feature depends on accurate fuel pin heat transfer coupling for its implementation.

### Design

The implementation will be performed without modifying the structure of the existing subchannel classes. A new class, inheriting from LiquidMetalSubChannel1PhaseProblem, will be created to handle the fuel pin heat conduction solution. This class will incorporate fuel pin material properties, solve the heat conduction within the pin, manage heat flux coupling at the pin surface, and override the default SCM solver to integrate these capabilities.

### Impact

This capability will provide an easy-to-use option for incorporating fuel pin heat transfer coupling in liquid metal-cooled fast reactor bundle analysis. It is essential for performing most transient subchannel analyses accurately. In addition, the upcoming Clad Damage Propagation (CDAP) model will require this feature to evaluate failure mechanisms driven by localized thermal conditions.

aydinka:
**Closes #30571**

## Reason
We need a fuel pin temperature solver that is energy-consistent with the subchannel thermal-hydraulics. This enables steady and transient prediction of fuel–gap–clad temperatures for liquid-metal cooled metallic fuel pins, and provides a foundation for future melting behavior and tighter pin↔channel coupling.

## Design
- **New Problem:** `PinTempSolver` (registered for both `MooseApp` and `SubChannelApp`) that solves 1-D **radial** conduction in cylindrical coordinates (fuel → gap → clad).
- **Coupling with subchannel:**
 
## Impact
- Adds a new `Problem` without changing existing APIs or behaviors.
- No effect unless selected in an input file or the subchannel module is loaded.
- Lays groundwork for regression tests and user docs in a follow-up PR.


kyriv-lab:
@aydinka , I think that at least one commit message should mention the issue #. I think that is why the precheck fails. So rename the first commit and add the issue number.

aydinka:
Updated the commit message to reflect the issue number.

kyriv-lab:
@aydinka Did you accidentally close the PR?

aydinka:
@kyriv1980 Not that I am aware. Can re-open or should I open another one?

kyriv-lab:
I think you should be able to re-open from this page. (@GiudGiud ?)

aydinka:
@kyriv1980 yes, it seems open now. Thanks.

kyriv-lab:
Great. That was it. Precheck is passing now.

rui-hu:
@kyriv1980 , please continue the review of the PR.

modules/subchannel/doc/content/source/problems/PinTempSolver.md
```diff
@@ -0,0 +1,21 @@
+# TriSubChannel1PhaseProblem
```
kyriv-lab: Change the name to PinTempSolver

modules/subchannel/doc/content/source/problems/PinTempSolver.md
```diff
@@ -0,0 +1,21 @@
...
+
+!syntax description /Problem/PinTempSolver
+
+## Overview
+
+!! Intentional comment to provide extra spacing
+
+This class solves for the steady state and transient U-Pu-Zr metal fuel pin temperature distribution coupled to subchannel flow solver.
```
kyriv-lab: Please add more information about how this solvers works. Is the model 2D or 3D, maybe write down the equations it solves, boundary conditions and assumptions, and the numerical scheme used.

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,124 @@
...
+
+  //  virtual void initializeSolutionPinTempSolver() override;
+  virtual void initializeSolutionPinTempSolver();
+
+  /// flag for whether jet is alive or not
+  std::vector<int> ijet;
+  /// the rod number for the originating jet
+  std::vector<int> rodjet;
+  /// peak CDF axial location for the clad failure
+  std::vector<int> peak_loc;
```
kyriv-lab:
```suggestion
  /// flag for whether jet is alive or not
  std::vector<unsigned int> ijet;
  /// the rod number for the originating jet
  std::vector<unsigned int> rodjet;
  /// peak CDF axial location for the clad failure
  std::vector<unsigned int> peak_loc;
```

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,124 @@
...
+  //  virtual void PinTempSolverDriver(Real dt, unsigned int pintemp_ss) override;
+  virtual void PinTempSolverDriver(Real dt, unsigned int pintemp_ss);
+  /// TempSolvCDAP
+  void TempSolverTR(Real dt, unsigned int iz, unsigned int i_pin, Real hcoef, Real tcool);
+
+  void TempSolverSS(unsigned int iz, unsigned int i_pin, Real hcoef, Real tcool);
+
+  void set_convective_bc();
+
+  Real computeAddedHeatPin(unsigned int i_ch, unsigned int iz) override;
```
kyriv-lab, line 155: add a description/comment of what each method does.

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,124 @@
...
+  Real _rco;
+  std::vector<Real> _r;
+  std::vector<Real> _dr;
+
+  std::vector<Real> _temp1;
+  std::vector<Real> _temp;
+  std::vector<Real> _temp0;
+  std::vector<Real> _qtrip;
+
+  std::vector<std::vector<Real>> _a;
```
kyriv-lab: same. Add comments of what everything is.

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+          auto w_perim = (*_w_perim_soln)(node);
+          auto Dh_i = 4.0 * S / w_perim;
+          auto Re = (((*_mdot_soln)(node) / S) * Dh_i / mu);
+          auto k = _fp->k_from_p_T((*_P_soln)(node) + _P_out, (*_T_soln)(node));
+          auto cp = _fp->cp_from_p_T((*_P_soln)(node) + _P_out, (*_T_soln)(node));
+          auto Pr = (*_mu_soln)(node)*cp / k;
+          auto Pe = Re * Pr;
+          auto pitch = _subchannel_mesh.getPitch();
+          auto ptod = pitch / _rco / 2.0;
+          auto Nu = 4.0 + 0.33 * pow(ptod, 3.8) * pow(Pe / 100.0, 0.86) + 0.16 * pow(ptod, 5);
```
kyriv-lab, line 219: This should be updated... see https://github.com/idaholab/moose/pull/31547.

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+      auto mu = (*_mu_soln)(node_chan);
+      auto S = (*_S_flow_soln)(node_chan);
+      auto w_perim = (*_w_perim_soln)(node_chan);
+      auto Dh_i = 4.0 * S / w_perim;
+      auto Re = (((*_mdot_soln)(node_chan) / S) * Dh_i / mu);
+      auto k = _fp->k_from_p_T((*_P_soln)(node_chan) + _P_out, (*_T_soln)(node_chan));
+      auto cp = _fp->cp_from_p_T((*_P_soln)(node_chan) + _P_out, (*_T_soln)(node_chan));
+      auto Pr = (*_mu_soln)(node_chan)*cp / k;
+      /// FIXME - model assumes HTC calculation via Dittus-Boelter correlation
+      auto Nu = 0.023 * std::pow(Re, 0.8) * std::pow(Pr, 0.4);
```
kyriv-lab, line 1043: This should be merged with https://github.com/idaholab/moose/pull/31547

kyriv-lab (commented):
I believe this solver should/could become an auxkernel instead of a child of the TriSubChannel1PhaseProblem.

kyriv-lab:
@GiudGiud could this Fuel-Pin heatconduction solver be turned into an auxkernel?

modules/subchannel/src/problems/PinTempSolver.C
GiudGiud: The way it will work is one of the two PRs will get merged first, and the other one will adapt. 

No need to update ahead of time

GiudGiud:
It really depends on:
- is the fluid flow solver influenced by the computations from the Tpin solver? How?
- if so, how often do they need to exchange information? CHT is often done with tight coupling, in which case the solver need to run in lockstep. But maybe this is more relaxed?
We could also have the SCMProblems execute auxkernels inside their iteration loop. It's a tad slower than SCM is, but it's not as slow as FE solver calculations

modules/subchannel/doc/content/source/problems/PinTempSolver.md
```diff
@@ -0,0 +1,68 @@
+# PinTempSolver
+
+!syntax description /Problem/PinTempSolver
+
+## Overview
+`PinTempSolver` provides a 1-D, axisymmetric **radial heat-conduction** model for metallic SFR pins that is tightly coupled to the Subchannel coolant solver. It computes the pin temperature distribution at every axial location and returns the **convective wall heat flux** to the subchannel energy equation. Conversely, it receives the **convective boundary condition** (HTC and bulk coolant temperature) from the subchannel solution. The class inherits from `TriSubChannel1PhaseProblem` so it runs inside the existing subchannel outer iteration.
```
kyriv1980, line 6: What is the user provided linear heat-rate in this case?

aydinka:
User-provided linear heat rate is used to compute the pin temperature
distribution only. The convective heat transfer at the fuel/cladding
interface provides power deposition to the coolant channel.

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,219 @@
...
+   * temperatures (typically with inlet/bulk values).
+   */
+  virtual void initializeSolutionPinTempSolver();
+
+  /** @name Jet override data (optional, per-pin)
+   *  Pins flagged as "jet" use these values instead of the averaged subchannel HTC/temperature.
+   *  @{
+   */
+  /// Flag indicating whether a jet is active on a pin (0/1)
+  std::vector<unsigned int> ijet;
```
kyriv-lab: is there information regarding which subchannel the jet is blowing into?

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,219 @@
...
+   *
+   * Expected parameters include geometry (fuel/clad radii), gap conductance, fuel composition
+   * (Pu, Zr weight fractions), porosity, optional solidus/liquidus/fusion-heat placeholders,
+   * and the number of fuel radial nodes.
+   */
+  static InputParameters validParams();
+
+protected:
+  /// Access to the subchannel triangular mesh (pins/channels/geometry)
+  TriSubChannelMesh & _tri_sch_mesh;
```
kyriv-lab, line 97:
```suggestion
```

aydinka:
Those are actually placeholders for CDAP. CDAP will have some more updates
on PinTempSolver.   Yes CDAP has those variables.

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,219 @@
...
+   * Calls the subchannel flow/thermal updates, initializes the pin solver,
+   * performs steady/transient radial solves per (pin, axial cell), updates wall heat
+   * fluxes for the subchannel energy equation, and publishes auxiliary fields.
+   */
+  virtual void externalSolve() override;
+
+  /**
+   * @brief Driver at the beginning of a time step.
+   * @param dt         Time step size [s]
+   * @param pintemp_ss Steady-state flag (0 = steady, 1 = transient)
```
kyriv-lab, line 111: could turn into a bool?

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,219 @@
...
+  void TempSolverTR(Real dt, unsigned int iz, unsigned int i_pin, Real hcoef, Real tcool);
+
+  /**
+   * @brief Steady-state radial temperature solve for a single (pin, axial) location.
+   * @param iz     Axial index
+   * @param i_pin  Pin index
+   * @param hcoef  Outer-wall HTC [W/m^2/K]
+   * @param tcool  Coolant reference temperature [K]
+   */
+  void TempSolverSS(unsigned int iz, unsigned int i_pin, Real hcoef, Real tcool);
```
kyriv-lab, line 135: Could you possibly combine the transient solver and the ss solver?

modules/subchannel/include/problems/PinTempSolver.h
```diff
@@ -0,0 +1,219 @@
...
+  Real HT9CladCP(Real temp);
+  /// HT9 cladding density rho_c [kg/m^3]
+  Real HT9CladRHO();
+  /** @} */
+
+  /** @name Solver mode and initialization flags @{ */
+  /// Initialization guard (0 = not initialized, 1 = initialized)
+  unsigned int pintemp_init = 0;
+  /// Steady-state flag (0 = steady solve, 1 = transient solve)
+  unsigned int pintemp_ss = 0;
```
kyriv-lab, line 179: turn into bool?

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+{
+}
+
+PinTempSolver::~PinTempSolver() {}
+
+void
+PinTempSolver::initializeSolutionPinTempSolver()
+{
+  _nrpin = _nrfuel + 2;
+  _rco = 0.5 * _subchannel_mesh.getPinDiameter();
```
kyriv-lab, line 64: better to use the Dpin solution for more generality.

kyriv-lab: (*_Dpin_soln)(pin_node)

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+      }
+    }
+  }
+}
+
+void
+PinTempSolver::set_convective_bc()
+{
+
+  auto dpin = 2.0 * _rco;
```
kyriv-lab, line 172: use dpin_soln when applicable.

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+        auto * node_in = _subchannel_mesh.getChannelNode(i_rod, 0);
+        auto T_in = (*_T_soln)(node_in);
+
+        _hcool_pin_ave(iz, i_rod) = 1.0e-15;
+        _tcool_pin_ave(iz, i_rod) = T_in;
+      }
+
+      for (unsigned int i_ch = 0; i_ch < _n_channels; ++i_ch)
+      {
+        _qbarconv_channel(iz, i_ch) = 0.0;
```
kyriv-lab, line 262: you already set to zero. Why set it again?

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+          }
+          else
+          {
+            frac = 1.0 / 4.0;
+          }
+
+          if (ijet[i_rod] == 1)
+          {
+            _qbarconv_channel(iz, i_ch) += frac * libMesh::pi * dpin * h_jet[i_rod] *
+                                           (_temp_pin[i_rod](iz, _nrfuel + 2) - temp_jet[i_rod]);
```
kyriv-lab: Can you set/use the Tpin here?

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+    {
+      _console << " node " << i << " _qtrip " << _qtrip[i] << " temp " << _temp[i] << " b " << _b[i]
+               << " a " << _a[i][i] << std::endl;
+    }
+    exit(0);
+  }
+
+  for (unsigned int i = 0; i < _nrpin + 1; i++)
+  {
+    _temp_pin[i_pin](iz, i) = _temp[i];
```
kyriv-lab, line 632: Can you also set Tpin here? (Tpin is the average pin surface temperature)

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+  }
+}
+
+void
+PinTempSolver::TempSolverTR(Real dt, unsigned int iz, unsigned int i_pin, Real hcoef, Real tcool)
```
kyriv-lab, line 638: Please try to combine if possible the transient with the ss solver.

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+  }
+
+  if (_pin_mesh_exist)
+  {
+    for (unsigned int i_pin = 0; i_pin < _n_pins; i_pin++)
+    {
+      for (unsigned int iz = 0; iz < _n_cells + 1; ++iz)
+      {
+        const auto * pin_node = _subchannel_mesh.getPinNode(i_pin, iz);
+        _Tpin_soln->set(pin_node, _temp_pin[i_pin](iz, _nrpin));
```
kyriv1980, line 1018: nice

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1091 @@
...
+    }
+  }
+
+  // Transient PinTempSolver
+  if (isTransient() > 0 && pintemp_ss == 1)
+  {
+    _console << "running PinTempSolverDriver for transient" << std::endl;
+    PinTempSolverDriver(_dt, pintemp_ss);
+  }
+
```
kyriv-lab, line 1014: It seems that you do the pintemp solution after the SCM solution... that's why i was thinking an aux kernel would work. But we can restructure later.

modules/subchannel/test/tests/problems/SFR/pin_temp_solverSS/pin_tempss.i
```diff
@@ -0,0 +1,319 @@
...
+  []
+[]
+
+[Postprocessors]
+  [TTC-27]
+    type = SubChannelPointValue
+    variable = T
+    index = 91
+    execute_on = 'TIMESTEP_END'
+    height = 0.322
```
kyriv-lab, line 232: I don't see how this test, test the Pin temp solver. How does the user get access to the pin temp solution? We want to see _temp_pin[i_pin](iz, i);

kyriv-lab:
@aydinka I can't activate the CIVET tests to run until the precheck test passes. From what i saw it should be an easy fix.

modules/subchannel/include/mesh/TriSubChannelMesh.h
```diff
@@ -153,6 +153,15 @@ class TriSubChannelMesh : public SubChannelMesh
...
 
+  // Channel centroid (x,y) as a Point; z=0 for convenience
+  Point getChannelCenter(unsigned int i_chan) const
+  {
+    return Point(_subchannel_position[i_chan][0], _subchannel_position[i_chan][1], 0.0);
+  }
+
+  // Pin center
+  Point getPinCenter(unsigned int i_pin) const { return _pin_position[i_pin]; }
+
```
kyriv-lab, line 164: I don't think we need those new getters.

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1095 @@
...
+    {
+
+      for (unsigned int i_rod = 0; i_rod < _n_pins; ++i_rod)
+      {
+
+        auto * node_in = _subchannel_mesh.getChannelNode(i_rod, 0);
+        auto T_in = (*_T_soln)(node_in);
+
+        _hcool_pin_ave(iz, i_rod) = 1.0e-15;
+        _tcool_pin_ave(iz, i_rod) = T_in;
```
kyriv-lab, line 257: This is not correct. Enthalpy mixing will change temperature even if there is no heating.

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1095 @@
...
+      }
+    }
+    else
+    {
+      auto * node_in = _subchannel_mesh.getChannelNode(i_ch, iz - 1);
+      auto * node_out = _subchannel_mesh.getChannelNode(i_ch, iz);
+      return ((*_q_prime_soln)(node_out) + (*_q_prime_soln)(node_in)) * dz / 2.0;
+    }
+  }
+  else
```
kyriv-lab, line 861: This means that you start your PinTemp solution assuming it has reached steady state and all  pin power goes into the coolant. Correct?

modules/subchannel/src/problems/PinTempSolver.C
kyriv-lab: Best to use T_soln for temperatures.

modules/subchannel/src/problems/PinTempSolver.C
```diff
@@ -0,0 +1,1095 @@
...
+          if (_ijet[i_rod] == 1)
+          {
+            _qbarconv_channel(iz, i_ch) += frac * libMesh::pi * dpin * _h_jet[i_rod] *
+                                           (_temp_pin[i_rod](iz, _nrfuel + 2) - _temp_jet[i_rod]);
+          }
+          else
+          {
+            _qbarconv_channel(iz, i_ch) +=
+                frac * libMesh::pi * dpin * _hcool_pin_ave(iz, i_rod) *
+                (_temp_pin[i_rod](iz, _nrfuel + 2) - _tcool_pin_ave(iz, i_rod));
```
kyriv-lab, line 298: So _qbarconv_channel(iz, i_ch) is generally equal to q_prime(iz, i_ch) unless in a case of power transient. Correct?

aydinka:
I need it for CDAP, going to upload today.

modules/subchannel/src/problems/PinTempSolver.C
kyriv-lab: One more reason to make this into an auxkernel.

modules/subchannel/src/problems/PinTempSolver.C
kyriv-lab: In case of a transient the  pintemp solution (PinTempSolverDriver)  must happen before the SCM solution?

aydinka:
Correct. This part I mean coupling part is critical. Please review well.

modules/subchannel/src/problems/PinTempSolver.C
kyriv-lab: At t=0. AddedheatPin uses the q_prime value so all heat goes into the coolant and we assume that we start from steady state. Then the PinTempSolver calculates  hw Tpin surface temperature and temperature distribution in the pin.

If we have a power transient. At the next time step power changes. But AddedheatPin uses still _qbarconv_channel(iz, i_ch) which is the same as q_prime at time zero. So the SCM solution will not change.  Then the PinTempSolver calculates  hw Tpin surface temperature and temperature distribution in the pin with the same boundary condition (The SCM solution hasn't changed yet) but a with a new power. That's when AddedheatPin gets updated due to the new temperature distribution. To be used in the next time step.

Correct?

modules/subchannel/src/problems/PinTempSolver.C
kyriv-lab: Note to self: Can create an auxkernel that executes at timestep_begin and time_step end. (solves for pin temperature distribution, updates q_prime, hw, addedheatpin etc...)

aydinka:
It is good to check this. My implementation is use q_prime_soln for steady
state and use _qbarconv for transients. If you find power is not properly
updated, we should discuss.

kyriv-lab:
@aydinka Please fix the precheck issue so that i can activate the testing. PS. please add your responses on my initial comments so that the discussion remains coherent. (reply to comment in the reply space: )
Screenshot 2025-09-23 at 10 40 07 AM

modules/subchannel/include/problems/PinTempSolver.h
kyriv-lab: actually i don't think you need this at all since in the base class we have :
  _dt = (isTransient() ? dt() : _one);
  _TR = isTransient();

PR #31501, opened 2025-09-11
Create ArrayNodalKernel

### LINKED ISSUE #31498: Add support for Nodal Array Kernels
### Motivation

MOOSE currently supports array variables, along with ArrayKernels, which can compute values that are assembled to their residuals. We have a need for a capability to compute contributions to ArrayVariables from nodes, similar to the current NodalKernel.

This request is driven by a need to develop more efficient cluster-dynamics models. These models can have potentially very large (up to tens of thousands or more) sets of variables describing densities of clusters of varying sizes. We have been doing this with large sets of variables, with each one tracking the density of a specific cluster size. We have a kernel associated with each variable, which leads to a large number of kernels. Based on discussions among developers, it sounds like this could be made significantly more efficient by using a single kernel that operates on an array variable. These computations are done at nodes, or even at a single point in the domain, so a nodal kernel to do this is needed.

### Design

This capability should enable us to compute the contribution to the residual, along with the associated Jacobian, for array variables at nodes. We would be interested in both hand-coded and AD-computed Jacobians, although the first priority should be supporting hand-coded Jacobians. 

### Impact

This should only add new capability.

lindsayad:
Closes #31498


lindsayad:
@bwspenc this is ready for review when you have time. This is just the hand-coded objects. The AD objects are in https://github.com/idaholab/moose/pull/31548

framework/doc/content/source/nodalkernels/ArrayReactionNodalKernel.md
```diff
@@ -0,0 +1,15 @@
+# ArrayReactionNodalKernel
+
+## Description
+
+This class is the same as [ReactionNodalKernel.md] except instead of being applied to a classic MOOSE variable, it's applied to an array variable.
```
bwspenc: 'classic MOOSE variable' might be a little confusing to a newcomer, but I'm not quite sure what the correct term is: 'standard field variable'?

framework/doc/content/source/nodalkernels/ArrayTimeDerivativeNodalKernel.md
```diff
@@ -0,0 +1,15 @@
+# ArrayTimeDerivativeNodalKernel
+
+## Description
+
+This class is the same as [TimeDerivativeNodalKernel.md] except instead of being applied to a classic MOOSE variable, it's applied to an array variable.
```
bwspenc: Same comment as before on 'classic MOOSE variable'

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -0,0 +1,98 @@
...
+   * Note: This is NOT what a user would normally want to override.
+   * Usually a user would override computeQpJacobian()
+   */
+  virtual void computeJacobian() override;
+
+  /**
+   * Compute the off-diagonal Jacobian at one node.
+   *
+   * Note: This is NOT what a user would normally want to override.
+   * Usually a user would override computeQpOffDiagJacobian()
```
bwspenc: Add `@param jvar ...`

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Gets the variable this is active on
+   * @return the variable
+   */
+  const MooseVariableFE<RealEigenVector> & variable() const override { return _var; }
+
+protected:
+  /**
+   * The user can override this function to compute the residual at a node.
```
bwspenc:
```suggestion
   * The user must override this function to compute the residual at a node.
   * @param residual Reference to the residual, computed by this method 
```

framework/include/nodalkernels/ArrayNodalKernel.h
bwspenc, line 49:
```suggestion
   * Usually a user would override computeQpOffDiagJacobian()
   * @param jvar Coupled variable for the off-diagonal Jacobian contribution
```

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -0,0 +1,98 @@
...
+   */
+  virtual void computeQpResidual(RealEigenVector & residual) = 0;
+
+  /**
+   * The user can override this function to compute the "on-diagonal"
+   * Jacobian contribution.  If not overriden,
+   * returns an array of 1s.
```
bwspenc, line 70:
```suggestion
   * returns an array of 1s.
   * @return On-diagonal Jacobian contribution
```

bwspenc: This seems a little odd that it's a RealEigenVector rather than a RealEigenMatrix. Is it just the coupling terms between each entry in the array with itself, and not the coupling between them? 

Looking at your example, it seems like the entries are decoupled from each other. For our use case, they will be coupled to each other. Maybe we need this to return a RealEigenMatrix to handle that?

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -0,0 +1,98 @@
...
+   */
+  virtual RealEigenVector computeQpJacobian();
+
+  /**
+   * This is the virtual that derived classes should override for
+   * computing an off-diagonal jacobian component.
```
bwspenc, line 77:
```suggestion
   * computing an off-diagonal jacobian component.
   * @param Coupled variable for the off-diagonal Jacobian contribution
   * @return Off-diagonal Jacobian contribution
```

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -0,0 +1,98 @@
...
+
+  /// Value of the unknown variable this is acting on
+  const ArrayVariableValue & _u;
+
+  /// Number of components of the array variable
+  const unsigned int _count;
+
+private:
+  /**
+   * Prepare our array ivar degrees of freedom
```
bwspenc:
```suggestion
   * Prepare our array variable degrees of freedom
```

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Class constructor.
+   * @param parameters The InputParameters for the object
+   */
+  static InputParameters validParams();
+
+  ArrayNodalKernel(const InputParameters & parameters);
+
+  /**
+   * Compute the residual at the current node.
```
bwspenc:
```suggestion
   * Compute and assemble the residual at the current node
```

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -0,0 +1,98 @@
...
+   *
+   * Note: This is NOT what a user would normally want to override.
+   * Usually a user would override computeQpResidual()
+   */
+  virtual void computeResidual() override;
+
+  /**
+   * Compute the Jacobian at one node.
```
bwspenc:
```suggestion
   * Compute and assemble the Jacobian at one node.
```

framework/src/nodalkernels/ArrayTimeDerivativeNodalKernel.C
```diff
@@ -0,0 +1,41 @@
...
+
+InputParameters
+ArrayTimeDerivativeNodalKernel::validParams()
+{
+  InputParameters params = ArrayNodalKernel::validParams();
+  params.set<MultiMooseEnum>("vector_tags") = "time";
+  params.set<MultiMooseEnum>("matrix_tags") = "system time";
+  params.addClassDescription(
+      "Forms the contribution to the residual and jacobian of the time derivative term from an ODE "
+      "being solved at all nodes.");
```
bwspenc: Maybe say something about the fact that this is for an array variable?

framework/src/nodalkernels/ArrayReactionNodalKernel.C
```diff
@@ -0,0 +1,44 @@
...
+
+#include "Function.h"
+
+registerMooseObject("MooseApp", ArrayReactionNodalKernel);
+
+InputParameters
+ArrayReactionNodalKernel::validParams()
+{
+  InputParameters params = ArrayNodalKernel::validParams();
+  params.addClassDescription("Implements a simple consuming reaction term at nodes");
```
bwspenc: Maybe say something about the fact that this is for an array variable?

bwspenc (changes requested):

framework/include/nodalkernels/ArrayNodalKernel.h
lindsayad: That would be a different concept than what we've adopted for neutronics, and I think I agree more with how it's implemented currently although I don't think what you're suggesting is crazy either.

In general I would think of array variables as a convenience and behind-the-scenes optimization. If you weren't treating your reacting species as an array variable then you would certainly be coding their coupling in off-diagonal Jacobian methods. So I think having that coupling treated in an off-diagonal Jacobian method for an array variable makes sense to me

bwspenc (approved):

PR #31505, opened 2025-09-12
Remove CRTP in Kokkos base classes through dispatcher registry

NamjaeChoi:
Refs #30655.


NamjaeChoi:
@hugary1995 This PR is based on your idea. You may be interested in reviewing it.

hugary1995:
Wow, this is impressive. I'm on vacation for the next 10 days or so, but I'll see if I can find some time to look at this. I glanced through the last commit, and it looks pretty good. My only suggestion is to move the non-device code into `.C` files since they are no longer templated. The header files should only contain signatures and the device code if possible.

In theory, it's also possible to move the device code implementation into the source file. But it's fairly involving, and I tried that once but failed.

NamjaeChoi:
> Wow, this is impressive. I'm on vacation for the next 10 days or so, but I'll see if I can find some time to look at this. I glanced through the last commit, and it looks pretty good. My only suggestion is to move the non-device code into `.C` files since they are no longer templated. The header files should only contain signatures and the device code if possible.
> 
> In theory, it's also possible to move the device code implementation into the source file. But it's fairly involving, and I tried that once but failed.

Which part are you referring to? Note that `*.K` is the source file for Kokkos stuff. It is to separate compile scope.

hugary1995:
I probably glanced too quickly. I checked again just now and I think you are doing things correctly.

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,185 @@
...
+  Dispatcher(const void * object)
+    : _functor_host(*static_cast<const Object *>(object)), _functor_device(_functor_host)
+  {
+  }
+  Dispatcher(const Dispatcher & functor)
+    : _functor_host(functor._functor_host), _functor_device(functor._functor_host)
+  {
+  }
+
+  virtual void parallelFor(const Policy & policy)
```
loganharbour:
```suggestion
  void parallelFor(const Policy & policy) override final
```

Right?

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,185 @@
...
+  template <typename Operation>
+  static bool hasUserMethod(const std::string & name)
+  {
+    auto it = getRegistry()._user_override.find(std::make_pair(typeid(Operation).name(), name));
+
+    return it == getRegistry()._user_override.end() ? false : it->second;
+  }
+
+  template <typename Operation>
+  static std::shared_ptr<DispatcherBase> build(const void * object, const std::string & name)
```
loganharbour:
```suggestion
  static std::unique_ptr<DispatcherBase> build(const void * object, const std::string & name)
```

I've worked towards make this the case in moose where possible - these should return unique_ptr, which can convert to shared_ptr. but the other way around doesn't work. thus, the base method should be unique_ptr

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,185 @@
...
+    mooseAssert(it != getRegistry()._dispatchers.end(),
+                "Kokkos functor dispatcher not registered.");
+
+    return it->second->build(object);
+  }
+
+private:
+  std::map<std::pair<std::string, std::string>, std::unique_ptr<DispatcherRegistryEntryBase>>
+      _dispatchers;
+  std::map<std::pair<std::string, std::string>, bool> _user_override;
```
loganharbour: Key should be able to be `std::pair<std::type_info, std::type_info>`

Can the user override flag not just be stored in the registry entry?

framework/include/kokkos/base/KokkosResidualObject.h
```diff
@@ -84,6 +85,14 @@ class ResidualObject : public ::ResidualObject,
    * Kokkos thread object
    */
   Thread _thread;
+  /**
+   * Kokkos functor dispatchers
+   */
+  ///@{
+  std::shared_ptr<DispatcherBase> _residual_dispatcher;
+  std::shared_ptr<DispatcherBase> _jacobian_dispatcher;
+  std::shared_ptr<DispatcherBase> _offdiag_jacobian_dispatcher;
```
loganharbour: Do these need to be shared?

NamjaeChoi: Yes, because `std::unique_ptr` is not copy constructible.

framework/include/kokkos/base/KokkosDispatcher.h
NamjaeChoi: Hmm key can be `std::pair<std::type_info, std::string>` but not both `std::type_info`, because I need to look up for the dispatcher using `MooseObject::type()`. See `KokkosKernel.K`

NamjaeChoi: User override can be stored in the registry entry and is actually better

framework/include/kokkos/base/KokkosResidualObject.h
NamjaeChoi: But I think I can still make them unique by explicitly not copying them in the copy constructor.

loganharbour:
Reminder that it's preferred to not squash changes from a review; it's not immediately clear what has changed

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,200 @@
...
+  Dispatcher(const void * object)
+    : _functor_host(*static_cast<const Object *>(object)), _functor_device(_functor_host)
+  {
+  }
+  Dispatcher(const Dispatcher & functor)
+    : _functor_host(functor._functor_host), _functor_device(functor._functor_host)
+  {
+  }
+
+  virtual void parallelFor(const Policy & policy) override final
```
loganharbour: not a required change but the request wasn't satisfied - `virtual` is redundant. it's common to include it when only `override` exists because a child can override it, while it's common to not include it when `override final` exists because children cannot override it

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,200 @@
...
+};
+
+class DispatcherRegistryEntryBase
+{
+public:
+  virtual ~DispatcherRegistryEntryBase() {}
+  virtual std::unique_ptr<DispatcherBase> build(const void * object) = 0;
+
+  void hasUserMethod(bool flag) { _has_user_method = flag; }
+  bool hasUserMethod() { return _has_user_method; }
```
loganharbour:
```suggestion
  virtual std::unique_ptr<DispatcherBase> build(const void * object) const = 0;

  void hasUserMethod(bool flag) { _has_user_method = flag; }
  bool hasUserMethod() const { return _has_user_method; }
```

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,200 @@
...
+class DispatcherRegistryEntry : public DispatcherRegistryEntryBase
+{
+public:
+  virtual std::unique_ptr<DispatcherBase> build(const void * object) override final
+  {
+    return std::make_unique<Dispatcher<Operation, Object>>(object);
+  }
+};
+
+class DispatcherRegistry
```
loganharbour, line 148: Is this unit testable at all

NamjaeChoi:
```
[----------] 1 test from KokkosDispatcherTest
[ RUN      ] KokkosDispatcherTest.dispatchKokkosObject
[       OK ] KokkosDispatcherTest.dispatchKokkosObject (0 ms)
[----------] 1 test from KokkosDispatcherTest (0 ms total)
```

loganharbour: Thank you!

loganharbour (approved):
@hugary1995 any concerns?

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,202 @@
...
+{
+
+using Policy = ::Kokkos::RangePolicy<ExecSpace, ::Kokkos::IndexType<ThreadID>>;
+
+class DispatcherBase
+{
+public:
+  virtual ~DispatcherBase() {}
+  virtual void parallelFor(const Policy & policy) = 0;
+};
```
hugary1995, line 38: Would be nice to add some docstrings.

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,202 @@
...
+
+template <typename Operation, typename Object>
```
hugary1995, line 46: Similarly, would be nice to explain the template parameters.

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,202 @@
...
+class Dispatcher : public DispatcherBase
+{
+public:
+  Dispatcher(const void * object)
+    : _functor_host(*static_cast<const Object *>(object)), _functor_device(_functor_host)
```
hugary1995, line 57: For my own reference, I wonder why we need to do a static cast here instead of just using `const Object * object` as the argument. I may find out why if I keep reading.

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  KOKKOS_FUNCTION void operator()(const ThreadID tid) const
+  {
+    _functor_device(Operation{}, tid, _functor_device);
+  }
+
+private:
+  const Object & _functor_host;
+  const Object _functor_device;
```
hugary1995, line 90: It took me a while to understand what these members do. I wonder if naming them `_host_fuctor` and `_device_functor` might help. I suppose adding some docstrings would also help.

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,202 @@
...
+};
+
+class DispatcherRegistryEntryBase
+{
+public:
+  virtual ~DispatcherRegistryEntryBase() {}
+  virtual std::unique_ptr<DispatcherBase> build(const void * object) const = 0;
+
+  void hasUserMethod(bool flag) { _has_user_method = flag; }
+  bool hasUserMethod() const { return _has_user_method; }
```
hugary1995, line 117: I think it would also be helpful to explain what user method means and why it's important to maintain this flag.

framework/include/kokkos/base/KokkosDispatcher.h
hugary1995: Okay, I see. You want to have a common build method in DispatcherRegistryEntryBase. This is fine. I think we should put this line of thinking in the docstring.

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,202 @@
...
+  DispatcherRegistry & operator=(DispatcherRegistry &&) = delete;
+
+  template <typename Operation, typename Object>
+  static void add(const std::string & name)
+  {
+    auto operation = std::type_index(typeid(Operation));
+
+    getRegistry()._dispatchers[std::make_pair(operation, name)] =
+        std::make_unique<DispatcherRegistryEntry<Operation, Object>>();
+  }
```
hugary1995, line 172: IIRC, the original MOOSE registry returns a dummy when adding to the registry so that we can declare a static variable to detect duplicate registration. Can the same be done here?

framework/include/kokkos/base/KokkosDispatcher.h
hugary1995: I think I understand what the flag does now. You want to use this to track if derived class has reimplemented the operator.

I think we should be able to check that statically with SFINAE. I made a simple example here https://godbolt.org/z/oYcYPYnMG Can you take a look and see if you can adapt it for the dispatcher?

hugary1995 (changes requested):
I took a closer look at the dispatcher code and had a few minor comments. Overall this looks great!

I also took this opportunity to try to understand more about the kokkos-moose implementation. I think you've done a great job in keeping the new syntax similar to the original. I really like it!

framework/include/kokkos/base/KokkosDispatcher.h
loganharbour: This is done in the macros; a static char with the classname in the variable name

hugary1995: Ahh, I see.

loganharbour:
I agree with the comment about docstrings. I actually had this pending in a review and looks like I forgot it...

framework/include/kokkos/base/KokkosDispatcher.h
NamjaeChoi: Another reason is that this is called in the base class with the base `this` pointer. Base class doesn't know its derived type.

framework/include/kokkos/base/KokkosDispatcher.h
NamjaeChoi: Thank you for the example and I also attempted to approach with SFINAE initially, but decided to take the current approach because the signatures can change by intermediate base class. For example, `KernelGrad` and `KernelValue` derive from `Kernel` but they use different signatures for the user methods. The check should be against those intermediate classes not the base class in this case. It will make the registration macro too fragmented.

hugary1995: Hmm, I see. That's a bummer. Can we perhaps check for all possible signatures, either in the type trait itself or in the macro?

NamjaeChoi: Possible but unsure if it's simpler than the current code.

NamjaeChoi: I think the current approach is also somewhat static. Templates and macros are pain, and their combination is even more painful.

hugary1995: I understand your concern. While I still have reservations, I wouldn't let this be the show stopper. I'll approve this PR if you add the missing docstrings.

NamjaeChoi: Docstrings were pushed. Hope those improve clarity

framework/doc/content/syntax/Kokkos/index.md
lindsayad, line 1: I am guilty of this myself too sometimes, but it would be good to split very long lines of text so that it's easier to see what diffs are when something changes

NamjaeChoi: How about I just break every sentence

NamjaeChoi: I think Kokkos documentation can be changed frequently for a while. I don't mind doing it here.

NamjaeChoi: Done

hugary1995 (approved):

framework/include/kokkos/base/KokkosDispatcher.h
```diff
@@ -0,0 +1,306 @@
...
+ */
+class DispatcherRegistryEntryBase
+{
+public:
+  virtual ~DispatcherRegistryEntryBase() {}
+  /**
+   * Build a dispatcher for this operation and functor
+   * @param object The pointer to the functor
+   */
+  virtual std::unique_ptr<DispatcherBase> build(const void * object) const = 0;
```
loganharbour, line 106: This is an aside and doesn't need to be done here (will merge anyway), but is it possible to use `std::any` here for the type erasure? So that we can get some runtime checking (at least in devel/dbg)?

We use this strategy of passing around a void pointer in DataIO and it's a bit atrocious

loganharbour (approved):
Beautiful docstrings.

framework/include/kokkos/base/KokkosDispatcher.h
NamjaeChoi: Attempted `std::any` but it doesn't work because `std::any` should exactly match the type. In this case, `this` pointer being passed is of the base class type, so the `std::any` container containing `this` can only be cast to the base class type.

PR #31521, opened 2025-09-15
Automatic weak form

dschwen:
## Reason
Prototyping models rapidly in MOOSE could be made easier. The step from a paper to a working simulation still requires substantial effort in building weak form expression or computing variational derivatives from energy expressions. In models wit vectorial contributions, such as anisotropic interfacial energies, these derivation can get tedious, and can require split formulations to accommodate finite element order.

## Design
An action system that parses and automatically performs variational derivatives and weak form derivations would be helpful. This is not a total pipe dream, as demonstarted by packages like GetFEM. Let's see how far along this road we can get with the help of an up-to-date code agent.

## Impact
Added capability and drastically simplified user experience.


mangerij:
this is very interesting! excited to see how this works.

PR #31479, opened 2025-09-10
Update conda packages

maythuwin1313:
## Reason
Conda packages need to be updated so that they are not behind.

## Design
Update pins in conda packages and add fixes as needed.

## Impact
Significant; package updates may cause things to break.

## Update summary

- Updated mpi and compilers in conda moose-mpi
- Updated VTK to 9.5.1 in conda moose-libmesh-vtk
- Updated libmesh submodule

PR #31522, opened 2025-09-15
Add workspaces to MooseServer by registering document resources

brandonlangley:
refs #22766

https://github.com/user-attachments/assets/831ace5c-91a8-4bbc-b4b4-025826403639

### DIFF
--- a/framework/include/base/MooseServer.h
+++ b/framework/include/base/MooseServer.h
@@ -55,21 +55,16 @@ private:
   bool parseDocumentForDiagnostics(wasp::DataArray & diagnosticsList);
 
   /**
-   * Update document text changes - specific to this server implemention.
-   * @param replacement_text - text to be replaced over the provided range
-   * @param start_line - starting replace line number ( zero-based )
-   * @param start_character - starting replace column number ( zero-based )
-   * @param end_line - ending replace line number ( zero-based )
-   * @param end_character - ending replace column number ( zero-based )
-   * @param range_length - length of replace range - server specific
-   * @return - true if the document text was updated successfully
+   * Add paths from includes and FileName parameters for client to watch.
    */
-  bool updateDocumentTextChanges(const std::string & replacement_text,
-                                 int start_line,
-                                 int start_character,
-                                 int end_line,
-                                 int end_character,
-                                 int range_length);
+  void addResourcesForDocument();
+
+  /**
+   * Recursively walk input to gather all FileName type parameter values.
+   * @param filename_vals - set to fill up with FileName parameter values
+   * @param parent - nodeview for recursive tree traversal starting point
+   */
+  void getFileNameTypeValues(std::set<std::string> & filename_vals, wasp::HITNodeView parent);
 
   /**
    * Gather document completion items - specific to this server implemention.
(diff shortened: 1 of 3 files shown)


brandonlangley:
@loganharbour -

Workspace updates (i.e., resource files of a document) is an [LSP feature](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles) the MooseServer previously did not support.

I recently updated the base ServerImpl class in WASP to handle this and register resource files for the client to watch.

This PR updates the MooseServer to add resource files that will get registered for the client to watch for any changes.

These are `!include` files and `FileName`, `FileNameNoExtension`, `MeshFileName`, `MatrixFileName` parameter values.

The client notifies the server when any watched file is modified, and the server then reprocesses the base document.

I added a test of a scenario modifying `!include` and `FileName` external resource files to verify diagnostics changes.

This also [fixes](https://github.com/idaholab/moose/pull/31522/commits/aafac09ed20101d77fa1e169d946eb67e0d290db) the MooseServer empty document crash we talked about by returning early if no parse tree root exists.

Could you review this update and merge it if you approve? FYI @dschwen this also improves the VS Code integration.

Some examples of updated resource files causing the base document to be reprocessed is shown in the video below:

https://github.com/user-attachments/assets/1d856aa1-292b-4e48-b7f6-edfddaf2ea46

loganharbour:
For the sake of all of the updates, we can incorporate the wasp bump in https://github.com/idaholab/moose/pull/31479.

brandonlangley:
@loganharbour -

> For the sake of all of the updates, we can incorporate the wasp bump in #31479.

So #31479 is also going to include this [wasp submodule update](https://github.com/idaholab/moose/pull/31522/commits/21e8e78bb66da383735f2c382eb13b52ecc18926) needed here?

```diff
--- a/framework/contrib/wasp
+++ b/framework/contrib/wasp
@@ -1 +1 @@
-Subproject commit 78e8705d3743cee12cb0b024a51a98db2c3713bc
+Subproject commit 02960f1b223fd8a481f07aa9684856eb42c640ac
```

brandonlangley:
@loganharbour -

I also just added a [commit](https://github.com/idaholab/moose/pull/31522/commits/1a9cadb631be5d6676b8e5056834f26b17408cbc) here fixing a MooseServer input formatter issue noted by @lindsayad on Slack last month.

The MooseServer document formatter logic did not handle `[Block/SubBlock]` shorthand path declarations correctly.

It would previously take any input components that were declared using shorthand path notation syntax, such as this:

```
[Physics/SolidMechanics/QuasiStatic]
  param = value
[]
```

and explicitly expand all hierarchy levels of those components as well as add unnecessary blank lines to look like this:

```
[Physics]

  [SolidMechanics]

    [QuasiStatic]
      param = value
    []
  []
[]
```

I fixed this issue and, since it was a small update, included the fix here along with unit testing added for that scenario.

loganharbour:
@brandonlangley drop your wasp updates in this PR. The submodule update is in master. Should be good to go after that change.

brandonlangley:
@loganharbour -

> drop your wasp updates in this PR. The submodule update is in master. Should be good to go after that change.

I just rebased on devel and dropped my WASP, conda, and versioner bumps. So this should be good to merge now.

loganharbour (approved):

PR #31527, opened 2025-09-15
Select unicode support and pythonized precheck

dschwen:
## Reason
Allowing select unicode characters required for plain text math equations will allow for richer documentation strings and code/input comments. Also moving the precheck script into the moose repository allows developers perform prechecks before making a PR and rewriting the checks in python improves maintainability (homogenizes repo languages by reducing shell script + perl code mix).

## Design
Specify select code characters and unicode groups. I'd like to also add a unicode normalization script to avoid having to support many similar looking characters (such as Middle Dot (·), Bullet Operator (∙), and Dot Operator (⋅), or U+00B5 vs U+03BC (micro vs mu)).

## Impact
Added capability, richer comments.


dschwen:
This obviously fails the precheck for two reasons.

1. classified/proprietary match
2. unicode symbols in the comments for the unicode check

GiudGiud (commented):
I think we should unit test the new precheck

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+import os
+import re
+import sys
+import stat
+import subprocess
+import textwrap
+from pathlib import Path
+from typing import Iterable, List, Tuple, Optional
+
+# --------------------------- Helpers ---------------------------
```
GiudGiud, line 31: so should we define new helpers for each file that uses them?
these seem more general than just pre_check

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+        CompletedProcess object with command result
+    """
+    return subprocess.run(cmd, capture_output=capture, text=text, check=check)
+
+def git_files(*patterns: str) -> List[str]:
+    """
+    Return repo-tracked files matching glob patterns.
+
+    Args:
+        *patterns: Glob patterns to match (e.g., "*.py", "*.[Ch]")
```
GiudGiud: are the patterns inclusive (*py *ch gives all for both) or exclusive (*py *ch gives nothing)
let s mention it (if only just for me)

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+
+    Returns:
+        List of file paths, excluding contrib/ directory
+    """
+    if not patterns:
+        patterns = ('',)
+    cmd = ["git", "ls-files", "-z", *patterns]
+    cp = run(cmd)
+    if cp.returncode != 0:
+        return []
```
GiudGiud: is there a return code for no match?
If ls-files fails for another reason we would want to error

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+        log_to: Ending commit reference
+
+    Returns:
+        Newline-separated list of matching commit messages
+    """
+    try:
+        cp = run(["git", "log", f"{log_from}..{log_to}"], capture=True, text=True)
+        text = cp.stdout or ""
+    except Exception:
+        text = ""
```
GiudGiud: probably want to error here too

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+    pat = re.compile(r"(?:moose/(?:issues|pull)/)|#\d{1,}")
```
GiudGiud: this is for repeated use but we only run this once on PRs?

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+
+    Returns:
+        List of files containing tabs, prefixed with tabs
+    """
+    bad = []
+    seen = set()
+    for f in files:
+        if f in seen:
+            continue
+        seen.add(f)
```
GiudGiud: why not just sort and unique the input list?
I guess what matters is, how long does the precheck take with this script rather than the old one?

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+    - printf() calls
+    - sleep() calls (except in SlowProblem.C)
+    - print_trace (except in MooseError files)
+
+    Returns:
+        List of files containing banned keywords, prefixed with tabs
+    """
+    bad = []
+    seen = set()
+    files = git_files("*.[Chi]")
```
GiudGiud:
```suggestion
    files = git_files("*.[Ch]")
```

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+    for f in files:
+        if f in seen:
+            continue
+        seen.add(f)
```
GiudGiud: does git_files return duplicates?

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+
+    Returns:
+        List of files containing classified keywords, prefixed with tabs
+    """
+    bad = []
+    seen = set()
+    files = git_files("*.[Chi]", "*.py")
+    pat_prop = re.compile(r"p\s*r\s*o\s*p\s*r\s*i\s*e\s*t\s*a\s*r\s*y", re.IGNORECASE)
+    pat_class = re.compile(r"c\s*l\s*a\s*s\s*s\s*i\s*f\s*i\s*e\s*d", re.IGNORECASE)
+    for f in files:
```
GiudGiud: we should add "OUO" to that list as well as CUI  and CNUI

GiudGiud: can be another PR

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+    return bad
+
+def windows_line_endings() -> List[str]:
+    """Check for Windows-style line endings (CRLF).
+
+    Returns:
+        List of files with Windows line endings, prefixed with tabs
+    """
+    bad = []
+    files = git_files("*.[Chi]", "*.py")
```
GiudGiud: should we include md files there too?

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+
+    Searches for variations of 'proprietary' and 'classified' that might
+    indicate restricted content. Ignores pre_check.py itself.
+
+    Returns:
+        List of files containing classified keywords, prefixed with tabs
+    """
+    bad = []
+    seen = set()
+    files = git_files("*.[Chi]", "*.py")
```
GiudGiud: should we include md files there too?

scripts/pre_check.py
```diff
@@ -0,0 +1,868 @@
...
+    "EQUALS SIGN",                   # =
+    "NOT EQUAL TO",                  # ≠
+    "ALMOST EQUAL TO",               # ≈
+    "IDENTICAL TO",                  # ≡
+    "PROPORTIONAL TO",               # ∝
+    "INFINITY",                      # ∞
+    "SQUARE ROOT",                   # √
+    "CIRCLED DOT OPERATOR",          # ⊙
+    "CIRCLED PLUS",                  # ⊕
+    "ELEMENT OF",                    # ∈
```
GiudGiud: move below?

scripts/pre_check.py
GiudGiud: what's been our design so far for the python side of moose wrt re-using code

scripts/pre_check.py
dschwen: What do you mean by "repeated use"?

scripts/pre_check.py
dschwen: Files is an iterable. Sure, I could make a list, sort it and uniq it. Would that be faster than using the seen set? That's not obvious to me. Either way this looks like a micro-optimization at best.

scripts/pre_check.py
dschwen: yes

scripts/pre_check.py
dschwen: no. That would mean we could not use the word proprietary or classified in a documentation.
https://github.com/idaholab/moose/blob/cffa4836ce665633ae9c563c12ad98baaed583c1/modules/peridynamics/doc/content/modules/peridynamics/PeridynamicModels.md?plain=1#L8

scripts/pre_check.py
dschwen: no

scripts/pre_check.py
GiudGiud: yes hence the second part of that comment. Any concern with performance with this pre-check, did you check?

dschwen: New pre-check: 18s
Old pre-check: 1:56s

scripts/pre_check.py
GiudGiud: not sure anymore

GiudGiud (approved):

loganharbour (commented):
This won't work with the apps. Everything is hard coded for moose.

- Should use argparse for parsing arguments
- Shouldn't have "idaholab/moose" in it anywhere
- Should optionally take a commit as the base to compare against, which is what civet will use. the merge base logic is unfortunately more complex due to how to we do things with moose merges
- The default base should be devel, so that when you don't provide a base it'll still work with the apps

scripts/pre_check.py
```diff
@@ -187,6 +188,16 @@ def find_bad_executables() -> list[str]:
     return bad
 
 
+def style_files() -> list[str]:
```
GiudGiud: clang format does not catch this?

dschwen: black?

loganharbour: Bad Daniel.

This check checks C++ files. It checks for something that clang format already fixes.

dschwen: Yeah, that's just straight converted from the old pre check and it's disabled in civet. We could remove it.

scripts/pre_check.py
```diff
@@ -0,0 +1,726 @@
...
+def _fmt_files(files: list[str]) -> str:
+    return "\n".join(f"\t{f}" for f in files)
+
+
+@dataclass
+class CheckResult:
+    """Validation result and output strings for one precheck."""
+
+    enabled: bool
+    files: list[str]
```
GiudGiud: is this the list of checked files or the list of files left to check or the list of files that failed the check?

I think comments would make it clear what the attributes mean

scripts/pre_check.py
```diff
@@ -0,0 +1,726 @@
...
+
+    def files(self, *patterns: str) -> list[str]:
+        """Return cached touched file paths matching glob patterns."""
+        return [path for _, path in self.entries(*patterns)]
+
+    def files_for_tabs_except_input(self) -> list[str]:
+        return self.files("*.[Ch]", "*.py")
+
+    def files_for_tabs(self) -> list[str]:
+        return self.files("*.[Chi]", "*.py")
```
GiudGiud, line 353: I would be fine with .md and tests for most of these too. 
not for this PR

scripts/pre_check.py
```diff
@@ -0,0 +1,726 @@
...
+        self.check_tabs = os.environ.get("CHECK_TABS", "1") == "1"
+        self.check_tabs_except_input = (
+            os.environ.get("CHECK_TABS_EXCEPT_INPUT_FILES", "0") == "1"
+        )
+        self.check_classified = os.environ.get("CHECK_CLASSIFIED", "1") == "1"
+        self.check_unicode = os.environ.get("CHECK_UNICODE", "0") == "1"
+        self.check_include_guards = os.environ.get("CHECK_INCLUDE_GUARDS", "0") == "1"
+        self.check_windows = os.environ.get("CHECK_WINDOWS_FILES", "1") == "1"
+        self.check_cpp_whitespace = os.environ.get("CHECK_CPP_WHITESPACE", "0") == "1"
+        self.check_f_whitespace = os.environ.get("CHECK_F_WHITESPACE", "0") == "1"
```
GiudGiud, line 296: let's move those two down they don't trigger a test they are options

also we should assert we dont have check_whitespace = false with check_f_whitespace = true

GiudGiud (approved):

dschwen:
Nope x2

black --config pyproject.toml --workers 1 .

formats over half a dozen file, but none that are changed in this PR.

scripts/pre_check.py
dschwen: removed

scripts/pre_check.py
dschwen: 🤖 Keeping these local on purpose — `mooseutils.gitutils.git_ls_files()` doesn't expose the `-z`/`--stage`/`--diff-filter` options this script needs, and `pre_check.py` is meant to run standalone (git hook/CI) without pulling in the mooseutils package. Open to a shared low-level git-listing helper later if more callers need the same thing.

scripts/pre_check.py
dschwen: 🤖 Added a comment: `files` holds the files that failed the check, not all files it ran against.

scripts/pre_check.py
dschwen: 🤖 Done — moved `check_cpp_whitespace`/`check_f_whitespace` below the check-enable flags with a comment noting they're modifiers, and added `assert not (self.check_f_whitespace and not self.check_whitespace)`.

scripts/tests/test_pre_check.py
```diff
@@ -0,0 +1,517 @@
...
+
+    @patch("pre_check.read_text")
+    def test_classified_keywords(self, mock_read_text):
+        """Test classified_keywords() detects proprietary/classified"""
+        check = make_precheck([("100644", "file1.C"), ("100644", "pre_check.py")])
+        mock_read_text.side_effect = lambda f: (
+            "p r o p r i e t a r y" if f == "file1.C" else "proprietary"
+        )
+        result = check.classified_keywords()
+        self.assertEqual(result, ["file1.C"])
```
GiudGiud, line 218: shouldnt it have caught the second one too? especially the second one with no obfuscation

scripts/tests/test_pre_check.py
```diff
@@ -0,0 +1,517 @@
...
+    @patch("builtins.open", new_callable=mock_open, read_data="line1\nline2\n")
+    def test_trailing_whitespace_files_no_whitespace(self, mock_file):
+        """Test trailing_whitespace_files() with no trailing spaces"""
+        check = make_precheck([])
+        result = check.trailing_whitespace_files(["file1.py"])
+        self.assertEqual(result, [])
+
+    def test_no_newline_at_eof_files(self):
+        """Test no_newline_at_eof_files() detects missing newline"""
+        check = make_precheck([("100644", "file1.py"), ("100644", "file2.py")])
```
GiudGiud: probably should have only one file have the newline

scripts/tests/test_pre_check.py
GiudGiud: might as well do classified / OUO / UCNI etc there was the list in a recent post on teams Ai chat

scripts/tests/test_pre_check.py
```diff
@@ -0,0 +1,517 @@
...
+        self.assertTrue(pre_check._is_allowed_manual("∂"))  # Partial
+
+    def test_is_allowed_manual_superscripts(self):
+        """Test _is_allowed_manual() with superscripts/subscripts"""
+        self.assertTrue(pre_check._is_allowed_manual("²"))
+        self.assertTrue(pre_check._is_allowed_manual("₃"))
+
+    def test_is_allowed_manual_disallowed(self):
+        """Test _is_allowed_manual() with disallowed characters"""
+        self.assertFalse(pre_check._is_allowed_manual("😀"))  # Emoji
```
GiudGiud, line 327: tough times for millenials

scripts/tests/test_pre_check.py
GiudGiud: (can be done later from a feature request if preferred)

GiudGiud (commented):
probably can tell the agents to look at pre_check for the list of authorized symbols now. except there would be a token cost. 
so maybe just have the agent run pre-check before committing

scripts/tests/test_pre_check.py
dschwen: > shouldnt it have caught the second one too? especially the second one with no obfuscation

That test makes sure that the `pre_check.py` code itself is excluded from being flagged.

GiudGiud: ah smaht

GiudGiud (approved):

scripts/tests/test_pre_check.py
dschwen: fixed

PR #31534, opened 2025-09-16
Accommodate changes in PETSc 3.24.0

nmnobre:
## Reason
API changes in PETSc 3.24.0.

## Design
Add compatible function pointer cast.

## Impact
Fix compilations against PETSc 3.24.0.

### DIFF
--- a/framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
+++ b/framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
@@ -179,7 +179,7 @@ MFEMCutTransitionSubMesh::labelMesh(mfem::ParMesh & parent_mesh)
 mfem::Vector
 MFEMCutTransitionSubMesh::findFaceNormal(const mfem::ParMesh & mesh, const int & face)
 {
-  MFEM_ASSERT(mesh->SpatialDimension() == 3,
+  MFEM_ASSERT(mesh.SpaceDimension() == 3,
               "MFEMCutTransitionSubMesh only works in 3-dimensional meshes!");
   mfem::Vector normal;
   mfem::Array<int> face_verts;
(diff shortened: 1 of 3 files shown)


lindsayad (approved):

lindsayad:
failures are unrelated

framework/src/mfem/submeshes/MFEMCutTransitionSubMesh.C
```diff
@@ -179,7 +179,7 @@ MFEMCutTransitionSubMesh::labelMesh(mfem::ParMesh & parent_mesh)
 mfem::Vector
 MFEMCutTransitionSubMesh::findFaceNormal(const mfem::ParMesh & mesh, const int & face)
 {
-  MFEM_ASSERT(mesh->SpatialDimension() == 3,
+  MFEM_ASSERT(mesh.SpaceDimension() == 3,
```
loganharbour, line 182: How did this work before?!

nmnobre: It didn't. Compilations against debug builds of MFEM just failed.

loganharbour: Should we do a debug build of MFEM inside the CI containers so that we can use MFEM debug in https://civet.inl.gov/job/3221338/ too?

Just need to do that here https://github.com/idaholab/moose/blob/e605d5e8f6e883fa354b2344665ce057013a0bb5/apptainer/moose-dev.def#L338.

I honestly wouldn't object to requiring in the MOOSE linking that if we're doing moose METHOD=dbg that we require a Debug build of MFEM

loganharbour: But, I think that the above assertion should also just be switched to a `mooseAssert()` so it doesn't matter

PR #31537, opened 2025-09-16
Swap MFEM_ASSERT by mooseAssert

nmnobre:
I think you're absolutely right @loganharbour, let's get it out of the way. This is the _only_ occurrence of `MFEM_ASSERT` in MOOSE, we just let it slip in.


loganharbour (approved):

loganharbour:
Apptainer moose-openmpi and docker moose-dev-openmpi failures are unrelated; due to a package update waiting to get from devel to master.

PR #31538, opened 2025-09-16
Improve Kokkos initialization settings

NamjaeChoi:
Refs #30655.

Kokkos currently prints three types of warnings depending on the configuration:

```
Kokkos::OpenMP::initialize WARNING: OMP_PROC_BIND environment variable not set
  In general, for best performance with OpenMP 4.0 or better set OMP_PROC_BIND=spread and OMP_PLACES=threads
  For best performance with OpenMP 3.1 set OMP_PROC_BIND=true
  For unit testing set OMP_PROC_BIND=false

MPI detected: For OpenMP binding to work as intended, MPI ranks must be bound to exclusive CPU sets.

Kokkos::Cuda::initialize WARNING: running kernels compiled for compute capability 8.0 on device with compute capability 8.9 , this will likely reduce potential performance.
```

I don't have a strong opinion on whether to disable those or add environment variable as instructed.

### DIFF
--- a/framework/src/kokkos/base/KokkosMooseInit.K
+++ b/framework/src/kokkos/base/KokkosMooseInit.K
@@ -27,6 +27,13 @@ MooseInit::initKokkos()
 {
   Kokkos::InitializationSettings settings;
 
+  // Explicitly set the number of threads consistently with MOOSE
+  settings.set_num_threads(libMesh::n_threads());
+
+  // Only print warnings on the head process
+  if (comm().rank())
+    settings.set_disable_warnings(true);
+
 #ifdef MOOSE_ENABLE_KOKKOS_GPU
 
   unsigned int num_kokkos_gpus = Kokkos::num_devices();


NamjaeChoi:
@loganharbour

loganharbour:
Take a look at https://civet.inl.gov/job/3221338/ for example, when we call the app in "Check moose test capabilities"

I'm not the biggest fan of having this here. What's the most consistent solution?

NamjaeChoi:
Looks like the first two warnings are cleared when `OMP_PROC_BIND` is set. The third warning might actually inform important performance information to users. If I disable warning completely, the third warning will be gone too. So let's add `OMP_PROC_BIND` in the environment then.

loganharbour:
So you're thinking that we should set:

```
OMP_PROC_BIND=spread
OMP_PLACES=threads
```

?

loganharbour:
@roystgnr will the above cause any issues with our use of openmp in libmesh?

NamjaeChoi:
> So you're thinking that we should set:
> 
> ```
> OMP_PROC_BIND=spread
> OMP_PLACES=threads
> ```
> 
> ?

Works for me.

lindsayad (approved):

roystgnr:
@loganharbour I wish I knew!

It doesn't look like those settings should be able cause any correctness problems.

I'd actually expect `spread` to be helpful with performance, since we try to keep threads working on only their own data as much as possible ... but my intuition alone isn't as good as my-intuition plus benchmarks would be, which in turn probably isn't as good as benchmarks alone would be.  :wink:

I could try a `LIBMESH_BENCHMARK` run shortly, but on a single-socket workstation it won't be completely informative; NUMA performance impact might be different.

loganharbour:
We can have @maythuwin1313 add an export for the variables in https://github.com/idaholab/moose/pull/31538#issuecomment-3299552587 to https://github.com/idaholab/moose/pull/31479

loganharbour (approved):

roystgnr:
On my workstation those environment variables had no noticeable effect.  Benchmark suite ran half a percent slower, which is pretty much in the noise.

loganharbour:
FYI setting

```
OMP_PROC_BIND=spread
OMP_PLACES=threads
```

failed horribly for moose. Timeouts everywhere. Looked at a build and processes were being throttled. Presumably by MPI.

NamjaeChoi:
Hmm..

PR #31539, opened 2025-09-16
Mark some MFEM tests as valgrind heavy

NamjaeChoi:
## Reason
The following MFEM Valgrind tests fail deterministically due to timeout:

```
[603.6s]  TIMEOUT mfem/submeshes.MFEMDomainSubMeshBoundary FAILED (TIMEOUT) [previous results: TIMEOUT]
[603.4s]  TIMEOUT mfem/submeshes.MFEMGeometryCutTransitionSubMesh FAILED (TIMEOUT) [previous results: TIMEOUT]
```

They need to be marked as Valgrind-heavy.

## Design
Mark the tests as Valgrind-heavy.

## Impact
Avoid Valgrind failure.

### DIFF
--- a/test/tests/mfem/submeshes/tests
+++ b/test/tests/mfem/submeshes/tests
@@ -55,6 +55,7 @@
     capabilities = 'mfem'
     compute_devices = 'cpu cuda'
     recover = false
+    valgrind = heavy
   []
   [MFEMVacuumCutTransitionSubMesh]
     design = 'syntax/MFEM/ClosedCoilMagnetostatic.md'
@@ -78,5 +79,6 @@
     compute_devices = 'cpu cuda'
     recover = false
     restep = false
+    valgrind = heavy
   []  
 []


loganharbour:
@nmnobre these tests don't take particularly long usually; is there something goofy with mfem being unhappy with valgrind? We've ran across this with other libs as well

loganharbour:
Sorry - didn't mean to close.... wrong button :(

nmnobre:
> @nmnobre these tests don't take particularly long usually; is there something goofy with mfem being unhappy with valgrind? We've ran across this with other libs as well

Errrm, on my machine, with debug builds for both MOOSE and MFEM:
```
$ ./run_tests --dbg -C tests/mfem/submeshes/ -t
Parsing application capabilities... 1.58 seconds
mfem/submeshes.MFEMDomainSubMesh ........................................................... OK [1.734s]
mfem/submeshes.MFEMBoundarySubMeshTransfer ................................................. OK [1.828s]
mfem/submeshes.MFEMDomainSubMeshTransfer ................................................... OK [1.741s]
mfem/submeshes.MFEMCoupledVariableSource ................................................... OK [2.061s]
mfem/submeshes.MFEMGeometryCutTransitionSubMesh ....................................... RUNNING [9.994s]
mfem/submeshes.MFEMGeometryCutTransitionSubMesh ................................. [FINISHED] OK [21.81s]
mfem/submeshes.MFEMVacuumCutTransitionSubMesh .............................................. OK [6.992s]
mfem/submeshes.MFEMDomainSubMeshBoundary .............................................. RUNNING [9.994s]
mfem/submeshes.MFEMDomainSubMeshBoundary ........................................ [FINISHED] OK [19.26s]
```

At least an order of magnitude heavier - so not terribly surprising.
These tests are new (#31405) thus why we hadn't marked them heavy before.
@NamjaeChoi, I'll trigger a sanity run of Valgrind, hope you don't mind.

loganharbour (approved):
Good pending valgrind target success

NamjaeChoi:
I think the Valgrind recipe already has MFEM. Do we need a separate WIP recipe?

loganharbour:
No - we're just making sure that it passes with heavy

nmnobre:
> I think the Valgrind recipe already has MFEM. Do we need a separate WIP recipe?
> Don't need to run this; it's the same as the normal valgrind target now

Oh I didn't know that, apologies. Since when does that "now" apply? Can we remove the WIP recipe then?

> The following MFEM Valgrind tests fail deterministically due to timeout:

If that were true, surely devel would have been stuck? You can see both those tests passed in https://civet.inl.gov/job/3221209? Both took fairly long to run, but still comfortably below the 10min threshold.

NamjaeChoi:
> > I think the Valgrind recipe already has MFEM. Do we need a separate WIP recipe?
> > Don't need to run this; it's the same as the normal valgrind target now
> 
> Oh I didn't know that, apologies. Since when does that "now" apply? Can we remove the WIP recipe then?
> 
> > The following MFEM Valgrind tests fail deterministically due to timeout:
> 
> If that were true, surely devel would have been stuck? You can see both those tests passed in https://civet.inl.gov/job/3221209? Both took fairly long to run, but still comfortably below the 10min threshold.

Maybe I was just unlucky with the machines assigned to me, but I got stuck a few times straight.

nmnobre:
> Maybe I was just unlucky with the machines assigned to me, but I got stuck a few times straight.

Okay, fair enough - they're clearly on the heavier side that's for sure :)
And, to be quite honest, I don't really think there's any difference between heavy or non-heavy Valgrind tests (besides the timeout), they're run under the same circumstances (on push to devel), and moving some tests to the "Valgrind heavy" recipe is actually better for load balancing, since the "Valgrind" recipe is currently some 3x slower.

loganharbour:
I posted about this in the dev-ops channel in slack. Our new build clients are about twice as fast for most operations. 

This causes issues with timeouts. I might need to scale the timeout time for them to be less, but I haven't thought about that too much yet.

nmnobre (approved):
As discussed, even if not strictly needed, this should accelerate "push to devel" events on CIVET since it distributes some of the load out of the "Valgrind" recipe to the "Valgrind heavy" recipe, which despite the name is actually, overall, considerably lighter.

loganharbour:
Cancelled recipe not needed.

PR #31542, opened 2025-09-16
Use parameter type_info for index in registry

### LINKED ISSUE #31541: Use `type_index` in `ParameterRegistry`
### Motivation

https://github.com/libMesh/libmesh/pull/4235 in libMesh enables getting the `type_info` from a parameter value. This should be used in the index within `ParameterRegistry` instead of the string type.

### Design

Use `std::type_index` from the parameter as the registry map in `ParameterRegistry`.

### Impact

Should be none. Will be slightly faster than string keys.

loganharbour:
Closes #31541

Requires https://github.com/idaholab/moose/pull/31479

### DIFF
--- a/framework/include/parser/ParameterRegistry.h
+++ b/framework/include/parser/ParameterRegistry.h
@@ -11,10 +11,13 @@
 
 #include <functional>
 #include <map>
+#include <typeinfo>
+#include <typeindex>
 
 #include "libmesh/parameters.h"
 
 #include "MooseError.h"
+#include "MooseUtils.h"
 
 #ifdef MOOSE_UNIT_TEST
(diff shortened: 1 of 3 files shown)


roystgnr (approved):
Huh.  I was skeptical about the `unordered_map` switch, but `hash<type_index>` is actually in the standard, we're not just getting lucky.

loganharbour:
> Huh. I was skeptical about the `unordered_map` switch, but `hash<type_index>` is actually in the standard, we're not just getting lucky.

Hah. I had the same initial concern until I read the description of `type_index` which explicitly calls out unordered containers:

> ... can be used as index in associative and unordered associative containers

PR #31548, opened 2025-09-17
AD array variables/kernels for finite elements

lindsayad:
Closes #31506

### DIFF
--- a/framework/doc/content/source/kernels/ArrayDiffusion.md
+++ b/framework/doc/content/source/kernels/ArrayDiffusion.md
@@ -26,7 +26,7 @@ We can rearrange the fully expanded [eq:weak-array-diffusion] into
 \begin{equation}
 (\nabla \vec{u}^\ast, \mathbf{D} \nabla \vec{u}) = \sum_{e} \sum_{i=1}^{N_{\text{dof}}} \sum_{\text{qp}=1}^{N_{qp}} (|J|w)_{\text{qp}} \vec{w}_p\vec{u}_i^\ast \underline{\mathbf{D}_{\text{qp}} \nabla \vec{u}_{\text{qp}} \cdot \nabla b_{i,\text{qp}}},
 \end{equation}
-where the underlined term is the array provided by [ArrayDiffusion::computeQpResidual](ArrayDiffusion.C).
+where the underlined term is the array provided by [ArrayDiffusion::computeQpResidual](/ArrayDiffusion.C).
 The element, shape function and quadrature point summations are taken care of by MOOSE.
 It is noted that since test functions are arbitrary, $\vec{u}_i^\ast$ can be viewed as an index indicator with which the local residual goes into the global residual vector.
 Note that $\vec{a}\vec{b}$ represents element-wise multiplication, i.e. $\vec{a}\vec{b}$ is equal to vector whose *i*th element is $a_i \times b_i$, where $a$ and $b$ are two generic vectors.
@@ -43,7 +43,7 @@ With some further transformation, the kernel becomes
 \begin{equation}
 (\nabla \vec{u}^\ast, \mathbf{D} \nabla \vec{u}) = \sum_{e} \sum_{i=1}^{N_{\text{dof}}} \sum_{j=1}^{N_{\text{dof}}} \sum_{\text{qp}=1}^{N_{qp}} (|J|w)_{\text{qp}} \vec{w}_p\vec{u}_i^\ast \underline{\mathbf{D}_{\text{qp}} \nabla b_{j,\text{qp}} \cdot \nabla b_{i,\text{qp}}} \vec{u}_j,
 \end{equation}
-where the underlined part is the local Jacobian evaluated by [ArrayDiffusion::computeQpJacobian](ArrayDiffusion.C) and [ArrayDiffusion::computeQpOffDiagJacobian](ArrayDiffusion.C).
+where the underlined part is the local Jacobian evaluated by [ArrayDiffusion::computeQpJacobian](/ArrayDiffusion.C) and [ArrayDiffusion::computeQpOffDiagJacobian](/ArrayDiffusion.C).
 
 ## Example Input Syntax
 
--- a/framework/include/bcs/ADArrayDirichletBC.h
+++ b/framework/include/bcs/ADArrayDirichletBC.h
@@ -1,0 +1,31 @@
+//* This file is part of the MOOSE framework
+//* https://mooseframework.inl.gov
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ADNodalBC.h"
+
+/**
+ * Boundary condition of a Dirichlet type
+ *
+ * Sets the value in the node
+ */
+class ADArrayDirichletBC : public ADArrayNodalBC
+{
+public:
+  static InputParameters validParams();
+
+  ADArrayDirichletBC(const InputParameters & parameters);
+
+protected:
+  virtual ADRealEigenVector computeQpResidual() override;
+
+  /// The value for this BC
+  const RealEigenVector & _values;
+};
--- a/framework/include/bcs/ADArrayDirichletBC.h
+++ b/framework/include/bcs/ADArrayDirichletBC.h
@@ -14,7 +14,7 @@
 /**
  * Boundary condition of a Dirichlet type
  *
- * Sets the value in the node
+ * Sets the value at the node for each component of the array variable
  */
 class ADArrayDirichletBC : public ADArrayNodalBC
 {
--- a/framework/include/kernels/ADArrayDiffusion.h
+++ b/framework/include/kernels/ADArrayDiffusion.h
@@ -1,0 +1,23 @@
+//* This file is part of the MOOSE framework
+//* https://www.mooseframework.org
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ADArrayKernel.h"
+
+class ADArrayDiffusion : public ADArrayKernel
+{
+public:
+  static InputParameters validParams();
+
+  ADArrayDiffusion(const InputParameters & parameters);
+
+protected:
+  virtual void computeQpResidual(ADRealEigenVector & residual) override;
+};
--- a/framework/include/kernels/ADArrayDiffusion.h
+++ b/framework/include/kernels/ADArrayDiffusion.h
@@ -11,6 +11,9 @@
 
 #include "ADArrayKernel.h"
 
+/**
+ * Diffusion term for an array variable
+ */
 class ADArrayDiffusion : public ADArrayKernel
 {
 public:
--- a/framework/include/kernels/ADArrayKernel.h
+++ b/framework/include/kernels/ADArrayKernel.h
@@ -13,6 +13,9 @@
 #include "MooseVariableInterface.h"
 #include "ADFunctorInterface.h"
 
+/**
+ * Base class for array variable (equation) kernels using automatic differentiation
+ */
 class ADArrayKernel : public KernelBase,
                       public MooseVariableInterface<RealEigenVector>,
                       public ADFunctorInterface
@@ -66,6 +69,7 @@ protected:
   /// Number of components of the array variable
   const unsigned int _count;
 
+  // Pre-allocated vector for storing the AD array residual of the kernel
   std::vector<ADReal> _local_ad_re;
 
 private:
--- a/framework/include/kernels/ADArrayKernel.h
+++ b/framework/include/kernels/ADArrayKernel.h
@@ -1,0 +1,77 @@
+//* This file is part of the MOOSE framework
+//* https://www.mooseframework.org
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "KernelBase.h"
+#include "MooseVariableInterface.h"
+#include "ADFunctorInterface.h"
+
+class ADArrayKernel : public KernelBase,
+                      public MooseVariableInterface<RealEigenVector>,
+                      public ADFunctorInterface
+{
+public:
+  static InputParameters validParams();
+
+  ADArrayKernel(const InputParameters & parameters);
+
+  virtual void computeResidual() override;
+  virtual void computeJacobian() override;
+  virtual void computeOffDiagJacobian(unsigned int jvar) override;
+  virtual void jacobianSetup() override;
+
+  virtual const ArrayMooseVariable & variable() const override { return _var; }
+
+protected:
+  /**
+   * Compute this Kernel's contribution to the residual at the current quadrature point,
+   * to be filled in \p residual.
+   */
+  virtual void computeQpResidual(ADRealEigenVector & residual) = 0;
+
+  /**
+   * Put necessary evaluations depending on qp but independent on test functions here
+   */
+  virtual void initQpResidual() {}
+
+  /// This is an array kernel so we cast to a ArrayMooseVariable
+  ArrayMooseVariable & _var;
+
+  /// the current test function
+  const ArrayVariableTestValue & _test;
+
+  /// gradient of the test function
+  const ArrayVariableTestGradient & _grad_test;
+  const MappedArrayVariablePhiGradient & _array_grad_test;
+
+  /// the current shape functions
+  const ArrayVariablePhiValue & _phi;
+
+  /// gradient of the shape function
+  const ArrayVariablePhiGradient & _grad_phi;
+
+  /// Holds the solution at current quadrature points
+  const ADArrayVariableValue & _u;
+
+  /// Holds the solution gradient at current quadrature points
+  const ADArrayVariableGradient & _grad_u;
+
+  /// Number of components of the array variable
+  const unsigned int _count;
+
+  std::vector<ADReal> _local_ad_re;
+
+private:
+  /// Work vector for residual and diag jacobian
+  ADRealEigenVector _work_vector;
+
+  /// Cache variable to prevent multiple invocations of Jacobian computation for one element (recall that AD computes the Jacobian at all once for all variable couplings)
+  const Elem * _my_elem = nullptr;
+};
--- a/framework/include/kernels/ADArrayReaction.h
+++ b/framework/include/kernels/ADArrayReaction.h
@@ -11,6 +11,9 @@
 
 #include "ADArrayKernel.h"
 
+/**
+ * Array kernel for adding a reaction term
+ */
 class ADArrayReaction : public ADArrayKernel
 {
 public:
@@ -25,6 +28,6 @@ protected:
   const MaterialProperty<Real> * _r;
   /// array diffusion coefficient
   const MaterialProperty<RealEigenVector> * const _r_array;
-  /// matrix diffusion coefficient
+  /// matrix diffusion coefficient (notably for anisotropic diffusion)
   const MaterialProperty<RealEigenMatrix> * const _r_2d_array;
 };
--- a/framework/include/kernels/ADArrayReaction.h
+++ b/framework/include/kernels/ADArrayReaction.h
@@ -1,0 +1,30 @@
+//* This file is part of the MOOSE framework
+//* https://www.mooseframework.org
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ADArrayKernel.h"
+
+class ADArrayReaction : public ADArrayKernel
+{
+public:
+  static InputParameters validParams();
+
+  ADArrayReaction(const InputParameters & parameters);
+
+protected:
+  virtual void computeQpResidual(ADRealEigenVector & residual) override;
+
+  /// scalar diffusion coefficient
+  const MaterialProperty<Real> * _r;
+  /// array diffusion coefficient
+  const MaterialProperty<RealEigenVector> * const _r_array;
+  /// matrix diffusion coefficient
+  const MaterialProperty<RealEigenMatrix> * const _r_2d_array;
+};
--- a/framework/include/kernels/ADArrayReaction.h
+++ b/framework/include/kernels/ADArrayReaction.h
@@ -25,9 +25,9 @@ protected:
   virtual void computeQpResidual(ADRealEigenVector & residual) override;
 
   /// scalar diffusion coefficient
-  const MaterialProperty<Real> * _r;
+  const ADMaterialProperty<Real> * _r;
   /// array diffusion coefficient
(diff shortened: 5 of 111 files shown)


lindsayad:
griffin patch at https://github.inl.gov/ncrc/griffin/pull/2812

lindsayad:
@bwspenc I dunno if you want to review this one too?

framework/include/kernels/ADArrayDiffusion.h
```diff
@@ -0,0 +1,23 @@
...
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ADArrayKernel.h"
+
+class ADArrayDiffusion : public ADArrayKernel
```
GiudGiud, line 17:
```suggestion
/**
 * Diffusion term for an array variable
 */
class ADArrayDiffusion : public ADArrayKernel
```

framework/include/bcs/ADArrayDirichletBC.h
```diff
@@ -0,0 +1,31 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ADNodalBC.h"
+
+/**
+ * Boundary condition of a Dirichlet type
+ *
+ * Sets the value in the node
```
GiudGiud:
```suggestion
 * Sets the value at the node for each component of the array variable
```

framework/src/bcs/ADArrayDirichletBC.C
```diff
@@ -0,0 +1,41 @@
...
+    paramError("values",
+               "Number of 'values' must equal number of variable components (",
+               _var.count(),
+               ").");
+}
+
+ADRealEigenVector
+ADArrayDirichletBC::computeQpResidual()
+{
+  return _u - _values;
```
GiudGiud, line 40: so we enforce this one strongly

do we have a place in the doco where we say we enforce DirichletBCs strongly ?

framework/include/kernels/ADArrayKernel.h
```diff
@@ -0,0 +1,77 @@
...
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "KernelBase.h"
+#include "MooseVariableInterface.h"
+#include "ADFunctorInterface.h"
+
+class ADArrayKernel : public KernelBase,
```
GiudGiud, line 19:
```suggestion
/**
 * Base class for array variable (equation) kernels using automatic differentiation
 */
class ADArrayKernel : public KernelBase,
```

framework/include/kernels/ADArrayKernel.h
```diff
@@ -0,0 +1,77 @@
...
+  /// Holds the solution at current quadrature points
+  const ADArrayVariableValue & _u;
+
+  /// Holds the solution gradient at current quadrature points
+  const ADArrayVariableGradient & _grad_u;
+
+  /// Number of components of the array variable
+  const unsigned int _count;
+
+  std::vector<ADReal> _local_ad_re;
```
GiudGiud, line 73:
```suggestion
  // Pre-allocated vector for storing the AD array residual of the kernel
  std::vector<ADReal> _local_ad_re;
```

framework/include/kernels/ADArrayReaction.h
```diff
@@ -0,0 +1,30 @@
...
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ADArrayKernel.h"
+
+class ADArrayReaction : public ADArrayKernel
```
GiudGiud, line 17:
```suggestion
/**
 * Array kernel for adding a reaction term
 */
class ADArrayReaction : public ADArrayKernel
```

framework/include/kernels/ADArrayReaction.h
```diff
@@ -0,0 +1,30 @@
...
+  ADArrayReaction(const InputParameters & parameters);
+
+protected:
+  virtual void computeQpResidual(ADRealEigenVector & residual) override;
+
+  /// scalar diffusion coefficient
+  const MaterialProperty<Real> * _r;
+  /// array diffusion coefficient
+  const MaterialProperty<RealEigenVector> * const _r_array;
+  /// matrix diffusion coefficient
```
GiudGiud:
```suggestion
  /// matrix diffusion coefficient (notably for anisotropic diffusion)
```

framework/include/nodalkernels/GenericArrayNodalKernel.h
```diff
@@ -0,0 +1,35 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "ArrayNodalKernel.h"
+#include "ADArrayNodalKernel.h"
+
+template <bool is_ad>
```
GiudGiud, line 18:
```suggestion
/**
 * Template for array nodal kernels, using or not automatic differentiation
 */
template <bool is_ad>
```

framework/include/utils/MooseTypes.h
```diff
@@ -308,7 +333,7 @@ struct OutputTools
 
   // DoF value type for the template class OutputType
   typedef typename Moose::DOFType<OutputType>::type OutputData;
-  typedef MooseArray<OutputData> DoFValue;
+  typedef MooseArray<OutputData> DofValue;
```
GiudGiud, line 338: suprised you dont have to change this in a million places

framework/include/utils/MooseTypes.h
```diff
@@ -378,6 +412,9 @@ typedef MooseArray<ADRealVectorValue> ADVariableGradient;
 typedef MooseArray<ADRealTensorValue> ADVariableSecond;
 typedef MooseArray<ADRealVectorValue> ADVectorVariableValue;
 typedef MooseArray<ADRealTensorValue> ADVectorVariableGradient;
+// typedef MooseArray<ADRealEigenVector> ADArrayVariableValue;
```
GiudGiud: if we had an ArrayCoupledReaction/Force kernel we d use this

framework/include/variables/MooseVariableData.h
```diff
@@ -425,22 +428,22 @@ class MooseVariableData : public MooseVariableDataBase<OutputType>
...
   /////////////////////////// DoF value getters /////////////////////////////////////
 
-  const DoFValue & dofValuesDot() const;
-  const DoFValue & dofValuesDotOld() const;
-  const DoFValue & dofValuesDotDot() const;
-  const DoFValue & dofValuesDotDotOld() const;
+  const DofValues & dofValuesDot() const;
+  const DofValues & dofValuesDotOld() const;
+  const DofValues & dofValuesDotDot() const;
+  const DofValues & dofValuesDotDotOld() const;
```
GiudGiud, line 434: ok you did have to change it in a bunch of places

framework/include/variables/MooseVariableFE.h
```diff
@@ -29,17 +29,19 @@ class MooseVariableFE;
 typedef MooseVariableFE<Real> MooseVariable;
 typedef MooseVariableFE<RealVectorValue> VectorMooseVariable;
 typedef MooseVariableFE<RealEigenVector> ArrayMooseVariable;
+// typedef MooseVariableFE<ADRealEigenVector> ADArrayMooseVariable;
```
GiudGiud:
```suggestion
```

GiudGiud: not sure why we want a special type? in therms of AD derivative values, it's still just "1" for the relevant dof and 0 for all others?

framework/src/bcs/ADArrayDirichletBC.C
```diff
@@ -0,0 +1,41 @@
...
+
+#include "ADArrayDirichletBC.h"
+
+registerMooseObject("MooseApp", ADArrayDirichletBC);
+
+InputParameters
+ADArrayDirichletBC::validParams()
+{
+  InputParameters params = ADArrayNodalBC::validParams();
+  params.addRequiredParam<RealEigenVector>("values",
```
GiudGiud, line 18: RealEigenVector is actually controllable?

framework/src/ics/InitialConditionTempl.C
```diff
@@ -51,6 +52,7 @@ InitialConditionTempl<T>::compute()
   // The following code is a copy from libMesh project_vector.C plus it adds some features, so we
   // can couple variable values
   // and we also do not call any callbacks, but we use our initial condition system directly.
+  // Eventually we shold try to fix things so we're not duplicating code
```
GiudGiud:
```suggestion
  // Eventually we should try to fix things so we're not duplicating code
```

framework/src/kernels/ADArrayReaction.C
```diff
@@ -0,0 +1,55 @@
...
+      "reaction_coefficient",
+      "The name of the reaction, can be scalar, vector, or matrix material property.");
+  params.addClassDescription(
+      "An automatic differentiation kernel for the array reaction operator ($u$)");
+  return params;
+}
+
+ADArrayReaction::ADArrayReaction(const InputParameters & parameters)
+  : ADArrayKernel(parameters),
+    _r(hasMaterialProperty<Real>("reaction_coefficient")
```
GiudGiud:
```suggestion
    _r(hasMaterialProperty<ADReal>("reaction_coefficient")
```
same for the others

ADArrayReaction should be ADArrayMatReaction and support AD everywhere

framework/src/nodalkernels/ADArrayNodalKernel.C
```diff
@@ -0,0 +1,68 @@
...
+    _work_vector(_count)
+{
+  addMooseVariableDependency(mooseVariable());
+}
+
+void
+ADArrayNodalKernel::computeResidual()
+{
+  if (!_var.isNodalDefined())
+    return;
```
GiudGiud, line 41: should we error in constructor then? why allow a non-nodal variable as the variable for a nodal kernel?

framework/src/nodalkernels/ArrayNodalKernel.C
```diff
@@ -29,41 +29,33 @@ ArrayNodalKernel::ArrayNodalKernel(const InputParameters & parameters)
...
 {
   addMooseVariableDependency(mooseVariable());
 }
 
-void
-ArrayNodalKernel::prepareDofs()
-{
-  const dof_id_type root_dof_idx = _var.nodalDofIndex();
-  std::iota(_work_dofs.begin(), _work_dofs.end(), root_dof_idx);
-}
```
GiudGiud, line 43: that was just an optimization right

framework/src/transfers/MultiAppVariableValueSamplePostprocessorTransfer.C
```diff
@@ -363,13 +363,14 @@ MultiAppVariableValueSamplePostprocessorTransfer::execute()
...
-                      "The variable must be a constant monomial with one DoF on an element");
+          mooseAssert(dof_indices.size() == _var.count(),
+                      "The variable must be a constant monomial with one DoF on an element per "
+                      "component.");
           mooseAssert(pp_values[_cached_multiapp_pos_ids[i]] != std::numeric_limits<Real>::max(),
                       "We should have pulled all the data we needed.");
           for (unsigned int c = 0; c < n_subapps / _apps_per_component; ++c)
           {
-            solution.set(dof_indices[0] + getVariableComponent(_cached_multiapp_pos_ids[i]),
+            solution.set(dof_indices[getVariableComponent(_cached_multiapp_pos_ids[i])],
```
GiudGiud, line 373: good catch

framework/src/variables/MooseVariableDataBase.C
```diff
@@ -522,6 +508,7 @@ MooseVariableDataBase<OutputType>::nodalValueArray(Moose::SolutionState state) c
       default:
         mooseError("No current support for PreviousNL for nodal value array");
     }
+    Moose::out << "_nodal_value_array.size() : " << _nodal_value_array.size() << std::endl;
```
GiudGiud:
```suggestion
```

framework/src/variables/MooseVariableDataBase.C
```diff
@@ -754,13 +724,14 @@ void
...
   bool is_transient = _subproblem.isTransient();
-
   libmesh_assert(_dof_indices.size());
 
   auto & dof_values = _vector_tags_dof_u[_solution_tag];
   _nodal_value = dof_values[0];
   _nodal_value_array[0] = _nodal_value;
 
+  // Moose::out<<"_nodal_value"<<_nodal_value <<std::endl;
+  // Moose::out<<"_nodal_value_aray"<<_nodal_value_array[0] <<std::endl;
```
GiudGiud:
```suggestion
```

framework/src/variables/MooseVariableDataBase.C
```diff
@@ -446,22 +439,15 @@ MooseVariableDataBase<RealEigenVector>::add(NumericVector<Number> & residual)
 {
   if (_has_dof_values)
   {
-    auto & dof_values = _vector_tags_dof_u[_solution_tag];
-    if (isNodal())
+    const auto & dof_values = _vector_tags_dof_u[_solution_tag];
+    mooseAssert(_dof_indices.size() % _count == 0,
+                "Dof indices should be cleanly divisible by the variable count");
+    const auto n_shapes = _dof_indices.size() / _count;
+    for (const auto indx : index_range(_dof_indices))
```
GiudGiud, line 446: we're losing the potential to vectorize over the vector in the nodal case. we had a clear access pattern and now we're accessing inside a vector to get the index

framework/include/utils/MooseTypes.h
lindsayad: I must have this somewhere else because I use it in `ADArrayKernel.h`

framework/src/bcs/ADArrayDirichletBC.C
lindsayad: They have it this way in the non-AD version. @YaqiWang do you leverage this? I'm not sure where I would check to see whether it's actually controllable?

framework/src/bcs/ADArrayDirichletBC.C
lindsayad: `DirichletBC.md`, which both array dirichlet BCs link to, states that 

> `DirichletBC` is the simplest type of `NodalBC`, and is used for
imposing so-called "essential" boundary conditions on systems of
partial differential equations (PDEs).  Such boundary conditions force
a particular set of degrees of freedom (DOFs) defined by the
`boundary` parameter to take on a single, controllable value.

framework/src/nodalkernels/ADArrayNodalKernel.C
lindsayad: This allows you to have first order variables on a second order mesh. `isNodalDefined()` is a by-node check

framework/src/nodalkernels/ArrayNodalKernel.C
lindsayad: This is no longer needed because libMesh now gives us the complete array dof indices without us needing to do anything in MOOSE

framework/src/variables/MooseVariableDataBase.C
lindsayad: The thing is that the old code had zero robustness. Even for a `isNodal()` `LAGRANGE_VEC` variable this would not have worked because the old `_dof_indices` behavior assumed a stride of 1 between variable components in the global dof index numbering but with a `LAGRANGE_VEC` that stride is no longer 1. We may lose some performance here but I much prefer robustness over speed

framework/src/kernels/ADArrayReaction.C
```diff
@@ -0,0 +1,55 @@
...
+      "The name of the reaction, can be scalar, vector, or matrix material property.");
+  params.addClassDescription(
+      "An automatic differentiation kernel for the array reaction operator ($u$)");
+  return params;
+}
+
+ADArrayReaction::ADArrayReaction(const InputParameters & parameters)
+  : ADArrayKernel(parameters),
+    _r(hasMaterialProperty<ADReal>("reaction_coefficient")
+           ? &getMaterialProperty<Real>("reaction_coefficient")
```
GiudGiud:
```suggestion
           ? &getMaterialProperty<ADReal>("reaction_coefficient")
```
should be the same for all types imo

framework/src/variables/MooseVariableDataBase.C
GiudGiud: > We may lose some performance here but I much prefer robustness over speed
we can have both we just need the "if" condition

is there a "stride" member attribute or routine we could use?

framework/src/bcs/ADArrayDirichletBC.C
GiudGiud: you d probably need a test.
I remeber vectors being referred to in controllable code, but I have never tested it either.

framework/src/nodalkernels/ADArrayNodalKernel.C
GiudGiud: ah my b

framework/include/variables/MooseVariableFE.h
```diff
@@ -35,11 +35,12 @@ typedef MooseVariableFE<RealEigenVector> ArrayMooseVariable;
  *
  * Each variable can compute nodal or elemental (at QPs) values.
  *
- * OutputType          OutputShape           OutputData
+ * OutputType          OutputShape           DofValue
  * ----------------------------------------------------
  * Real                Real                  Real
  * RealVectorValue     RealVectorValue       Real
  * RealEigenVector     Real                  RealEigenVector
+ * ADRealEigenVector   ADReal                ADRealEigenVector
```
GiudGiud:
```suggestion
```

GiudGiud: since you deleted it from the code

framework/doc/content/source/kernels/ArrayDiffusion.md
```diff
@@ -26,7 +26,7 @@ We can rearrange the fully expanded [eq:weak-array-diffusion] into
 \begin{equation}
 (\nabla \vec{u}^\ast, \mathbf{D} \nabla \vec{u}) = \sum_{e} \sum_{i=1}^{N_{\text{dof}}} \sum_{\text{qp}=1}^{N_{qp}} (|J|w)_{\text{qp}} \vec{w}_p\vec{u}_i^\ast \underline{\mathbf{D}_{\text{qp}} \nabla \vec{u}_{\text{qp}} \cdot \nabla b_{i,\text{qp}}},
 \end{equation}
-where the underlined term is the array provided by [ArrayDiffusion::computeQpResidual](ArrayDiffusion.C).
+where the underlined term is the array provided by [ArrayDiffusion::computeQpResidual](/ArrayDiffusion.C).
```
bwspenc: Just curious about what the purpose of the leading slash is. Is that to disambiguate ADArrayDiffusion.C and ArrayDiffusion.C?

bwspenc (commented):
I looked through this and I have no concerns.

Along with our discussion about the non-AD version of this, I assume this will pick up the coupling between array variable entries at a given node in the Jacobian. If so, we should be able to use this for what we want it for.

framework/include/kernels/ADArrayReaction.h
```diff
@@ -0,0 +1,33 @@
...
+
+protected:
+  virtual void computeQpResidual(ADRealEigenVector & residual) override;
+
+  /// scalar diffusion coefficient
+  const ADMaterialProperty<Real> * _r;
+  /// array diffusion coefficient
+  const ADMaterialProperty<RealEigenVector> * const _r_array;
+  /// matrix diffusion coefficient (notably for anisotropic diffusion)
+  const ADMaterialProperty<RealEigenMatrix> * const _r_2d_array;
```
GiudGiud:
```suggestion
  /// scalar reaction coefficient
  const ADMaterialProperty<Real> * _r;
  /// array reaction coefficient
  const ADMaterialProperty<RealEigenVector> * const _r_array;
  /// matrix of reaction coefficients (for cross reaction terms between array variables)
  const ADMaterialProperty<RealEigenMatrix> * const _r_2d_array;
```

framework/include/materials/GenericConstant2DArray.h
```diff
@@ -11,12 +11,13 @@
 
 #include "Material.h"
 
-class GenericConstant2DArray : public Material
+template <bool is_ad>
```
GiudGiud, line 17:
```suggestion
/**
 * Template for defining a constant array material property, using automatic differentiation or not
 */
template <bool is_ad>
```

lindsayad:
Yea with the AD version you don't have to worry about what goes where in compute Jacobian routines because you don't have to write them

framework/include/utils/MooseTypes.h
```diff
@@ -378,6 +414,7 @@ typedef MooseArray<ADRealVectorValue> ADVariableGradient;
 typedef MooseArray<ADRealTensorValue> ADVariableSecond;
 typedef MooseArray<ADRealVectorValue> ADVectorVariableValue;
 typedef MooseArray<ADRealTensorValue> ADVectorVariableGradient;
+
```
GiudGiud:
```suggestion
```

framework/src/kernels/ADArrayReaction.C
```diff
@@ -0,0 +1,55 @@
...
+registerMooseObject("MooseApp", ADArrayReaction);
+
+InputParameters
+ADArrayReaction::validParams()
+{
+  InputParameters params = ADArrayKernel::validParams();
+  params += ADFunctorInterface::validParams();
+  params.addRequiredParam<MaterialPropertyName>(
+      "reaction_coefficient",
+      "The name of the reaction, can be scalar, vector, or matrix material property.");
```
GiudGiud:
```suggestion
      "The name of the reaction (can be either a scalar, vector, or matrix) material property.");
```
I dont really know if this is better

framework/src/kernels/ADArrayReaction.C
```diff
@@ -0,0 +1,55 @@
...
+  params.addClassDescription(
+      "An automatic differentiation kernel for the array reaction operator ($u$)");
```
GiudGiud:
```suggestion
      "An automatic differentiation kernel for the array reaction operator ($r * u$)");
```

framework/src/variables/MooseVariableDataBase.C
```diff
@@ -620,66 +606,49 @@ MooseVariableDataBase<OutputType>::fetchDoFValues()
...
-      auto dof = _dof_indices[i];
-      for (unsigned int j = 0; j < _count; ++j)
-      {
-        dof_values[i](j) = sol(dof);
-        dof += n;
-      }
-    }
+    dof_values[i].resize(_count);
+    for (const auto j : make_range(_count))
+      dof_values[i](j) = sol(_dof_indices[j * nshapes + i]);
```
GiudGiud, line 618: so why is there no concern that the vector be an array of lagrange_vec and that for each shape we need to loop from 0 to _count * 3 in this routine?

GiudGiud: as in why is it ok not to use dof indices here

GiudGiud (commented):
last round

framework/src/variables/MooseVariableDataBase.C
lindsayad: we are using the `_dof_indices`. We're directly indexing into `_dof_indices` here. We mandate in libMesh that the array dof indexing has major ordering based on count and minor ordering based on shape

framework/src/variables/MooseVariableDataBase.C
lindsayad: the stride between the same shape function for different array variable components is the number of shapes for a single component, e.g. `n_shapes`

framework/src/variables/MooseVariableDataBase.C
lindsayad: We pick this convention because, for example, users can specify different scaling factors for different array variable components. It is easy to extract contiguous submatrices and subvectors corresponding to a given array variable component using the convention we've chosen for multiplication by these potentially disparate scaling factors

GiudGiud (approved):
looks good to me

lindsayad:
thanks for your ever present and solid reviews!

PR #31551, opened 2025-09-17
Fix Kokkos build for Mac

NamjaeChoi:
Refs #30655.

### DIFF
--- a/framework/src/kokkos/base/KokkosMooseInit.K
+++ b/framework/src/kokkos/base/KokkosMooseInit.K
@@ -56,6 +56,9 @@ MooseInit::initKokkos()
 
 #endif
 
+  // Set this environment variable if not set already to silence Kokkos warning
+  setenv("OMP_PROC_BIND", "false", false);
+
   Kokkos::initialize(settings);
 
   LibmeshPetscCall(PetscRegisterFinalize(finalizeKokkos));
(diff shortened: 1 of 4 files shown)


NamjaeChoi:
Looks like this problem is more than just linking.

NamjaeChoi:
@loganharbour Kokkos Mac works now. Decided to rely on libtool only for CPU builds.

framework/src/kokkos/base/KokkosMooseInit.K
```diff
@@ -56,6 +56,9 @@ MooseInit::initKokkos()
 
 #endif
 
+  // Set this environment variable if not set already to silence Kokkos warning
+  setenv("OMP_PROC_BIND", "false", false);
```
loganharbour, line 60: I'm not super comfortable with applying this globally

lindsayad: I think this is the right thing to do. It's not going to overwrite the users's environment and it makes the default openmp setting explicit in the environment which is what Kokkkos cares about.

All this line of code should do is remove the Kokkos warning. It should not impact how code executes in any way

loganharbour (approved):
Good enough for me.

loganharbour:
@NamjaeChoi can you please put up a civet_recipes PR that removes the WIP recipe and adds the kokkos configure to the Mac condo tests?

NamjaeChoi:
https://github.com/idaholab/civet_recipes/pull/2479

loganharbour:
Conda recipe will come once this makes it to devel so that it doesn't break PRs.

PR #31552, opened 2025-09-17
Allow MooseStringUtils to be built without libmesh and use in capabil…

loganharbour:
…ities


loganharbour:
Closed in https://github.com/idaholab/moose/pull/31557

PR #31568, opened 2025-09-18
Macro assisted MooseEnum / C++ enum class pairing

### LINKED ISSUE #26278: Use MagicEnum for enum reflection
## Reason

We have pairs of corresponding `MooseEnum`s and `enum class`es all over the code. It would be cool if we could construct a MooseEnum from an `enum class`, ensuring they never go out of sync.

## Design

https://github.com/Neargye/magic_enum would help with this. I brought this up a couple of years ago, but now our minimum compiler requirements have almost caught up (we have 8.5 and magic enum needs 9). Let's discuss if we can get there in the forseeable future.

## Impact

More robust code.

dschwen:
Closes #26278

Our minimum compiler requirements are now recent enough to support this (gee I hope it works with Intel...).

Anyways, this has been requested befor (e.g. by @tophmatthews). With this capability we can automatically create the option lists for `MooseEnum`s so that they are guaranteed to match a given C++ `enum class` (and there is no risk of them getting out of sync, which can introduce subtle errors, where parameter values are not what they seem).

I'll probably prune the magic enum contrib a bit, because we only really need one of the includes currently.

### DIFF
--- a/framework/doc/content/source/utils/MooseEnum.md
+++ b/framework/doc/content/source/utils/MooseEnum.md
@@ -1,0 +1,83 @@
+# MooseEnum
+
+`MooseEnum` and `MultiMooseEnum` are utility types used throughout MOOSE for parameters with a fixed
+set of valid string values. They provide validation, optional defaults, and a convenient way to map
+validated input values back to C++ enums in implementation code.
+
+## `MooseEnum`
+
+`MooseEnum` stores one selected value from a list of valid options. The constructor takes a
+space-delimited list of allowed names and may also take a default value:
+
+```cpp
+MooseEnum point_not_found_behavior("ERROR WARNING IGNORE", "IGNORE");
+params.addParam<MooseEnum>("point_not_found_behavior",
+                           point_not_found_behavior,
+                           "Behavior to use when a point cannot be located.");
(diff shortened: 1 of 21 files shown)


dschwen:
Actually the alternative is a macro that creates the `enum class` and a function returning the `MooseEnum` string initializer. That would be a lot fewer lines of code and guaranteed to be compiler friendly.

loganharbour (changes requested):
Noice.

We need to document this in the doco for MooseEnum. Add a newsletter entry too

loganharbour:
Ping @dschwen

loganharbour:
@dschwen where you at

dschwen:
> We need to document this in the doco for MooseEnum.

and where would that be?

GiudGiud:
There's at least the class docstring

Otherwise look under content/source/base in case anything fits

dschwen:
> There's at least the class docstring
> 
> Otherwise look under content/source/base in case anything fits

Yeah, no. Didn't you see df918dc ?

loganharbour (approved):
Re-running tests; they haven't ran in a week. Good otherwise

PR #31569, opened 2025-09-18
Enable builds of MFEM at multiple optimisation levels

nmnobre:
## Reason
We'd like to have (coexisting) `dbg` and `opt` builds of MFEM that MOOSE `dbg` and `opt` binaries can, respectively, link to. Debug builds include debugging info and assertions, which we are otherwise missing on (most notably on CIVET runs).

## Design
We compile from `devel`, `dbg` and `opt` versions of the MFEM library (as requested by the user, but all by default), corresponding respectively to the CMake build types `RelWithDebInfo`, `Debug`, and `Release`. We use `CMAKE_<CONFIG>_POSTFIX` to distinguish the different libraries and manually save different versions of `_config.hpp`, which we then use when building each MOOSE build. We don't compile examples anymore to recover some of the cost of compiling MFEM multiple times (to expedite this further, we could skip `devel` by default, thoughts?). Though irrelevant for MOOSE apps, we also create symlinks to the last built version of the MFEM library (note we can control this by writing the desired method last in `METHOD(S)`) to restore the default library and header names MFEM's `config.mk` and non-MOOSE apps will be expecting.

## Impact
Better debugging and code correctness checks. People will need to rerun `update_and_rebuild_mfem.sh` when this is merged because MOOSE now looks for the postfixed MFEM libraries and config header, which didn't exist before.

### DIFF
--- a/apptainer/app.def
+++ b/apptainer/app.def
@@ -167,7 +167,7 @@ Fingerprints: 0CFFCAB55E806363601C442D211817B01E0911DB
         export MOOSE_DIR=${MOOSE_DIR_REAL}
     fi
 
-    # Remove the libmesh methods from moose-dev that we do not need
+    # Remove the libmesh, mfem and neml2 methods from moose-dev that we do not need
     for method in opt oprof devel dbg; do
         if [ "$method" == "$METHOD" ]; then
             continue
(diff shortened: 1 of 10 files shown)


loganharbour (changes requested):
I have this very disgusting logic: https://github.com/idaholab/moose/blob/bfcd56ac91457f5414d1daf68dd3a52f34139c06/apptainer/app.def#L170

...which deletes the libs that we don't need when creating an installed app. moose-dev has all methods, app just has one (usually opt)

We need something like this for mfem now

loganharbour:
Added to https://github.com/idaholab/moose/pull/31650.

PR #31582, opened 2025-09-22
Store and Check restorable data

### LINKED ISSUE #31545: Add interface to query whether a variable was restarted or not
### Motivation

It would be useful to have a method that can be called to determine whether some variable was restarted or not. An example of where this would be useful: actions are being more heavily used in MOOSE applications to set problems up. An action may setup a transient problem that is meant to be restarted from some other calculation. The action can check that the app is restarting, but there is no way currently to check if specific variables were restarted. Currently the action can add the variables that were meant to be restarted, and if they were not in fact restarted, no error will be generated, the variable values will simply be zero.

### Design

Store information about what variables have been restarted and add an interface method that allows querying this information.

### Impact

Allow actions to check directly if a variable was restarted and then issue an error if the variable was supposed to be restarted for the problem being setup.

maythuwin1313:
Closes #31545


framework/include/restart/RestartableEquationSystems.h
```diff
@@ -169,6 +179,9 @@ class RestartableEquationSystems
   std::vector<const libMesh::DofObject *> _loaded_ordered_objects;
   /// The loaded header
   EquationSystemsHeader _loaded_header;
+
+  // To store restored variables
+  std::set<std::tuple<std::string, std::string, std::string>> _restored_variable_group;
```
loganharbour:
```suggestion
  /// The variables that were loaded in load(); [system name, vector name, variable name]
  std::set<std::tuple<std::string, std::string, std::string>> _loaded_variables;
```

framework/src/restart/RestartableEquationSystems.C
```diff
@@ -309,6 +309,28 @@ RestartableEquationSystems::load(std::istream & stream)
...
+    {
+      // std::cerr << "restarted_data_get " << std::get<0>(restore_data) << " "
+      //           << std::get<1>(restore_data) << " " << std::get<2>(restore_data) << " " <<
+      //           std::endl;
+      // std::cerr << "restarted_data " << system_name << " " << vector_name << " " << variable_name
+      //           << " " << std::endl;
+      return true;
+    }
+  }
+  return false;
```
loganharbour: This is a linear search. You can search instead using the `count` method:

```
std::tuple<std::string, std::string, std::string> key(system_name, vector_name, variable_name);
return _loaded_variables.count(key);

File: framework/src/restart/RestartableEquationSystems.C [hunk 3]
```diff
@@ -455,3 +492,20 @@ dataLoad(std::istream & stream, RestartableEquationSystems::VectorHeader & heade
...
+void
+to_json(nlohmann::json & json, const RestartableEquationSystems & res)
+{
+  std::cout << "json for RestartableEquationSystems" << std::endl;
+
+  nlohmann::json loaded_vars = nlohmann::json::array();
+
+  for (const auto & tup : res.getRestoredVariableGroup())
+  {
+    const auto & [system, vector, variable] = tup;
```
loganharbour:
```suggestion
  for (const auto & [system, vector, variable] : res.getRestoredVariableGroup())
  {
```

framework/include/restart/RestartableEquationSystems.h
```diff
@@ -142,6 +142,16 @@ class RestartableEquationSystems
    */
   void setLoadAllVectors(const bool load_all_vectors) { _load_all_vectors = load_all_vectors; }
 
+  bool isVariableRestored(const std::string & system_name,
+                          const std::string & vector_name,
+                          const std::string & variable_name) const;
+
+  const std::set<std::tuple<std::string, std::string, std::string>> &
+  getRestoredVariableGroup() const
```
loganharbour: With the name change below, let's also call this `getLoadedVariables()`

loganharbour (changes requested):
We also need to add a method to `FEProblemBase` which owns the `RestartableEquationSystems` object a method that lets you get a const reference to the `RestartableEquationSystems`. This will enable other objects to be able to query your new `getLoadedVariables` method

loganharbour:
FYI @jthano - with this you can query via [system, vector, variable] names if a variable is loaded

framework/include/restart/RestartableEquationSystems.h
maythuwin1313: Updated with suggested change

framework/src/restart/RestartableEquationSystems.C
maythuwin1313: Updated with suggested change

framework/src/restart/RestartableEquationSystems.C
maythuwin1313: Updated with suggested change

framework/include/restart/RestartableEquationSystems.h
maythuwin1313: Updated with suggested change

loganharbour (changes requested):
Please update the commit messages to reference the issue #31545 (one already exists for this)

framework/include/problems/FEProblemBase.h
```diff
@@ -244,6 +244,9 @@ class FEProblemBase : public SubProblem, public Restartable
                                                   const std::string & var_name) override;
   virtual libMesh::System & getSystem(const std::string & var_name) override;
 
+  // To reference to the RestartableEquationSystems
```
loganharbour:
```suggestion
  /// Get the RestartableEquationSystems object
```

For doxygen comments for method and variable declaration, we use the standard doxygen format which is `///`

framework/include/restart/RestartableEquationSystems.h
```diff
@@ -142,6 +143,15 @@ class RestartableEquationSystems
...
   void setLoadAllVectors(const bool load_all_vectors) { _load_all_vectors = load_all_vectors; }
 
+  bool isVariableRestored(const std::string & system_name,
+                          const std::string & vector_name,
+                          const std::string & variable_name) const;
+
+  const std::set<std::tuple<std::string, std::string, std::string>> & getLoadedVariables() const
+  {
+    return _loaded_variables;
+  }
```
loganharbour, line 155: These need docstring comments as well!

framework/include/restart/RestartableEquationSystems.h
```diff
@@ -169,6 +179,9 @@ class RestartableEquationSystems
   std::vector<const libMesh::DofObject *> _loaded_ordered_objects;
   /// The loaded header
   EquationSystemsHeader _loaded_header;
+
+  // The variables that were loaded in load(); [system name, vector name, variable name]
```
loganharbour:
```suggestion
  /// The variables that were loaded in load(); [system name, vector name, variable name]
```

test/tests/restart/restart_add_variable/add_variable_restart.i
```diff
@@ -117,12 +117,22 @@
...
+
 [Outputs]
   [./out]
     type = Exodus
     elemental_as_nodal = true
     execute_elemental_on = none
   [../]
+  [json]
+    type = JSON
+    execute_system_information_on = none
```
loganharbour, line 134:
```suggestion
    execute_system_information_on = none
    execute_on = INITIAL
```

We only load the data at the beginning of the solve - thus this value shouldn't change on timesteps (look at the json output, the data is repeated for each timestep). This `execute_on` input option controls when an object will run. With this, it only outputs at the beginning (which is all we care about!)

This will require that you "regold" (change the file in gold) the json file

loganharbour (changes requested):

test/tests/restart/restart_add_variable/add_variable_restart.i
maythuwin1313: Commit with suggested changes

framework/include/restart/RestartableEquationSystems.h
maythuwin1313: Commit with suggested changes

framework/include/restart/RestartableEquationSystems.h
maythuwin1313: Added docstring comments

framework/include/problems/FEProblemBase.h
maythuwin1313: Updated with suggested changes

loganharbour (approved):

PR #31594, opened 2025-09-24
Kokkos AuxKernel System

NamjaeChoi:
Refs #30655.


NamjaeChoi:
Want to see if this is problematic or will receive pushback.

NamjaeChoi:
All gold files for Kokkos tests are being generated with equivalent non-Kokkos inputs, just FYI in case you didn't know.

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+namespace Moose
+{
+namespace Kokkos
+{
+
+/**
+ * The base class for a user to derive their own Kokkos auxiliary kernels.
+ *
+ * The user should define computeValue() as inlined public method in their derived class (not
+ * virtual override). The signature of computeQpResidual() expected to be defined in the derived
```
lindsayad:
```suggestion
 * virtual override). The signature of computeValue() expected to be defined in the derived
```

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+ * class is as follows:
+ *
+ * @param qp The local quadrature point index
+ * @param datum The ResidualDatum object of the current thread
```
lindsayad: I wonder if we could come up with a more generic name. Maybe `AssemblyDatum`?

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+    solution[i] = 0;
+    load[i] = 0;
+
+    for (unsigned int j = 0; j < datum.n_dofs(); ++j)
+      mass[j + datum.n_dofs() * i] = 0;
+  }
+
+  for (unsigned int qp = 0; qp < datum.n_qps(); ++qp)
+  {
+    auto value = auxkernel.computeValue(qp, datum);
```
lindsayad:
```suggestion
    const auto value = auxkernel.computeValue(qp, datum);
```

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+
+    datum.reinit();
+
+    for (unsigned int i = 0; i < datum.n_dofs(); ++i)
+    {
+      auto t = datum.JxW(qp) * _test(datum, i, qp);
```
lindsayad:
```suggestion
      const auto t = datum.JxW(qp) * _test(datum, i, qp);
```

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+  setNodeSolution(value, datum);
+}
+
+KOKKOS_FUNCTION inline void
+AuxKernel::setElementSolution(const Real * const values,
+                              const ResidualDatum & datum,
+                              const unsigned int comp) const
+{
+  auto & sys = kokkosSystem(_kokkos_var.sys(comp));
+  auto var = _kokkos_var.var(comp);
```
lindsayad:
```suggestion
  auto var_num = _kokkos_var.var(comp);
```

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+  auto tag = _kokkos_var.tag();
+  auto elem = datum.elem().id;
+
+  for (unsigned int i = 0; i < datum.n_dofs(); ++i)
+    sys.getVectorDofValue(sys.getElemLocalDofIndex(elem, i, var), tag) = values[i];
```
lindsayad:
```suggestion
    sys.getVectorDofValue(sys.getElemLocalDofIndex(elem, i, var_num), tag) = values[i];
```

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+}
+
+KOKKOS_FUNCTION inline void
+AuxKernel::setNodeSolution(const Real value,
+                           const ResidualDatum & datum,
+                           const unsigned int comp) const
+{
+  auto & sys = kokkosSystem(_kokkos_var.sys(comp));
+  auto var = _kokkos_var.var(comp);
```
lindsayad:
```suggestion
  auto var_num = _kokkos_var.var(comp);
```

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+  auto tag = _kokkos_var.tag();
+  auto node = datum.node();
+
+  sys.getVectorDofValue(sys.getNodeLocalDofIndex(node, var), tag) = value;
```
lindsayad:
```suggestion
  sys.getVectorDofValue(sys.getNodeLocalDofIndex(node, var_num), tag) = value;
```

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+ *
+ * The user should define computeValue() as inlined public method in their derived class (not
+ * virtual override). The signature of computeQpResidual() expected to be defined in the derived
+ * class is as follows:
+ *
+ * @param qp The local quadrature point index
+ * @param datum The ResidualDatum object of the current thread
+ * @returns The value at the quadrature point
+ *
+ * KOKKOS_FUNCTION Real computeValue(const unsigned int qp, ResidualDatum & datum) const;
```
lindsayad:
```suggestion
 * KOKKOS_FUNCTION Real computeValue(const unsigned int qp, const ResidualDatum & datum) const;
```

lindsayad: can we mark most of these `ResidualDatum` references with `const`?

framework/include/kokkos/auxkernels/KokkosVariableTimeIntegrationAux.h
```diff
@@ -0,0 +1,84 @@
...
+    Real y0 = _coupled_vars[2](datum, qp);
+    Real y1 = _coupled_vars[1](datum, qp);
+    Real y2 = _coupled_vars[0](datum, qp);
+    Real term1 = (x2 - x0) * (y0 + (x2 - x0) * (y1 - y0) / (2.0 * (x1 - x0)));
+    Real term2 = (2.0 * x2 * x2 - x0 * x2 - x0 * x0 + 3.0 * x0 * x1 - 3.0 * x1 * x2) / 6.0;
+    Real term3 = (y2 - y1) / (x2 - x1) - (y1 - y0) / (x1 - x0);
+    integral_value = term1 + term2 * term3;
+  }
+
+  return integral_value;
```
lindsayad, line 83: Could we make a templated free function or something which both this class and the non-Kokkos class can leverage?

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
```diff
@@ -0,0 +1,317 @@
...
+  auxkernel.computeNodeInternal(auxkernel, datum);
+}
+
+template <typename Derived>
+KOKKOS_FUNCTION void
+AuxKernel::computeElementInternal(const Derived & auxkernel, ResidualDatum & datum) const
+{
+  Real solution[MAX_CACHED_DOF];
+  Real load[MAX_CACHED_DOF];
+  Real mass[MAX_CACHED_DOF * MAX_CACHED_DOF];
```
lindsayad:
```suggestion
  Real x[MAX_CACHED_DOF];
  Real b[MAX_CACHED_DOF];
  Real A[MAX_CACHED_DOF * MAX_CACHED_DOF];
```

framework/include/kokkos/base/KokkosUtils.h
```diff
@@ -50,6 +50,54 @@ find(const T & target, const T * const begin, const T * const end)
...
+/**
+ * Perform an in-place linear solve using Cholesky decomposition
+ * Matrix and right-hand-side vector are modified after this call
+ * @param A The row-major matrix
+ * @param x The solution vector
+ * @param b The right-hand-side vector
+ * @param n The system size
+ */
+KOKKOS_INLINE_FUNCTION void
+choleskySolve(Real * A, Real * x, Real * b, const unsigned int n)
```
lindsayad:
```suggestion
choleskySolve(Real * const A, Real * const x, Real * const b, const unsigned int n)
```

framework/include/kokkos/base/KokkosVariableValue.h
```diff
@@ -259,5 +214,70 @@ class VariableGradient
...
+    auto tag = _var.tag();
+
+    if (_dof)
+    {
+      unsigned int dof;
+
+      if (datum.isNodal())
+      {
+        auto node = datum.node();
+        dof = sys.getNodeLocalDofIndex(node, var);
```
lindsayad: Cubic hierarchics for instance can have multiple dof indices per node

lindsayad: I know the Lagrange logic is pervasive in MOOSE still 😢

framework/include/kokkos/systems/KokkosSystem.h
```diff
@@ -63,11 +63,16 @@ class System : public MeshHolder, public AssemblyHolder
    * @param vars The active MOOSE variables
    */
   void setActiveVariables(const std::set<MooseVariableFieldBase *> & vars);
+  /**
+   * Set the active variable tags
```
lindsayad: what are variable tags?

framework/include/problems/FEProblemBase.h
```diff
@@ -937,6 +937,12 @@ class FEProblemBase : public SubProblem, public Restartable
                                   const std::string & name,
                                   InputParameters & parameters);
 
+#ifdef MOOSE_KOKKOS_ENABLED
+  virtual void addKokkosAuxKernel(const std::string & kernel_name,
```
lindsayad, line 951:
```suggestion
  void addKokkosAuxKernel(const std::string & kernel_name,
```

framework/src/base/Moose.C
```diff
@@ -557,6 +560,8 @@ associateSyntaxInner(Syntax & syntax, ActionFactory & /*action_factory*/)
 #ifdef MOOSE_KOKKOS_ENABLED
   registerSyntaxTask("AddKokkosKernelAction", "KokkosKernels/*", "add_kernel");
   registerSyntaxTask("AddKokkosNodalKernelAction", "KokkosNodalKernels/*", "add_nodal_kernel");
+  registerSyntaxTask("AddKokkosKernelAction", "KokkosAuxKernels/*", "add_aux_kernel");
```
lindsayad, line 563: I believe per a conversation we had in our NEAMS MP meeting that we'll eventually put these under our standard syntax?

framework/src/kokkos/problems/KokkosFEProblemBase.K
```diff
@@ -104,6 +110,46 @@ FEProblemBase::getKokkosMaterialData(Moose::MaterialDataType type, const MooseOb
...
+      // provided in the Mesh block.  If that happened, update the value
+      // of use_displaced_mesh appropriately for this AuxKernel.
+      if (parameters.have_parameter<bool>("use_displaced_mesh"))
+        parameters.set<bool>("use_displaced_mesh") = false;
+    }
+
+    parameters.set<SubProblem *>("_subproblem") = this;
+    parameters.set<SystemBase *>("_sys") = _aux.get();
+    parameters.set<SystemBase *>("_nl_sys") = _solver_systems[0].get();
+  }
```
lindsayad: Maybe we need something similar to `setResidualObjectParamsAndLog`? A lot of duplication here

framework/src/kokkos/systems/KokkosAuxiliarySystem.K
```diff
@@ -0,0 +1,152 @@
...
+  {
+    TIME_SECTION("KokkosAuxKernel", 1);
+
+    // Compute auxiliary kernels
+
+    for (auto elemental : _kokkos_elemental_aux_storage[type].getActiveObjects())
+      elemental->compute();
+
+    for (auto nodal : _kokkos_nodal_aux_storage[type].getActiveObjects())
+      nodal->compute();
```
lindsayad: I would switch the order of these to be consistent with what is done for non-kokkos

framework/src/problems/FEProblemBase.C
```diff
@@ -1463,13 +1463,25 @@ FEProblemBase::initialSetup()
...
-    TIME_SECTION("computeMaterials", 2, "Computing Initial Material Properties");
+    if (_material_props.hasStatefulProperties() || _bnd_material_props.hasStatefulProperties() ||
+        _neighbor_material_props.hasStatefulProperties())
+    {
+      TIME_SECTION("computeMaterials", 2, "Computing Initial Material Properties");
+
+      initElementStatefulProps(*_mesh.getActiveLocalElementRange(), true);
+    }
+#ifdef MOOSE_KOKKOS_ENABLED
+    else if (_kokkos_material_props.hasStatefulProperties() ||
```
```suggestion
    if (_kokkos_material_props.hasStatefulProperties() ||
```
Or if you cannot have them both at once we need an error check somewhere. If the error check is not here then and at least an assert here

framework/src/systems/AuxiliarySystem.C
```diff
@@ -95,6 +100,14 @@ AuxiliarySystem::initialSetup()
     _elemental_array_aux_storage.sort(tid);
     _elemental_array_aux_storage.initialSetup(tid);
   }
+
+#ifdef MOOSE_KOKKOS_ENABLED
+  _kokkos_nodal_aux_storage.sort(0);
+  _kokkos_nodal_aux_storage.initialSetup(0);
+
+  _kokkos_elemental_aux_storage.sort(0);
+  _kokkos_elemental_aux_storage.initialSetup(0);
```
lindsayad:
```suggestion
  _kokkos_nodal_aux_storage.sort(/*tid=*/0);
  _kokkos_nodal_aux_storage.initialSetup(/*tid=*/0);

  _kokkos_elemental_aux_storage.sort(/*tid=*/0);
  _kokkos_elemental_aux_storage.initialSetup(/*tid=*/0);
```
and similarly elsewhere

test/tests/kokkos/auxkernels/copy_value_aux/tests
```diff
@@ -0,0 +1,20 @@
...
+    csvdiff = 'kokkos_copy_aux_out_results_0001.csv'
+    requirement = 'The Kokkos system shall be able to copy from a variable to another of the same finite element family and order.'
+    capabilities = 'kokkos'
+    compute_devices = 'cpu cuda'
+  []
+  [copy_old]
+    type = 'CSVDiff'
+    input = 'kokkos_copy_old_aux.i'
+    csvdiff = 'kokkos_copy_old_aux_out.csv'
+    requirement = 'The Kokkos system shall be able to copy from a variable OLD and OLDER state to an AuxVariable.'
```
lindsayad:
```suggestion
    requirement = 'The Kokkos system shall be able to copy from a variables old and older state to an auxiliary variable.'
```
We should keep implementation details out of requirements (generally) as in theory we develop the requirement before the code. I struggle with "Kokkos" in these requirements but I've allowed MFEM in the MFEM test spec requirements

test/tests/kokkos/auxkernels/pp_depend/kokkos_pp_depend.i
```diff
@@ -0,0 +1,80 @@
...
+    boundary = right
+    value = 1
+  []
+[]
+
+[Postprocessors]
+  [./t_pp]
+    type = FunctionValuePostprocessor
+    function = t_func
+  [../]
```
lindsayad:
```suggestion
  [t_pp]
    type = FunctionValuePostprocessor
    function = t_func
  []
```

test/tests/kokkos/auxkernels/pp_depend/tests
```diff
@@ -0,0 +1,12 @@
+[Tests]
+  issues = '#30655'
+  design = 'syntax/KokkosAuxKernels/index.md'
+  [test]
+    type = 'Exodiff'
+    input = 'kokkos_pp_depend.i'
+    exodiff = 'kokkos_pp_depend_out.e'
+    requirement = 'The Kokkos system shall be able to execute a UserObject before the execution of an AuxKernel if the latter depends on the former.'
```
lindsayad:
```suggestion
    requirement = 'The Kokkos system shall be able to execute user objects before the execution of an auxiliary kernel if the latter depends on the former.'
```

framework/doc/content/syntax/KokkosAuxKernels/index.md
```diff
@@ -0,0 +1,58 @@
+# Kokkos AuxKernels System
+
+!if! function=hasCapability('kokkos')
+
+Before reading this documentation, consider reading the following materials first for a better understanding of this documentation:
+
+- [AuxKernels System](syntax/AuxKernels/index.md) to understand the MOOSE kernel system,
```
lindsayad:
```suggestion
- [AuxKernels System](syntax/AuxKernels/index.md) to understand the MOOSE auxiliary kernel system,
```

framework/doc/content/syntax/KokkosAuxKernels/index.md
```diff
@@ -0,0 +1,58 @@
...
+The Kokkos-MOOSE auxiliary kernels can be created by inheriting `Moose::Kokkos::AuxKernel` and registering them with `registerKokkosAuxKernel()`.
+There is no vector or array version at this moment.
+The signature of the `computeValue()` hook method is now as follows:
+
+```cpp
+KOKKOS_FUNCTION Real computeValue(const unsigned int qp, ResidualDatum & datum) const;
+```
+
+To reiterate, the hook method no longer uses virtual dispatch.
+Instead, it should be statically defined in the derived class as +*inlined public*+ method.
```
lindsayad, line 23: Do you prefer `KOKKOS_FUNCTION inline` over `KOKKOS_INLINE_FUNCTION`?

framework/doc/content/syntax/KokkosAuxKernels/index.md
```diff
@@ -0,0 +1,58 @@
...
+         caption=The `KokkosVariableTimeIntegrationAux` header file.
+
+!listing framework/src/kokkos/auxkernels/KokkosVariableTimeIntegrationAux.K id=kokkos-time-integration-aux-source language=cpp
+         caption=The `KokkosVariableTimeIntegrationAux` source file.
+
+## Advanced Topics
+
+The computational loops are defined in `computeElementInternal()` and `computeNodeInternal()` for elemental and nodal variables, respectively, and they call the hook method `computeValue()`.
+Depending on the type of operations, however, you may be able to write more efficient kernels by customizing the loops directly.
+For example, what `KokkosCopyValueAux` does is simply copying an auxiliary variable to another, and the shape function family and order are always identical between the two variables.
```
lindsayad:
```suggestion
For example, what `KokkosCopyValueAux` does is simply copy one auxiliary variable to another, and the shape function family and order are always identical between the two variables.
```

lindsayad (changes requested):

lindsayad:
A general concern I have is the object duplication. Because what if someone comes along and fixes a bug or adds a feature in the legacy class? Or what if someone does the same for the kokkos class? We have to consider updating the code in multiple places. What I tell myself to make myself feel better is that in the future the kokkos objects will become *the* objects because they can be configured to execute on arbitrary backends including cpu and so in the long term we won't have this duplication. @NamjaeChoi do you have any thoughts on this?

NamjaeChoi:
> A general concern I have is the object duplication. Because what if someone comes along and fixes a bug or adds a feature in the legacy class? Or what if someone does the same for the kokkos class? We have to consider updating the code in multiple places. What I tell myself to make myself feel better is that in the future the kokkos objects will become _the_ objects because they can be configured to execute on arbitrary backends including cpu and so in the long term we won't have this duplication. @NamjaeChoi do you have any thoughts on this?

That is a legitimate concern and I'm also concerned about Kokkos objects lagging features in the long run. Not a serious concern at this moment though because Kokkos objects are currently catching up the original features. And at the same time, I don't think Kokkos objects having the full set of features is realistic, although they are designed to resemble the original objects. Kokkos objects replacing original objects seems ambitious.

lindsayad:
So you think we will have duplicates in perpetuity?

NamjaeChoi:
I believe so

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
NamjaeChoi: I like it

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
NamjaeChoi: This can't be const because some APIs are non-const, like `JxW()` and `xyz()`. They are computed (or loaded) and cached on demand.

framework/include/kokkos/systems/KokkosSystem.h
NamjaeChoi: I distinguish them as follows:
Variable tags - Tags associated with coupled variables which are read-only (and QP values might be required)
Solution tags - Tags associated with solution vectors that should be written

framework/src/base/Moose.C
NamjaeChoi: I will have a PR later that will change this all at once.

framework/doc/content/syntax/KokkosAuxKernels/index.md
NamjaeChoi: Just a personal preference.

framework/include/kokkos/base/KokkosVariableValue.h
NamjaeChoi: Hmm that's problematic. With my poor FEM background I liked the nodal logic in MOOSE.

framework/include/problems/FEProblemBase.h
NamjaeChoi: Why others are marked as virtual?

framework/include/kokkos/auxkernels/KokkosVariableTimeIntegrationAux.h
NamjaeChoi: If so, where should it go?

NamjaeChoi: By the way, that's a bit tricky because that function will have to be `KOKKOS_FUNCTION`, which is undefined outside of Kokkos scope. Might need dirty preprocessors.

framework/include/problems/FEProblemBase.h
lindsayad: Some of that is leveraged by the `DumpObjectsProblem`. I'd say either implement this new virtual in `DumpObjectsProblem` (probably best) or drop the `virtual`

framework/include/kokkos/systems/KokkosSystem.h
lindsayad: are they both vector tags?

framework/include/kokkos/auxkernels/KokkosAuxKernel.h
lindsayad: 👍

framework/include/kokkos/auxkernels/KokkosVariableTimeIntegrationAux.h
lindsayad: Yea maybe we just have to live with the duplication

framework/include/kokkos/base/KokkosVariableValue.h
lindsayad: @roystgnr and I really don't like it haha. So we've been slowly trying to eradicate it

lindsayad: You could add a `component` argument to this API and default it to 0 although I I generally prefer not to default things

NamjaeChoi: I think I haven't considered such case at all. Might need to visit several places to fix it and change the underlying data structure. What else can have multiple DOFs per node?

framework/include/kokkos/systems/KokkosSystem.h
NamjaeChoi: Yes

framework/include/kokkos/base/KokkosVariableValue.h
NamjaeChoi: Or probably I will just error out for now whenever such case is detected.

lindsayad: `HERMITE`, `LAGRANGE_VEC` are a couple other examples

framework/include/kokkos/systems/KokkosSystem.h
lindsayad: Is this in the documentation somewhere? We would definitely want documentation since we don't have a different enumerator in `VectorTagType` for a variable tag. Should we introduce one? Could it be helpful for sanity/error checking?

NamjaeChoi: I will document it in more detail in Kokkos system.

framework/include/kokkos/base/KokkosVariableValue.h
NamjaeChoi: If those DOFs that belong to the same node are always contiguous for a variable, my life would be a bit easier. Is it contiguous?

roystgnr: DoFs that belong to the same node are currently always contiguous for a variable.  `obj.dof_number(sysnum, varnum, comp) == obj.dof_number(sysnum, varnum, 0) + comp` is currently always true for any valid inputs (unsigned `sysnum < obj.n_systems()`, `varnum < obj.n_vars(sysnum)`, `comp < obj.n_comp(sysnum, varnum)`).

Don't make that assumption without adding an assertion to the contrary everywhere you're making it, though.  We've had requests for alternate numbering options (e.g. to make corresponding dofs for different components in an array variable contiguous) in the past, and we're not promising to never implement any in the future, and the only *guaranteed* behavior is `obj.dof_number(sysnum, varnum, comp) == obj.dof_number(sysnum, varnum, comp)`

NamjaeChoi: So are things like those considered as "nodal" in MOOSE? Can they use nodal BCs and nodal kernels?

lindsayad: Unfortunately not. But that doesn't mean we shouldn't try to do better in the present 😄

NamjaeChoi: Okay, that means I don't need to fix those Kokkos objects for now.

lindsayad: No I would say not. @roystgnr what does the constraint matrix look like for Dirichlet boundary conditions in libmesh when there are collocated degrees of freedom?

With respect to nodal kernels... those should really only be used for ordinary differential equations that we want to locate at nodes and shouldn't be used in conjunction with ordinary kernels. E.g. finite elements shouldn't be involved. So we should only allow something whose degree of freedom values are the values at the nodes

roystgnr: The constraint matrix row for a Dirichlet constraint just ends up empty, regardless of basis!

Even if the DirichletBoundary constraint is for a more complicated basis, we still end up solving projection equations until we get the same sort of `u_i = h_i`.  (although not necessarily the exact same `h_i` - if we have a quadratic non-Lagrange element then we use quadrature on boundary edges and faces to L2 project a specified dirichlet boundary function into the finite element space, whereas with a quadratic Lagrange element we just evaluate at nodes)

For heterogeneous constraints the constraint system looks like u=Cy+h, so for a Dirichlet `u_i`, every `C_ij` needs to be 0; `u_i` only depends on `h_i`, not on any unconstrained DoF.

But bear in mind that you're specifying "in libMesh" here.  We still don't have any option to use libMesh Dirichlet boundaries in MOOSE, AFAIK.

framework/src/kokkos/systems/KokkosSystem.K
```diff
@@ -150,7 +150,14 @@ System::setupDofs()
         _dof_map.dof_indices(node, dof_indices, var);
 
         if (dof_indices.size())
+        {
+          for (unsigned int i = 1; i < dof_indices.size(); ++i)
+            if (dof_indices[i] != dof_indices[i - 1] + 1)
+              mooseError("Kokkos system error: a variable has multiple DOFs on a node, but the DOF "
+                         "indices are discontiguous. This is not supported.");
```
lindsayad, line 157: Nice check!

framework/include/kokkos/systems/KokkosSystem.h
lindsayad: Have you added this yet? I think this is my only unresolved conversation at this point

NamjaeChoi: It was added in another place:
```
  /**
   * List of active tags
   * Variable tags are vector tags associated with coupled variables whose quadrature point values
   * need to be computed
   * Solution tags are vector tags associated with solution vectors to be written - required by
   * auxiliary kernels
   * Residual tags are vector tags associated with residual vectors to be written - required by
   * residual objects
   */
  ///@{
  Array<TagID> _active_variable_tags;
  Array<TagID> _active_solution_tags;
  Array<TagID> _active_residual_tags;
  Array<TagID> _active_matrix_tags;
  ///@}
```

lindsayad: This makes it seem like solution tags are for vector writing, but they're traditionally used to denote vectors for reading. Take this code from `MooseVariableDataBase::fetchDoFValues()`
```c++
  for (auto tag : _required_vector_tags)
    if (_need_vector_tag_u[tag] || _need_vector_tag_grad[tag] || _need_vector_tag_dof_u[tag])
      if ((_subproblem.vectorTagType(tag) == Moose::VECTOR_TAG_RESIDUAL &&
           _subproblem.safeAccessTaggedVectors()) ||
          _subproblem.vectorTagType(tag) == Moose::VECTOR_TAG_SOLUTION)
      {
        // tag is defined on problem but may not be used by a system
        // the grain tracker requires being able to read from solution vectors that we are also in
        // the process of writing :-/
        // Note: the extra vector tags are also still not closed when a TagVectorAux uses them
        if (_sys.hasVector(tag) /* && _sys.getVector(tag).closed()*/)
        {
          auto & vec = _sys.getVector(tag);
          _vector_tags_dof_u[tag].resize(n);
          vec.get(_dof_indices, &_vector_tags_dof_u[tag][0]);
        }
      }
```

NamjaeChoi: Actually, the solution vector being written doesn't even have a tag. So I had to associate it separately.

NamjaeChoi: Any better name?

NamjaeChoi: Changed the names to be more consistent with MOOSE

lindsayad (approved):
Love the last commit, thank you

NamjaeChoi:
Let's go

PR #31603, opened 2025-09-25
Fix XFEM 3d connected nodes issue

### LINKED ISSUE #31602: Incorrectly connected-nodes issue with 3D XFEM
### Bug Description

In some cases, the nodes are incorrectly connected near the cutting plane. This suggests there are bugs or missing corner cases in the XFEM cutting algorithm, particularly in how child elements are connected and how nodes are duplicated.

### Steps to Reproduce

Image

Left is the correct result. Right is the result obtained using the latest MOOSE. 

### Impact

This will cause wrong displacement and stress field. 

### [Optional] Diagnostics

_No response_

jiangwen84:
closes #31602

### DIFF
--- a/modules/xfem/src/efa/EFAElement3D.C
+++ b/modules/xfem/src/efa/EFAElement3D.C
@@ -1011,6 +1011,9 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
       EFAPoint normal(0.0, 0.0, 0.0);
       EFAPoint orig(0.0, 0.0, 0.0);
 
+      std::vector<EFAPoint> normal_v;
+      std::vector<EFAPoint> orig_v;
+
       if (cut_plane_nodes.size())
       {
         for (unsigned int i = 0; i < cut_plane_nodes[0].size(); ++i)
@@ -1028,6 +1031,7 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
           }
           cut_plane_points.push_back(coor);
         }
+
         for (unsigned int i = 0; i < cut_plane_points.size(); ++i)
           orig += cut_plane_points[i];
         orig /= cut_plane_points.size();
@@ -1039,13 +1043,32 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
 
         for (unsigned int i = 0; i < cut_plane_points.size(); ++i)
         {
+          int iplus1 = i < cut_plane_points.size() - 1 ? i + 1 : 0;
+          int iplus2 = i + 2;
+          if (iplus2 + 2 - cut_plane_points.size() >= 0)
+            iplus2 = iplus2 + 2 - cut_plane_points.size();
+
+          EFAPoint center(0.0, 0.0, 0.0);
+
+          center = cut_plane_points[i] + cut_plane_points[iplus1] + cut_plane_points[iplus2];
+          center /= 3;
+
+          orig_v.push_back((center));
+        }
+
+        for (unsigned int i = 0; i < cut_plane_points.size(); ++i)
+        {
           unsigned int iplus1 = i < cut_plane_points.size() - 1 ? i + 1 : 0;
-          EFAPoint ray1 = cut_plane_points[i] - center;
-          EFAPoint ray2 = cut_plane_points[iplus1] - center;
+          EFAPoint ray1 = cut_plane_points[i] - orig;
+          EFAPoint ray2 = cut_plane_points[iplus1] - orig;
           normal += ray1.cross(ray2);
+          normal_v.push_back(ray1.cross(ray2));
         }
         normal /= cut_plane_points.size();
         Xfem::normalizePoint(normal);
+
+        for (unsigned int i = 0; i < normal_v.size(); ++i)
+          Xfem::normalizePoint(normal_v[i]);
       }
 
       // get child element's nodes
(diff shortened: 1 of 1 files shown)


jiangwen84:
@bwspenc

bwspenc:
Thanks for submitting, @jiangwen84! I'll start working on this.

bwspenc:
@jiangwen84 I just updated the description in the PR to reference the issue you created. I think we need to reference that in one of the commit messages to pass precheck, though. I'm going to see if I can push a commit that does that now.

bwspenc:
@jiangwen84 I added a commit that just made some whitespace changes and included an issue number. That got it past precheck so we can see if this breaks tests.

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1011,6 +1011,9 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
       EFAPoint normal(0.0, 0.0, 0.0);
       EFAPoint orig(0.0, 0.0, 0.0);
 
+      std::vector<EFAPoint> normal_v;
+      std::vector<EFAPoint> orig_v;
```
bwspenc: `orig_v` is not used. Delete?

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1011,6 +1011,9 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
       EFAPoint normal(0.0, 0.0, 0.0);
```
bwspenc: With these changes, `normal` is not used. Delete.

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1037,15 +1041,34 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
           center += cut_plane_points[i];
         center /= cut_plane_points.size();
 
+        for (unsigned int i = 0; i < cut_plane_points.size(); ++i)
```
bwspenc: This whole block can be deleted if `orig_v` is not needed (as it appears is the case).

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1037,15 +1041,34 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
...
+        {
+          int iplus1 = i < cut_plane_points.size() - 1 ? i + 1 : 0;
+          int iplus2 = i + 2;
+          if (iplus2 + 2 - cut_plane_points.size() >= 0)
```
bwspenc: This is causing a compile error that says this comparison is always true. I'm impressed that the compiler caught that! That's irrelevant if you delete this code block, though.

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1037,15 +1041,34 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
...
+            iplus2 = iplus2 + 2 - cut_plane_points.size();
+
+          EFAPoint center(0.0, 0.0, 0.0);
```
bwspenc: This variable is shadowing the variable with the same name outside this block. Again, deleting the block resolves this issue.

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1037,15 +1041,34 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
...
+
+          orig_v.push_back((center));
+        }
+
         for (unsigned int i = 0; i < cut_plane_points.size(); ++i)
         {
           unsigned int iplus1 = i < cut_plane_points.size() - 1 ? i + 1 : 0;
-          EFAPoint ray1 = cut_plane_points[i] - center;
-          EFAPoint ray2 = cut_plane_points[iplus1] - center;
+          EFAPoint ray1 = cut_plane_points[i] - orig;
```
bwspenc: My intuition is that it would be better to make the rays go between the current point and the two neighbors on either side. Something like this:
```
EFAPoint ray1 = cut_plane_points[i-1] -cut_plane_points[i];
EFAPoint ray2 = cut_plane_points[i+1] -cut_plane_points[i];
```
You would need to adjust the indices for when they go outside the size of cut_plane_points, and also check for the case where those rays are collinear, but this seems like it would give a more meaningful nodal normal.

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1054,9 +1077,18 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
         EFAPoint p(0.0, 0.0, 0.0);
         p = _local_node_coor[j];
         EFAPoint origin_to_point = p - orig;
+
+        bool use_temp = false;
```
bwspenc: This variable could use a better name to improve readability. Maybe `node_outside_fragment`?

modules/xfem/src/efa/EFAElement3D.C
```diff
@@ -1054,9 +1077,18 @@ EFAElement3D::createChild(const std::set<EFAElement *> & CrackTipElements,
...
+          if (origin_to_point * normal_v[k] > Xfem::tol)
+          {
+            use_temp = true;
+            break;
+          }
+
         if (_fragments.size() == 1 && !shouldDuplicateForCrackTip(CrackTipElements))
           childElem->setNode(j, _nodes[j]); // inherit parent's node
-        else if (origin_to_point * normal < Xfem::tol)
+        else if (!use_temp)
```
bwspenc: Why are we relying these geometric checks to see if the node is in the fragment? We have the list of nodes in the fragment in its data structure. Can't we just check to see if it's in that set of nodes?

Is the purpose of this to deal with higher-order elements? If that's the case, maybe we could do both the geometric check and see if it's in the list of nodes?

bwspenc (changes requested):

bwspenc (approved):
This addresses the issues I raised -- thanks!

bwspenc:
A few tests were failing, but it looks like they were system issues so I invalidated them. We'll see if they work. I'm going to try out this latest version on the test case to see if it still fixes the issue. I'll also investigate the other way of determining if nodes are in the fragment, as we discussed, and will maybe submit another PR to add that.

bwspenc:
I just confirmed that this updated version still does resolve the large test case. I'll merge this when the tests pass.

jiangwen84:
> A few tests were failing, but it looks like they were system issues so I invalidated them. We'll see if they work. I'm going to try out this latest version on the test case to see if it still fixes the issue. I'll also investigate the other way of determining if nodes are in the fragment, as we discussed, and will maybe submit another PR to add that.

I will take a look at alternative way to determine nodes location.

PR #31632, opened 2025-09-30
Fix dropped arguments bug

dschwen:
## Reason
A `.bashrc_apptainer` file in the user's home directory triggers a bug which leads to all but the first arguments in an apptainer exec call to be dropped. This for example breaks `moose-dev-exec`.

## Design
Turns out that a typo in `apptainer/files/mpi/.singularity.d/env/89-user-overrides.sh` is responsible for this. 

## Impact
Fixing this will enable the currently broken override capability.


dschwen:
This is not a high priority fix, so I suggest we merge it and not have a dedicated conda stack update for it.

dschwen:
Oh, cool, that's not an option apparently...

PR #31634, opened 2025-09-30
Make `getUNodal` function in `ConservativeAdvection`

### LINKED ISSUE #31633: Allow u_nodal to be modified in `ConservativeAdvection`
### Motivation

`ConservativeAdvection` uses `_u_nodal` in the upwinding formulation. I have a downstream application that is using the log transform of `u`, thus needs to plug in `exp(u_nodal)` into upwinding. Instead of making function `ConservativeAdvection::fullUpwind` virtual (which takes an enum and makes inheritance difficult) I would rather make a virtual getter function for `_u_nodal` I can pass a exp transformed value in downstream apps.

### Design

Make `virtual GenericReal<is_ad> getUNodal(unsigned int n) { return _u_nodal[n]; }` in `ConservativeAdvection`.

### Impact

More downstream application flexibility.

tophmatthews:
Closes #31633

### DIFF
--- a/framework/include/kernels/ConservativeAdvection.h
+++ b/framework/include/kernels/ConservativeAdvection.h
@@ -31,7 +31,7 @@ protected:
   virtual Real computeQpOffDiagJacobian(unsigned int jvar) override;
   virtual void computeResidual() override;
   virtual void computeJacobian() override;
-  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) { return _u_nodal[n]; }
+  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) const { return _u_nodal[n]; }
 
   /// Material property multiplied against the velocity to scale advection strength
   const GenericMaterialProperty<Real, is_ad> & _scalar;
--- a/framework/include/kernels/ConservativeAdvection.h
+++ b/framework/include/kernels/ConservativeAdvection.h
@@ -31,6 +31,7 @@ protected:
   virtual Real computeQpOffDiagJacobian(unsigned int jvar) override;
   virtual void computeResidual() override;
   virtual void computeJacobian() override;
+  virtual GenericReal<is_ad> getUNodal(unsigned int n) { return _u_nodal[n]; }
 
   /// Material property multiplied against the velocity to scale advection strength
   const GenericMaterialProperty<Real, is_ad> & _scalar;
--- a/framework/include/kernels/ConservativeAdvection.h
+++ b/framework/include/kernels/ConservativeAdvection.h
@@ -31,7 +31,7 @@ protected:
   virtual Real computeQpOffDiagJacobian(unsigned int jvar) override;
   virtual void computeResidual() override;
   virtual void computeJacobian() override;
-  virtual GenericReal<is_ad> getUNodal(unsigned int n) { return _u_nodal[n]; }
+  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) { return _u_nodal[n]; }
 
   /// Material property multiplied against the velocity to scale advection strength
   const GenericMaterialProperty<Real, is_ad> & _scalar;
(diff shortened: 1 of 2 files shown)


framework/include/kernels/ConservativeAdvection.h
```diff
@@ -31,6 +31,7 @@ class ConservativeAdvectionTempl : public GenericKernel<is_ad>
   virtual Real computeQpOffDiagJacobian(unsigned int jvar) override;
   virtual void computeResidual() override;
   virtual void computeJacobian() override;
+  virtual GenericReal<is_ad> getUNodal(unsigned int n) { return _u_nodal[n]; }
```
loganharbour:
```suggestion
  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) { return _u_nodal[n]; }
```

AD containers are big - we definitely don't want to copy them

loganharbour (dismissed):

tophmatthews:
Ready for re-review

tophmatthews:
wait...I don't want to return a ref, since downstream I want something like this,
```
  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) override
  {
    return std::exp(this->_u_nodal[n]);
  }
  ```
and this leads to `warning: returning reference to local temporary object [-Wreturn-stack-address]`

**Comment by tophmatthews** [comment]
Okay, if I turn the `_upwinding` into a bool instead of an enum, which is valid as far as I can tell since it's either `FULL` or `NONE`, then I can now override `computeResidual` and `computeJacobian` to fill a container that holds the exponent of `_u_nodal`:
```
template <bool is_ad>
void
LogConservativeAdvectionTempl<is_ad>::computeJacobian()
{
  if (this->_upwinding)
    for (const auto n : make_range(_exp_u_nodal.size()))
      _exp_u_nodal[n] = protectedExp(this->_u_nodal[n]);
  ConservativeAdvectionTempl<is_ad>::computeJacobian();
}
```
which can then be called via `getUNodal`, e.g.,
```
  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) override
  {
    return _exp_u_nodal[n];
  }
```

**Comment by tophmatthews** [comment]
Ready for re-review @loganharbour @lindsayad

**Comment by lindsayad** [comment]
> Okay, if I turn the `_upwinding` into a bool instead of an enum

You should still be able to make comparisons with an enum. I prefer the enum because it allows the opportunity for expanding the list of enumerators in the future. Full upwinding is only first order accurate. There's also second order upwinding and many other types of upwind schemes one can construct

**Comment by tophmatthews** [comment]
oh I thought I couldn't inherit enums. I'll try again...

**Comment by tophmatthews** [comment]
indeed, I was thinking of something different. fixed it now. Thanks @lindsayad !

File: framework/include/kernels/ConservativeAdvection.h [hunk 2]
```diff
@@ -31,6 +31,7 @@ class ConservativeAdvectionTempl : public GenericKernel<is_ad>
   virtual Real computeQpOffDiagJacobian(unsigned int jvar) override;
   virtual void computeResidual() override;
   virtual void computeJacobian() override;
+  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) { return _u_nodal[n]; }
```
lindsayad:
```suggestion
  virtual const GenericReal<is_ad> & getUNodal(const std::size_t n) const { return _u_nodal[n]; }
```

lindsayad (commented):
Just the one suggestion

framework/include/kernels/ConservativeAdvection.h
tophmatthews: oh whoops!

lindsayad (approved):

tophmatthews:
This is ready to go in right? What else can I do to help?

PR #31525, opened 2025-09-15
Former MFEM quadrature name bug

Heinrich-BR:
Fixes #31524

PR #31635, opened 2025-09-30
Switch from double to real_t and improve support for different FE bases

nmnobre:
## Reason
We should be able to fully support all of MFEM's floating point formats.
The user should be able to successfully select the basis type for all four supported FE collections.

## Design
Code refactoring, including extra careful parameter checks and processing.

## Impact
Better support for different floating point precisions and FE collections with non-default basis types.
Though this will affect #31525 (as it'll conflict), it's completely orthogonal and separate.
Indeed, I've explicitly commented the code to explicitly mark what needs removing in #31525.

### DIFF
--- a/framework/src/mfem/fespaces/MFEMScalarFESpace.C
+++ b/framework/src/mfem/fespaces/MFEMScalarFESpace.C
@@ -18,38 +18,35 @@ MFEMScalarFESpace::validParams()
 {
   InputParameters params = MFEMSimplifiedFESpace::validParams();
   params.addClassDescription("Convenience class to construct scalar finite element spaces.");
-  MooseEnum fec_types("H1 L2", "H1", true);
+  MooseEnum fec_types("H1 L2", "H1");
   params.addParam<MooseEnum>("fec_type", fec_types, "Specifies the family of FE shape functions.");
-  params.addParam<std::string>(
-      "basis", "GaussLobatto", "Specifies the quadrature basis used for scalar elements.");
+  MooseEnum basis_types("GaussLegendre GaussLobatto Positive OpenUniform ClosedUniform "
+                        "OpenHalfUniform Serendipity ClosedGL IntegratedGLL",
+                        "GaussLobatto");
+  params.addParam<MooseEnum>(
+      "basis",
+      basis_types,
+      "Specifies the basis used for scalar elements. H1 spaces require a closed basis "
+      "(GaussLobatto Positive ClosedUniform Serendipity ClosedGL)");
 
   return params;
 }
 
 MFEMScalarFESpace::MFEMScalarFESpace(const InputParameters & parameters)
-  : MFEMSimplifiedFESpace(parameters), _fec_type(parameters.get<MooseEnum>("fec_type"))
+  : MFEMSimplifiedFESpace(parameters), _fec_type(getParam<MooseEnum>("fec_type"))
 {
 }
 
 std::string
 MFEMScalarFESpace::getFECName() const
 {
-  const char b = mfem::BasisType::GetChar(getBasis(getParam<std::string>("basis")));
-  std::string basis(1, b);
+  std::string basis =
+      _fec_type == "L2"
+          ? "_T" + std::to_string(getParam<MooseEnum>("basis"))
+          : "@" + std::string({mfem::BasisType::GetChar(getParam<MooseEnum>("basis"))});
 
-  // This is to get around an MFEM bug where if you pass the full name of the default element type,
-  // it crashes
-  if (_fec_type == "H1")
-    basis = (basis == "G" ? "" : "@" + basis);
-  else if (_fec_type == "L2")
-  {
-    if (basis != "g")
-      mooseInfo("L2 finite element space only supports GaussLegendre basis. Ignoring " +
-                getParam<std::string>("basis") +
-                " basis choice and using GaussLegendre instead.\n");
-
-    basis = "";
-  }
+  // This is to get around an MFEM bug (to be removed in #31525)
+  basis = (basis == "@G" || basis == "_T0") ? "" : basis;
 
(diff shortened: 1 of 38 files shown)


framework/src/mfem/fespaces/MFEMScalarFESpace.C
```diff
@@ -18,38 +18,35 @@ MFEMScalarFESpace::validParams()
...
-  params.addParam<std::string>(
-      "basis", "GaussLobatto", "Specifies the quadrature basis used for scalar elements.");
+  MooseEnum basis_types("GaussLegendre GaussLobatto Positive OpenUniform ClosedUniform "
+                        "OpenHalfUniform Serendipity ClosedGL IntegratedGLL",
+                        "GaussLobatto");
+  params.addParam<MooseEnum>(
+      "basis",
+      basis_types,
+      "Specifies the basis used for scalar elements. H1 spaces require a closed basis "
+      "(GaussLobatto Positive ClosedUniform Serendipity ClosedGL)");
```
lindsayad, line 30: Do we need an error check for this or is MFEM going to do that for us?

framework/src/mfem/fespaces/MFEMVectorFESpace.C
```diff
@@ -20,14 +20,24 @@ MFEMVectorFESpace::validParams()
...
+  MooseEnum closed_basis_types("GaussLobatto=1 ClosedUniform=4 Serendipity=6 ClosedGL=7",
+                               "GaussLobatto");
+  params.addParam<MooseEnum>("closed_basis",
+                             closed_basis_types,
+                             "Specifies the closed basis used for ND and RT elements.");
+  MooseEnum basis_types("GaussLegendre GaussLobatto Positive OpenUniform ClosedUniform "
+                        "OpenHalfUniform Serendipity ClosedGL IntegratedGLL",
+                        "GaussLegendre");
+  params.addParam<MooseEnum>(
+      "open_basis", basis_types, "Specifies the open basis used for ND and RT elements.");
```
lindsayad, line 34: The parameter name seems misleading if we can also pass closed basis types

framework/src/mfem/fespaces/MFEMScalarFESpace.C
nmnobre: There's this, yes:
```
         MFEM_VERIFY(Quadrature1D::CheckClosed(pt_type) !=
                     Quadrature1D::Invalid,
                     "unsupported BasisType: " << BasisType::Name(btype));
```

I thought about it I'll be honest - I really like that `MooseEnum` input parameters can automatically reject out-of-bounds values and that's part of the reason for this whole PR, in particular for the vector spaces (though MFEM will do similar verifications in all cases anyway) - but I can trap the error in our constructor if we think that'd be best?

framework/src/mfem/fespaces/MFEMVectorFESpace.C
nmnobre: I hear you, but changing that would conflict directly with MFEM's naming scheme (and the distinction with the `basis` parameter allows us to choose another default for that case). Here's the MFEM code that allows this:
```
int Quadrature1D::CheckOpen(int type)
{
   switch (type)
   {
      case GaussLegendre:
      case GaussLobatto:
      case OpenUniform:
      case ClosedUniform:
      case OpenHalfUniform:
      case ClosedGL:
         return type; // all types can work as open
      default:
         return Invalid;
   }
}
```

An open basis, _can_ be open, that's the key. Whereas a closed basis _must_ be closed (to establish continuity across elements).

framework/src/mfem/fespaces/MFEMScalarFESpace.C
lindsayad: It's ok with me if there's an error check in MFEM. It removes a maintenance burden for us if we can rely on the upstream library to do error checking for state validity of its objects. Attempting to keep up with enum options can be a pain and brittle. For instance I believe we still maintain a database of PETSc options although I've tried to gradually remove that ... there's no chance of us keeping up with the ever changing landscape of PETSc options and so error checking a user based on what is our likely out-of-date database of PETSc options is brittle and can just be straight-up wrong.

Things are a little different here since the number of enumerators is much smaller and will change at a slow rate. The value of giving users clear parameter documention through use of the enum I think outweighs the maintanance cost.

TLDR
1) I like enums for user parameter documentation 
2) No ned to add a duplicate error check in our constructor if we are getting it in MFEM

framework/src/mfem/fespaces/MFEMVectorFESpace.C
lindsayad: I wish that they called it something like `interpolation_basis` and `quadrature_basis` instead

framework/src/mfem/fespaces/MFEMScalarFESpace.C
nmnobre: 💯

framework/src/mfem/fespaces/MFEMVectorFESpace.C
lindsayad: That last comment assumes I am understanding the MFEM notation correctly. If I am could we change the naming of our MOOSE parameters to what I suggested?

nmnobre: Errrm... I'm afraid I'm not a fan of those names. Except for `Positive` and `IntegratedGLL`, all bases are interpolatory (nodal), some at quadrature points, some at uniform points along [0,1]. ND and RT bases are products of closed and open bases. No matter your FE basis choice though, integration will still happen by default at Gauss-Legendre quadrature points.

lindsayad (approved):

lindsayad:
failures are unrelated

lindsayad:
@nmnobre merge when you're ready (presuming other test targets pass)

PR #31644, opened 2025-09-30
Read general matrices

roystgnr:
## Reason
The only current option for MOOSE constraint operator matrix input is "ASCII Matlab script" (inefficient at large sizes).  This adds support for PETSc binaries (of the same PetscIndex and PetscScalar types), which is more efficient at the expense of legibility and portability, and HDF5 CSR matrix storage, which is more efficient while retaining the portability and most of the legibility.

## Design
Call `read()` instead of `read_matlab()`

## Impact
If anyone is relying on Matlab-script matrices with a non `.m` file extension, they'll have to rename those files.  Otherwise, no adverse impact, just a new feature.

Refs #18768


GiudGiud (approved):

loganharbour:
Broke sweeps: https://civet.inl.gov/event/258863/

PR #31646, opened 2025-09-30
Test Results Summary

### LINKED ISSUE #31645: Add Summary of Test Results Compared to Base
### Motivation

When adding or removing tests, there is currently no summary generated that compares the current test results to a baseline. This makes it difficult for developers and reviewers to quickly understand what has changed in a test run. A structured comparison summary would improve visibility into test changes and reduce manual effort during code reviews

### Design

Generate and display a structured summary of test result differences compared to a baseline. 
Removed tests: Tests that existed in the baseline but are no longer present.
Added tests: New tests introduced in the current run, along with their runtime.
Add unit tests to validate

### Impact

Save reviewers time by summarizing test changes at a glance.
Reduce the chance of missing important test additions or removals.

maythuwin1313:
Closes #31645

### DIFF
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+import argparse
+from typing import Tuple
+
+from TestHarness.resultsreader.reader import TestHarnessResultsReader
+from TestHarness.resultsreader.results import TestHarnessResults, TestName
+from tabulate import tabulate
+
+class TestHarnessResultsSummary:
+    def __init__(self, database: str):
+        self.reader = TestHarnessResultsReader(database)
+
+    @staticmethod
+    def parseArgs() -> argparse.Namespace:
+        parser = argparse.ArgumentParser(description='Produces a summary from test harness results')
+        parent = argparse.ArgumentParser(add_help=False)
+
+        parser.add_argument(
+            'database',
+            type=str,
+            help='The name of the database')
+        
+        action_parser = parser.add_subparsers(dest='action', help='Action to perform')
+        action_parser.required = True
+
+        pr_parser = action_parser.add_parser(
+            'pr',
+            parents=[parent],
+            help='Performs a summary for a pull request'
+        )
+        
+        pr_parser.add_argument(
+            'event_id',
+            type=int,
+            help='The event ID')
+        
+        return parser.parse_args()
+  
+    @staticmethod
+    def diff_table(results: TestHarnessResults, base_names: set[TestName],
+                   head_names: set[TestName]) -> Tuple[list, list]:
+        removed_names = base_names - head_names
+        if removed_names:
+            removed_table = list(removed_names)
+        else:
+            removed_table = None
+
+        add_names = head_names - base_names
+        if add_names:
+            added_table = []
+            for test_name in add_names:
+                test_result = results.get_test(test_name.folder, test_name.name)
+                added_table.append([str(test_name), test_result.run_time])
+        else:
+            added_table = None
+        
+        return removed_table, added_table
+    
+    def pr_test_names(self, **kwargs):
+        event_id = kwargs['event_id']
+        assert isinstance(event_id, int)
+
+        results = self.reader.getEventResults(kwargs['event_id'])
+        test_names = set(results.test_names)
+        
+        base_sha = results.base_sha
+        base_results = self.reader.getCommitResults(base_sha)
+        base_test_names = set(base_results.test_names)
+
+        return results, test_names, base_test_names
+    
+
+    def build_summary(self, removed_table: list, added_table: list) -> str:
+        summary = []
+
+        summary.append("\nRemoved Tests:")
+        if removed_table:
+            table = [[str(test_name)] for test_name in removed_table]
+            summary.append(tabulate(table, headers=["Test Name"], tablefmt="github"))
+        else:
+            summary.append("No Removed Tests")
+
+        summary.append("\nNew Tests:")
+        if added_table:
+            summary.append(tabulate(added_table, headers=["Test Name", "Run Time"], tablefmt="github"))
+        else:
+            summary.append("No New Tests")
+
+        return "\n".join(summary)
+
+
+    def pr(self, **kwargs) -> str:
+        results, head_names, base_names = self.pr_test_names(**kwargs)
+        removed_table, added_table = self.diff_table(results, base_names, head_names)
+        print(self.build_summary(removed_table, added_table))
+
+    def main(self, **kwargs):
+        action = kwargs['action']
+        getattr(self, action)(**kwargs)
+        
+if __name__ == '__main__':
+    args = TestHarnessResultsSummary.parseArgs()
+    TestHarnessResultsSummary(args.database).main(**vars(args))
+
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -92,7 +92,6 @@ class TestHarnessResultsSummary:
     def pr(self, **kwargs) -> str:
         results, head_names, base_names = self.pr_test_names(**kwargs)
         if not base_names:
-            print("Skipping Test Summary Results")
             return
         removed_table, added_table = self.diff_table(results, base_names, head_names)
         print(self.build_summary(removed_table, added_table))
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -198,15 +198,16 @@ class TestHarnessResultsSummary:
         assert isinstance(run_time_rate_floor, (float, int))
         #Extract removed tests
         removed_names = base_names - head_names
-        removed_table = [[str(testname)] for testname in removed_names] if removed_names else None
-
+        removed_table = [['`'+str(test_name)+'`'] for test_name in removed_names] if removed_names else None
+        #removed_table = [str(test_name) for test_name in removed_names] if removed_names else None
         #Extract new tests
         add_names = head_names - base_names
         if add_names:
             added_table = []
             for test_name in add_names:
                 test_result = head_results.get_test(test_name.folder, test_name.name)
-                added_table.append([str(test_name), test_result.run_time])
+                formatted_test_name = '`'+str(test_name)+'`'
+                added_table.append([formatted_test_name, test_result.run_time])
         else:
             added_table = None
         #Extract same tests
@@ -226,9 +227,10 @@ class TestHarnessResultsSummary:
                     relative_runtime = abs(head_result.run_time - base_result.run_time) / base_result.run_time
                     #Check if relative run time rate is higher than threadshold, then it will put in the result
                     if relative_runtime >= run_time_rate_floor:
+                        formatted_test_name = '`'+str(test_name)+'`'
                         same_table.append(
                             [
-                                str(test_name),
+                                formatted_test_name,
                                 base_result.run_time,
                                 head_result.run_time,
                                 f'{relative_runtime:.2%}'
@@ -247,6 +249,7 @@ class TestHarnessResultsSummary:
         assert isinstance(removed_table,(list, NoneType))
         format_removed_table = ["### Removed Tests:"]
         if removed_table:
+            assert all(isinstance(removed_test, list) for removed_test in removed_table)
             format_removed_table.append(
                 tabulate(
                     removed_table,
@@ -265,6 +268,7 @@ class TestHarnessResultsSummary:
         assert isinstance(added_table,(list, NoneType))
         format_added_table = ["### New Tests:"]
         if added_table:
+            assert all(isinstance(added_test, list) for added_test in added_table)
             format_added_table.append(
                 tabulate(
                     added_table,
@@ -286,6 +290,7 @@ class TestHarnessResultsSummary:
         assert isinstance(same_table,(list, NoneType))
         format_same_table = ["### Same Tests that exceed relative run time rate:"]
         if same_table:
+            assert all(isinstance(same_test, list) for same_test in same_table)
             format_same_table.append(
                 tabulate(
                     same_table,
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -19,7 +19,7 @@ class TestHarnessResultsSummary:
             'database',
             type=str,
             help='The name of the database')
-        
+
         action_parser = parser.add_subparsers(dest='action', help='Action to perform')
         action_parser.required = True
 
@@ -28,14 +28,14 @@ class TestHarnessResultsSummary:
             parents=[parent],
             help='Performs a summary for a pull request'
         )
-        
+
         pr_parser.add_argument(
             'event_id',
             type=int,
             help='The event ID')
-        
+
         return parser.parse_args()
-  
+
     @staticmethod
     def diff_table(results: TestHarnessResults, base_names: set[TestName],
                    head_names: set[TestName]) -> Tuple[list, list]:
@@ -54,7 +54,7 @@ class TestHarnessResultsSummary:
         else:
             added_table = None
         return removed_table, added_table
-    
+
     def pr_test_names(self, **kwargs):
         event_id = kwargs['event_id']
         assert isinstance(event_id, int)
@@ -65,9 +65,9 @@ class TestHarnessResultsSummary:
         test_names = set(results.test_names)
         base_sha = results.base_sha
         assert isinstance(base_sha, str)
-        
+
         base_results = self.reader.getCommitResults(base_sha)
-        if not isinstance(base_results, TestHarnessResults):         
+        if not isinstance(base_results, TestHarnessResults):
             print(f"\nComparison not available: no baseline results found for base SHA {base_sha}")
             return results, test_names, None
         base_test_names = set(base_results.test_names)
@@ -93,7 +93,7 @@ class TestHarnessResultsSummary:
         results, head_names, base_names = self.pr_test_names(**kwargs)
         if not base_names:
             print("Skipping Test Summary Results")
-            return 
+            return
         removed_table, added_table = self.diff_table(results, base_names, head_names)
         print(self.build_summary(removed_table, added_table))
 
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -177,7 +177,7 @@ class TestHarnessResultsSummary:
 
     def _build_added_table(self, add_names: set[TestName], head_results: StoredResult) -> Optional[List[List]]:
         """
-        Build a table of newly added test names, optionally including runtime.
+        Build a table of added test names, optionally including runtime.
 
         Parameters
         ----------
@@ -309,7 +309,7 @@ class TestHarnessResultsSummary:
         removed_table : Optional[List[List]]
             A list of removed test name, each containing a formatted test name.
         added_table : Optional[List[List]]
-            A sorted list of newly added test names and their runtime.
+            A sorted list of added test names and their runtime.
             - the formatted test name (str)
             - the runtime as a string formatted to two decimal places, or "None" if not available.
         same_table : Optional[List[List]]
@@ -367,7 +367,7 @@ class TestHarnessResultsSummary:
         Parameters
         ----------
         title : str
-            The section title to display above the table (e.g., "### Removed Tests").
+            The section title to display above the table (e.g., "### Removed tests").
         table_data : list of list or None
             The table content, where each inner list represents a row.
         headers : list of str
@@ -395,14 +395,14 @@ class TestHarnessResultsSummary:
 
     def build_summary(self, removed_table: list, added_table: list, same_table: list) -> str:
         """
-        Build a summary report of removed, newly added, and same tests with optional runtime results.
+        Build a summary report of removed, added, and same tests with optional runtime results.
 
         Parameters
         ----------
         removed_table : list of list or None
             A list of removed test name, each containing a formatted test name.
         added_table : list or None
-            A sorted list of newly added test names and their optional runtime.
+            A sorted list of added test names and their optional runtime.
             - the formatted test name (str)
             - the runtime as a string formatted to two decimal places, or "None" if not available.
         same_table : list of list or None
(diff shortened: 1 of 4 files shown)


python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
+import argparse
+from typing import Tuple
+
+from TestHarness.resultsreader.reader import TestHarnessResultsReader
+from TestHarness.resultsreader.results import TestHarnessResults, TestName
+from tabulate import tabulate
+
+class TestHarnessResultsSummary:
```
loganharbour, line 21: Before I review fully, let's add some doc strings to each function that describe what the function does

loganharbour (changes requested):

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,107 @@
...
+            'event_id',
+            type=int,
+            help='The event ID')
+
+        return parser.parse_args()
+
+    @staticmethod
+    def diff_table(results: TestHarnessResults, base_names: set[TestName],
+                   head_names: set[TestName]) -> Tuple[list, list]:
+        removed_names = base_names - head_names
```
loganharbour, line 323: Assert at the top that the arguments are the expected types

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,107 @@
...
+            summary.append(tabulate(added_table, headers=["Test Name", "Run Time"], tablefmt="github"))
+        else:
+            summary.append("No New Tests")
+        return "\n".join(summary)
+
+    def pr(self, **kwargs) -> str:
+        results, head_names, base_names = self.pr_test_names(**kwargs)
+        if not base_names:
+            print("Skipping Test Summary Results")
+            return
```
loganharbour: Do we need to print this here? Isn't there already a print in pr_test_names?

python/TestHarness/tests/test_resultssummary.py
```diff
@@ -0,0 +1,148 @@
...
+
+EVENT_ID = os.environ.get('CIVET_EVENT_ID')
+if EVENT_ID is not None:
+    EVENT_ID = int(EVENT_ID)
+
+class TestResultsSummary(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @unittest.skipUnless(HAS_AUTH, "Skipping because authentication is not available")
```
loganharbour: Similarly, we should have a simple doc string for each test method that describes what the test is testing

loganharbour (changes requested):

python/TestHarness/resultssummary/summary.py
maythuwin1313: Oh yes, I will remove this print.

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,196 @@
...
+        pr_parser.add_argument(
+            'event_id',
+            type=int,
+            help='The event ID')
+
+        return parser.parse_args()
+
+    @staticmethod
+    def diff_table(results: TestHarnessResults, base_names: set[TestName],
+                   head_names: set[TestName]) -> Tuple[list, list]:
```
loganharbour:
```suggestion
                   head_names: set[TestName]) -> Tuple[Optional[list], Optional[list]]:
```

you might need to import `Optional` typing

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        Compare test names between the base and current head, and return
+        the difference
+
+        Parameters
+        ----------
+        results : TestHarnessResults
+            An object that provides access to test results.
```
loganharbour:
```suggestion
            An object that provides access to test results for the head
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,196 @@
...
+        Returns
+        -------
+        summary : str
+            A formatted string summarizing removed and new tests using GitHub-style format
+        """
+        assert isinstance(removed_table,(list,NoneType))
+        assert isinstance(added_table,(list,NoneType))
+
+        summary = []
+        summary.append("\nRemoved Tests:")
```
loganharbour: We should add these to include the header markdown syntax, like:

```
### Removed tests
```

which will display as:

### Removed tests

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,196 @@
...
+            summary.append(tabulate(added_table, headers=["Test Name", "Run Time"], tablefmt="github"))
+        else:
+            summary.append("No New Tests")
+        return "\n".join(summary)
+
+    def pr(self, **kwargs) -> str:
+        results, head_names, base_names = self.pr_test_names(**kwargs)
+        assert isinstance(head_names,set)
+        assert isinstance(base_names,(set,NoneType))
+        if not base_names:
```
loganharbour: Shouldn't this be `base_names is None`?

`if not base_names`

will be true if `base_names` is none _or_ empty

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python3
```
loganharbour, line 1: Add the python header that we use in all moose python files

loganharbour (changes requested):

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+        - action : str
+            The action to perform (e.g., 'pr').
+        - event_id : int
+            The event ID (required for the 'pr' action).
+        --run-time-floor : float
+            Set minimum threshold for the head test run time
+            (default: 1 s)
+        --run-time-rate-floor : float
+            Set minimum relative run time ratio between base and head
+            (default: 0.5 i.e 50%)
```
loganharbour: I don't think we need the description here

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+        )
+
+        return parser.parse_args()
+
+    def get_commit_results(self, commit: str) -> Optional[StoredResult]:
+        return self.reader.getCommitResults(commit)
+
+    def get_event_results(self, event_id: int) -> Optional[StoredResult]:
+        return self.reader.getEventResults(event_id)
+
```
loganharbour, line 96: Let's add a docstring here. We don't need to describe the parameters, but something like

```
The results associated with a commit.

This is a separate function so that it can be
mocked in unit tests
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+    def pr_test_names(self, **kwargs):
```
loganharbour: Let's just define `event_id` and `out` as actual arguments.

Also, you can declare the return type of this function with something like
```
def foo() -> Tuple[StoredResult, StoredResult, ...]:
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+        assert isinstance(same_table,(list,NoneType))
+
+        summary = []
+        summary.append(self._format_removed_table(removed_table))
+        summary.append(self._format_added_table(added_table))
+        summary.append(self._format_same_table(same_table))
+
+        return "\n".join(summary)
+
+    def summary_output_file(self, output_result: str, out: str) -> None:
```
loganharbour:
```suggestion
    def write_output(self, output_result: str, out: str) -> None:
```

Let's just rename this function so it's a little clearer as to what it does. And rename the `out` argument to `out_file` because right now it's hard to distinguish.

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+            self.summary_output_file(error_msg, out)
+            raise SystemExit(error_msg)
+
+        test_names = set(head_results.test_names)
+        base_sha = head_results.base_sha
+        assert isinstance(base_sha, str)
+
+        base_results = self.get_commit_results(base_sha)
+        if not isinstance(base_results, StoredResult):
+            no_base =f"Comparison not available: no baseline results found for base SHA {base_sha}"
```
loganharbour:
```suggestion
            no_base = f"Base results not available for {base_sha}"
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+            The set of test names from the current head commit.
+
+        Optional Parameters
+        -------------------
+        --run-time-floor : float
+            The runtime at which to not check for a difference
+            (default: 1 s)
+        --run-time-rate-floor : float
+            The runtime rate at which to not attach in same_table summary
+            (default: 0.5 i.e 50%)
```
loganharbour, line 287:
```suggestion
        Optional Parameters
        -------------------
        run_time_floor : float
            The runtime at which to not check for a difference
            (default: 1 s)
        run_time_rate_floor : float
            The runtime rate at which to not attach in same_table summary
            (default: 0.5 i.e 50%)
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+        run_time_rate_floor = kwargs.pop('run_time_rate_floor', 0.5)
+        assert isinstance(run_time_rate_floor, (float, int))
+
+        removed_names = base_names - head_names
+        if removed_names:
+            removed_table = []
+            for test_name in removed_names:
+                removed_table.append(str(test_name))
+        else:
+            removed_table = None
```
loganharbour:
```suggestion
        removed_table = [str(v) for v in removed_names] if removed_table else None
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+            same_table = []
+            for test_name in same_names:
+                base_result = base_results.get_test(test_name.folder, test_name.name)
+                head_result = head_results.get_test(test_name.folder, test_name.name)
+                if (head_result.run_time is None or
+                    base_result.run_time is None or
+                    head_result.run_time < head_run_time_floor):
+                    continue
+                else:
+                    relative_runtime = abs(head_result.run_time - base_result.run_time)/ base_result.run_time
```
loganharbour:
```suggestion
                    relative_runtime = abs(head_result.run_time - base_result.run_time) / base_result.run_time
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,395 @@
...
+        None
+            This method does not return anything. It performs a file write operation.
+        """
+        try:
+            with open(out, 'w') as f:
+                f.write(output_result)
+        except Exception as e:
+            print(f"Failed to write to {out}: {e}")
+
+    def pr(self, **kwargs) -> str:
```
loganharbour: Let's also explicitly make `event_id` and `out` parameters here, in addition to kwargs

loganharbour (changes requested):
I haven't looked at the unit tests yet. I can once these comments are addressed

maythuwin1313:
@loganharbour suggested changes have been updated.

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,396 @@
...
+    def get_event_results(self, event_id: int) -> Optional[StoredResult]:
+        """
+        The results associated with a event_id.
+
+        This is a separate function so that it can be
+        mocked in unit tests
+        """
+        return self.reader.getEventResults(event_id)
+
+    def pr_test_names(self, event_id:int, out:str) -> Tuple[Optional[StoredResult],Optional[StoredResult],
```
loganharbour:
```suggestion
    def pr_test_names(self, event_id: int, out: str) -> Tuple[Optional[StoredResult],Optional[StoredResult],
```

Fixes spacing. Let's also rename `out` to `out_file` for consistency (remember to change it in the docstring, too)

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,396 @@
...
+            self.write_output(error_msg, out)
+            raise SystemExit(error_msg)
+
+        test_names = set(head_results.test_names)
+        base_sha = head_results.base_sha
+        assert isinstance(base_sha, str)
+
+        base_results = self.get_commit_results(base_sha)
+        if not isinstance(base_results, StoredResult):
+            no_base =f"Base results not available for {base_sha}"
```
loganharbour:
```suggestion
            no_base = f"Base results not available for {base_sha}"
```

loganharbour (changes requested):

loganharbour (changes requested):
- We need to add command line option to disable the run time comparison, so that we can disable it if it isn't reliable or useful
- Output the runtimes with just two values after the decimal (i.e., 1.12). We don't care about printing all the extra numbers
- The runtime change should include a + or a - for if the runtime increased or decreased in the head
- The runtime changes should be sorted from increasing to decreasing order (from tests that got longer to tests that got shorter)

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+        pr_parser.add_argument(
+            'event_id',
+            type = int,
+            help = 'The event ID')
+
+        pr_parser.add_argument(
+            '--run-time-floor',
+            type = float,
+            default = 1,
+            help = 'Sets a minimum threshold for the head test run time (seconds)'
```
loganharbour:
```suggestion
            help = 'The minimum threshold for a test to checked for a run time difference'
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+        )
+
+        pr_parser.add_argument(
+            '--run-time-rate-floor',
+            type = float,
+            default = 0.5,
+            help = 'Sets a minimum relative run time ratio between base and head'
+        )
```
loganharbour, line 72:
```suggestion
        pr_parser.add_argument(
            '--run-time-ratio',
            type = float,
            default = 0.5,
            help = 'The ratio by which to report a test's difference in run time'
        )
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+        return self.reader.getEventResults(event_id)
+
+    def pr_test_names(self, event_id: int, out_file: str) -> Tuple[Optional[StoredResult], Optional[StoredResult],
+            Optional[set], Optional[set]]:
+        """
+        Retrieve test names and base test names for a pull request event.
+
+        This method fetches the test results for a given pull request event ID,
+        extracts the test names and base SHA, and retrieves the corresponding
+        base test names from the base.
```
loganharbour:
```suggestion
        test names from the base.
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+        no_run_time_comparison : bool
+            If True, skip runtime comparison and only include test names.
+
+        Returns
+        -------
+        list of list
+        A sorted list of lists where each sublist contains:
+        - the formatted test name (str)
+        - the runtime as a string formatted to two decimal places, or "None" if not available.
+        If `no_run_time_comparison` is True, only test names are included.
```
loganharbour:
```suggestion
        List[List]:
            A sorted list of lists where each sublist contains:
              - the formatted test name (str)
              - the runtime as a string formatted to two decimal places, or "None" if not available.
            If `no_run_time_comparison` is True, only test names are included.
```

I think this is the correct formatting, right?

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+        """
+        assert isinstance(add_names, set)
+        assert isinstance(head_results, StoredResult)
+        assert isinstance(no_run_time_comparison, bool)
+        added_table = []
+        if no_run_time_comparison:
```
loganharbour: I'm confused here. For "added" tests, there should be nothing to compare to, right? And we should always be able to at least show the run time of the added tests.

Which means for the added table - there should be no branches and we shouldn't care about whether or not comparison is enabled. We always have listing of new tests and then their times.

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+            Returns None if no tests meet the criteria.
+        """
+        assert isinstance(same_names, set)
+        assert isinstance(base_results, StoredResult)
+        assert isinstance(head_results, StoredResult)
+        assert isinstance(head_run_time_floor, (float, int))
+        assert isinstance(run_time_rate_floor, (float, int))
+
+        same_table = []
+        #disable run time comparison
```
loganharbour: Is this comment old?

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+        head_results : StoredResult
+            Object providing access to test results for the head.
+        head_run_time_floor : float
+            Minimum runtime in the head commit to consider for comparison.
+        run_time_rate_floor : float
+            Minimum relative runtime change (as a fraction) to include in the table.
+
+        Returns
+        -------
+        list of list or None
```
loganharbour:
```suggestion
        Optional[List[List]]
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,546 @@
...
+
+        removed_table, added_table, same_table = self.diff_table(
+            base_results,
+            head_results,
+            base_names,
+            head_names,
+            **kwargs
+        )
+        summary_result = self.build_summary(removed_table, added_table, same_table)
+        print(summary_result)
```
loganharbour: Is this print an old debug print that should be removed?

loganharbour (changes requested):
One other thing - for all comments there should be a space after the comment.

For example:

```
#This is a comment
```

should be:

```
# This is a comment
```

python/TestHarness/resultssummary/summary.py
maythuwin1313: yes, I will update it.

python/TestHarness/resultssummary/summary.py
maythuwin1313: oh noted. I thought when we disable runtime comparison option, we don't want to see new added run time as well. I will update to correct logic.

python/TestHarness/resultssummary/summary.py
maythuwin1313: It's for the final message to be able to print in console as well. 
Do we want to display only in output file or both console and output file?

maythuwin1313: Removed

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,539 @@
...
+                ["Test", "Head(s)"],
+                "None"
+            )
+        )
+        # Format same table
+        summary.append(
+            self._format_table(
+                "### Run time changes",
+                same_table,
+                ["Test", "Base(s)", "Head(s)", "+/-"],
```
loganharbour:
```suggestion
                ["Test", "Base (s)", "Head (s)", "+/-"],
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,539 @@
...
+                ["Test"],
+                "None"
+            )
+        )
+        # Format added table
+        summary.append(
+            self._format_table(
+                "### New Tests",
+                added_table,
+                ["Test", "Head(s)"],
```
loganharbour:
```suggestion
                ["Test", "Time (s)"],
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,539 @@
...
+        action = kwargs.pop('action')
+
+        if action == 'pr':
+            self.pr(
+                event_id = kwargs.pop('event_id'),
+                out_file = kwargs.pop('out_file'),
+                **kwargs
+            )
+        else:
+            getattr(self, action)(**kwargs)
```
loganharbour:
```suggestion
            raise NotImplementedError(f'Action {action} not implemented')
```

python/TestHarness/tests/tests
```diff
@@ -346,4 +346,10 @@
     requirement = 'The system shall be able to import python utilities when running tests.'
     issues = '#31457'
   []
+  [resultssummary]
+    type = PythonUnitTest
+    input = test_resultssummary.py
+    requirement = 'The system shall provide summary of test results compared to base for pr event'
```
loganharbour:
```suggestion
    requirement = 'The system shall provide a system that summaries the changes in tests within a pull request event`
```

python/TestHarness/tests/test_resultssummary.py
```diff
@@ -0,0 +1,746 @@
...
+        self.assertIn('Removed Tests', has_test_build_summary)
+        self.assertIn('New Tests', has_test_build_summary)
+        self.assertIn('Run time changes', has_test_build_summary)
+        self.assertIn('Test', has_test_build_summary)
+        self.assertIn('Base(s)', has_test_build_summary)
+        self.assertIn('Head(s)', has_test_build_summary)
+        self.assertIn('+/-', has_test_build_summary)
+        self.assertIn(str(MOCKED_TEST_NAME), has_test_build_summary)
+        self.assertIn(str(10), has_test_build_summary)
+        self.assertIn(str(17), has_test_build_summary)
```
loganharbour:
```suggestion
        self.assertIn('10.00', has_test_build_summary)
        self.assertIn('17.00', has_test_build_summary)
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,539 @@
...
+                "### Removed Tests",
+                removed_table,
+                ["Test"],
+                "None"
+            )
+        )
+        # Format added table
+        summary.append(
+            self._format_table(
+                "### New Tests",
```
loganharbour:
```suggestion
                "### Added tests",
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -0,0 +1,539 @@
...
+        """
+        assert isinstance(removed_table, (list, NoneType))
+        assert isinstance(added_table, (list, NoneType))
+        assert isinstance(same_table, (list, NoneType))
+
+        summary = []
+        # Format removed table
+        summary.append(
+            self._format_table(
+                "### Removed Tests",
```
loganharbour:
```suggestion
                "### Removed tests",
```

loganharbour (changes requested):
Tests look great! This is just about done.

Tomorrow - we should have a single test that tries to read a PR (from the database, if available) and otherwise this is good.

loganharbour (approved):

PR #31647, opened 2025-10-01
Our new test is erroring on more than 2 ranks

roystgnr:
This looks like a libMesh-level bug to me; let's cap at 2 procs until I can get a fix downstream.

Refs #18768

This should unbreak next.


loganharbour (approved):

roystgnr:
Oh, I should also ref #31644, where I introduced the breakage.

loganharbour:
Merging despite test completion; sweeps are passing

PR #31652, opened 2025-10-01
fix UMAT energy quantities

### LINKED ISSUE #31651: UMAT energy quantities should be treated like state variables
### Bug Description

Currently UMAT energy quantities (elastic strain energy, plastic dissipation, and creep dissipation) are treated as material property references and the value that comes out of the UMAT call gets passed back in for the next call. The Abaqus user manual states that these values "should be updated to the corresponding energy values at the end of the increment". Effectively they are to be treated as state variables since the value that goes in should the *old* value and get updated at the end of the increment. 

Typically the job of these quantities is to *accumulate* energy over the step, so they are added to each time the UMAT is called. If the same value is passed in after each iteration (as is currently done now), the quantities accumulate to unphysical values.

### Steps to Reproduce

Check the values after a few timesteps, they will be orders of magnitude larger than expected (verifed with Abaqus).

### Impact

Only impacts analysis, does not impact solution.

### [Optional] Diagnostics

_No response_

jessecarterMOOSE:
Treats the UMAT energy quantities effectively like state variables per the Abaqus manual. 

This has been verified to match Abaqus output. Note the values that Abaqus reports are *element-integrated*, where the integration is done over the *displaced* element, including the out-of-plane log strain. So to match Abaqus, the element-integrated value is needed (or the element-averaged value multiplied by the element volume), and in the case of generalized plane strain, multiplied by $exp\left(\epsilon_{33}\right)$ (exponential of the out-of-plane strain value).

closes #31651

@dschwen @GiudGiud

### DIFF
--- a/modules/solid_mechanics/src/materials/abaqus/AbaqusUMATStress.C
+++ b/modules/solid_mechanics/src/materials/abaqus/AbaqusUMATStress.C
@@ -80,8 +80,11 @@ AbaqusUMATStress::AbaqusUMATStress(const InputParameters & parameters)
     _state_var(declareProperty<std::vector<Real>>(_base_name + "state_var")),
     _state_var_old(getMaterialPropertyOld<std::vector<Real>>(_base_name + "state_var")),
     _elastic_strain_energy(declareProperty<Real>(_base_name + "elastic_strain_energy")),
+    _elastic_strain_energy_old(getMaterialPropertyOld<Real>(_base_name + "elastic_strain_energy")),
     _plastic_dissipation(declareProperty<Real>(_base_name + "plastic_dissipation")),
+    _plastic_dissipation_old(getMaterialPropertyOld<Real>(_base_name + "plastic_dissipation")),
     _creep_dissipation(declareProperty<Real>(_base_name + "creep_dissipation")),
+    _creep_dissipation_old(getMaterialPropertyOld<Real>(_base_name + "creep_dissipation")),
     _material_timestep(declareProperty<Real>(_base_name + "material_timestep_limit")),
     _rotation_increment(
         getOptionalMaterialProperty<RankTwoTensor>(_base_name + "rotation_increment")),
@@ -171,6 +174,11 @@ AbaqusUMATStress::initQpStatefulProperties()
 
   // Initialize total rotation tensor
   _total_rotation[_qp] = _R;
+
+  // Initialize energy quantities
+  _elastic_strain_energy[_qp] = 0.0;
+  _plastic_dissipation[_qp] = 0.0;
+  _creep_dissipation[_qp] = 0.0;
 }
 
 void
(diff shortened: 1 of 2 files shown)


modules/solid_mechanics/src/materials/abaqus/AbaqusUMATStress.C
```diff
@@ -269,6 +277,11 @@ AbaqusUMATStress::computeQpStress()
   for (const auto i : make_range(_aqNSTATV))
     _aqSTATEV[i] = _state_var_old[_qp][i];
 
+  // Recvoer "old" energy quantities
+  _aqSSE = _elastic_strain_energy_old[_qp];
+  _aqSPD = _plastic_dissipation_old[_qp];
+  _aqSCD = _creep_dissipation_old[_qp];
```
dschwen:
```suggestion
  _elastic_strain_energy[_qp] = _elastic_strain_energy_old[_qp];
  _plastic_dissipation[_qp] = _plastic_dissipation_old[_qp];
  _creep_dissipation[_qp] = _creep_dissipation_old[_qp];
```

modules/solid_mechanics/src/materials/abaqus/AbaqusUMATStress.C
```diff
@@ -338,9 +351,9 @@ AbaqusUMATStress::computeQpStress()
   _umat(_aqSTRESS.data(),
         _aqSTATEV.data(),
         _aqDDSDDE.data(),
-        &_elastic_strain_energy[_qp],
-        &_plastic_dissipation[_qp],
-        &_creep_dissipation[_qp],
+        &_aqSSE,
+        &_aqSPD,
+        &_aqSCD,
```
dschwen: revert this change

modules/solid_mechanics/src/materials/abaqus/AbaqusUMATStress.C
```diff
@@ -377,6 +390,11 @@ AbaqusUMATStress::computeQpStress()
   for (int i = 0; i < _aqNSTATV; ++i)
     _state_var[_qp][i] = _aqSTATEV[i];
 
+  // Update energy quantities
+  _elastic_strain_energy[_qp] = _aqSSE;
+  _plastic_dissipation[_qp] = _aqSPD;
+  _creep_dissipation[_qp] = _aqSCD;
+
```
dschwen:
```suggestion
```

dschwen (changes requested):
See comments and remove the   `_aqSSE`, `_aqSPD`, `_aqSCD` members. Should make it a bit simpler.

modules/solid_mechanics/src/materials/abaqus/AbaqusUMATStress.C
jessecarterMOOSE: If I did this then I would need to initialize the old properties to zero. Not sure how to do that because the references are const @dschwen

jessecarterMOOSE: unless material properties are initialized to zero automatically.......

dschwen (approved):

Issue #7659, opened 2016-08-31
Verify that a function such as "x" doesn't conflict with a named function "x"


permcody:
### Description of the enhancement or error report

It's possible to define a function or a function name such as "x" that could conflict. Right now this is allowed in MOOSE but perhaps it should at least be an warning if not an error.
### Rationale for the enhancement or information for reproducing the error

Difficult to validate input
### Identified impact

(i.e. Internal object changes, limited interface changes, public API change, or a list of specific applications impacted)

NEAMS Workbench

brianmoose:
I am not sure I understand what the problem is.
@permcody Do you have an example?

dschwen:
Right now we assume that 

`function = x`

Is a `[Function/x]` object. If it does not exist a ParsedFunction with the (valid) expression `x` is created on the fly.

dschwen:
The behavior is well defined as it is, but it may lead to confusing scenarios.

brianmoose:
OK, so where would the warning/error come into play since both of these are valid things to do? If you have a function name that is also the name of a variable (or other reserved word, like `pi`)?

If you did have `[Function/x]` then you couldn't have a ParsedFunction `function = x`. Is that ever a problem?

dschwen:
There would have to be a test that tries to parse names of `[Functions]` sub-blocks to see if they are valid parsed function expressions and issue a warning. 

The problem would come up if you have 

```
[Functions]
  [./v]
  [../]
  [./w]
  [../]
  [./x]
  [../]
[]

..

[ICs]
  [./blah]
    type = FunctionIC
    variable = ramp
    function = x
  [../]
[]
```

And you either

* accidentally delete the `x` function
* have copied the IC from some other file and expect it to give you a linear slope in x direction but forgot that there already was an `x` Function in the input

lindsayad:
I contend that the warning added in the action doesn't really address the title of this issue. There is no check for conflict, just a check that there is the possibility of a conflict

PR #31653, opened 2025-10-01
Kokkos Function System

NamjaeChoi:
Refs #30655.

Functors are not dispatched but plugged inside the user's parallel code, so I can't think of a way to achieve static polymorphism here. I think I will have to introduce virtual functions minimally in the wrapper class, but any idea to avoid it would be welcome, if you have any. Of course, if the user specifies the exact object type instead of using the abstract type, they will be able to evaluate functors statically.

Concerns are performance and SYCL, but regarding SYCL, Intel is adding an extension to allow virtual functions on GPU. We may be able to leverage it when the time comes to support SYCL. I don't think we are interested in supporting SYCL for backends other than Intel GPU.


NamjaeChoi:
@lindsayad My next thing to look into would be this. I will start from functions. Do you have any thoughts on this?

lindsayad:
If we use virtuals, should we consider just adapting the current functor system?

NamjaeChoi:
The problem with that is we can't construct MOOSE objects directly on device. If an object that has virtual methods is constructed on host, the vtable will be populated with host function pointers. What I'm trying to do here is to define a thin wrapper that has virtual methods and is light enough to be built on device, and let the wrapper call the wrapped functor methods statically.

framework/include/kokkos/base/KokkosFunctor.h
```diff
@@ -0,0 +1,46 @@
...
+  Functor(const Functor & functor);
+  virtual ~Functor();
+
+  KOKKOS_FUNCTION void hello() const { _wrapper_device->hello(); }
+
+private:
+  /**
+   * Pointer to the host functor wrapper
+   */
+  std::shared_ptr<FunctorWrapperHostBase> _wrapper_host;
```
lindsayad: what is this needed for?

framework/include/kokkos/base/KokkosFunctorWrapper.h
```diff
@@ -0,0 +1,139 @@
...
+  // Allocate device wrapper on device using placement new to populate vtable with device pointers
+  ::Kokkos::parallel_for(
+      1, KOKKOS_LAMBDA(const int) { new (wrapper_device) FunctorWrapperDevice<Object>(); });
+
+  // Allocate storage for functor on device
+  _functor_device =
+      static_cast<Object *>(::Kokkos::kokkos_malloc<ExecSpace::memory_space>(sizeof(Object)));
+
+  // Let device wrapper point to the copy
+  ::Kokkos::Impl::DeepCopy<MemSpace, ::Kokkos::HostSpace>(
```
lindsayad, line 144: Why are we using functions in `Impl`? Does `deep_copy` not work?

lindsayad (commented):
I think this looks ok to me

framework/include/kokkos/base/KokkosFunctorWrapper.h
NamjaeChoi: This directly corresponds to `cudaMemcpy` or its equivalent. `Kokkos::deep_copy` does more than just memcpy, like it copy-constructs a clone on the host first, checks whether the data type is trivially copyable, etc.

NamjaeChoi: And many times it complains for complicated class types like our objects.

NamjaeChoi:
Still working... just want to run test suite once

NamjaeChoi:
Need to add a few more tests, docstrings, and MOOSE docs, but should be good enough to start reviewing.

lindsayad:
going to unsubscribe for now. Mention me when you want more input or it's ready for review

NamjaeChoi:
@lindsayad I might have to hold this. Virtual functions on device seem to require RDC to be fully functional.

NamjaeChoi:
The code is currently in a state where CUDA will work when we have RDC.

NamjaeChoi:
I don't want to indefinitely hold this PR for RDC. I will temporarily disable using virtual dispatch for GPU.

NamjaeChoi:
@lindsayad You can review this. I temporarily blocked the code path that requires RDC with a runtime error as the ETA of it is unknown.

framework/doc/content/syntax/KokkosFunctions/index.md
```diff
@@ -0,0 +1,68 @@
...
+The signatures of the hook methods are defined as follows:
+
+```cpp
+KOKKOS_FUNCTION Real value(Real t, Real3 p) const;
+KOKKOS_FUNCTION Real3 vectorValue(Real t, Real3 p) const;
+KOKKOS_FUNCTION Real3 gradient(Real t, Real3 p) const;
+KOKKOS_FUNCTION Real3 curl(Real t, Real3 p) const;
+KOKKOS_FUNCTION Real div(Real t, Real3 p) const;
+KOKKOS_FUNCTION Real timeDerivative(Real t, Real3 p) const;
+KOKKOS_FUNCTION Real timeIntegral(Real t1, Real t2, Real3 p) const;
```
lindsayad, line 24: Might it be better to pass a reference here for `p`? If we were doing a lot of reads of `p` I could see a strong argument to put it in register instead of doing a global read, but at least the first class I looked at (tabulated) it looked like you just did a single read of `p`

framework/doc/content/syntax/KokkosFunctions/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+In fact, Kokkos-MOOSE already has the code path for acquiring functions as a type-agnostic abstract type and using virtual dispatch to evaluate actual functions.
+There is another non-template API `getKokkosFunction` which returns an abstract type `Moose::Kokkos::Function`.
+It is not the base class of your function but a wrapper class that contains your function and provides shims to call the function methods.
+
+However, that code path is currently blocked for GPU backends, and you will hit a runtime error when you attempt to use it.
+While our target GPU backends (CUDA, HIP, and Intel SYCL) support virtual functions on GPU, it requires the relocatable device code (RDC) option to be enabled during compilation, which the current MOOSE build system is lacking.
+Kokkos imposes a restriction to have a consistent RDC configuration across the whole software stack, but PETSc, from which MOOSE is currently acquring Kokkos, requires significant changes in its build system to enable RDC.
+Namely, the current code is functional under the assumption that the MOOSE software stack can be built with the RDC option, but it involves a major rework from our upstream dependency.
```
lindsayad:
```suggestion
The non-templated `getKokkosFunction` code path is hypothetically functional under the assumption that the MOOSE software stack can be built with the RDC option, but this hypothetical will only become a reality after a PETSc build system rework.
```

framework/doc/content/syntax/KokkosFunctions/index.md
```diff
@@ -0,0 +1,68 @@
...
+The PETSc and MOOSE team are looking into this issue, and it is expected to be resolved in a foreseeable future.
```
lindsayad:
```suggestion
The PETSc and MOOSE team are looking into this issue, and it is expected to be resolved in the foreseeable future.
```

framework/doc/content/source/kokkos/auxkernels/KokkosFunctionAux.md
```diff
@@ -0,0 +1,18 @@
+# KokkosFunctionAux
+
+!if! function=hasCapability('kokkos')
+
+This is the Kokkos version of [FunctionAux](FunctionAux.md). See the original document for details.
+
+## Example Syntax
+
+!listing test/tests/kokkos/functions/piecewise_constant/kokkos_piecewise_constant.i start=[coef_aux] end=[] include-end=true
```
lindsayad:
```suggestion
!listing test/tests/kokkos/functions/piecewise_constant/kokkos_piecewise_constant.i block=KokkosAuxKernels
```

framework/doc/content/source/kokkos/functions/KokkosConstantFunction.md
```diff
@@ -0,0 +1,18 @@
+# KokkosConstantFunction
+
+!if! function=hasCapability('kokkos')
+
+This is the Kokkos version of [ConstantFunction](ConstantFunction.md). See the original document for details.
+
+## Example Syntax
+
+!listing test/tests/kokkos/functions/constant_function/kokkos_constant_function.i start=[constant] end=[] include-end=true
```
lindsayad:
```suggestion
!listing test/tests/kokkos/functions/constant_function/kokkos_constant_function.i block=KokkosFunctions
```

framework/doc/content/source/kokkos/functions/KokkosPiecewiseConstant.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Example Syntax
+
+!listing test/tests/kokkos/functions/piecewise_constant/kokkos_piecewise_constant.i start=[func_x_y] end=[] include-end=true
+
+!listing test/tests/kokkos/functions/piecewise_constant/kokkos_piecewise_constant.i start=[func_xy_data] end=[] include-end=true
+
+!listing test/tests/kokkos/functions/piecewise_constant/kokkos_piecewise_constant.i start=[func_csv] end=[] include-end=true
+
+!listing test/tests/kokkos/functions/piecewise_constant/kokkos_piecewise_constant.i start=[func_json] end=[] include-end=true
```
lindsayad:
```suggestion
!listing test/tests/kokkos/functions/piecewise_constant/kokkos_piecewise_constant.i block=KokkosFunctions
```
I've grown suspicious that we're missing some doc hooks?

framework/doc/content/syntax/KokkosFunctions/index.md
NamjaeChoi: The compiler can still choose to spill `p` or use register if taking it as value, but I think it should always go through global memory if we use reference.

NamjaeChoi: I suggest to copy scalars and small structs whenever possible.

framework/doc/content/source/kokkos/functions/KokkosPiecewiseConstant.md
NamjaeChoi: What do you mean?

framework/doc/content/syntax/KokkosFunctions/index.md
lindsayad: kk. Yea we don't have to dwell on this too much. Device profiling tools could help us change our minds in the future if need be

framework/doc/content/source/kokkos/functions/KokkosPiecewiseConstant.md
lindsayad: Did you just not want to show the top-level block before? I was worried that there was a documentation issue if you used the syntax I suggested. But the fact that you made the changes I suggested makes me not worry

framework/include/functions/FunctionInterface.h
```diff
@@ -71,13 +81,72 @@ class FunctionInterface
...
+   * @return The copy of the Kokkos function of the abstract type with name associated with the
+   * parameter 'name'
+   */
+  Moose::Kokkos::Function getKokkosFunction(const std::string & name) const;
+
+  /**
+   * Get a Kokkos function of an abstract type with a given name
+   * @param name The name of the Kokkos function to retrieve
+   * @return The copy of the Kokkos function of the abstract type with name 'name'
+   */
```
lindsayad, line 104: Let's mention in these doc strings that these functions will error if Kokkos is configured with a GPU backend

framework/include/functions/FunctionInterface.h
```diff
@@ -71,13 +81,72 @@ class FunctionInterface
...
+   * Determine if the Kokkos function exists
+   * @param name The name of the Kokkos function
+   * @return True if the Kokkos function exists
+   */
+  bool hasKokkosFunctionByName(const FunctionName & name) const;
+#endif
+
 private:
+#ifdef MOOSE_KOKKOS_ENABLED
+  /// Helper function to retrieve a Kokkos function of concrete type
```
lindsayad: Looks abstract to me

framework/include/functions/PiecewiseTabularBase.h
```diff
@@ -31,27 +32,9 @@ class PiecewiseTabularBase : public PiecewiseBase
...
-
-  /// Builds data from 'xy_data' parameter.
-  void buildFromXY();
-
-  /// Boolean to keep track of whether the data has been loaded
-  bool _raw_data_loaded;
+  using PiecewiseTabularInterface::buildFromFile;
+  using PiecewiseTabularInterface::buildFromJSON;
+  using PiecewiseTabularInterface::buildFromXandY;
+  using PiecewiseTabularInterface::buildFromXY;
```
lindsayad, line 39: this shouldn't be required if the base is not template dependent

framework/include/kokkos/actions/AddKokkosFunctionAction.h
```diff
@@ -0,0 +1,26 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "KokkosObjectAction.h"
+
+/**
+ * Adds a Kokkos function for GPU computation
```
lindsayad:
```suggestion
 * Adds a Kokkos function for device computation
```

framework/include/kokkos/base/KokkosFunctor.h
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Moose
+{
+namespace Kokkos
+{
+
+/**
+ * The abstract class that provides polymorphic interfaces for a functor
+ */
+class Functor final
```
lindsayad: I am a little confused. My current belief is that this code is not used at all? If so can we remove it, split it into a WIP PR that waits on RDC in PETSc? I'd prefer not to have uncovered code sitting around

framework/doc/content/source/kokkos/functions/KokkosPiecewiseConstant.md
NamjaeChoi: I just wanted to I show all the four possible data formats, but your change also serves the same purpose.

framework/include/kokkos/functions/KokkosConstantFunction.h
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
+
+#include "KokkosFunction.h"
+
+/**
+ * Class that represents constant function
+ */
+class KokkosConstantFunction : public Moose::Kokkos::FunctionBase
+{
+  using Real3 = Moose::Kokkos::Real3;
```
lindsayad: I think it's ok for these class-scoped typedefs to be public

framework/include/kokkos/functions/KokkosFunction.h
```diff
@@ -0,0 +1,197 @@
...
+   */
+  FunctionBase(const FunctionBase & object);
+
+  /**
+   * Evaluate a scalar value at point (t,x,y,z)
+   * @param t The time
+   * @param p The location in space (x,y,z)
+   * @returns The scalar value evaluated at the time and location
+   */
+  KOKKOS_FUNCTION Real value(Real /* t */, Real3 /* p */) const
```
lindsayad, line 51: what would happen if these declarations/definitions weren't here?

framework/include/kokkos/base/KokkosFunctor.h
lindsayad: Well actually it seems that this is just a duplicate of `Moose::Kokkos::Function` at this point?

framework/include/kokkos/functions/KokkosPiecewiseBase.h
```diff
@@ -0,0 +1,41 @@
...
+ * for such data formulation.
+ */
+class KokkosPiecewiseBase : public Moose::Kokkos::FunctionBase
+{
+public:
+  static InputParameters validParams();
+
+  KokkosPiecewiseBase(const InputParameters & parameters);
+
+  KOKKOS_FUNCTION auto functionSize() const { return _raw_x.size(); }
```
lindsayad: I'd like to know the return type here

framework/include/kokkos/functions/KokkosPiecewiseBase.h
```diff
@@ -0,0 +1,41 @@
...
+  KOKKOS_FUNCTION Real domain(const unsigned int i) const { return _raw_x[i]; }
+  KOKKOS_FUNCTION Real range(const unsigned int i) const { return _raw_y[i]; }
+
+  /**
+   * Provides a means for explicitly setting the x and y data. This must
+   * be called in the constructor of inherited classes.
+   */
+  virtual void setData(const std::vector<Real> & x, const std::vector<Real> & y);
```
lindsayad: Is this overridden currently?

framework/include/kokkos/functions/KokkosPiecewiseConstant.h
```diff
@@ -0,0 +1,103 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "KokkosPiecewiseTabularBase.h"
+#include "KokkosUtils.h"
+
+using Moose::Kokkos::Real3;
```
lindsayad: Let's scope this in the class as you did earlier, not have it dangle in the global namespace

framework/include/kokkos/functions/KokkosPiecewiseConstant.h
```diff
@@ -0,0 +1,103 @@
...
+
+KOKKOS_FUNCTION inline Real
+KokkosPiecewiseConstant::value(Real t, Real3 p) const
+{
+  using Moose::Kokkos::Utils::sign;
+
+  const Real x = _has_axis ? p(_axis) : t;
+
+  const auto len = functionSize();
+  const Real tolerance = 1.0e-14;
```
lindsayad:
```suggestion
  constexpr Real tolerance = 1.0e-14;
```
My understanding is that all the device compilers we care about support this

framework/include/kokkos/functions/KokkosPiecewiseTabularBase.h
```diff
@@ -0,0 +1,43 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "KokkosPiecewiseBase.h"
+
+#include "PiecewiseTabularInterface.h"
+
+/**
+ * Function base which provides a piecewise approximation to a provided (x,y) point data set via
```
lindsayad:
```suggestion
 * Piecewise tabular base class which provides a piecewise approximation to a provided (x,y) point data set via
```

framework/include/kokkos/functions/KokkosPiecewiseTabularBase.h
```diff
@@ -0,0 +1,43 @@
...
+
+protected:
+  /// function value scale factor
+  Moose::Kokkos::Scalar<const Real> _scale_factor;
+
+private:
+  using PiecewiseTabularInterface::buildFromFile;
+  using PiecewiseTabularInterface::buildFromJSON;
+  using PiecewiseTabularInterface::buildFromXandY;
+  using PiecewiseTabularInterface::buildFromXY;
```
lindsayad, line 39: Again feel like we should be able to remove this without a template dependent base?

framework/include/problems/FEProblemBase.h
```diff
@@ -644,6 +650,26 @@ class FEProblemBase : public SubProblem, public Restartable
   virtual bool hasFunction(const std::string & name, const THREAD_ID tid = 0);
   virtual Function & getFunction(const std::string & name, const THREAD_ID tid = 0);
 
+#ifdef MOOSE_KOKKOS_ENABLED
+  virtual void addKokkosFunction(const std::string & type,
+                                 const std::string & name,
+                                 InputParameters & parameters);
+  virtual bool hasKokkosFunction(const std::string & name);
```
lindsayad, line 668: doxygen plz. Might have skipped over it if you hadn't provided nice doxygen for your other new methods 😆

framework/include/problems/FEProblemBase.h
```diff
@@ -644,6 +650,26 @@ class FEProblemBase : public SubProblem, public Restartable
...
+   * @returns The copy of the Kokkos function in the abstract type
+   */
+  virtual Moose::Kokkos::Function getKokkosFunction(const std::string & name);
+  /**
+   * Get a Kokkos function in a concrete type
+   * @param name The Kokkos function name
+   * @returns The reference of the Kokkos function in the concrete type
+   */
+  template <typename T>
+  T & getKokkosFunction(const std::string & name);
```
lindsayad, line 682: why does the former return a copy but the latter returns a reference?

framework/src/functions/PiecewiseTabularInterface.C
```diff
@@ -0,0 +1,239 @@
...
+        "been set. This is not allowed. Please check the parameter groups in the "
+        "documentation for the list of parameters for each data input method.");
+}
+
+void
+PiecewiseTabularInterface::buildFromFile(const libMesh::Parallel::Communicator & comm)
+{
+  // Input parameters
+  const auto & data_file_name = _parameters.get<FileName>("data_file");
+  const MooseEnum format = _parameters.get<MooseEnum>("format");
```
lindsayad:
```suggestion
  const MooseEnum & format = _parameters.get<MooseEnum>("format");
```

framework/src/kokkos/actions/AddKokkosFunctionAction.K
```diff
@@ -0,0 +1,36 @@
...
+{
+}
+
+void
+AddKokkosFunctionAction::act()
+{
+  FunctionParserBase<Real> fp;
+  std::string vars = "x,y,z,t,NaN,pi,e";
+  if (fp.Parse(_name, vars) == -1) // -1 for success
+    mooseWarning("Function name '" + _name + "' could evaluate as a ParsedFunction");
```
lindsayad: I don't understand this warning. The actual name of the function in the input file is parsable? This warning should add more context if it remains. Like "name parses to a function... we don't use the name itself as a function; hopefully that was not your intent"

framework/src/kokkos/base/KokkosFunctor.K
lindsayad, line 1: as stated previously, I feel like this code can go?

test/tests/kokkos/functions/constant_function/tests
```diff
@@ -0,0 +1,12 @@
+[Tests]
+  issues = '#30655'
+  design = 'KokkosConstantFunction.md'
+  [test]
+    type = 'Exodiff'
+    input = 'kokkos_constant_function.i'
+    exodiff = 'kokkos_constant_function_out.e'
+    requirement = 'The Kokkos function system shall include a constant function.'
```
lindsayad: Mention ability to retrieve abstract type

framework/include/functions/PiecewiseTabularBase.h
NamjaeChoi: These were private functions originally. I'm doing this to make these not accessible in derived classes.

framework/include/functions/FunctionInterface.h
NamjaeChoi: Probably this comment was misleading. It's used in the interface API that returns a concrete type.

framework/include/kokkos/base/KokkosFunctor.h
NamjaeChoi: Yeah this code isn't used yet. `Functor` will have a common interface that is different from `Function` for different types of underlying objects, but not implemented yet.

framework/include/kokkos/functions/KokkosFunction.h
NamjaeChoi: Since these functions are called by virtual functions in `FunctionWrapperDevice`, it will error.

framework/include/problems/FEProblemBase.h
NamjaeChoi: The former returns a lightweight wrapper that merely contains two pointers. This one returns the actual instance of a function.

framework/include/kokkos/functions/KokkosPiecewiseBase.h
NamjaeChoi: No

NamjaeChoi: This was basically copied from PiecewiseBase, but now I think of it, having this as virtual doesn't make sense because calling virtual functions in a constructor isn't safe. Correct me if I'm wrong.

test/tests/kokkos/functions/constant_function/tests
NamjaeChoi: Hmm the context is a bit weird.

framework/src/kokkos/actions/AddKokkosFunctionAction.K
NamjaeChoi: So the issue linked to this warning is https://github.com/idaholab/moose/issues/7659. Do you want to improve the message?

framework/include/kokkos/functions/KokkosFunction.h
lindsayad: A compiler error?

framework/include/kokkos/functions/KokkosPiecewiseBase.h
lindsayad: you are correct. vtables are disabled during construction

framework/include/kokkos/functions/KokkosFunction.h
NamjaeChoi: Yeah

framework/include/kokkos/functions/KokkosPiecewiseBase.h
NamjaeChoi: So the current code in `PiecewiseLinearBase` is technically wrong

framework/src/kokkos/actions/AddKokkosFunctionAction.K
lindsayad: I would change it to an error in your new action. And then yea I think we could improve the message string in both the old action (which remains a warning in case someone somewhere is relying on the behavior) and your new action to describe the possible conflict outlined in #7659

test/tests/kokkos/functions/constant_function/tests
lindsayad: I would just like it mentioned somewhere given that you have a distinct requirement regarding the concrete type

framework/include/kokkos/functions/KokkosPiecewiseBase.h
NamjaeChoi: Attempted to fix the original class but a public app was relying on this behavior. But I just deleted that function from my class. It's a bad design.

NamjaeChoi:
Any other comments?

framework/src/kokkos/actions/AddKokkosFunctionAction.K
```diff
@@ -0,0 +1,39 @@
...
+void
+AddKokkosFunctionAction::act()
+{
+  FunctionParserBase<Real> fp;
+  std::string vars = "x,y,z,t,NaN,pi,e";
+  if (fp.Parse(_name, vars) == -1) // -1 for success
+    mooseError("Function name '",
+               _name,
+               "' can be interpreted as a ParsedFunction expression. It becomes ambiguous and "
+               "can lead to an undefined behavior.");
```
lindsayad:
```suggestion
               "' can be interpreted as a parsed function expression. This can be problematic. "
               "As an example you may name a function 'x' whose functional form is 'xy'. You probably "
               "wouldn't do this, but let's assume for the sake of argument. You might also write somewhere "
               "in a consumer object 'function = x'. Well, MOOSE supports direct construction of parsed functions "
               "from `FunctionName` parameters, so is this consumer going to end up using the functional form "
               "'x' or 'xy'? It is undefined behavior.");
```

NamjaeChoi: 😮

lindsayad (approved):

PR #31669, opened 2025-10-02
Adding moose-python conda package

zachmprince:
Refs #31668

### DIFF
--- a/conda/python/meta.yaml
+++ b/conda/python/meta.yaml
@@ -1,0 +1,59 @@
+# Making a Change to this package?
+# REMEMBER TO UPDATE the .yaml files for the following packages:
+#   moose/conda_build_config.yaml
+# As well as any directions pertaining to modifying those files.
+{% set version = "2025.10.02" %}
+
+package:
+  name: moose-python
+  version: {{ version }}
+
+source:
+  - path: setup.py
+  - path: ../../python/mooseutils
+    folder: src/mooseutils
+  - path: ../../python/MooseControl
+    folder: src/MooseControl
+  - path: ../../python/mms
+    folder: src/mms
+  - path: ../../modules/geochemistry/python
+    folder: src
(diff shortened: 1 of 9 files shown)


scripts/tests/test_versioner.py
```diff
@@ -150,5 +153,254 @@ def testMatchDate(self):
         self.assertEqual(Versioner.match_date('xxx2025.04.04xxx'), (2025, 4, 4))
         self.assertEqual(Versioner.match_date('20.01.01'), None)
 
+class TestVerify(unittest.TestCase):
```
loganharbour, line 156: Can you please add docstrings for the tests that briefly describe what they're doing and how

scripts/versioner.py
```diff
@@ -269,7 +269,7 @@ def colorize(value, color):
                 if not changes and brief:
                     continue
                 change_text = 'BEHIND' if changes else colorize('OK', 'GREEN')
-                entries.append((package.name, change_text, to_path, template_path))
+                entries.append([package.name, change_text, to_path, template_path])
```
loganharbour, line 272: Why tuple to list?

loganharbour (changes requested):

conda/python/meta.yaml
```diff
@@ -0,0 +1,59 @@
+# Making a Change to this package?
+# REMEMBER TO UPDATE the .yaml files for the following packages:
+#   moose/conda_build_config.yaml
+# As well as any directions pertaining to modifying those files.
+{% set version = "2025.10.02" %}
+
+package:
+  name: moose-python
```
milljm, line 8: I think this might conflict with the internal name I already used in `moose-python` with the MPI package: https://github.com/idaholab/moose/blob/ee0740de41d6d9f880bc7bc261a6d100a8c92a40/conda/mpi/meta.yaml#L109

PR #31672, opened 2025-10-03
Use common serialization, move result storage to moose

### LINKED ISSUE #31671: Simpify resultsreader testing
### Motivation

The current resultsreader testing relies on different types of manual serialization and deserialization. There should be a common method for serialization and deserialization that is used for all tests.

### Design

Add a `serialize` and `deserialize` method used in `TestHarnessResults` that is used in unit testing.

### Impact

No new capability, expanded unit testing.

### LINKED ISSUE #31675: Move resultsreader storage to MOOSE
### Motivation

The storage of results in a database for the resultsreader currently lives outside of MOOSE, which makes it difficult to version control.

### Design

Move the database storage within the TestHarness so that it is controlled along side moose.

### Impact

May break CI while introducing the changes, but otherwise this will improve testing and reliability.

loganharbour:
Closes #31671 and closes #31675

- Added `CIVETStore` object, which takes TestHarness JSON output and adds it to a mongo database. This capability existed outside of MOOSE, which made it very difficult to keep in sync. Adding it within MOOSE also made it possible to have unit tests for the full workflow (run -> store -> read).
- Renamed `resultsreader` module to `resultsstore`.
- Added and tested the ability to store tests within the result document, avoiding the need to (when test entries are small in size) create thousands of small mongodb test documents (makes mongo slower).
- Added serialization routines for the result objects to simplify unit testing.


python/TestHarness/resultsstore/auth.py
```diff
@@ -0,0 +1,82 @@
...
+    # The host name
+    host: str
+    # The username
+    username: str
+    # The password
+    password: str
+    # The port
+    port: Optional[int] = None
+
+@staticmethod
```
zachmprince:
```suggestion
```

python/TestHarness/resultsstore/auth.py
```diff
@@ -0,0 +1,82 @@
...
+    if auth_file is None:
+        return None
+    try:
+        with open(auth_file, 'r') as f:
+            values = yaml.safe_load(f)
+        return Authentication(**values)
+    except Exception as e:
+        raise Exception(f"Failed to load credentials from '{auth_file}'") from e
+
+@staticmethod
```
zachmprince:
```suggestion
```

python/TestHarness/resultsstore/civetstore.py
```diff
@@ -0,0 +1,541 @@
...
+
+from TestHarness.resultsstore.auth import Authentication, load_authentication, has_authentication
+
+NoneType = type(None)
+
+# The max size that a result entry can be before its tests are
+# stored separately in a tests document
+MAX_RESULT_SIZE = 5.0 # MB
+
+@staticmethod
```
zachmprince, line 32:
```suggestion
```

python/TestHarness/resultsstore/civetstore.py
```diff
@@ -0,0 +1,541 @@
...
+def compress_dict(value: dict) -> bytes:
+    """
+    Compresses dict into binary.
+    """
+    assert isinstance(value, dict)
+    return zlib.compress(json.dumps(value).encode('utf-8'))
+
+@staticmethod
```
zachmprince, line 40:
```suggestion
```

python/TestHarness/resultsstore/civetstore.py
```diff
@@ -0,0 +1,541 @@
...
+from bson.binary import Binary
+import zlib
+
+from TestHarness.resultsstore.auth import Authentication, load_authentication, has_authentication
+
+NoneType = type(None)
+
+# The max size that a result entry can be before its tests are
+# stored separately in a tests document
+MAX_RESULT_SIZE = 5.0 # MB
```
zachmprince, line 30: Would it make any sense to allow control of this via a environment variable?

python/TestHarness/resultsstore/civetstore.py
```diff
@@ -0,0 +1,541 @@
...
+            if tests:
+                # Accumulate test data to be inserted, and also keep
+                # track of the names in this order so that we can
+                # in the main results object replace each test
+                # with the corresponding ID of the test within
+                # the separate tests collection
+                insert_tests = []
+                insert_test_names = []
+                for folder_name, folder_entry in result['tests'].items():
+                    for test_name, test_entry in folder_entry['tests'].items():
```
zachmprince, line 464: It might be nice to have some sort of iterator utility for the results. That way we can reduce some of these nested loops. Just something to consider next time your digging into this.

python/TestHarness/resultsstore/storedresults.py
```diff
@@ -591,9 +693,138 @@ def get_test(self, folder_name: str, test_name: str) -> TestHarnessTestResult:
...
+            raise KeyError(f'Failed to load test results for _id={missing}')
+
+    @property
+    def tests(self) -> list[StoredTestResult]:
+        """
+        Get all of the test results.
+
+        Will load all tests that have not been loaded from the database.
+        """
+        self.load_all_tests()
```
zachmprince, line 743: This function seems a bit heavy to put in a class property. Does it need to be called every time `tests` is accessed?

python/TestHarness/resultsstore/storedresults.py
```diff
@@ -591,9 +693,138 @@ def get_test(self, folder_name: str, test_name: str) -> TestHarnessTestResult:
...
+
+                # Convert time if it exists (removed after
+                # civet_version 3)
+                if 'time' in test_entry:
+                    test_entry['time'] = datetime.fromisoformat(test_entry['time'])
+
+        return data
+
+    @staticmethod
+    def deserialize_build(data: dict) -> dict:
```
zachmprince: This seems to return a `StoredResult` not a `dict`

python/TestHarness/tests/TestHarnessTestCase.py
```diff
@@ -145,6 +146,30 @@ def runTests(self, *args,
 
         return result
 
+    def runTestsCached(self, *args, **kwargs) -> RunTestsResult:
```
zachmprince, line 149: You may consider using [this](https://docs.python.org/3/library/functools.html) if you don't want to deal with maintaining your own cache.

python/TestHarness/tests/test_resultsstore_civetstore.py
```diff
@@ -0,0 +1,527 @@
...
+
+        self.assertEqual(
+            CIVETStore.parse_ssh_repo('git@othergithub.host.com:org/repo.git'),
+            ('othergithub.host.com', 'org', 'repo')
+        )
+
+        with self.assertRaises(ValueError):
+            CIVETStore.parse_ssh_repo('foo')
+
+    def getGetCIVETRepoUrl(self):
```
zachmprince: I think you need to prefix this method with `test` to have it run.

python/TestHarness/tests/test_resultsstore_civetstore.py
```diff
@@ -0,0 +1,527 @@
...
+        gold = self.buildHeaderGold(base_sha, base_env)
+        gold['civet']['event_url'] = f'github.com/idaholab/moose/commit/{base_env["CIVET_HEAD_SHA"]}'
+        gold['event_cause'] = 'push'
+        gold['pr_num'] = None
+        gold['time'] = result['time']
+
+        self.assertEqual(result, gold)
+
+    @unittest.skipUnless(os.environ.get('CIVET_EVENT_CAUSE', '').startswith('Push'),
+                         f"Skipping because not on a CIVET push")
```
zachmprince:
```suggestion
                         "Skipping because not on a CIVET push")
```

python/TestHarness/tests/test_resultsstore_civetstore.py
```diff
@@ -0,0 +1,527 @@
...
+
+    def checkResult(self, base_sha: str, env: dict, result: dict, stored_results: dict,
+                    stored_tests: Optional[list], build_kwargs: dict = {}):
+        """
+        Combined helper for comparing a test harness result to a set
+        of stored results and tests.
+
+        Each test entry is compared directly to the test harness result entry,
+        where vaues are modified/removed as needed based on the arguments
+        that are passed the build method via build_kwargs
```
zachmprince:
```suggestion
        where values are modified/removed as needed based on the arguments
        that are passed to the build method via build_kwargs.
```

python/TestHarness/tests/test_resultsstore_civetstore.py
```diff
@@ -0,0 +1,527 @@
...
+                        self.assertIsInstance(stored_test_id, ObjectId)
+                        find_test = [doc for doc in stored_tests if doc['_id'] == stored_test_id]
+                        self.assertEqual(len(find_test), 1)
+                        stored_test = find_test[0]
+                        modified_test_entry['_id'] = stored_test_id
+                        modified_test_entry['result_id'] = result_id
+                    else:
+                        self.assertIsInstance(stored_test_id, int)
+                        stored_test = stored_tests[stored_test_id]
+                    pass
```
zachmprince: What does this `pass` here do exactly?

python/TestHarness/tests/test_resultsstore_civetstore.py
```diff
@@ -0,0 +1,527 @@
...
+        store_args = [TEST_DATABASE, results, base_sha]
+        build_kwargs = {'env': env}
+        result_id, test_ids = CIVETStore().store(*store_args, **build_kwargs, **kwargs)
+        self.assertIsInstance(result_id, ObjectId)
+        self.assertIsInstance(test_ids, (list, type(None)))
+
+        stored_result, stored_tests = self.getStoredResult(result_id, test_ids)
+        self.checkResult(base_sha, env, results, stored_result, stored_tests, build_kwargs=kwargs)
+
+    @unittest.skipUnless(HAS_AUTH, f"Skipping because authentication is not available")
```
zachmprince:
```suggestion
    @unittest.skipUnless(HAS_AUTH, "Skipping because authentication is not available")
```

python/TestHarness/tests/test_resultsstore_civetstore.py
```diff
@@ -0,0 +1,527 @@
...
+    def testStore(self):
+        """
+        Test the store() method with no additional arguments and the
+        tests contained within the result (result is small enough)
+        """
+        self.runTestStore()
+
+    @unittest.skipUnless(HAS_AUTH, f"Skipping because authentication is not available")
```
zachmprince:
```suggestion
    @unittest.skipUnless(HAS_AUTH, "Skipping because authentication is not available")
```

python/TestHarness/tests/test_resultsstore_civetstore.py
```diff
@@ -0,0 +1,527 @@
...
+    def testStoreSeparateTests(self):
+        """
+        Test the store() method with no additional arguments and the
+        tests stored separate from the results
+        """
+        self.runTestStore(max_result_size=1e-6)
+
+    @unittest.skipUnless(HAS_AUTH, f"Skipping because authentication is not available")
```
zachmprince:
```suggestion
    @unittest.skipUnless(HAS_AUTH, "Skipping because authentication is not available")
```

python/TestHarness/tests/test_resultsstore_combined.py
```diff
@@ -0,0 +1,245 @@
...
+                result = reader.getPRResults(pr_num)
+            elif load_type == 'event':
+                result = reader.getEventResults(event_id)
+            elif load_type == 'commit':
+                result = reader.getCommitResults(head_sha)
+            self.assertIsNotNone(result)
+            self.compareResults(result, env, base_sha, id)
+
+    @unittest.skipUnless(HAS_STORE_AUTH, f"Skipping because store auth is not available")
+    @unittest.skipUnless(HAS_READER_AUTH, f"Skipping because reader auth is not available")
```
zachmprince:
```suggestion
    @unittest.skipUnless(HAS_STORE_AUTH, "Skipping because store auth is not available")
    @unittest.skipUnless(HAS_READER_AUTH, "Skipping because reader auth is not available")
```

python/TestHarness/tests/test_resultsstore_combined.py
```diff
@@ -0,0 +1,245 @@
...
+            self.compareResults(test_result.results, env, base_sha, result_id)
+
+        except:
+            self.deleteDocuments(result_id)
+            raise
+
+        self.deleteDocuments(result_id)
+
+    @unittest.skipUnless(HAS_STORE_AUTH, f"Skipping because store auth is not available")
+    @unittest.skipUnless(HAS_READER_AUTH, f"Skipping because reader auth is not available")
```
zachmprince:
```suggestion
    @unittest.skipUnless(HAS_STORE_AUTH, "Skipping because store auth is not available")
    @unittest.skipUnless(HAS_READER_AUTH, "Skipping because reader auth is not available")
```

python/TestHarness/tests/test_resultsstore_storedresults.py
```diff
@@ -0,0 +1,523 @@
...
+
+                # Fake values from HPC
+                test_values['timing']['hpc_queued'] = HPC_QUEUED_TIME
+
+                # Delete requested key
+                if delete_test_key:
+                    current = test_values
+                    for key in delete_test_key[:-1]:
+                        current = current[key]
+                    del current[delete_test_key[-1]]
```
zachmprince, line 132: This is pretty clever!

python/TestHarness/tests/test_resultsstore_storedresults.py
```diff
@@ -0,0 +1,523 @@
...
+        """
+        captured_result = self.captureResult(test_in_results=False)
+        test_data = captured_result.test_data
+        result_data = captured_result.result_data
+
+        # Mock getting the test results from mongodb
+        def get_test_data(id):
+            find_test = [v for v in test_data.values() if v['_id'] == id]
+            return find_test[0] if find_test else None
+        patch_find_test_data.side_effect = get_test_data
```
zachmprince:
```suggestion
        patch_find_test_data.side_effect = lambda id: next((v for v in test_data.values() if v['_id'] == id), None)
```

zachmprince (changes requested):
Looks good. Just a few suggestions and some small change requests.

python/TestHarness/resultsstore/civetstore.py
loganharbour: It's a command line argument, this is just the default value:
```
        parser.add_argument('--max-result-size', type=float, default=MAX_RESULT_SIZE,
                            help='Max size of a result for tests to be stored within it')
```

It only needs to exist like this in the event that `build()` is called without kwargs (like in unit testing), so that the default for `build()` and the default from the CLI args is the same

python/TestHarness/resultsstore/civetstore.py
loganharbour: I agree. I'm really sick of typing this particular thing out. It's difficult because sometimes I'm modifying the dict (where you really can't just use an iterator) and sometimes I'm not

python/TestHarness/resultsstore/storedresults.py
loganharbour: I thought about this - but it's only expensive once. The alternative is that I could get rid of `tests` and just make a method called `get_tests()`

python/TestHarness/tests/TestHarnessTestCase.py
loganharbour: That's a can of worms. Example? :D

python/TestHarness/tests/test_resultsstore_storedresults.py
loganharbour: I really struggled with how to describe this in the docstring... but it's really convenient being able to pass it nested

python/TestHarness/resultsstore/storedresults.py
loganharbour: Went ahead and changed this to `get_tests()`; I agree

loganharbour:
@zachmprince I did two more things:

- Separated out a function in one of the unit tests so that it could be used by other tests (like what May is working on) in https://github.com/idaholab/moose/pull/31672/commits/dad89a7bfff529aad7abc51773ba4fe58539871d
- Instead of using the same database for all tests, separated them into one for each test class just for easier debugging and more separation in https://github.com/idaholab/moose/pull/31672/commits/20a2fdce3dc16b8ae13e9f6b1b1d1716b67981e8 and https://github.com/idaholab/moose/pull/31672/commits/ef84765073b3f3c2c55ea3b74dc21dc4f5bf3fb4

Everything else addresses your review.

python/TestHarness/resultsstore/civetstore.py
zachmprince: I saw that later and forgot to delete the comment. My bad!

python/TestHarness/tests/TestHarnessTestCase.py
zachmprince: I believe you can just do this:
```python
import functools

...

    @functools.cache
    def runTests(self, *args,
                       tmp_output: bool = True,
                       no_capabilities: bool = True,
                       capture_results: bool = True,
                       exit_code: int = 0,
                       run: bool = True,
                       tests: Optional[dict[str, dict]] = None) -> RunTestsResult:
```

But I'm unsure what would happen if attributes in `self` change. I

zachmprince (approved):

loganharbour:
Open pronghorn failure is related and will be patched in the app.

PR #31708, opened 2025-10-10
Basic Kokkos Array BLAS Operations

NamjaeChoi:
Refs #30655.

### DIFF
--- a/framework/include/kokkos/base/KokkosArray.h
+++ b/framework/include/kokkos/base/KokkosArray.h
@@ -1144,6 +1144,45 @@ public:
 
     return this->operator[](i0 - _d[0]);
   }
+
+  /**
+   * Device BLAS operations
+   */
+  ///@{
+  /**
+   * Perform \p a * \p x \p op \p b * \p y and write the result to this array
+   * @param accumulate Whether to accumulate or overwrite the result
+   */
+  template <typename U = T>
+  typename std::enable_if<std::is_same<U, Real>::value, void>::type
+  axby(const U a,
+       const Array<U, 1> & x,
+       const char op,
+       const U b,
+       const Array<U, 1> & y,
+       const bool accumulate = false);
+  /**
+   * Scale \p x with \p a and write the result to this array
+   */
+  template <typename U = T>
+  typename std::enable_if<std::is_same<U, Real>::value, void>::type scal(const U a,
+                                                                         const Array<U, 1> & x);
+  /**
+   * Scale this array with \p a
+   */
+  template <typename U = T>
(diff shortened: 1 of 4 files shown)


framework/include/kokkos/base/KokkosArray.h
```diff
@@ -1144,6 +1144,45 @@ class Array<T, 1> : public ArrayBase<T, 1>
...
+  /**
+   * Device BLAS operations
+   */
+  ///@{
+  /**
+   * Perform \p a * \p x \p op \p b * \p y and write the result to this array
+   * @param accumulate Whether to accumulate or overwrite the result
+   */
+  template <typename U = T>
+  typename std::enable_if<std::is_same<U, Real>::value, void>::type
```
lindsayad, line 1157: Why have all this SFINAE if we're only allowing one possible type? Can't we just remove the templating and change `U` to `Real`?

NamjaeChoi: You mean
`typename std::enable_if<std::is_same<T, Real>::value, void>::type`
?

NamjaeChoi: Wouldn't that prevent instantiating the whole array whenever `T` is not `Real`?

lindsayad: No, I mean just change the signature (remove templating entirely) to
```
  void
  axby(const Real a,
       const Array<Real, 1> & x,
       const char op,
       const Real b,
       const Array<Real, 1> & y,
       const bool accumulate = false);

NamjaeChoi: What if `T` is not `Real`? This BLAS APIs change the array itself.

lindsayad: What are you trying to achieve? Are you only trying to have this API if `T` is `Real`? Because this SFINAE pattern won't accomplish that

NamjaeChoi: Want to give out reasonable compile error messages when somebody attempts to use them for arrays that are not `Real`.

NamjaeChoi: If I don't have SFINAE, the compiler will try to instantiate them and can throw some confusing messages.

lindsayad (approved):

framework/include/kokkos/base/KokkosArray.h
lindsayad: I see

PR #31716, opened 2025-10-14
Fix MFEMBoundarySubMesh attributes

nmnobre:
## Reason
For submeshes created from boundaries, the domain attributes are the parent's boundary attributes.
Also, the boundary of the new submesh is also numbered at one plus the parent's max boundary attribute.
I'm trying to document this is mfem/mfem#5068, [bf21910](https://github.com/mfem/mfem/pull/5068/commits/bf21910fe46f59c975dc4ba8b044a0dfb9dff904).

## Design
Minor adjustment in attribute set creation and assignment for this particular case. 

## Impact
Correct names for submesh attributes.

### DIFF
--- a/framework/src/mfem/submeshes/MFEMBoundarySubMesh.C
+++ b/framework/src/mfem/submeshes/MFEMBoundarySubMesh.C
@@ -22,6 +22,7 @@ MFEMBoundarySubMesh::validParams()
 
   params.addClassDescription("Class to construct an MFEMSubMesh formed from the subspace of the "
                              "parent mesh restricted to the set of user-specified boundaries.");
+  params.addParam<BoundaryName>("submesh_boundary", "Name to assign submesh boundary.");
   return params;
 }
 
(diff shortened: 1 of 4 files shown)


lindsayad (approved):

framework/src/mfem/submeshes/MFEMDomainSubMesh.C
```diff
@@ -38,14 +38,8 @@ MFEMDomainSubMesh::buildSubMesh()
...
-
-  if (isParamSetByUser("submesh_boundary"))
-  {
-    const BoundaryName & submesh_boundary = getParam<BoundaryName>("submesh_boundary");
-    _submesh->bdr_attribute_sets.CreateAttributeSet(submesh_boundary);
-    _submesh->bdr_attribute_sets.AddToAttributeSet(submesh_boundary,
-                                                   getMesh().bdr_attributes.Max() + 1);
-  }
+  _submesh->bdr_attribute_sets.CreateAttributeSet(getParam<BoundaryName>("submesh_boundary"))
+      .Append(getMesh().bdr_attributes.Max() + 1);
```
alexanderianblair: This will likely crash if two `MFEMDomainSubMesh` objects are left with the default name, since `CreateAttributeSet` includes an `MFEM_VERIFY` call to check a new attribute set is not registered with the same name as an existing one.

Worth checking, or keeping within the `isParamSetByUser` guard to avoid unexpected behaviour

framework/src/mfem/submeshes/MFEMBoundarySubMesh.C
```diff
@@ -36,8 +37,9 @@ MFEMBoundarySubMesh::buildSubMesh()
 {
   _submesh = std::make_shared<mfem::ParSubMesh>(
       mfem::ParSubMesh::CreateFromBoundary(getMesh(), getBoundaryAttributes()));
-  _submesh->attribute_sets.attr_sets = getMesh().attribute_sets.attr_sets;
-  _submesh->bdr_attribute_sets.attr_sets = getMesh().bdr_attribute_sets.attr_sets;
+  _submesh->attribute_sets.attr_sets = getMesh().bdr_attribute_sets.attr_sets;
+  _submesh->bdr_attribute_sets.CreateAttributeSet(getParam<BoundaryName>("submesh_boundary"))
+      .Append(getMesh().bdr_attributes.Max() + 1);
```
alexanderianblair: See other comment regarding `CreateAttributeSet` - worth adding an `isParamSetByUser` guard to avoid the default name causing issues when multiple `MFEMBoundarySubMesh` objects are in use

alexanderianblair (changes requested):

alexanderianblair (approved):

Issue #31729, opened 2025-10-15
XFEM near-tip enrichment for crack propagation


donggiang:
### Motivation

New features allow crack tip to be placed anywhere inside elements whose solutions are enriched. Consequently, it allows correct computation of stress and contour integrals such as J and C(t).

### Design

**Extended XFEM Repetition Logic in FEProblemBase**
A new flag has been added to the FEProblemBase class, complementing the existing mechanism that detects changes in the background mesh. This enhancement ensures that XFEM solutions are repeated not only when the background mesh changes, but also when the cutter mesh advances by an arbitrary increment. Additionally, this resolves a visualization lag present in the previous implementation.

**New Material Class: ComputeCrackTipEnrichmentIncrementalStrain**
Introduces a new material class designed to enrich the solution field around crack tips while supporting nonlinear material behavior. 

### Impact

Improve the robustness and accuracy of the XFEM module

PR #31730, opened 2025-10-15
XFEM near-tip enrichment for crack propagation

donggiang:
refs #31729

## Reason
New features allow crack tip to be placed anywhere inside elements whose solutions are enriched. Consequently, it allows correct computation of stress and contour integrals such as J and C(t).

## Design
Extended XFEM Repetition Logic in FEProblemBase
A new flag has been added to the FEProblemBase class, complementing the existing mechanism that detects changes in the background mesh. This enhancement ensures that XFEM solutions are repeated not only when the background mesh changes, but also when the cutter mesh advances by an arbitrary increment. Additionally, this resolves a visualization lag present in the previous implementation.

New Material Class: ComputeCrackTipEnrichmentIncrementalStrain
Introduces a new material class designed to enrich the solution field around crack tips while supporting nonlinear material behavior.

## Impact
Improve the robustness and accuracy of the XFEM module


framework/src/problems/FEProblemBase.C
```diff
@@ -8276,18 +8276,27 @@ FEProblemBase::updateMeshXFEM()
...
+    crack_front_advanced = _xfem->didNearTipEnrichmentChange();
+    if (crack_front_advanced)
+      _console << "\nXFEM update complete: Mesh not modified but advancing crack changed near-tip "
+                  "enrichment"
+               << std::endl;
     else
       _console << "\nXFEM update complete: Mesh not modified" << std::endl;
   }
-  return updated;
+  return (updated | crack_front_advanced);
```
```suggestion
  return (updated || crack_front_advanced);
```

bwspenc:
You'll need to reference an issue number. I would do it both in the title of your commit message (something like "ref #xxxxxx") as well as in the PR description, which you can edit.

framework/src/problems/FEProblemBase.C
```diff
@@ -8276,18 +8276,27 @@ FEProblemBase::updateMeshXFEM()
           /*intermediate_change=*/false, /*contract_mesh=*/true, /*clean_refinement_flags=*/false);
 
     updated = _xfem->update(_time, _nl, *_aux);
-    if (updated)
+    if (updated) // if subcrical crack advancement, do not need to reinitialize solution
```
bwspenc:
```suggestion
    if (updated)
```

framework/src/problems/FEProblemBase.C
```diff
@@ -8276,18 +8276,27 @@ FEProblemBase::updateMeshXFEM()
...
     {
       meshChanged(
           /*intermediate_change=*/false, /*contract_mesh=*/true, /*clean_refinement_flags=*/false);
       _xfem->initSolution(_nl, *_aux);
       restoreSolutions();
       _console << "\nXFEM update complete: Mesh modified" << std::endl;
     }
+  }
+  bool crack_front_advanced = false;
```
bwspenc, line 8439:
```suggestion
  // Changing near-tip enrichment requires repeating the solution, even if the mesh is not updated
  bool crack_front_advanced = false;
```

modules/solid_mechanics/include/materials/ComputeIncrementalStrainBase.h
```diff
@@ -40,4 +40,5 @@ class ComputeIncrementalStrainBase : public ComputeStrainBase
   const MaterialProperty<RankTwoTensor> & _total_strain_old;
 
   std::vector<const MaterialProperty<RankTwoTensor> *> _eigenstrains_old;
+  MaterialProperty<RankTwoTensor> & _grad_disp_rate;
```
bwspenc:
```suggestion
  /// Rate of change of the displacement gradient
  MaterialProperty<RankTwoTensor> & _grad_disp_rate;
```

modules/solid_mechanics/include/userobjects/CrackFrontDefinition.h
```diff
@@ -41,6 +41,7 @@ class CrackFrontDefinition : public GeneralUserObject, public BoundaryRestrictab
   virtual void finalize() override;
   virtual void execute() override;
 
+  void updateCrackFrontPoints();
```
bwspenc:
```suggestion
    /**
     * Obtain the updated set of crack points from the CrackFrontPointsProvider and 
     * perform other needed initialization for that set of crack front points. This is only
     * to be called if a CrackFrontPointsProvider is used.
     */
    void updateCrackFrontPoints();

```

modules/solid_mechanics/src/userobjects/CrackFrontDefinition.C
```diff
@@ -357,25 +357,26 @@ CrackFrontDefinition::initialSetup()
   }
 }
 
+void
+CrackFrontDefinition::updateCrackFrontPoints()
+{
+  _crack_front_points =
```
bwspenc:
```suggestion
  mooseAssert(_crack_front_points_provider, "_crack_front_points_provider is NULL");
  _crack_front_points =
```

modules/xfem/doc/content/source/materials/ComputeCrackTipEnrichmentIncrementalStrain.md
```diff
@@ -0,0 +1,14 @@
+# ComputeCrackTipEnrichmentIncrementalStrain
+
+!syntax description /Materials/ComputeCrackTipEnrichmentIncrementalStrain
+
+# Description
+
+The displacement field contains both the standard and the enrichment solution. The
+`ComputeCrackTipEnrichmentIncrementalStrain` is inherited from '`ComputeCrackTipEnrichmentSmallStrain` to allow nonlinear material models, and is applicable only to small-strain problems.
```
bwspenc:
```suggestion
With XFEM, the displacement field contains both the standard and optionally, the near-tip enrichment solution. This Material calculates the strains for near-tip enrichment from both the standard and enrichment solutions. This model is incremental, which allows for the use of nonlinear material models, but is applicable only to small-strain problems. This is a drop-in replacement for the standard strain calculators.
```

modules/xfem/doc/content/source/materials/ComputeCrackTipEnrichmentIncrementalStrain.md
```diff
@@ -0,0 +1,14 @@
+# ComputeCrackTipEnrichmentIncrementalStrain
```
bwspenc, line 1: Should we rename this Material to `ComputeCrackTipEnrichmentIncrementalSmallStrain`?

modules/xfem/include/base/XFEM.h
```diff
@@ -322,6 +322,7 @@ class XFEM : public XFEMInterface
   CutSubdomainID getCutSubdomainID(const GeometricCutUserObject * gcuo,
                                    const Elem * cut_elem,
                                    const Elem * parent_elem = nullptr) const;
+  virtual bool didNearTipEnrichmentChange() override;
```
bwspenc, line 331:
```suggestion

  /**
   * Check whether the near-tip enrichment changed during the most recent call to
   * XFEM::update().
   * @return true if it changed
   */
  virtual bool didNearTipEnrichmentChange() override;
```

modules/xfem/include/materials/ComputeCrackTipEnrichmentIncrementalStrain.h
```diff
@@ -0,0 +1,76 @@
...
+  /// enrichment displacement variables
+  std::vector<std::vector<MooseVariableFEBase *>> _enrich_variable;
+
+  /// the current shape functions
+  const VariablePhiValue & _phi;
+
+  /// gradient of the shape function
+  const VariablePhiGradient & _grad_phi;
+
+  const MaterialProperty<RankTwoTensor> & _mechanical_strain_old;
```
bwspenc, line 51: Add doxygen descriptions of all these member variables

modules/xfem/include/materials/ComputeCrackTipEnrichmentIncrementalStrain.h
```diff
@@ -0,0 +1,76 @@
...
+  const MaterialProperty<RankTwoTensor> & _total_strain_old;
+  MaterialProperty<RankTwoTensor> & _grad_disp_tensor;
```
bwspenc: I think we can avoid making _grad_disp_tensor and _grad_disp_tensor_old material properties. Look at the usage of `_grad_disp_old in ADComputeIncrementalStrainBase.C for an example.

modules/xfem/src/materials/ComputeCrackTipEnrichmentIncrementalStrain.C
```diff
@@ -0,0 +1,172 @@
...
+#include "libmesh/quadrature.h"
+
+registerMooseObject("XFEMApp", ComputeCrackTipEnrichmentIncrementalStrain);
+
+InputParameters
+ComputeCrackTipEnrichmentIncrementalStrain::validParams()
+{
+  InputParameters params = ComputeStrainBase::validParams();
+  params.addClassDescription(
+      "Computes the crack tip enrichment at a point within an incremental strain formulation.");
```
bwspenc:
```suggestion
      "Computes the crack tip enrichment strain for an incremental small-strain formulation.");
```

modules/xfem/src/materials/ComputeCrackTipEnrichmentIncrementalStrain.C
```diff
@@ -0,0 +1,172 @@
...
+    _enrich_disp(3),
+    _grad_enrich_disp(3),
+    _grad_enrich_disp_old(3),
+    _enrich_variable(4),
+    _phi(_assembly.phi()),
+    _grad_phi(_assembly.gradPhi()),
+    _mechanical_strain_old(getMaterialPropertyOld<RankTwoTensor>(_base_name + "mechanical_strain")),
+    _total_strain_old(getMaterialPropertyOld<RankTwoTensor>(_base_name + "total_strain")),
+    _grad_disp_tensor(declareProperty<RankTwoTensor>(_base_name + "grad_disp_tensor")),
+    _small_strain(declareProperty<RankTwoTensor>(_base_name + "small_strain")),
```
bwspenc: I'd like to avoid declaring a property called `small_strain` because we don't do that for the non-enriched strain, and this could be confusing to users. I think we can just use `total_strain` for this. We can discuss this.

modules/xfem/src/materials/ComputeCrackTipEnrichmentIncrementalStrain.C
```diff
@@ -0,0 +1,172 @@
...
+    _BI.resize(8); // HEX8
+
+  for (unsigned int i = 0; i < _BI.size(); ++i)
+    _BI[i].resize(4);
+}
+
+void
+ComputeCrackTipEnrichmentIncrementalStrain::computeProperties()
+{
+  FEType fe_type(Utility::string_to_enum<Order>("first"),
```
bwspenc, line 79: Does this only work with first-order elements?

modules/xfem/include/materials/ComputeCrackTipEnrichmentIncrementalStrain.h
```diff
@@ -0,0 +1,76 @@
...
+//* https://www.mooseframework.org
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+// #include "ComputeStrainBase.h"
```
bwspenc:
```suggestion
```

modules/xfem/src/materials/ComputeCrackTipEnrichmentIncrementalStrain.C
```diff
@@ -0,0 +1,172 @@
...
+  : ComputeIncrementalStrainBase(parameters),
+    EnrichmentFunctionCalculation(&getUserObject<CrackFrontDefinition>("crack_front_definition")),
+    _enrich_disp(3),
+    _grad_enrich_disp(3),
+    _grad_enrich_disp_old(3),
+    _enrich_variable(4),
+    _phi(_assembly.phi()),
+    _grad_phi(_assembly.gradPhi()),
+    _mechanical_strain_old(getMaterialPropertyOld<RankTwoTensor>(_base_name + "mechanical_strain")),
+    _total_strain_old(getMaterialPropertyOld<RankTwoTensor>(_base_name + "total_strain")),
```
bwspenc, line 45: We already have `_total_strain_old` defined in `ComputeIncrementalStrainBase`.

modules/xfem/src/userobjects/MeshCut2DUserObjectBase.C
```diff
@@ -279,14 +279,61 @@ MeshCut2DUserObjectBase::findOriginalCrackFrontNodes()
...
+          // Check if the point lies within the edge segment bounds
+          Real dot_product = point_vector * edge_vector;
+          Real edge_length_squared = edge_vector * edge_vector;
+
+          if (dot_product >= 0 && dot_product <= edge_length_squared)
+            return true;
+        }
+      }
+      else
+        mooseError("Current implementation only supports 2D problem!\n");
```
bwspenc:
```suggestion
        mooseError("MeshCut2DUserObjectBase currently only supports linear elements.\n");
```

modules/xfem/test/tests/crack_tip_test/tests
```diff
@@ -0,0 +1,22 @@
+[Tests]
+  issues = '#31729'
+  design = 'ComputeCrackTipEnrichmentIncrementalStrain.md'
+  requirement = 'The XFEM Module shall have the capability to allow crack tip'
```
bwspenc: You'll need a separate requirement for each test, and those requirements should specifically state the capabilities tested.

bwspenc (changes requested):

modules/xfem/src/materials/ComputeCrackTipEnrichmentIncrementalStrain.C
donggiang: yes, the number of enriched functions only support linear element

modules/xfem/doc/content/source/materials/ComputeCrackTipEnrichmentIncrementalStrain.md
donggiang: We should not. It has the same structure as `ComputeIncrementalStrainBase`, however I removes rotational tensor part for simplicity at this stage.

modules/xfem/include/materials/ComputeCrackTipEnrichmentIncrementalStrain.h
donggiang: I removed this variable.

modules/xfem/src/materials/ComputeCrackTipEnrichmentIncrementalStrain.C
donggiang: I removed this variable.

modules/xfem/test/tests/crack_tip_test/tests
donggiang: I added two seperate requirements for each test.

donggiang:
@bwspenc please review again!

PR #31732, opened 2025-10-15
Drop support for calling deprecated BoundaryInfo::build_side_list()

jwpeterson:
The deprecated version of this function will be removed in libMesh/libmesh#4276, so let's see if any MOOSE apps are still calling it.

cc: @roystgnr

### DIFF
--- a/framework/include/mesh/MooseMesh.h
+++ b/framework/include/mesh/MooseMesh.h
@@ -278,16 +278,8 @@ public:
   void buildNodeListFromSideList();
 
   /**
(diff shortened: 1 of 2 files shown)


roystgnr (changes requested):
Can we just remove this (version of this) method entirely now?  If anybody is still trying to use it then I'd prefer to have them discover the problem at compile time rather than their users at run time.

jwpeterson:
FYI @roystgnr I completely removed the function in question and this PR still passes.

roystgnr (approved):

PR #31739, opened 2025-10-16
Results Summary Improvements

### LINKED ISSUE #31738: `ResultsSummary` improvements
### Motivation

Looking to add a few quality of life improvements to the results summary comments.

### Design

- Increase the default "floor" for whether or not to show test times as changed to 2 seconds. This means that tests won't be checked for run-time differences unless the head test takes more than 2 seconds.
- Ensure that formatting in tables is consistent; currently some times are reported as "0.8" for example, but they should be reported as "0.80" (exactly 2 values after the decimal; `:.2f` formatting with python)
- When a comparison is available, show the commit hash of the comparison and the URL to CIVET job that produced those tests

### Impact

New capabilities.

maythuwin1313:
Closes #31738

### DIFF
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -160,7 +160,7 @@ class TestHarnessResultsSummary:
 
     def _sort_test_times_key(test_table_row: list, test_time_col_index: int) -> tuple:
         """
-        Generate a sorting key for a row based on column index which test time is located
+        Generate a sorting key for each test table row based on test time value
         Parameters:
         ----------
         test_table_row : list
@@ -190,7 +190,7 @@ class TestHarnessResultsSummary:
 
     def sort_test_times(self, test_table: list[list], test_time_col_index: int) -> list[list]:
         """
-        Sort a list of test results rows based on the test time column using custom sort.
+        Sort a list of test table based on the test time values using custom sort.
         The sorting logic priorities as follow:
             1. Numeric values (sorted in descending order).
             2. The string 'SKIP' (sorted after numeric values).
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -60,7 +60,7 @@ class TestHarnessResultsSummary:
         pr_parser.add_argument(
             '--run-time-floor',
             type = float,
-            default = 2,
+            default = 2.0,
             help = 'The minimum threshold for a test to checked for a run time difference'
         )
 
@@ -143,7 +143,7 @@ class TestHarnessResultsSummary:
         base_results = self.get_commit_results(base_sha)
         # If no base, display message and return None for base
         if not isinstance(base_results, StoredResult):
-            no_base = f"Base results not available for {base_sha}"
+            no_base = f"Base results not available for {base_sha[:7]}"
             # Write no base message in out_file path
             self.write_output(no_base, out_file)
             return None, head_results, None, test_names
--- a/python/TestHarness/resultssummary/summary.py
+++ b/python/TestHarness/resultssummary/summary.py
@@ -182,14 +182,19 @@ class TestHarnessResultsSummary:
         test_table = []
         for test_name in test_names:
             test_result = test_results.get_test(test_name.folder, test_name.name)
-            # if run time is None, it will display "" in Time column
-            test_result_run_time = (
-            f'{test_result.run_time:.2f}' if test_result.run_time is not None else ""
-            )
+            # Test_result status is SKIP, run time will show as SKIP
+            # Run time is None, run time will show ''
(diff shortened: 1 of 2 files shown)


python/TestHarness/resultssummary/summary.py
```diff
@@ -158,6 +158,24 @@ def _format_test_name(test_name):
         """
         return f'`{str(test_name)}`'
 
+    def _sort_key(self, column_index):
```
loganharbour: Let's call this `_sort_test_times` because I think that's a bit clearer? Can you also add type hinting?

python/TestHarness/resultssummary/summary.py
```diff
@@ -158,6 +158,24 @@ def _format_test_name(test_name):
...
+        """
+        def __sort_key(row):
+            value = row[column_index]
+            if value.replace('.', '', 1).isdigit():
+                return (0, -float(value))
+            elif value == 'SKIP':
+                return (1, 0)
+            else:
+                return (2, 0)
+        return __sort_key
```
loganharbour:
```suggestion
            return (2, 0)
```

Looks like you have an extra return here

python/TestHarness/resultssummary/summary.py
```diff
@@ -182,19 +200,22 @@ def _build_diff_table(self, test_names: set[TestName], test_results: StoredResul
...
-            )
+            # Test_result status is SKIP, run time will show as SKIP
+            # Run time is None, run time will show ''
+            if test_result.status is not None and \
+                test_result.status_value == 'SKIP':
+                run_time = 'SKIP'
+            elif test_result.run_time is not None:
+                run_time = f'{test_result.run_time:.2f}'
+            else:
+                run_time = ''
```
loganharbour:
```suggestion
            # Test is skipped, so show time as SKIP
            if test_result.status is not None and \
                test_result.status_value == 'SKIP':
                run_time = 'SKIP'
            # Test has a run time
            elif test_result.run_time is not None:
                run_time = f'{test_result.run_time:.2f}'
            # Test does not have a run time
            else:
                run_time = ''
```

python/TestHarness/resultssummary/summary.py
```diff
@@ -242,7 +263,10 @@ def _build_same_table(self, same_names: set[TestName],
                 head_result.run_time < head_run_time_floor:
                 continue
             # Calculate relative runtime ratio between base and head
-            relative_runtime = (head_result.run_time - base_result.run_time) / base_result.run_time
+            elif base_result.run_time == 0:
+                relative_runtime = head_result.run_time
+            else:
+                relative_runtime = float((head_result.run_time - base_result.run_time) / base_result.run_time)
```
loganharbour: If the test doesn't have a run time before, we can't really compare can we? `relative_runtime` is the percentage, right?

loganharbour (changes requested):

python/TestHarness/resultssummary/summary.py
maythuwin1313: oh yes. sorry always miss to add type hinting. will add it.

python/TestHarness/resultssummary/summary.py
maythuwin1313: It has another get key function inside. I will update to refractor the function, and also make the function name to be clear.

python/TestHarness/resultssummary/summary.py
maythuwin1313: yeah it make sense. I will skip test.

python/TestHarness/resultssummary/summary.py
```diff
@@ -158,6 +158,65 @@ def _format_test_name(test_name):
...
+        ----------
+        test_table_row : list
+            A list representing a row of data from the test table
+        test_time_col_index : int
+            The index of the column containing the test time value
+
+        Returns:
+        ----------
+        sorting_key : tuple
+            A tempory sorting key:
```
loganharbour:
```suggestion
            A temprorary sorting key:
```

You should use this in vscode:

Screenshot 2025-10-21 at 12 30 27 PM

It's pretty great for spelling; catches my poor spelling all of the time

python/TestHarness/resultssummary/summary.py
```diff
@@ -158,6 +158,65 @@ def _format_test_name(test_name):
         """
         return f'`{str(test_name)}`'
 
+    def _sort_test_times_key(self, test_table_row: list, test_time_col_index: int) -> tuple:
```
loganharbour: You can:

```
from typing import Tuple
```

and then do:

```
... -> Tuple[type1, type2]:
```

for the types in the tuple

loganharbour (changes requested):
Small changes. Otherwise, looks great!

loganharbour (approved):

Issue #31755, opened 2025-10-20
Rewrite MooseControl to be pluggable, add slurm execution support


loganharbour:
### Motivation

There is currently a need to run the MooseControl with a slurm process on a remote host. This is very difficult as it stands because the branching logic in MooseControl is rigid.

The execution portion of the MooseControl (starting and finishing the process, etc) should be separated out so that it can be more modular.

### Design

Separate out how the process is started and finished in the MooseControl into a different class.

Add the ability for MOOSE to be spawned and communicated with on a remote HPC host using the slurm scheduler.

### Impact

Breaking changes to the current MooseControl.

PR #31756, opened 2025-10-20
Rework runner strategy in `MooseControl`, add testing

loganharbour:
Refs #31755

- Renames the `MooseControl` python module to `moosecontrol`
- Moves initialization and finalization in `MooseControl` to a separate runner class


python/moosecontrol/validation.py
```diff
@@ -0,0 +1,184 @@
...
+        """Name of the WebServerControl."""
+        return self.data["control_name"]
+
+    @property
+    def execute_on_flags(self) -> list[str]:
+        """Execute on flags the WebServerControl is listening on."""
+        return self.data["execute_on_flags"]
+
+
+@staticmethod
```
zachmprince:
```suggestion
```

python/moosecontrol/validation.py
```diff
@@ -0,0 +1,184 @@
...
+    # Check for bad statuses
+    response.raise_for_status()
+
+    return WebServerControlResponse(response=response, _data=data)
+
+
+def check_response_data(
+    ws_response: WebServerControlResponse,
+    expected: list[Tuple[str, Type | Any | Tuple[Type]]],
+    optional: Optional[list[Tuple[str, Type | Any | Tuple[Type]]]] = None,
```
zachmprince: Is this what you meant to do? Having `Any` kinda negates any of the other options.
```suggestion
    expected: list[Tuple[str, Type | Tuple[Type]]],
    optional: Optional[list[Tuple[str, Type | Tuple[Type]]]] = None,
```

python/moosecontrol/validation.py
```diff
@@ -0,0 +1,184 @@
...
+
+    Additional Parameters
+    ---------------------
+    optional : list[Tuple[str, Type | Tuple[Type]]]:
+        List of optional key name -> type/types.
+
+    """
+    assert isinstance(ws_response, WebServerControlResponse)
+    if optional is None:
+        optional = []
```
zachmprince, line 153: Why not have the argument default just be a `[]` instead of `None`?

python/moosecontrol/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+        marker = pytest.mark.skip(reason="--no-moose")
+        for item in items:
+            if "moose" in item.keywords:
+                item.add_marker(marker)
+
+
+@pytest.fixture(scope="session")
+def moose_exe(request) -> Optional[str]:
+    """Get the --moose-exe command line argument."""
+    return request.config.getoption("--moose-exe")
```
zachmprince: Would it be possible to test that the executable exists and works here before trying to run all the tests depending on it? Maybe auto mark them as fail?

python/moosecontrol/moosecontrol.py
```diff
@@ -0,0 +1,613 @@
...
+        require_status : Optional[int]
+            Check that the status code is this if set; default 200.
+
+        Returns
+        -------
+        WebServerControlResponse:
+            The combined response, along with the JSON data if any.
+
+        """
+        assert self.initialized
```
zachmprince: With all the new fancy exceptions, could you make one for this?

python/moosecontrol/runners/portrunner.py
```diff
@@ -0,0 +1,85 @@
...
+
+from .baserunner import BaseRunner
+
+DEFAULT_HOST = "http://localhost"
+
+
+class PortRunner(BaseRunner):
+    """Runner that interacts with an already-running webserver over a port."""
+
+    def __init__(self, port: int, host: str = DEFAULT_HOST, **kwargs):
```
zachmprince:
```suggestion
    def __init__(self, port: int, host: str = DEFAULT_HOST, *args, **kwargs):
```

python/moosecontrol/runners/portrunner.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        assert isinstance(port, int)
+        assert isinstance(host, str)
+
+        # The port
+        self._port: int = port
+        # The host
+        self._host: str = host
+
+        super().__init__(**kwargs)
```
zachmprince:
```suggestion
        super().__init__(*args, **kwargs)
```

python/moosecontrol/runners/socketrunner.py
```diff
@@ -0,0 +1,127 @@
...
+
+from .baserunner import BaseRunner
+
+logger = getLogger("SocketRunner")
+
+
+class SocketRunner(BaseRunner):
+    """Runner that interacts with an already-running webserver over a socket."""
+
+    def __init__(self, socket_path: str, **kwargs):
```
zachmprince:
```suggestion
    def __init__(self, socket_path: str, *args, **kwargs):
```

python/moosecontrol/runners/socketrunner.py
```diff
@@ -0,0 +1,127 @@
...
+            See BaseRunner.__init__().
+
+        """
+        # The path to the socket
+        self._socket_path: str = socket_path
+
+        # Whether or not we've used the socket (used in cleanup())
+        self._socket_used: bool = False
+
+        super().__init__(**kwargs)
```
zachmprince:
```suggestion
        super().__init__(*args, **kwargs)
```

python/moosecontrol/runners/socketrunner.py
```diff
@@ -0,0 +1,127 @@
...
+
+        If the path doesn't exist, returns False. If the path
+        exist but isn't a socket, raises FileNotFoundError.
+        """
+        if not os.path.exists(socket_path):
+            return False
+
+        mode = os.stat(socket_path).st_mode
+        if not stat.S_ISSOCK(mode):
+            raise FileNotFoundError(f"Path {socket_path} is not a socket")
```
zachmprince, line 78: Should this just return `False` instead of throwing? Or is something really wrong when the file exists but is not configured for a socket?

python/moosecontrol/runners/socketrunner.py
```diff
@@ -0,0 +1,127 @@
...
+
+        def socket_exists():
+            return self.socket_exists(socket_path)
+
+        if not socket_exists():
+            logger.info(f"Waiting for connection socket {socket_path}...")
+            self.initialize_poll(socket_exists)
+        self._socket_used = True
+
+        logger.info(f"Found connection socket {socket_path}")
```
zachmprince, line 103: Should this just be put in `initialize_start`? Otherwise, `initialize_start` will be run twice: once here and once in `super().initialize()`.

python/moosecontrol/runners/subprocessportrunner.py
```diff
@@ -0,0 +1,122 @@
...
+    """Runner that spawns a MOOSE process and connects to it via a socket."""
+
+    def __init__(
+        self,
+        command: list[str],
+        moose_control_name: str,
+        port: Optional[int] = None,
+        directory: str = DEFAULT_DIRECTORY,
+        use_subprocess_reader: bool = True,
+        **kwargs,
```
zachmprince, line 36:
```suggestion
        *args,
        **kwargs,
```

python/moosecontrol/runners/subprocessportrunner.py
```diff
@@ -0,0 +1,122 @@
...
+            moose_control_name=moose_control_name,
+            directory=directory,
+            use_subprocess_reader=use_subprocess_reader,
+        )
+
+        # Find an available port if one was not provided
+        if port is None:
+            port = PortRunner.find_available_port()
+
+        PortRunner.__init__(self, port=port, **kwargs)
```
zachmprince:
```suggestion
        PortRunner.__init__(self, port=port, *args, **kwargs)
```

python/moosecontrol/runners/subprocessportrunner.py
```diff
@@ -0,0 +1,122 @@
...
+            The data to be passed to /initialize.
+
+        """
+        self.initialize_start()
+
+        if not self.port_is_available(self.port):
+            raise ConnectionRefusedError(f"Port {self.port} is already used")
+
+        # Start the subprocess
+        SubprocessRunnerInterface.initialize(self)
```
zachmprince, line 109: Can this be in `initialize_start` instead?

python/moosecontrol/runners/socketrunner.py
```diff
@@ -0,0 +1,127 @@
...
+
+"""Defines the SocketRunner."""
+
+import os
+import stat
+from logging import getLogger
+
+from moosecontrol.requests_unixsocket import Session
+
+from .baserunner import BaseRunner
```
zachmprince:
```suggestion
from moosecontrol.runners.baserunner import BaseRunner
```

python/moosecontrol/runners/utils/timedpoller.py
```diff
@@ -0,0 +1,191 @@
...
+    >>>> should_exit = lambda: True # fake criteria
+    >>>>
+    >>>> # Poll until should_exit is true, which can be done
+    >>>> # multiple times. If the timeout time is reached,
+    >>>> # these calls to poll() will raise TimedPollTimeout
+    >>>> helper.poll(should_exit)
+    >>>> helper.poll(should_exit)
+    >>>>
+    >>>> # Finalize timing if it succeeds
+    >>>> helper.end()
```
zachmprince:
```suggestion
    >>> helper = TimedPoller(poll_time=0.1, timeout_time=10)
    >>> helper.start()
    >>>
    >>> # Develop some criteria on when to exit
    >>> should_exit = lambda: True # fake criteria
    >>>
    >>> # Poll until should_exit is true, which can be done
    >>> # multiple times. If the timeout time is reached,
    >>> # these calls to poll() will raise TimedPollTimeout
    >>> helper.poll(should_exit)
    >>> helper.poll(should_exit)
    >>>
    >>> # Finalize timing if it succeeds
    >>> helper.end()
```

python/moosecontrol/runners/interfaces/subprocessrunnerinterface.py
```diff
@@ -0,0 +1,237 @@
...
+            logger.info("Reader thread has ended")
+
+    @abstractmethod
+    def get_additional_command(self) -> list[str]:
+        """
+        Get the additional arguments to start the subprocess with.
+
+        Must be overridden.
+        """
+        raise NotImplementedError  # pragma: no cover
```
zachmprince, line 186: I don't think it is necessary for this to be abstract. You could just return `[]`. Therefore, the class probably doesn't need to be abstract.

python/moosecontrol/tests/common.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def tearDown(self):
+        """
+        Tear down a test.
+
+        Raises an exception if warning logs exist
+        and they are not allowed and cleans up
+        the temporary directory.
+        """
+        super().setUp()
```
zachmprince:
```suggestion
        super().tearDown()
```

python/moosecontrol/tests/common.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def assert_log_size(self, num: int):
+        """Assert that there are exactly the given number of logs."""
+        records = self._caplog.records
+        num_records = len(records)
+        if num_records != num:
+            raise AssertionError(
+                f"Num logs {num_records} != {num}; "
+                f"present logs:\n{self.printable_logs()}"
+            )
```
zachmprince:
```suggestion
        self.assertEqual(
            num_records,
            num,
            f"Num logs {num_records} != {num}; present logs:\n{self.printable_logs()}",
        )
```

python/moosecontrol/tests/common.py
```diff
@@ -0,0 +1,346 @@
...
+            self.assertEqual(record.name, name)
+        self.assertEqual(record.message, message)
+        self.assertEqual(record.levelname, levelname)
+
+    def assert_in_log(
+        self,
+        message: str,
+        name: Optional[str] = None,
+        levelname: str = "INFO",
+        after_index: Optional[int] = None,
```
zachmprince: I think you make this default to -1.

python/moosecontrol/tests/common.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def assert_no_warning_logs(self):
+        """Assert that no logs of level WARNING or higher exist."""
+        for record in self._caplog.records:
+            if record.levelname == "WARNING":
+                raise AssertionError(
+                    f"Warning found in logs:\n"
+                    f"Name: {record.name}\n"
+                    f'Message: "{record.message}"'
+                )
```
zachmprince:
```suggestion
        for record in self._caplog.records:
            self.assertNotEqual(
                record.levelname,
                "WARNING",
                f"Warning found in logs:\n"
                f"Name: {record.name}\n"
                f'Message: "{record.message}"',
            )
```

python/moosecontrol/tests/common.py
```diff
@@ -0,0 +1,346 @@
...
+    response.url = url
+    if data is not None:
+        response.headers = {"content-type": "application/json"}
+        if isinstance(response, Response):
+            response._content = dumps(data).encode("utf-8")
+        else:
+            response.json.return_value = data
+
+
+@staticmethod
```
zachmprince:
```suggestion
```

python/moosecontrol/tests/common.py
```diff
@@ -0,0 +1,346 @@
...
+    **kwargs : dict
+        See set_fake_response().
+    """
+    response = MagicMock()
+    response.__enter__.return_value = response
+    set_fake_response(response, **kwargs)
+    return response
+
+
+@staticmethod
```
zachmprince:
```suggestion
```

python/moosecontrol/tests/common.py
```diff
@@ -0,0 +1,346 @@
...
+  type = Steady
+[]
+"""
+
+# Keyword arguments to pass to the BaseRunner for live tests
+LIVE_BASERUNNER_KWARGS = {
+    "initialize_timeout": 60,
+    "poll_time": 0.01,
+    "poke_poll_time": 0.01,
+}
```
zachmprince: This will allow the linter to pickup the description of the attribute.
```suggestion
LIVE_BASERUNNER_KWARGS = {
    "initialize_timeout": 60,
    "poll_time": 0.01,
    "poke_poll_time": 0.01,
}
"Keyword arguments to pass to the BaseRunner for live tests"
```

python/moosecontrol/tests/test_moosecontrol.py
```diff
@@ -0,0 +1,709 @@
...
+import os
+from copy import deepcopy
+from getpass import getuser
+from numbers import Number
+from socket import getfqdn
+from typing import Any, Optional, Tuple
+from unittest.mock import MagicMock, PropertyMock, patch
+
+import pytest
+from common import (
```
zachmprince: I would recommend using a relative path here as to clarify where this module is coming from.

python/moosecontrol/tests/test_moosecontrol.py
```diff
@@ -0,0 +1,709 @@
...
+            )
+        self.assertEqual(control.runner, runner)
+        self.assertEqual(control.poll_time, control.runner.poll_time)
+        self.assertFalse(control.initialized)
+
+    def test_init_quiet_and_verbose(self):
+        """Test __init__() with quiet and verbose set."""
+        with self.assertRaises(ValueError) as e:
+            MooseControl(BaseRunnerTest(), quiet=True, verbose=True)
+        self.assertEqual(str(e.exception), "Cannot set quiet=True and verbose=True")
```
zachmprince:
```suggestion
        with self.assertRaisesRegex(
            ValueError, "Cannot set quiet=True and verbose=True"
        ):
            MooseControl(BaseRunnerTest(), quiet=True, verbose=True)
```

python/moosecontrol/tests/test_moosecontrol.py
```diff
@@ -0,0 +1,709 @@
...
+        class MooseControlDerived(MooseControl):
+            def additional_initialize_data(self):
+                return {bad_key: "foo"}
+
+        with self.assertRaises(KeyError) as e:
+            MooseControlDerived(BaseRunnerTest()).get_initialize_data()
+        self.assertEqual(
+            str(e.exception),
+            f"'Cannot override entry \"{bad_key}\" in additional_initialize_data()'",
+        )
```
zachmprince:
```suggestion
        with self.assertRaisesRegex(
            KeyError,
            f"'Cannot override entry \"{bad_key}\" in additional_initialize_data\(\)'",
        ):
            MooseControlDerived(BaseRunnerTest()).get_initialize_data()
```

python/moosecontrol/tests/test_moosecontrol.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def test_initialize(self):
+        """
+        Test initialize().
+
+        Should call the underlying runner's initialize() and wait()
+        and query the daat from initialize.
```
zachmprince:
```suggestion
        and query the data from initialize.
```

python/moosecontrol/tests/test_moosecontrol.py
```diff
@@ -0,0 +1,709 @@
...
+            self.assertEqual(value, get_value)
+            self.assertGetPaths(["waiting", "get/dt"])
+            self.assert_log_size(1)
+            self.assert_log_message(0, "Getting simulation timestep", levelname="DEBUG")
+
+    def test_set_controllable_scalar_bad_type(self):
+        """Test set_controllable_scalar() with a bad type."""
+        with self.assertRaises(TypeError) as e:
+            self.control.set_controllable_scalar("unused", "unused", (Number,), "foo")
+        self.assertEqual(str(e.exception), "Type str is not of allowed type(s) Number")
```
zachmprince:
```suggestion
        with self.assertRaisesRegex(
            TypeError, r"Type str is not of allowed type\(s\) Number"
        ):
            self.control.set_controllable_scalar("unused", "unused", (Number,), "foo")
```

python/moosecontrol/tests/test_moosecontrol.py
```diff
@@ -0,0 +1,709 @@
...
+        with self.assertRaises(TypeError) as e:
+            self.control.set_controllable_vector(
+                "unused",
+                "unused",
+                (Number,),
+                ["foo"],
+            )
+        self.assertEqual(
+            str(e.exception), "At index 0: type list is not of allowed type(s) Number"
+        )
```
zachmprince:
```suggestion
        with self.assertRaisesRegex(
            TypeError, r"At index 0: type list is not of allowed type\(s\) Number"
        ):
            self.control.set_controllable_vector(
                "unused",
                "unused",
                (Number,),
                ["foo"],
            )
```

python/moosecontrol/tests/test_moosecontrol.py
```diff
@@ -0,0 +1,709 @@
...
+    def test_set_realeigenmatrix_non_1D_2D(self):
+        """Tests set_realeigenmatrix() when the value is not 1D or 2D."""
+        value = [[[1]]]
+        with self.assertRaises(ValueError) as e:
+            self.run_test_set_controllable(
+                value,
+                "RealEigenMatrix",
+                "realeigenmatrix",
+            )
+        self.assertEqual(str(e.exception), "Value not convertible to a 1- or 2-D array")
```
zachmprince:
```suggestion
        with self.assertRaisesRegex(
            ValueError, "Value not convertible to a 1- or 2-D array"
        ):
            self.run_test_set_controllable(
                value,
                "RealEigenMatrix",
                "realeigenmatrix",
            )
```

python/moosecontrol/tests/test_runners_baserunner.py
```diff
@@ -0,0 +1,500 @@
...
+from moosecontrol.runners import BaseRunner
+from moosecontrol.runners.baserunner import (
+    DEFAULT_INITIALIZE_TIMEOUT,
+    DEFAULT_POKE_POLL_TIME,
+    DEFAULT_POLL_TIME,
+)
+from requests import Session
+from requests.exceptions import ConnectionError
+
+setup_moose_python_path()
```
zachmprince: Doesn't this need to be before the `moosecontrol` imports?

python/moosecontrol/tests/test_runners_baserunner.py
```diff
@@ -0,0 +1,500 @@
...
+
+
+def patch_baserunner(name: str, **kwargs):
+    """Patch a method in the BaseRunner."""
+    return patch(f"moosecontrol.runners.BaseRunner.{name}", **kwargs)
+
+
+def check_baserunner_cleanup_live(
+    test: MooseControlTestCase,
+    runner: BaseRunner,
```
zachmprince: This argument isn't used.

python/moosecontrol/tests/test_runners_baserunner.py
```diff
@@ -0,0 +1,500 @@
...
+            poll_time=poll_time, initialize_timeout=initialize_timeout
+        )
+        runner.initialize_start()
+        with self.assertRaises(InitializeTimeout) as e_cm:
+            runner.initialize_poll(lambda: False)
+        self.assertEqual(
+            str(e_cm.exception),
+            f"Initialization timed out after "
+            f"{runner._initialize_poller.total_time:.2f} seconds",
+        )
```
zachmprince:
```suggestion
        with self.assertRaisesRegex(
            InitializeTimeout,
            "Initialization timed out after "
            f"{runner._initialize_poller.total_time:.2f} seconds",
        ):
            runner.initialize_poll(lambda: False)
```

python/doc/content/python/moosecontrol/index.md
```diff
@@ -0,0 +1,158 @@
+# MooseControl
```
zachmprince, line 1: This documentation is amazing. The only thing I would like to see is how users would play with the logger. Particularly, how would they configure their own log level for the control and subprocess and how would they pipe it to a file.

framework/src/actions/StartWebServerControlAction.C
```diff
@@ -0,0 +1,36 @@
...
+  : Action(parameters)
+{
+}
+
+void
+StartWebServerControlAction::act()
+{
+  for (auto & control_ptr : _problem->getControlWarehouse().getObjects())
+    if (auto wsc_ptr = dynamic_cast<WebServerControl *>(control_ptr.get()))
+      wsc_ptr->startServer({});
```
zachmprince, line 35: Can there be more than one `WebServerControl`? This seems like a good opportunity to throw something for that.

zachmprince (changes requested):
It's so beautiful!

python/moosecontrol/runners/socketrunner.py
loganharbour: If the file exists but isn't a socket we can never connect to it, so there's nothing else to do

python/moosecontrol/runners/interfaces/subprocessrunnerinterface.py
loganharbour: It doesn't really make sense to use this class on its own, but I guess

python/moosecontrol/validation.py
loganharbour: Any is on purpose here. While it negates the other type, the literal value `Any` in the list means allow any type.

python/moosecontrol/validation.py
loganharbour: https://docs.astral.sh/ruff/rules/mutable-argument-default/

python/moosecontrol/runners/socketrunner.py
loganharbour: `initialize_start()` is fine to run multiple times. If already started, it does nothing.

python/moosecontrol/runners/subprocessportrunner.py
loganharbour: All start does is start timing for the timeout. It needs to be started before doing anything that could timeout. Doesn't make sense to override initialize_start

python/moosecontrol/runners/interfaces/subprocessrunnerinterface.py
loganharbour: Actually yeah, you always need to specify an override to how to connect to the server in the spanwed process. It should always be set. If someone wants to not set it, they can have it really return `[]` in derived if desired.

python/moosecontrol/tests/common.py
loganharbour: I don't like this. There's no need to check it if it's not set and checking for `-1` as being not set is more confusing than checking if not none.

python/moosecontrol/tests/common.py
loganharbour: I don't really like changing these, as throwing `AssertionError` is exactly what unittest does. And it adds in additional context in the message that isn't necessary. But I'm too lazy to fight.

python/doc/content/python/moosecontrol/index.md
loganharbour: Working on this

framework/src/actions/StartWebServerControlAction.C
loganharbour: Working on this

python/moosecontrol/runners/socketrunner.py
zachmprince: Wouldn't this be the same sequence of events for if the file didn't exist at all, where you just return `False`?

python/moosecontrol/tests/common.py
zachmprince: You wouldn't need to do a check at all, the `after_index` would end up just being 0, including all of the responses as you have right now with `None`.

python/moosecontrol/runners/socketrunner.py
loganharbour: No because then it would keep checking the same file over and over again, even though it's never gonna change. We don't wanna wait 60 seconds for a timeout if the file isn't gonna change

framework/src/actions/StartWebServerControlAction.C
loganharbour: Actually, this is very valid. You could have multiple if you really wanted to. Maybe a derived one listening on a different port. It probably won't happen, but it's realistic.

python/doc/content/python/moosecontrol/index.md
loganharbour: I'm not quite sure I want to address this yet. I would still like to do more work on setting up the logger configuration to make it easier. The vebose and quiet is just the basic config

zachmprince (approved):

PR #31763, opened 2025-10-21
Drop includes of libmesh auto_ptr.h compatibility header

jwpeterson:
This file is almost empty in current libmesh master, and will soon be completely empty/removed. No files in MOOSE should be including it any longer.

Part of an ongoing deprecated code cleanup, (see also: #31732).


jwpeterson:
FYI @roystgnr, hopefully this change is uncontroversial.

roystgnr (approved):

PR #31781, opened 2025-10-24
Ignore untracked changes in mfem/conduit submodules

nmnobre:
To be stolen by someone who can actually edit `.gitmodules`.

## Reason
When we build and install conduit and mfem with default settings, we populate build and install directories within the respective submodule trees. A `git status`  will then show `modified:   framework/contrib/conduit (untracked content)` which is slightly distracting (so much red!).

## Design
Simply ignore untracked changed in `.gitmodules` for these two submodules. I've also changed the urls for all repos hosted on GitHub for consistency.

## Impact
Fewer git warnings (less red!).


loganharbour (approved):

loganharbour:
@nmnobre you should be good now

PR #31784, opened 2025-10-24
Excluding invalid ids when generating positions with an EEID

### LINKED ISSUE #31783: Invalid IDs of an EEID should not be considered when used for generating positions
### Motivation

It could be possible users only assign valid ids to an EEID (extra element integer) for part of a domain. When it is used for generating a position in `ElementGroupCentroidPositions`, we should not generate a position for it.

### Design

To add a one line protection in the code checking whether the retrieved ids are valid.

### Impact

Minor cleanup. Hopefully users is not relying the wrong behavior.

YaqiWang:
Closes #31783.

### DIFF
--- a/framework/include/positions/ElementGroupCentroidPositions.h
+++ b/framework/include/positions/ElementGroupCentroidPositions.h
@@ -44,8 +44,8 @@ public:
   /// The indices of the extra element ids to use to make groups
   std::vector<unsigned int> _extra_id_indices;
   /// The particular extra id values, for each extra id of interest
-  std::vector<std::vector<unsigned int>> _extra_id_group_indices;
+  std::vector<std::vector<dof_id_type>> _extra_id_group_indices;
 
 private:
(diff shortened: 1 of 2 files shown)


lindsayad (changes requested):
It seems like you need to convert a few more types to `dof_id_type`, e.g. the local variable `ids` should be `std::set<dof_id_type> ids;` and `_extra_id_group_indices` should be `std::vector<std::vector<dof_id_type>>`

YaqiWang:
> It seems like you need to convert a few more types to `dof_id_type`, e.g. the local variable `ids` should be `std::set<dof_id_type> ids;` and `_extra_id_group_indices` should be `std::vector<std::vector<dof_id_type>>`

Yes. Should have been addressed.

lindsayad:
looks like you'll need to change the parameter type as well

lindsayad (approved):

PR #31813, opened 2025-10-30
Prepare MOOSE for MetaPhysicL 2.0

### LINKED ISSUE #31812: Prepare MOOSE for MetaPhysicL 2.0
### Motivation

MetaPhysicL 2.0 supports compilation with Kokkos and also introduces some API breaking changes

### Design

Protect code with MetaPhysicL versions checks

### Impact

Allow continuous integration to be green while the MetaPhysicL update makes its way through libMesh to MOOSE

lindsayad:
Closes #31812

### DIFF
--- a/framework/include/limiters/Limiter.h
+++ b/framework/include/limiters/Limiter.h
@@ -170,10 +170,11 @@ public:
             const VectorValue<T> * grad_phi_downwind,
             const RealVectorValue & dCD) const
   {
+    using std::max;
     const auto grad_elem = (*grad_phi_upwind) * dCD;
     const auto grad_face = (*grad_phi_downwind) * dCD;
     const auto grad_ratio = grad_elem / (grad_face + 1e-10);
-    return std::max(2.0 * grad_ratio - 1.0, 0.0);
+    return max(2.0 * grad_ratio - 1.0, 0.0);
   };
 
   /**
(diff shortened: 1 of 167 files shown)


dschwen (approved):

dschwen:
So metaphysicl 2.0 does not overload members of the `std` namespace anymore? Are we going to drop all explicit `std::` qualifiers? Are we going to be one of those `using std;` projects? ;-)

lindsayad:
> So metaphysicl 2.0 does not overload members of the `std` namespace anymore?

Correct, we won't be violating the standard anymore 😄 

> Are we going to drop all explicit `std::` qualifiers?

We'll need to in places where we're using dual numbers. People can keep it around where the types are builtins as I did in many places in this PR

> Are we going to be one of those `using std;` projects? ;-)

`using std::foo` or `using namespace std`? If the former, I guess we will be partially. I didn't know there was a reputation for projects `using std` haha

roystgnr:
> I didn't know there was a reputation for projects `using std` haha

`using namespace std;` in a source file or a private header you only include from source files is IMHO a heuristic for "I uploaded my class project to github but I'm never looking it again and neither should you" code ... but it's surprisingly common even in high quality codebases; it's all over Eigen tests and benchmarks, for instance.

`using namespace std;` in global scope in a header file you expose to users is evil ... I say, as I grep the MOOSE codebase and find `modules/misc/include/postprocessors/GeneralSensorPostprocessor.h`.  At least that's only going into two misc postprocessor .C files ... but I *am* seeing .edu in the commit data, so maybe double-check my heuristic before using those.

PR #31817, opened 2025-10-30
csv sampler code and test 

### LINKED ISSUE #31810: CSVSamplerReporter
### Motivation

I want to read csv file of xyz coord and sample variable values at those locations.

### Design

Follow the pattern used in optimization and line value sampler

### Impact

Easy sampling data from simulations.

BozoVazic:
closes #31810


lindsayad:
@GiudGiud I feel like you've done a decent amount with csv. Would you mind looking at this?

GiudGiud (changes requested):
you can already do this with these two objects:
https://mooseframework.inl.gov/source/positions/FilePositions.html to load the points and
https://mooseframework.inl.gov/source/vectorpostprocessors/PositionsFunctorValueSampler.html to sample the functor (=variable, function, spatial user object, functor matprop)

lynnmunday:
@BozoVazic Move this PR to falcon.  This will still be easier to use with comma separated files and the sorting vector.  Maybe we can add those parts to the filePositions later.

PR #31821, opened 2025-10-30
Exception coherence before comm in ElementJacobianDamper

lindsayad:
Refs discussion on https://github.com/idaholab/moose/discussions/31700


GiudGiud (approved):
I see why you did this and it's the correct fix.

this sucks long term. We basically need to compile two lists:
- routines that can throw
- routines that need to be in sync
and add these catch everywhere one precedes the other

PR #31832, opened 2025-11-03
Rewrite resultsstore test loading strategy, switch to pytest

### LINKED ISSUE #31831: Speed up resultsstore serialization and deserialization
### Motivation

With the work that @maythuwin1313 is doing on developing a web application for results, it would be useful if we could more quickly serialize and deserialize test results for the purpose of caching.

### Design

Add the option to disable type validation checking in object construction in the stored results objects. Don't build each test up front unless explicitly desired.

### Impact

User experience should not change. A few more options will be added to the reader.

loganharbour:
Closes https://github.com/idaholab/moose/issues/31831

This is in support of performance improvements for the web application that will parse these databases.

- Changes all arguments for getting a test to a `TestName` instead of two strings (folder and test name)
- Add `query_test` for getting a test if it is available, otherwise None
- Do not load tests with each result by default, moving test loading to a `ResultCollection`
- Optimize serialize and deserialize methods
- Don't connect to the client/database unless needed
- Rewrite unit tests and add pytest support


python/TestHarness/resultsstore/reader.py
```diff
@@ -9,160 +9,236 @@
...
+        """
+        if self._client is None:
+            auth = self._authentication
+            assert isinstance(auth, Authentication)
+            self._client = pymongo.MongoClient(
+                auth.host,
+                username=auth.username,
+                password=auth.password,
+                timeoutMS=self._timeout * 1000,
+            )
```
zachmprince, line 178: Does this need to pass in the `port` attribute from the `auth` as well?

python/TestHarness/resultsstore/reader.py
```diff
@@ -201,85 +324,85 @@ def iterateLatestPushResults(
...
+                if i > 1:
+                    assert result.id < self._latest_push_results[i - 1].id
 
-            if i < len(self._latest_push_results):
-                yield self._latest_push_results[i]
+                yield result
                 i += 1
+            # No more results, and we didn't find any more
             else:
+                self._found_final_push_event = True
```
zachmprince, line 350: Is this supposed to be reset somewhere? Seems like the loop would break early if the the iterator finished once and was called again.

python/TestHarness/resultsstore/reader.py
```diff
@@ -201,85 +324,85 @@ def iterateLatestPushResults(
...
         # Search for the value
-        filter = {index: {"$eq": value}}
-        data = self._db.results.find_one(filter, sort=self.mongo_sort_id)
+        docs = self._find_results({index: {"$eq": value}}, limit=1)
 
         # No such thing
-        if data is None:
+        if not docs:
             cache[value] = None
             return None
```
zachmprince, line 395: I don't think this cache is ever utilized. Since the value is `None` the logic above will think this value was never cached, which indeed it is here.

python/TestHarness/resultsstore/resultcollection.py
```diff
@@ -0,0 +1,456 @@
...
+    def __init__(
+        self,
+        results: list[StoredResult],
+        database_getter: Callable[[], Database],
+    ):
+        """
+        Initialize state.
+
+        Arguments:
+        ---------
```
zachmprince: For some reason, these arguments are not showing up in the linting, but if it is changed to
```suggestion
        Parameters
        ----------
```
it does show up.

python/TestHarness/resultsstore/testdatafilters.py
```diff
@@ -0,0 +1,57 @@
...
+    STATUS = "status"
+
+    """Load the tester enter in tests."""
+    TESTER = "tester"
+
+    """Load the timing entry in tests."""
+    TIMING = "timing"
+
+    """Load the validation entry in tests."""
+    VALIDATION = "validation"
```
zachmprince: The docstrings should go below the attribute
```suggestion
    ALL = "all"
    """Load all data in tests."""

    HPC = "hpc"
    """Load HPC data in tests."""

    STATUS = "status"
    """Load the status entry in tests."""

    TESTER = "tester"
    """Load the tester enter in tests."""

    TIMING = "timing"
    """Load the timing entry in tests."""

    VALIDATION = "validation"
    """Load the validation entry in tests."""
```

python/TestHarness/resultsstore/testdatafilters.py
```diff
@@ -0,0 +1,57 @@
...
+
+def filter_as_iterable(
+    test_filter: Union[Iterable[TestDataFilter], TestDataFilter],
+) -> Iterable[TestDataFilter]:
+    """
+    If applicable, convert a single filter into an iterable filter.
+
+    Useful for methods that can take a single filter or muliple filters.
+    """
+    return [test_filter] if isinstance(test_filter, TestDataFilter) else test_filter
```
zachmprince: Should there be an `assert` to make sure that the it is a iterable class if it is not a `TestDataFilter`?

python/TestHarness/resultsstore/utils.py
```diff
@@ -13,38 +12,98 @@
 import json
 import zlib
 from dataclasses import dataclass
-from typing import Any, Iterator
+from typing import Any, Iterator, Optional, Tuple, Type
+
+
+@dataclass
+class MongoPath:
```
zachmprince: Does this need to be a `dataclass`?

python/TestHarness/resultsstore/utils.py
```diff
@@ -13,38 +12,98 @@
...
+
+        """
+        return MongoPath(*self.path, *args)
 
 
 @dataclass(frozen=True)
 class TestName:
     """Name for a test (folder and name)."""
 
+    __test__ = False  # prevents pytest collection
```
zachmprince, line 67: You can configure your `pytest.ini` to only collect classes that inherit from `unittest.TestCase`

python/TestHarness/resultsstore/utils.py
```diff
@@ -54,30 +113,87 @@ def results_test_entry(results: dict, name: TestName) -> Any:
...
 
 
 def results_set_test_value(results: dict, name: TestName, value: Any):
     """Set the test value for a given test in TestHarness JSON results."""
     results["tests"][name.folder]["tests"][name.name] = value
 
 
+@dataclass
+class ResultsFolderIterator:
+    """Iterator for a test folder in TestHarness results JSON output."""
```
zachmprince, line 139: This really isn't an iterator anymore, huh? Do you still like the name?

python/TestHarness/tests/resultsstore/common.py
```diff
@@ -0,0 +1,349 @@
...
+
+def base_civet_env() -> Tuple[str, dict]:
+    """Build a random base CIVET environment for testing."""
+    # Dummy sha to use for the base commit
+    base_sha = random_git_sha()
+    # Dummy civet environment
+    env = {
+        "APPLICATION_REPO": APPLICATION_REPO,
+        "CIVET_BASE_SHA": random_git_sha(),
+        "CIVET_BASE_SSH_URL": "git@github.com:idaholab/moose.git",
```
zachmprince: Would this be valid?
```suggestion
        "CIVET_BASE_SSH_URL": APPLICATION_REPO + ".git",
```

zachmprince (changes requested):
Beautiful testing setup.

python/TestHarness/resultsstore/reader.py
loganharbour: It's supposed to exit early if that's the case. If we find the end (the oldest result), there's never anything else to find.

python/TestHarness/resultsstore/utils.py
loganharbour: It is an iterator? It represents a single value in something to be iterated through

zachmprince (approved):

loganharbour:
Will require a patch to openpronghorn; failure is related.

PR #31857, opened 2025-11-06
Libmesh exception handling

roystgnr:
Refs #31856 - this is the "medium-term" step need to make MOOSE compatible with future fixes at the libMesh level.

MOOSE will now catch std::exception and check what() for keywords like Jacobian and singular to identify recoverable exceptions from degenerate maps regardless of libMesh exception type.

While I was at it I noticed a `MeshDiagnosticsGenerator` exception-catching issue for which @GiudGiud might want to review my fix, as well as a couple places where we were using a `try`/`catch` as a bit of a hack where an `if()` would do.

### DIFF
--- a/framework/include/loops/ComputeFVFluxThread.h
+++ b/framework/include/loops/ComputeFVFluxThread.h
@@ -329,7 +329,7 @@ ThreadedFaceLoop<RangeType>::operator()(const RangeType & range, bool bypass_thr
       // Continue if we find a libMesh degenerate map exception, but
       // just throw for any real error
       if (!strstr(e.what(), "Jacobian") && !strstr(e.what(), "singular"))
-        throw e;
+        throw;
 
       mooseException("We caught a libMesh exception: ", e.what());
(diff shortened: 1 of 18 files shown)


lindsayad (changes requested):
I made a change in the last week to `ElementJacobianDamper.C` introducing another catch block for element Jacobian issues. That could also use your clean up to use `strstr` and to not avoid slicing the derived type information from `e`. You might have to rebase to see it

GiudGiud (approved):
bug fix in counting negative vol elements looks good to me
other changes too

lindsayad (approved):

loganharbour:
This broke next: https://civet.inl.gov/job/3368192/

roystgnr:
I'll see if I can reproduce

roystgnr:
I can replicate this, at least; I'm still trying to figure out why it fails in parallel but not serial, but I should have a fix shortly.

PR #31858, opened 2025-11-06
Kokkos UserObjects and Postprocessors System

NamjaeChoi:
Refs #30655.


framework/include/kokkos/base/KokkosUtils.h
```diff
@@ -104,7 +104,7 @@ choleskySolve(Real * const A, Real * const x, Real * const b, const unsigned int
     Real sum = b[i];
 
     for (unsigned int j = i + 1; j < n; ++j)
-      sum -= A[i + n * j] * b[j];
+      sum -= A[i + n * j] * x[j];
```
NamjaeChoi, line 105: I must be drunk when implementing this. And I have no idea why this worked with the element auxkernels tested thus far.

NamjaeChoi:
@lindsayad The code is pretty much ready for review. Looking for some good UserObject tests. UserObjects are really exotic and not easy to find something that can be easily ported except for some dummy test objects.

framework/doc/content/syntax/KokkosUserObjects/index.md
```diff
@@ -0,0 +1,124 @@
...
+KOKKOS_FUNCTION void execute(Datum & datum) const;
+```
+
+For other CPU APIs, Kokkos-MOOSE user objects share the same interface with the original MOOSE objects except `threadJoin()`, which is undefined in Kokkos-MOOSE user objects as they have separate parallel loops dispatched by Kokkos.
+
+!alert note
+Kokkos-MOOSE user objects are always executed after the original MOOSE user objects, so the original MOOSE user objects cannot depend on the Kokkos-MOOSE user objects.
+
+!alert note
+Kokkos-MOOSE user objects are executed without any predefined order.
```
loganharbour: Do we have a way to alert the user when they to couple to things that the dependencies aren't tracked?

framework/include/problems/FEProblemBase.h
```diff
@@ -1340,20 +1354,59 @@ class FEProblemBase : public SubProblem, public Restartable
...
+                                   const std::string & name,
+                                   InputParameters & parameters);
+
+  /**
+   * Get the Kokkos user object by its name
+   * @param name The name of the Kokkos user object being retrieved
+   * @return Const reference to the Kokkos user object
+   */
+  template <class T>
+  T & getKokkosUserObject(const std::string & name) const
```
loganharbour: I'm not the biggest fan that this is const and we're getting a non-const value. Do we need non-const values? We made this mistake a while ago in moose and we can't come back from it. Would be nice to not make it here.

Also - while I'm here - as I implement getters like this, I've been doing it like they do in libmesh.

You have a `queryXXX` method that gives you a pointer (that can be null).  Then you have a `getXXX` method that calls `queryXXX`, errors if it's null and otherwise dereferences. and then the `hasXXX` method just calls `queryXXX` and checks if it is null. Means that all of your actual checking will live in a single function, the `queryXXX` method.

framework/include/problems/FEProblemBase.h
```diff
@@ -1340,20 +1354,59 @@ class FEProblemBase : public SubProblem, public Restartable
...
+   */
+  bool hasKokkosUserObject(const std::string & name) const;
+#endif
+
+  /**
+   * Check for name collision between different user objects
+   * @param name The object name being added
+   * @param type The object type being added
+   */
+  void checkUserObjectNameCollision(const std::string & name, const std::string & type) const;
```
loganharbour, line 1401: Is there not a better way to do this?

framework/src/actions/AddUserObjectAction.C
```diff
@@ -27,5 +27,10 @@ AddUserObjectAction::AddUserObjectAction(const InputParameters & params) : Moose
 void
 AddUserObjectAction::act()
 {
-  _problem->addUserObject(_type, _name, _moose_object_pars);
+#ifdef MOOSE_KOKKOS_ENABLED
+  if (_moose_object_pars.isParamValid(MooseBase::kokkos_object_param))
```
loganharbour: Feel free to put this logic for this parameter in the parameters class. Something like `_moose_object_pars.isKokkosObject()` or something if you want

framework/src/kokkos/postprocessors/KokkosElementPostprocessor.K
```diff
@@ -0,0 +1,31 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#include "KokkosElementPostprocessor.h"
+
+namespace Moose
+{
+namespace Kokkos
+{
```
loganharbour: Still really not a fan of
```
namespace Moose::Kokkos
{
```

huh? :D

framework/src/kokkos/problems/KokkosFEProblemBase.K
```diff
@@ -344,3 +402,159 @@ FEProblemBase::reinitKokkosMaterials()
...
 }
+
+void
+FEProblemBase::computeKokkosUserObjectsInternal(const ExecFlagType & type,
+                                                const Moose::AuxGroup & group,
+                                                TheWarehouse::Query & primary_query)
+{
+  try
+  {
+    TIME_SECTION("computeUserObjects", 1, "Computing User Objects");
```
loganharbour: This named section already exists?

test/include/kokkos/auxkernels/KokkosCoupledAux.h
```diff
@@ -0,0 +1,52 @@
...
+public:
+  static InputParameters validParams();
+
+  KokkosCoupledAux(const InputParameters & parameters);
+
+  KOKKOS_FUNCTION Real computeValue(const unsigned int qp, Datum & datum) const;
+
+protected:
+  const Real _value;
+  const int _operator;
```
loganharbour, line 26: Please add some docstrings here. `int` for operator isn't immediately apparent

loganharbour (changes requested):
I'm still a bit concerned about the user-object dependency stuff. is it possible to couple to other UOs even though we can't do anything about execution ordering?

framework/doc/content/syntax/KokkosUserObjects/index.md
NamjaeChoi: This sentence was misleading and was changed to the following before you submitted your review:
`Different types of Kokkos-MOOSE user objects are executed without any predefined order (dependencies are still respected).`
Does this resolve your issue?

framework/include/problems/FEProblemBase.h
NamjaeChoi: I think `const` is fine.

> You have a queryXXX method that gives you a pointer

Isn't this what `TheWarehouse` is doing?

framework/include/problems/FEProblemBase.h
NamjaeChoi: Open to suggestions

framework/src/kokkos/postprocessors/KokkosElementPostprocessor.K
NamjaeChoi: Habit from C++11. All changed

loganharbour (approved):
No concerns from me. The UO collision stuff is very dirty, but we should probably make a change in the warehouse to resolve that

NamjaeChoi:
@lindsayad I'll wait for another week in case you have comments (since this is not a small PR) and then merge it.

loganharbour (changes requested):
Can you please add an entry to the newsletter for this? Please feel free to create a new section for kokkos content. It'd be nice for us to more simply share the progress on this with the community

NamjaeChoi:
Wanna re-approve? Since Alex won't be reviewing for a while I think I will just merge

loganharbour (approved):

PR #31880, opened 2025-11-10
Fix `next` MeshDiagnosticsGenerator parallel+dbg case

### LINKED ISSUE #31879: Regression in MeshDiagnosticsGenerator in `next`
### Bug Description

The updates in #31857 missed an important exception type, an assertion that can be thrown in `dbg` mode from `inverse_map`, including in FE side-reinit on 2D elements.

### Steps to Reproduce

See https://civet.inl.gov/job/3368192/ - one entry in the regression test suite for MeshDiagnosticsGenerator now breaks in parallel in `dbg` mode

### Impact

This is blocking `next`->`devel` merges; fix to come shortly.

### [Optional] Diagnostics

_No response_

roystgnr:
This fixes #31879 for me, and extends the fix to any other code blocks that might be hit by the same issue.

### DIFF
--- a/framework/include/loops/ComputeFVFluxThread.h
+++ b/framework/include/loops/ComputeFVFluxThread.h
@@ -328,7 +328,8 @@ ThreadedFaceLoop<RangeType>::operator()(const RangeType & range, bool bypass_thr
     {
       // Continue if we find a libMesh degenerate map exception, but
       // just throw for any real error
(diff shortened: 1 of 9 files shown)


GiudGiud (approved):

roystgnr:
Force-push on that last commit to hide my shame, but basically I got cut-and-paste happy and accidentally reverted the proper negation on an if condition.  This should be good to go once CI is happy.

PR #31896, opened 2025-11-12
Kokkos VectorPostprocessors and Reporters

NamjaeChoi:
Refs #30655.


framework/doc/content/syntax/KokkosUserObjects/index.md
```diff
@@ -36,27 +36,14 @@ A Kokkos-MOOSE general user object should be registered with the standard `regis
...
+Postprocessors, vector postprocessors, and reporters often perform calculations that aggregate data.
+Such aggregation calculation is also known as reduction operation.
+If your calculation is considered more suitable for a reduction operation than an ordinary parallel operation, you can register your user object with `registerKokkosReducer()` instead of `registerKokkosUserObject()`.
+This makes your user object a `Reducer` object, which uses a different parallelization scheme and interface from regular parallel objects.
 
-The aggregate calculations, also known as reduction operations, are not trivial on GPU which performs a massively parallel computation, as the data race should be carefully managed.
+Reduction operations are not trivial on GPU due to its massively parallel nature that requires the data race to be carefully managed.
 Therefore, you cannot directly perform reduction operations on your own variables.
 Instead, the redcution operations should be performed on a preallocated buffer defined by the reducer.
```
lindsayad:
```suggestion
Instead, the reduction operations should be performed on a preallocated buffer defined by the reducer.
```

framework/doc/content/syntax/KokkosUserObjects/index.md
```diff
@@ -99,26 +86,49 @@ See the following source codes of `KokkosIntegralPostprocessor` for an example o
...
+
+!listing framework/src/kokkos/vectorpostprocessors/KokkosExtraIDIntegralVectorPostprocessor.K id=kokkos-extra-id-vpp-source language=cpp
+         caption=The `KokkosExtraIDIntegralVectorPostprocessor` source file.
+
+## User-defined APIs and Virtual Functions
+
+While user objects are intended to embrace user-defined APIs, Kokkos-MOOSE user objects currently require GPU APIs to not rely on virtual dispatch.
+Namely, you should always retrieve your user objects in their concrete types if you intend to use your own GPU APIs.
+Using virtual functions on GPU has two prerequisites: enabling the relocatable device code (RDC) option and constructing objects on GPU.
+The RDC is option is currently disabled in Kokkos-MOOSE due to the restrictions imposed by upstream packages (see [the discussions on this page](syntax/KokkosFunctions/index.md#kokkos_rdc)), and its resolution is being actively worked on.
```
lindsayad:
```suggestion
The RDC option is currently disabled in Kokkos-MOOSE due to the restrictions imposed by upstream packages (see [the discussions on this page](syntax/KokkosFunctions/index.md#kokkos_rdc)), and its resolution is being actively worked on.
```

framework/include/kokkos/bcs/KokkosIntegratedBC.h
```diff
@@ -135,6 +129,35 @@ class IntegratedBC : public IntegratedBCBase
   }
   ///@}
 
+  /**
+   * Get the function pointer of the default computeQpJacobianShim()
+   * @returns The function pointer
+   */
+  template <typename Derived>
+  static auto defaultJacobianShim()
```
lindsayad, line 142: I'd like some more explanation of what this is for. Might also be reasonable to have these changes in their own PR

framework/include/kokkos/reporters/KokkosElementStatistics.h
```diff
@@ -0,0 +1,78 @@
...
+
+class KokkosElementStatistics : public Moose::Kokkos::ElementReporter
+{
+public:
+  static InputParameters validParams();
+
+  KokkosElementStatistics(const InputParameters & parameters);
+
+  template <typename Derived>
+  KOKKOS_FUNCTION void executeShim(const Derived & reducer, Datum & datum, Real * result) const;
```
lindsayad: What makes it a shim? I must have missed this in the mardown

framework/include/kokkos/reporters/KokkosElementStatistics.h
```diff
@@ -0,0 +1,78 @@
...
+  result[2] += source[2];
+  result[3] += source[3];
+  result[4] += source[4];
+}
+
+KOKKOS_FUNCTION inline void
+KokkosElementStatistics::init(ReducerLoop, Real * result) const
+{
+  result[0] = std::numeric_limits<Real>::min();
+  result[1] = std::numeric_limits<Real>::max();
```
lindsayad: From my MetaPhysicL work I didn't think `std::numeric_limits` was safe to use on device

framework/include/kokkos/reporters/KokkosElementStatistics.h
NamjaeChoi: There is a document: https://mooseframework.inl.gov/syntax/Kokkos/index.html#kokkos_shim. In this case I'm using it to call `computeValue` of derived class.

framework/include/kokkos/reporters/KokkosElementStatistics.h
NamjaeChoi: Did you notice any weird behavior? Because these `numeric_limits` are `constexpr`

framework/include/kokkos/reporters/KokkosElementStatistics.h
NamjaeChoi: Actually now I'm thinking that it might be better to just make hook methods be template functions instead of having shims. Might be still not late to change design.

NamjaeChoi: Which do you prefer between having shims vs making hooks be function template? https://github.com/idaholab/moose/pull/32512/changes/30c3dbf41433cb1b2b582f95a8b91afbdef3737f

lindsayad: This certainly looks a lot cleaner from a framework maintenance perspective, but I think it's a slight regression in the application developer's experience. However, our application developers have generally shown little issue with simple templates like this. For instance at one point I removed all templating from AD objects in MOOSE but over time our application developers have introduced their own `template <bool is_ad>` templates all over the place.

So overall, yes, I think what you've done in 30c3dbf is a win. What do you think?

NamjaeChoi: I'm also leaning towards removing shims and make users be able to do casting directly in the hooks

NamjaeChoi: The PR is already open as draft based on this PR. Logan asked to open a separate issue but once this PR is merged, that one can move as well.

framework/src/kokkos/vectorpostprocessors/KokkosExtraIDIntegralVectorPostprocessor.K
```diff
@@ -0,0 +1,221 @@
...
+  _props.copyToDevice();
+}
+
+void
+KokkosExtraIDIntegralVectorPostprocessor::initialSetup()
+{
+  _unique_vpp_ids.create(_unique_vpp_id_map.size());
+
+  for (const auto elem : _mesh.getMesh().active_local_element_ptr_range())
+    _unique_vpp_ids[kokkosMesh().getContiguousElementID(elem)] = _unique_vpp_id_map[elem->id()];
```
lindsayad: This isn't going to work out if the object is block restricted right?

framework/src/kokkos/vectorpostprocessors/KokkosExtraIDIntegralVectorPostprocessor.K
```diff
@@ -0,0 +1,221 @@
...
+  {
+    const auto id_type = _extra_id[i];
+    auto & p = declareVector("Level-" + std::to_string(i) + "-" + id_type);
+    // collect local map of local vpp id to extra id
+    std::map<dof_id_type, dof_id_type> extra_ids;
+    for (const auto elem : _mesh.getMesh().active_local_element_ptr_range())
+    {
+      if (!hasBlocks(elem->subdomain_id()))
+        continue;
+      auto vpp_id = _unique_vpp_id_map[elem->id()];
```
lindsayad:
```suggestion
      auto vpp_id = libmesh_map_find(_unique_vpp_id_map, elem->id());
```
I always prefer doing this for sanity checking over using non-const `operator[]`

lindsayad (changes requested):
I think the registration situation is confusing. I feel like every concrete Kokkos user object should need
```
registerKokkosUserObject
```
I don't like the seeming mixture of purposes of registration macros where currently `registerKokkosUserObject` also registers `DefaultLoop/parallel_for` capability. And `registerKokkosReducer` also registers `parallel_reduce` capability. I think every user object should have the same number of registrations, which is perhaps 2. One to register the user object and then one (or two I guess if it can do both) registration that registers to ability to do `parallel_for`/`parallel_reduce`.

framework/src/kokkos/vectorpostprocessors/KokkosExtraIDIntegralVectorPostprocessor.K
NamjaeChoi: ~~I think block restriction won't matter. This does some extra work for the blocks this object is not covering, but basically this is just converting a map to an array for fast device access.~~ No actually you are correct. `_unique_vpp_id_map` is not generated for blocks not covered by this object.

NamjaeChoi:
> I think the registration situation is confusing. I feel like every concrete Kokkos user object should need
> 
> ```
> registerKokkosUserObject
> ```
> 
> I don't like the seeming mixture of purposes of registration macros where currently `registerKokkosUserObject` also registers `DefaultLoop/parallel_for` capability. And `registerKokkosReducer` also registers `parallel_reduce` capability. I think every user object should have the same number of registrations, which is perhaps 2. One to register the user object and then one (or two I guess if it can do both) registration that registers to ability to do `parallel_for`/`parallel_reduce`.

Hopefully the updated design makes sense. There is now only a single registration macro `registerKokkosUserObject`. Registers both dispatcher and reducer. The hook method for reducers was renamed to `reduce()`. Behavior will be chosen depending on whether users have overriden (redefined) `execute()` or `reduce()`. Their default (empty) implementations including `init()` and `join()` should now be provided to prevent compile error, so users will have to be more careful not to forget redefining them.

framework/include/kokkos/bcs/KokkosIntegratedBC.h
lindsayad: I guess you're going to put the shim removal PR in first so then this diff will disappear?

framework/include/kokkos/postprocessors/KokkosExtremeValueBase.h
```diff
@@ -92,16 +92,16 @@ KokkosExtremeValueBase<Base>::join(typename Base::DefaultLoop,
...
 KOKKOS_FUNCTION void
-KokkosExtremeValueBase<Base>::init(typename Base::DefaultLoop, Real * result) const
+KokkosExtremeValueBase<Base>::init(typename Base::ReducerLoop, Real * result) const
 {
   if (_type == ExtremeType::MAX || _type == ExtremeType::MAX_ABS)
   {
-    result[0] = -std::numeric_limits<Real>::max();
-    result[1] = -std::numeric_limits<Real>::max();
+    result[0] = Kokkos::Experimental::finite_min_v<Real>;
+    result[1] = Kokkos::Experimental::finite_min_v<Real>;
```
lindsayad, line 100: cool!

framework/include/kokkos/vectorpostprocessors/KokkosExtraIDIntegralVectorPostprocessor.h
```diff
@@ -0,0 +1,165 @@
...
+  const unsigned int _nprop;
+  /// Name of material properties
+  const std::vector<MaterialPropertyName> _prop_names;
+  /// Extra IDs in use
+  const std::vector<ExtraElementIDName> _extra_id;
+  /// Number of extra IDs in use
+  const unsigned int _n_extra_id;
+  /// Map of element ids to parsed vpp ids
+  std::unordered_map<dof_id_type, dof_id_type> _unique_vpp_id_map;
+  Array<Array<dof_id_type>> _unique_vpp_ids;
```
lindsayad, line 61: doxygen

framework/include/kokkos/vectorpostprocessors/KokkosExtraIDIntegralVectorPostprocessor.h
```diff
@@ -0,0 +1,165 @@
...
+  VariableValue _var_values;
+  /// Material properties to be integrated
+  Array<MaterialProperty<Real>> _props;
+  /// Vector holding the volume of extra IDs
+  Array<Real> _volumes;
+  /// Vectors holding integrals over extra IDs
+  Array<Array<Real>> _integrals;
+  /// Vectors holding extra IDs
+  std::vector<VectorPostprocessorValue *> _extra_ids;
+  /// Size of each vector
```
lindsayad:
```suggestion
  /// Size of the vector
```

framework/include/kokkos/bcs/KokkosIntegratedBC.h
NamjaeChoi: No, the shim removal PR will follow this PR.

NamjaeChoi:
Let me know when you are done with review.

framework/include/kokkos/bcs/KokkosIntegratedBC.h
NamjaeChoi: Do you want me to put the shim removal first? How far is this PR from merging?

lindsayad:
I am done with review. This PR is close to approve/merge so makes sense to follow the shim removal PR with this one

NamjaeChoi:
Excellent

NamjaeChoi:
I'm confused. Shim first or this one first? Doing shim first will require some non-trivial rebasing.

lindsayad (approved):

lindsayad:
this one first

PR #31897, opened 2025-11-12
Pack multiple column and transfer packed vectors to multi-output gaussian process trainer

lin-yang-ly:
This PR create a object in Stochastic Module Tools which can pack multiple columns from user-generated data and transfer packed data to multi-output gaussian process trainer.

Ref. #31820


lin-yang-ly:
@somu15 Please let me know if you have any suggestions for it!

PR #31911, opened 2025-11-13
Fix next broken by #31761

miaoyinb:
closes #31910


miaoyinb:
@loganharbour FYI

roystgnr (approved):
This is good enough for now, under the "next is broken" circumstances.

In the long run, these are at best workarounds, not fixes.  The long-term fix for any incompatibility of the form "X doesn't work unless Y happens first" should be "X checks if Y has been done and does Y itself if not" in cases where that's cheap enough, and should at least include "the X Doxygen mentions the requirement for Y" and "X asserts Y has been done" (i.e. not in opt mode) otherwise.  "We do Y first to save ourselves and woe unto whoever tries to use X next" just fixes the immediate problem at a future cost.

loganharbour (commented):
I deferred to @roystgnr here; good with me.

Issue #5079, opened 2015-05-19
Add clobberall target


dschwen:
This will make clobbering behave more like cleaning, with ```clobber``` operating from ```$(CURDIR)``` and ```clobberall``` iterating over all dependent app dirs.

jwpeterson:
BTW, @dschwen reports that the speed of make clobber is nearly doubled by not making the `-exec rm` system calls...

-delete:
```
real        0m10.515s
```

-exec rm:
```
real        0m18.253s
```

PR #31932, opened 2025-11-17
Update `Moose::MFEM::TimeDependentEquationSystem` to use state variables instead of time derivatives as trial vars

### LINKED ISSUE #31931: Solve for variable state at next timestep rather than time derivatives in Moose::MFEM::TimeDependentEquationSystem
### Motivation

Currently, `Moose::MFEM::TimeDependentEquationSystem` assembles transient problems for a vector of variable time derivatives (du/dt) rather than directly for variables at the next timestep ($u_{n+1}$), for compatibility with the `mfem::ODESolver` used for time integration. However, this comes with a number of drawbacks:

- Essential boundary conditions are most commonly expressed in terms of $u$ rather than $du/dt$; currently, we must convert between these during assembly of the `TimeDependentEquationSystem` in order to apply the correct Dirichlet conditions on $du/dt$. This will be a significant challenge when handling Dirichlet conditions for $u$ in nonlinear problems, and causes issues for the generalisation of the time integration scheme used when solving MFEM problems.
- Requiring trial variables to be time derivatives adds complexity for the user when wanting to solve differential-algebraic mixed systems of equations, as seen by the use of the `time_integrated_heat_flux` variable to label equations/kernel contributions in the mixed heat equation example. 
- This differs from the libMesh-based assembly methods in MOOSE, introducing a point of divergence between the different backends.

However, [support for stage-state based solvers](https://github.com/mfem/mfem/pull/5079) is being added upstream into MFEM, removing this constraint, and providing an opportunity for significantly simplifying EquationSystem and improving maintainability for transient problems.

### Design

Change `Moose::MFEM::TimeDependentEquationSystem` to set up problems described in terms of the state of the system at the next timestep, instead of for time derivatives. Prior to [MFEM PR#5079](https://github.com/mfem/mfem/pull/5079) being merged upstream, a patched ODESolver for supporting time integration using a backwards Euler integration scheme with a state-state solve provided by the `Moose::MFEM::ProblemOperator::ImplicitSolve` method would allow testing without loss of current capabilities.

### Impact

For the first iteration of the PR, only input files using the MFEM backend imposing Dirichlet conditions directly on time derivatives would need changing to be described in terms of the state variable(s) instead; only the mixed heat example satisfies this case. Otherwise, this change is restricted primarily to `Moose::MFEM::TimeDependentEquationSystem`.

In subsequent contributions, the requirement for a variable to be constructed to store the time derivative of all MFEM state variables could be relaxed, reducing output file sizes when they are not used or required. Similarly, support for a wider range of time integrators can easily be added.

alexanderianblair:
Closes #31931


loganharbour:
+250k/-250k scares me

alexanderianblair:
Almost all of that diff is in [6fcf645](https://github.com/idaholab/moose/pull/31932/commits/6fcf645d8cdbb6908daa8b96f55b11f820e8fda0) from re-golding the `.vtu` outputs of transient examples that include an integrated BC that provides bilinear form contributions - which is effectively those tests using the `MFEMConvectiveHeatFluxBC`. 

The actual diff on the results at each timestep is small, but the previous version was incorrectly adding contributions from these integrated BCs to `td_blf` in `TimeDependentEquationSystem::BuildBilinearForms()`, so erred for re-golding to fix the outputs rather than relaxing diff tolerances here, as this should help prevent a similar issue arising in any future re-factor

framework/doc/content/source/mfem/equation_systems/TimeDependentEquationSystem.md
```diff
@@ -0,0 +1,28 @@
...
+!equation
+\left(\mathcal{T}(\dot{u}), v\right)_{\Omega} + {\left(\mathcal{L}(u), v\right)_{\Omega}
+=\left(f,v\right)_{\Omega}\,\,\,\forall v \in V}
+
+Contributions to $\left(\mathcal{T}(\phi_j), \varphi_i\right)_{\Omega}$ are given by
+time derivative kernels such as
+ [MFEMTimeDerivativeMassKernel](source/mfem/kernels/MFEMTimeDerivativeMassKernel.md)
+
+Currently, transient problems in MOOSE's wrapping of MFEM are solved using an implicit backwards
+Euler method.
```
lindsayad, line 17: Do we have an issue/is there a plan for different time integration schemes? We have someone within the NEAMS program who has talked about leveraging or implementing (I can't remember which) exponential time integrators in MFEM

framework/include/mfem/equation_systems/EquationSystem.h
```diff
@@ -134,33 +79,35 @@ class EquationSystem : public mfem::Operator
...
+  /// Form linear system and jacobian operator based on on- and off-diagonal bilinear form
+  /// contributions, populate solution and RHS vectors of true DoFs, and apply constraints
+  virtual void FormLinearSystem(mfem::OperatorHandle & op,
+                                mfem::BlockVector & trueX,
+                                mfem::BlockVector & trueRHS);
+  virtual void
+  FormSystem(mfem::OperatorHandle & op, mfem::BlockVector & trueX, mfem::BlockVector & trueRHS);
+  virtual void FormLegacySystem(mfem::OperatorHandle & op,
+                                mfem::BlockVector & trueX,
+                                mfem::BlockVector & trueRHS);
```
lindsayad: Doxygen for this please (what makes it different from non-legacy?)

framework/include/mfem/equation_systems/EquationSystem.h
```diff
@@ -184,19 +132,45 @@ class EquationSystem : public mfem::Operator
...
+  /// Pointers to coupled variables not part of the reduced EquationSystem.
+  Moose::MFEM::GridFunctions _eliminated_variables;
+  /// Names of all test variables corresponding to linear forms in this equation system
+  std::vector<std::string> _test_var_names;
+  /// Pointers to finite element spaces associated with test variables.
+  std::vector<mfem::ParFiniteElementSpace *> _test_pfespaces;
+  /// Pointers to finite element spaces associated with coupled variables.
+  std::vector<mfem::ParFiniteElementSpace *> _coupled_pfespaces;
+
+  // Components of weak form. // Named according to test variable
```
lindsayad:
```suggestion
  // Components of weak form, named according to test variable
```

framework/include/mfem/equation_systems/TimeDependentEquationSystem.h
```diff
@@ -0,0 +1,52 @@
...
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#ifdef MOOSE_MFEM_ENABLED
+
+#pragma once
+#include "EquationSystem.h"
```
lindsayad, line 14:
```suggestion

#include "EquationSystem.h"
```
I think @nmnobre just tried to make this uniform

framework/include/mfem/equation_systems/TimeDependentEquationSystem.h
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Moose::MFEM
+{
+/**
+ * Class to store weak form components for time dependent PDEs
+ */
+class TimeDependentEquationSystem : public EquationSystem
+{
+public:
+  friend class TimeDomainEquationSystemProblemOperator;
```
lindsayad: I'd move this to private since friends are not part of an API

framework/include/mfem/equation_systems/TimeDependentEquationSystem.h
```diff
@@ -0,0 +1,52 @@
...
+
+protected:
+  /// Set trial variable names from subset of coupled variables that have an associated test variable.
+  virtual void SetTrialVariableNames() override;
+  virtual void BuildBilinearForms() override;
+  virtual void BuildMixedBilinearForms() override;
+  virtual void EliminateCoupledVariables() override;
+
+  /// Coefficient for timestep scaling
+  mfem::ConstantCoefficient _dt_coef;
```
lindsayad: It's not clear to me immediately what this means. Is this a coefficient scaling a suggested dt? Or is it dt? Or what?

framework/include/mfem/problem_operators/TimeDependentEquationSystemProblemOperator.h
```diff
@@ -12,6 +12,8 @@
 #pragma once
 #include "TimeDomainProblemOperator.h"
 #include "EquationSystemInterface.h"
+#include "TimeDependentEquationSystem.h"
+#include "MFEMBackwardEulerStateSolver.h"
```
lindsayad, line 17: With the refactoring I understand the new inclusion of `TimeDependentEquationSystem.h`, but usually I expect some class definition code change to require a new header include like `MFEMBackwardEulerStateSolver.h`

framework/include/mfem/timeintegrators/MFEMBackwardEulerStateSolver.h
```diff
@@ -0,0 +1,37 @@
...
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#ifdef MOOSE_MFEM_ENABLED
+
+#pragma once
```
lindsayad, line 12:
```suggestion
#pragma once

```

framework/include/mfem/timeintegrators/MFEMBackwardEulerStateSolver.h
```diff
@@ -0,0 +1,37 @@
...
+#include "mfem.hpp"
+#include "libmesh/restore_warnings.h"
+
+namespace Moose::MFEM
+{
+
+/// Modified version of the MFEM backward Euler solver solving for the state of
+/// the time dependent variable at the next timestep directly, instead of the
+/// time derivative.
+/// To be replaced if/when https://github.com/mfem/mfem/pull/5079 is merged.
```
lindsayad, line 21:
```suggestion
///
/// To be replaced if/when https://github.com/mfem/mfem/pull/5079 is merged.
```

framework/include/mfem/timeintegrators/MFEMBackwardEulerStateSolver.h
```diff
@@ -0,0 +1,37 @@
...
+class MFEMBackwardEulerStateSolver : public mfem::ODESolver
+{
+protected:
+  mfem::Vector k;
```
lindsayad: Let's keep order
```
public:
protected:
private:
```

framework/src/mfem/equation_systems/TimeDependentEquationSystem.C
```diff
@@ -0,0 +1,217 @@
...
+          ->Register(coupled_var_time_integral_name, std::move(kernels));
+    }
+    _td_kernels_map.GetRef(test_var_name)
+        .Get(coupled_var_time_integral_name)
+        ->push_back(std::move(kernel));
+  }
+  else
+  {
+    EquationSystem::AddKernel(kernel);
+  }
```
lindsayad:
```suggestion
    EquationSystem::AddKernel(kernel);
```

framework/src/mfem/equation_systems/TimeDependentEquationSystem.C
```diff
@@ -0,0 +1,217 @@
...
+}
+
+void
+TimeDependentEquationSystem::BuildBilinearForms()
+{
+  // Register bilinear forms
+  for (const auto i : index_range(_test_var_names))
+  {
+    auto test_var_name = _test_var_names.at(i);
```
lindsayad:
```suggestion
    const auto & test_var_name = _test_var_names.at(i);
```

framework/src/mfem/equation_systems/TimeDependentEquationSystem.C
```diff
@@ -0,0 +1,217 @@
...
+{
+  // Register mixed bilinear forms. Note that not all combinations may
+  // have a kernel.
+
+  // Create mblf for each test/coupled variable pair with an added kernel.
+  // Mixed bilinear forms with coupled variables that are not trial variables are
+  // associated with contributions from eliminated variables.
+  for (const auto i : index_range(_test_var_names))
+  {
+    auto test_var_name = _test_var_names.at(i);
```
lindsayad:
```suggestion
    const auto & test_var_name = _test_var_names.at(i);
```

framework/include/mfem/equation_systems/EquationSystem.h
lindsayad: (Presumably this is a reference to MFEM's legacy integrators)

framework/doc/content/source/mfem/equation_systems/TimeDependentEquationSystem.md
alexanderianblair: We do indeed - this PR is (in part) to lay the groundwork ahead of [MFEM PR #5079](https://github.com/mfem/mfem/pull/5079/files#diff-50af48cc3fc39a0cac6b79f564a8cb47a8a356818309468298fb4f105004f82b), where MFEM's ODE solvers are more widely being updated to support systems solved in terms of the state at the next timestep, instead of requiring the equation system to solve for the vector of time derivatives.

The difficulty prior to this PR was that Dirichlet conditions had to be expressed in terms of the time derivative du/dt when provided by the user in terms of the state variable u (as is most common). Once this and [MFEM PR #5079](https://github.com/mfem/mfem/pull/5079/files#diff-50af48cc3fc39a0cac6b79f564a8cb47a8a356818309468298fb4f105004f82b) are merged, extending support to other integration schemes should be simple - adding MOOSE wrappers for MFEM ODESolvers and then using the user-defined ODESolver specifying the desired integration scheme when setting up the `ProblemOperator`.

framework/include/mfem/equation_systems/TimeDependentEquationSystem.h
alexanderianblair: Fair - updated!

framework/include/mfem/timeintegrators/MFEMBackwardEulerStateSolver.h
alexanderianblair: Indeed - updated

framework/include/mfem/equation_systems/TimeDependentEquationSystem.h
alexanderianblair: This is just the timestep, but in previous iterations we needed to store it as a coefficient for the purpose of combining coefficients when scaling integrators by `dt`.

However - I don't believe it's used in this form anywhere now, so I've converted it to `mfem::real_t` and updated the doxygen to reflect its role as the timestep for the `TimeDependentEquationSystem`, as I think that's clearer

framework/include/mfem/equation_systems/EquationSystem.h
alexanderianblair: It is indeed - I've now renamed these methods (and added associated Doxygen) to `FormSystemOperator` and `FormSystemMatrix` to better reflect their usage. These new names are intended to be consistent with that used in the [MFEM documentation](https://mfem.org/howto/block_operators_matrices/) - the difference is whether or not we're explicitly setting up a `HypreParMatrix` or not here.

framework/doc/content/source/mfem/equation_systems/TimeDependentEquationSystem.md
lindsayad: Nice. Let's make a MOOSE issue for this soon if it doesn't already exist

lindsayad (approved):

framework/doc/content/source/mfem/equation_systems/TimeDependentEquationSystem.md
```diff
@@ -0,0 +1,28 @@
...
+For transient problems, time derivatives $\dot{u}$ of the trial variables $u$ will also be
+present in the weak form.
+
+!equation
+\left(\mathcal{T}(\dot{u}), v\right)_{\Omega} + {\left(\mathcal{L}(u), v\right)_{\Omega}
+=\left(f,v\right)_{\Omega}\,\,\,\forall v \in V}
+
+Contributions to $\left(\mathcal{T}(\phi_j), \varphi_i\right)_{\Omega}$ are given by
+time derivative kernels such as
+ [MFEMTimeDerivativeMassKernel](source/mfem/kernels/MFEMTimeDerivativeMassKernel.md)
```
nmnobre:
```suggestion
 [MFEMTimeDerivativeMassKernel.md].
```

framework/src/mfem/problem_operators/EquationSystemProblemOperator.C
```diff
@@ -34,9 +32,8 @@ EquationSystemProblemOperator::Solve()
 {
   GetEquationSystem()->BuildJacobian(_true_x, _true_rhs);
 
-  if (_problem_data.jacobian_solver->isLOR() && _equation_system->_test_var_names.size() > 1)
+  if (_problem_data.jacobian_solver->isLOR() && _equation_system->_trial_var_names.size() > 1)
     mooseError("LOR solve is only supported for single-variable systems");
-
   _problem_data.jacobian_solver->updateSolver(
       *_equation_system->_blfs.Get(_equation_system->_test_var_names.at(0)),
```
nmnobre: 1) The space removal is inconsistent with the other equation system problem operators;
2) We seem to be mixing `_equation_system` and `GetEquationSystem()` somewhat arbitrarily (see time domain class counterpart as well);
3) We have `TrialVarNames()` and `TestVarNames()` available, the latter of which is now unused. These classes are friends of `EquationSystem` anyway, maybe we should either remove the accessors or add the missing accessors so they don't need to be friends?
4) Here we changed to check the trial variables, but not in the time domain class;
5) The check is on trial variables, but we then update the solver by passing the zeroth test variable name;

framework/src/mfem/problem_operators/TimeDomainEquationSystemProblemOperator.C
```diff
@@ -28,34 +27,54 @@ TimeDomainEquationSystemProblemOperator::Init(mfem::BlockVector & X)
...
+        _problem_data.time_derivative_map.getTimeDerivativeName(trial_var_name);
+    auto * trial_var = _problem_data.gridfunctions.Get(trial_var_name);
+    auto * trial_var_time_derivative = _problem_data.gridfunctions.Get(time_derivative_name);
+
+    *trial_var_time_derivative = *trial_var;
+  }
   // Advance time step of the MFEM problem. Time is also updated here, and
   // _problem_operator->SetTime is called inside the ode_solver->Step method to
   // update the time used by time dependent (function) coefficients.
   _problem_data.ode_solver->Step(_problem_data.f, _problem.time(), _problem.dt());
```
nmnobre: Conceptually, I'd prefer us to do something like the following:
```suggestion
  _problem_data.ode_solver->Step(const_cast<mfem::Vector &>(*_trial_true_vector), _problem.time(), _problem.dt());
```
Or, otherwise, `Init()` should take no input.

framework/include/mfem/problem_operators/TimeDomainEquationSystemProblemOperator.h
```diff
@@ -13,6 +13,8 @@
 
 #include "TimeDomainProblemOperator.h"
```
nmnobre: Shall we rename `TimeDomain*ProblemOperator` to `TimeDependent*ProblemOperator`?

nmnobre (changes requested):
This is awesome!

framework/include/mfem/equation_systems/EquationSystem.h
```diff
@@ -219,15 +201,19 @@ EquationSystem::ApplyDomainBLFIntegrators(
...
     auto kernels = kernels_map.GetRef(test_var_name).GetRef(trial_var_name);
     for (auto & kernel : kernels)
     {
-      mfem::BilinearFormIntegrator * integ = kernel->createBFIntegrator();
-      if (integ != nullptr)
+      mfem::BilinearFormIntegrator * integ =
+          scale_factor.has_value()
+              ? new ScaleIntegrator(kernel->createBFIntegrator(), scale_factor.value(), true)
+              : kernel->createBFIntegrator();
+      if (integ != nullptr && kernel->createBFIntegrator() != nullptr)
```
nmnobre: Memory leak. Likewise below in `ApplyBoundaryBLFIntegrators`. I'll activate the Valgrind recipe which will hopefully flag this more explicitly.

alexanderianblair: Thanks for catching this - pushed update that I believe should fix this

framework/src/mfem/problem_operators/EquationSystemProblemOperator.C
alexanderianblair: 1. Indeed - now made consistent with the other equation system problem operators by removal there too - the intent is to keep this block logically associated with the jacobian solver update. 
2. We are indeed - I've now updated this to use `GetEquationSystem()` consistently.
3. I would be a fan of adding the remaining accessors for the bilinear form and the tdof list of essential constraints required for LOR solver setup - I would much rather avoid having the friend association here. I would prefer to keep that to a separate PR though, since I think the LOR solver setup should have some broader cleanup (not least because updating the ess_tdof_list argument of updateSolver to be a const ref instead will require making diffs to all solvers). 
That said - I have now renamed the existing accessors to `GetTestVarNames()` and `GetTrialVarNames()` to make the access relationship clear.
4. Now updated to consistently use `GetTestVarNames()`
5. Now using `GetTestVarNames()`

framework/src/mfem/problem_operators/TimeDomainEquationSystemProblemOperator.C
alexanderianblair: Updated to use `_trial_true_vector`, but by making it non-const (as it's updated in `Step` anyway) rather than using `const_cast`

framework/include/mfem/problem_operators/TimeDomainEquationSystemProblemOperator.h
alexanderianblair: We should indeed, now we're using 'complex' for describing frequency domain problems - `TimeDependentXXX` is more consistent with its base classes too. Updated.

framework/src/mfem/problem_operators/EquationSystemProblemOperator.C
nmnobre: I've added an item to our internal project tracker for (3).

framework/include/mfem/problem_operators/TimeDependentEquationSystemProblemOperator.h
nmnobre: @alexanderianblair can you expand a bit on the need for `MFEMBackwardEulerStateSolver`? I understand we need to do something to circumvent MFEM from adding what it supposes is a time derivative, but now isn't, so far so good.

But we don't need a new derived class to do this, we can copy the old solution to a temporary vector and then subtract this from the new vec and divide by dt. Example: https://github.com/nmnobre/moose/commit/62fe00d099e4f50d19907569183aa0a489cb5620.

Or we could derive from `BackwardEulerSolver` directly. Example: https://github.com/nmnobre/moose/commit/ce37b684f82bc0d51e57ba8cf4d6b1002c0fcb6b.

Going forward, since `MFEMTransient` inherits from `TransientBase`, which has this `scheme` parameter, could we simply not piggy back from this - maybe expanding the available options to cover the MFEM range - and map string to mfem classes and call it a day?

alexanderianblair: > @alexanderianblair can you expand a bit on the need for MFEMBackwardEulerStateSolver? I understand we need to do something to circumvent MFEM from adding what it supposes is a time derivative, but now isn't, so far so good.

This PR changes the effect of the `TimeDependentProblemOperator::ImplicitSolve` method to solve for the 'next' state of the system, rather than the next effective vector of time derivatives. Inside the `Step` method of the MFEM ODESolver in use, `ImplicitSolve` is called one or more times depending on the time integration scheme being employed. In `mfem::BackwardEulerSolver`, it is assumed that `ImplicitSolve` updates the vector of time derivatives (`k=du/dt`), and `mfem::BackwardEulerSolver::Step` updates the state of `u` accordingly (using a backwards Euler update). `MFEMBackwardEulerStateSolver` updates the `Step` method when `ImplicitSolve` updates the state of the system (`k=u(t+dt)`) instead of the vector of time derivatives. `MFEMBackwardEulerStateSolver` will be unnecessary once `mfem::ODESolver` is updated to support state-state solves, since its functionality will be entirely covered by `mfem::BackwardEulerSolver`.

> But we don't need a new derived class to do this, we can copy the old solution to a temporary vector and then subtract this from the new vec and divide by dt. Example: https://github.com/nmnobre/moose/commit/1261a7488c752cc57a6c2bc99c67815ec7d9fedb

This won't work as far as I can tell, since this is still calling the old `mfem::BackwardEulerSolver::Step` method, which assumes `k=du/dt`.

> Or we could derive from BackwardEulerSolver directly. Example: https://github.com/nmnobre/moose/commit/ce37b684f82bc0d51e57ba8cf4d6b1002c0fcb6b

This should work fine, as it overrides the `Step` method.

> Going forward, since MFEMTransient inherits from TransientBase, which has this scheme parameter, could we simply not piggy back from this - maybe expanding the available options to cover the MFEM range - and map string to mfem classes and call it a day?

Yes, this should work well once those `mfem::ODESolvers` have state-state solve support and we add a suitable `TimeDependentProblemOperator::Mult` method for explicit updates!

alexanderianblair: > Or we could derive from BackwardEulerSolver directly. Example: https://github.com/nmnobre/moose/commit/ce37b684f82bc0d51e57ba8cf4d6b1002c0fcb6b.

Update - cherry-picked this - I like the smaller footprint here

nmnobre: We are 100% on the same page here. :)

> This won't work as far as I can tell

I've tested it, and it compiles and passes all tests. BUT it's ugly, confusing, hard to read and requires an extra vector allocation... Basically if we use the mfem integrator, we basically end up with x(t) + dt * x(t + dt) stored in our trial vector, so by doing the sequence of subtraction and division suggested, you recover x(t + dt). But, as I was saying, I agree, this is one of those trying-to-be-too-clever-for-our-own-good-type solutions we shouldn't use.

> Update - cherry-picked this - I like the smaller footprint here

I'm very happy we agree on this though - it's just (much) cleaner than the above. :) When the time comes, we ditch the header and replace the wrapper class with MFEM's class and we are good to go.

nmnobre (approved):

framework/doc/content/source/mfem/equation_systems/TimeDependentEquationSystem.md
alexanderianblair: Added in #32036

framework/include/mfem/problem_operators/TimeDependentEquationSystemProblemOperator.h
alexanderianblair: > Basically if we use the mfem integrator, we basically end up with x(t) + dt * x(t + dt) stored in our trial vector, so by doing the sequence of subtraction and division suggested, you recover x(t + dt)

Ah - I'm with you - yes, that would work for this backwards Euler case, but would need to be rewritten for later `mfem::ODESolvers` - I'm happy with this version, as it'll be easier to extend later on, as you say!

PR #31938, opened 2025-11-18
Shifted Cohesive Zone Method: Add geometric search infrastructure, distance calculation components, and App updates

ChengHauYang:
close #31937 

## Reason
The Shifted Boundary Method (SBM) is an immersed boundary method that applies boundary conditions on non-boundary-fitted meshes. It relies on Nitsche’s method to weakly impose boundary conditions, where the “shifted” aspect involves transferring the boundary condition from the true boundary to a surrogate boundary using a Taylor-series-based correction.

To account for the discrepancy between the true and surrogate boundaries, the method requires precise computation of the distance vector from the surrogate boundary to the true boundary and the true normal on the projected point. Currently, MOOSE lacks a unified infrastructure to handle these geometric queries for mesh-based boundaries. As a result, we require a geometric backend for SBM, which enables efficient nearest-neighbor searches (via KDTree) and provides a consistent interface for evaluating distances/normals to both analytical level sets and discrete surface meshes.

The Shifted Cohesive Zone Method (SCZM) is a special kind of SBM that focuses on applying a shifted boundary condition to the interface problem. In the SCZM, we consider the jump of the variable across the interface to become the Neumann boundary condition we applied through SBM.

This is a first step toward a complete SCZM implementation in MOOSE. Follow-up PRs will build on this infrastructure to add:
- An in–out test to mark elements as IN, OUT, or INTERSECTED and define the surrogate domain/boundary.
- Interface kernels to apply the SCZM.

## Design
This PR introduces the core geometric architecture for SBM:

1. Geometric Search Infrastructure (`SBMSurfaceMeshBuilder`)
   - A `UserObject` that processes a named surface mesh.
   - Builds a KDTree of boundary element centroids for fast spatial lookups.
   - Creates lightweight wrappers (`SBMBndEdge2`, `SBMBndTri3`) to handle element-specific geometry (normals, ray intersections) without full mesh overhead.

2. Distance Interfaces
   - `UnsignedDistanceToSurfaceMesh`: A MOOSE `Function` that queries the builder's KDTree to provide spatially continuous distance fields and gradients.
   - `ShortestDistanceToSurface`: A `UserObject` that aggregates multiple surfaces (both analytical functions and mesh-based distances) to determine the absolute minimum distance and true normal for a given point.
   - `DistanceAux`: An `AuxKernel` for visualizing the computed distances.

3. Shared Utilities (`SBMUtils`)
   - Provides centralized logic for extracting distance vectors and normals from both `ParsedFunction` (level sets) and mesh-based objects.
   - Includes topology validation tools (e.g., `checkWatertightness`) to ensure surface mesh integrity.

## Impact
- New Capabilities: Enables SBM simulations to run on complex geometries defined by mesh files, not just analytical equations.
- Architecture: Establishes the "Geometric Engine" that future Nitsche-based kernels will query for boundary data.


modules/shifted_boundary_method/doc/content/source/auxkernels/DistanceAux.md
```diff
@@ -0,0 +1,21 @@
+# DistanceAux
+
+`DistanceAux` evaluates the unsigned distance from an element centroid to the closest
+surface described by a [`ShortestDistanceToSurface`](userobjects/ShortestDistanceToSurface.md)
+user object. The auxiliary kernel calculates the Euclidean distance from an element's
+centroid to the boundary represented by the `distance_to_surface` object.
```
hugary1995: From the description it's not clear to me whether this aux kernel itself is responsible for calculating the distance, or is it just outputting values to the aux variable?

modules/shifted_boundary_method/doc/content/source/modules/shifted_boundary_method/index.md
```diff
@@ -0,0 +1,17 @@
+# Shifted Boundary Method Module
+
+The Shifted Boundary Method (SBM) module provides a flexible framework for solving
+partial differential equations on complex geometries without requiring boundary-fitted
+meshes. Instead of conforming the mesh to curved or intricate boundaries, SBM embeds
+the geometry inside a background mesh and shifts the boundary conditions from the
+true interface to surrogate locations on nearby mesh faces. This approach preserves
+accuracy, avoids costly remeshing, and enables robust treatment of Dirichlet, Neumann,
+and Robin conditions.
```
hugary1995:
```suggestion
and Robin boundary conditions.
```

modules/shifted_boundary_method/doc/content/source/SBMBndElement/SBMBndEdge2.md
hugary1995, line 1: The documentation of internal data structures belong to doxygen, not moosedocs. I don't think we need this markdown file.

modules/shifted_boundary_method/doc/content/source/SBMBndElement/SBMBndElementBase.md
hugary1995, line 1: Same. This belongs to doxygen, not moosedocs.

modules/shifted_boundary_method/doc/content/source/SBMBndElement/SBMBndTri3.md
hugary1995, line 1: Same. This belongs to doxygen, not moosedocs.

modules/shifted_boundary_method/doc/content/source/utils/SBMUtils.md
hugary1995, line 1: This belongs to doxygen, not moosedocs.

modules/shifted_boundary_method/doc/moosedocs.py
hugary1995, line 1: @cticenhour , do we need a new copy of the moosedocs.py file for each module? I wonder if this can just be a symlink. I suppose @ChengHauYang generated this using the stork.sh script.

modules/shifted_boundary_method/include/auxkernels/DistanceAux.h
```diff
@@ -7,24 +7,22 @@
 //* Licensed under LGPL 2.1, please see LICENSE for details
 //* https://www.gnu.org/licenses/lgpl-2.1.html
 
-#ifdef MOOSE_MFEM_ENABLED
-
 #pragma once
 
-#include "MFEMComplexEssentialBC.h"
+#include "AuxKernel.h"
+#include "ShortestDistanceToSurface.h"
```
hugary1995: Remove include, use forward declaration.

modules/shifted_boundary_method/include/functions/UnsignedDistanceToSurfaceMesh.h
```diff
@@ -0,0 +1,54 @@
...
+  UnsignedDistanceToSurfaceMesh(const InputParameters & parameters);
+
+  /// find closest boundary element using KDTree
+  SBMBndElementBase & closestBoundaryElem(const Point & p) const;
+
+  /// find closest boundary element and return its distance vector
+  RealVectorValue distanceVectorToSurface(const Point & p) const;
+
+  /// Return unsigned distance value
+  virtual Real value(Real t, const Point & p) const override;
```
hugary1995: virtual is redundant.

modules/shifted_boundary_method/include/functions/UnsignedDistanceToSurfaceMesh.h
```diff
@@ -0,0 +1,54 @@
...
+
+  /// Return unit direction of distance vector (derivative)
+  virtual RealGradient gradient(Real t, const Point & p) const override;
```
hugary1995: virtual is redundant.

modules/shifted_boundary_method/include/functions/UnsignedDistanceToSurfaceMesh.h
```diff
@@ -0,0 +1,54 @@
...
+
+  /// Return surface normal vector on boundary
```
hugary1995: Perhaps this can be clarified. Is this the surface normal corresponding to the closest point on the surface?

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
```diff
@@ -0,0 +1,30 @@
...
+#include "SBMBndElementBase.h"
+
+/// Derived class for 2-node edge elements
+class SBMBndEdge2 : public SBMBndElementBase
+{
+public:
+  /// Constructor
+  explicit SBMBndEdge2(const Elem * elem);
+
+  /// Check if a line (a-b) intersects this edge
```
hugary1995: Line or line segment?

I would prefer to abstract out the definition of line (or line segment, whichever you use).

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
```diff
@@ -0,0 +1,30 @@
...
+  bool intercepted(const Point & a, const Point & b) const override;
+
+  /// Compute the bounding ball (circle) of the edge
+  std::pair<Point, Real> computeBoundingBall() const override;
```
hugary1995: Similarly, I'd like to abstract out the definition of a ball.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndElementBase.h
```diff
@@ -0,0 +1,91 @@
...
+   * This is a pure virtual function and must be implemented in derived classes.
+   */
+  virtual std::pair<Point, Real> computeBoundingBall() const = 0;
+
+  /**
+   * Compute the longest diagonal of the projected bounding box onto a given plane.
+   * @param plane_normal: The normal vector of the plane onto which the bounding box is projected.
+   * @return The length of the longest diagonal of the projected rectangle.
+   */
+  Real getProjectedBoundingBoxDiagonal(const Point & plane_normal) const;
```
hugary1995: Similarly, I prefer abstracting out the definition of a plane.

If we need a lot of geometrical entities, feel free to bring in dependency on a 3rd party lightweight geometry library.

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
```diff
@@ -0,0 +1,83 @@
...
+class SBMSurfaceMeshBuilder : public GeneralUserObject
+{
+public:
+  static InputParameters validParams();
+  SBMSurfaceMeshBuilder(const InputParameters & parameters);
+
+  virtual void initialSetup() override;
+  virtual void initialize() override {}
+  virtual void execute() override {}
+  virtual void finalize() override {}
```
hugary1995, line 26: virtual is redundant.

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
```diff
@@ -0,0 +1,83 @@
...
+
+  const KDTree & getKDTree() const { return *_kd_tree; }
+
+  const std::vector<std::unique_ptr<SBMBndElementBase>> & getBoundaryElements() const
+  {
+    return _boundary_elements;
+  }
+
+  const std::vector<Point> & getCentroids() const { return _centroids; }
```
hugary1995: For one-liners like these, it's okay to keep them in the header. But I personally still prefer to put them into the translation unit.

hugary1995: They are also missing doxygen comments.

modules/shifted_boundary_method/include/userobjects/ShortestDistanceToSurface.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+class ShortestDistanceToSurface : public ElementUserObject
+{
+public:
+  static InputParameters validParams();
+  ShortestDistanceToSurface(const InputParameters & parameters);
+  virtual void execute() override {};
+  virtual void finalize() override {}
+  virtual void initialize() override {}
+  virtual void threadJoin(const UserObject &) override {}
```
hugary1995: virtual is redundant.

modules/shifted_boundary_method/include/utils/SBMUtils.h
```diff
@@ -0,0 +1,56 @@
...
+bool checkWatertightnessFromRawElems(const std::vector<const Elem *> & bd_elements);
+
+/// Build a list of distance functions based on names specified in input.
+std::vector<const Function *>
+buildDistanceFunctions(const std::vector<FunctionName> & function_names,
+                       const FunctionInterface & function_provider);
+
+/// Compute the distance vector contributed by a single function.
+RealVectorValue
+distanceVectorFromFunction(const Function * func, const libMesh::Point & pt, Real t);
```
hugary1995: This is not any function, right? It must be a distance function.

modules/shifted_boundary_method/src/auxkernels/DistanceAux.C
```diff
@@ -0,0 +1,49 @@
...
+InputParameters
+DistanceAux::validParams()
+{
+  InputParameters params = AuxKernel::validParams();
+
+  params.addParam<UserObjectName>("distance_to_surface",
+                                  "UserObject that provides distance calculations.");
+
+  params.addClassDescription("Computes the distance using "
+                             "geometry data provided by the 'distance_to_surface' UserObject.");
```
hugary1995: This object is much more limited than this description, right? It is always calculating the distance from element centroids to the surface.

hugary1995 (changes requested):

modules/shifted_boundary_method/doc/moosedocs.py
ChengHauYang: Yes. I think this is generated automatically.

modules/shifted_boundary_method/doc/content/source/auxkernels/DistanceAux.md
ChengHauYang: I tried to fix this *md to be

```
`DistanceAux` outputs the distance from an element centroid to the closest
surface calculated by a [`ShortestDistanceToSurface`](userobjects/ShortestDistanceToSurface.md)
user object.

```

modules/shifted_boundary_method/doc/content/source/modules/shifted_boundary_method/index.md
ChengHauYang: Fixed

modules/shifted_boundary_method/include/auxkernels/DistanceAux.h
ChengHauYang: Fixed

modules/shifted_boundary_method/include/functions/UnsignedDistanceToSurfaceMesh.h
ChengHauYang: Fixed

modules/shifted_boundary_method/include/functions/UnsignedDistanceToSurfaceMesh.h
ChengHauYang: Fixed

modules/shifted_boundary_method/include/functions/UnsignedDistanceToSurfaceMesh.h
ChengHauYang: Sure.

We now have a better Doxygen docstring.:
```
  /// Return the spatial gradient of the unsigned distance function.
  /// The gradient is a unit vector pointing in the direction of decreasing
  /// distance, i.e., from the query point toward the closest point on the surface.
  RealGradient gradient(Real t, const Point & p) const override;

  /// Return the true boundary surface normal evaluated at the closest point obtained
  /// by projecting the query point onto the true boundary.
  RealVectorValue surfaceNormal(const Point & p) const;
```

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
ChengHauYang: Fixed to be a line segment.

abstarting is still WIP

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
ChengHauYang: Fixed

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
ChengHauYang: 1. Put them inside the translation unit now.
2. Add the missing doxygen comments.

modules/shifted_boundary_method/include/userobjects/ShortestDistanceToSurface.h
ChengHauYang: Fixed

modules/shifted_boundary_method/include/utils/SBMUtils.h
ChengHauYang: Fixed the docstring to be better now.

modules/shifted_boundary_method/src/auxkernels/DistanceAux.C
ChengHauYang: We have a better class description now:
```
  params.addClassDescription(
      "Creates a distance field based on the 'distance_to_surface' UserObject.");
```

modules/shifted_boundary_method/doc/moosedocs.py
ChengHauYang: Oh. Sorry. I actually copied moosedocs.py from another module. I think.

modules/shifted_boundary_method/doc/content/source/SBMBndElement/SBMBndEdge2.md
ChengHauYang: I have rewritten the history to delete this file.

modules/shifted_boundary_method/doc/content/source/SBMBndElement/SBMBndElementBase.md
ChengHauYang: Same. I have rewritten the history to delete this file.

modules/shifted_boundary_method/doc/content/source/SBMBndElement/SBMBndTri3.md
ChengHauYang: Same. I have rewritten the history to delete this file.

modules/shifted_boundary_method/doc/content/source/utils/SBMUtils.md
ChengHauYang: Same. I have rewritten the history to delete this file.

framework/include/geometries/Ball.h
hugary1995, line 1: Can we put Ball somewhere in the framework? As I think it is generically useful, not just for SBM.

modules/shifted_boundary_method/src/SBMBndElement/SBMBndEdge2.C
```diff
@@ -19,42 +20,32 @@ SBMBndEdge2::SBMBndEdge2(const Elem * elem) : SBMBndElementBase(elem)
...
-    return true;
+  if (!collinear)
+    return true; // Non-collinear case: the result from edge_seg.intersect(line_segment, ip)
+                 // is reliable and sufficient to confirm an intersection.
 
-  return false;
+  const bool overlap = line_segment.contains_point(p1) || line_segment.contains_point(p2) ||
+                       edge_seg.contains_point(line_segment.start()) ||
+                       edge_seg.contains_point(line_segment.end());
+  return overlap;
```
hugary1995: Does `LineSegment::contains_point` not consider the cases of overlapping colinear line segments?

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
```diff
@@ -11,18 +11,21 @@
...
 class SBMBndEdge2 : public SBMBndElementBase
 {
 public:
   /// Constructor
   explicit SBMBndEdge2(const Elem * elem);
 
-  /// Check if a line segment (a-b) intersects this edge
-  bool intercepted(const Point & a, const Point & b) const override;
+  /// Check if a line segment intersects this edge
+  bool intercepted(const LineSegment & line_segment) const override;
```
hugary1995: I guess "intercepted" is a bit of an uncommon usage. Can we rename it to intersect?

modules/shifted_boundary_method/src/SBMBndElement/SBMBndTri3.C
```diff
@@ -53,12 +55,15 @@ SBMBndTri3::computeBoundingBall() const
   }
 
   Real radius = std::sqrt(max_sq_dist) * 1.01; // add epsilon buffer
-  return {centroid, radius};
+  return Ball(centroid, radius);
 }
 
 bool
-SBMBndTri3::intercepted(const Point & a, const Point & b) const
+SBMBndTri3::intercepted(const LineSegment & line_segment) const
```
hugary1995: Let's make a free function to detect intersection between a triangle and a line segment. This member function can just call that free function.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
```diff
@@ -11,18 +11,21 @@
 
 #include "SBMBndElementBase.h"
 
+class Ball;
+class LineSegment;
+
 /// Derived class for 2-node edge elements
 class SBMBndEdge2 : public SBMBndElementBase
```
hugary1995: Can we let SBMBndEdge2 be a LineSegment, i.e.
```
class SBMBndEdge2 : public SBMBndElementBase, public LineSegment
```

modules/shifted_boundary_method/include/SBMBndElement/SBMBndTri3.h
```diff
@@ -9,28 +9,31 @@
...
 #pragma once
 
+class LineSegment;
+
 #include "SBMBndElementBase.h"
 
+class Ball;
+
 /// Derived class for 3-node triangular elements (Tri3)
 class SBMBndTri3 : public SBMBndElementBase
```
hugary1995: Similarly, can we let SBMBndTri3 be a Triangle? i.e.
```
class SBMBndTri3 : public SBMBndElementBase, public Triangle
```
and of course we will need to write a brief Triangle class/struct.

hugary1995 (changes requested):
Some picky requests. But I suggest we put all the generic geometry utilities into framework. Perhaps start a new subfolder called "geometries". There is currently a utility header called `GeometryUtils.h` which we can relocate into "framework/geometries".

framework/include/geometries/Ball.h
ChengHauYang: Yes. Put inside "framework/geometries" as well.

modules/shifted_boundary_method/src/SBMBndElement/SBMBndEdge2.C
ChengHauYang: `LineSegment::intersect` does not check overlapping colinear line segments. I also fixed `LineSegment::intersect` so we no longer need to reimplement the logic here.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
ChengHauYang: Sure. Fixed to be `intersect `

modules/shifted_boundary_method/src/SBMBndElement/SBMBndTri3.C
ChengHauYang: Sure. Right now, this free function is inside `Triangle.[C,h]`.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
ChengHauYang: Sure. This is in the commit 5e22a5c.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndTri3.h
ChengHauYang: Sure. This is in the commit 5e22a5c.

modules/shifted_boundary_method/doc/moosedocs.py
hugary1995: moosedocs.py in other modules are symlinks. This one is no exception.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
```diff
@@ -0,0 +1,30 @@
...
+
+/// Derived class for 2-node edge elements
+class SBMBndEdge2 : public SBMBndElementBase, public LineSegment
+{
+public:
+  /// Constructor
+  explicit SBMBndEdge2(const Elem * elem);
+
+  /// Check if a line segment intersects this edge
+  bool intersect(const LineSegment & line_segment) const override;
```
hugary1995: Why do we override it?

modules/shifted_boundary_method/include/SBMBndElement/SBMBndElementBase.h
```diff
@@ -0,0 +1,101 @@
...
+  /// Constructor takes a pointer to a boundary element
+  SBMBndElementBase(const Elem * elem);
+
+  /// Virtual destructor to ensure proper cleanup in derived classes
+  virtual ~SBMBndElementBase() = default;
+
+  /**
+   * Check if the given line segment intersects this boundary element.
+   */
+  virtual bool intersect(const LineSegment & line_segment) const = 0;
```
hugary1995: I see, so you need an abstract interface for querying line segment intersection. Why not create a base class for geometries (LineSegment, Triangle, Ball, etc.) and move this intersect method to that base class?

modules/shifted_boundary_method/include/SBMBndElement/SBMBndTri3.h
```diff
@@ -0,0 +1,36 @@
...
+
+/// Derived class for 3-node triangular elements (Tri3)
+class SBMBndTri3 : public SBMBndElementBase, public Triangle
+{
+public:
+  /// Constructor
+  explicit SBMBndTri3(const Elem * elem);
+
+  /// Check if a line segment intersects this triangle
+  bool intersect(const LineSegment & line_segment) const override;
```
hugary1995: same comment on overriding.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndTri3.h
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Compute the bounding ball (sphere) of the triangle
+   * This function computes an approximate bounding ball of a triangle by taking the triangle's
+   * centroid as the center, and setting the radius to be slightly larger than the farthest vertex
+   * distance from the centroid.
+   */
+  Ball computeBoundingBall() const override;
```
hugary1995: bounding ball method could also be moved to that geometry base class.

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
```diff
@@ -0,0 +1,82 @@
...
+
+  /// The flag to check the watertightness of the mesh.
+  bool _check_watertightness;
+
+  /// whether we want to build a kd-tree or not
+  bool _build_kd_tree;
+
+private:
+  /// The dimension of the embedding mesh (2D or 3D).
+  unsigned int _dim_embedding_mesh;
```
hugary1995: They can/should all be const, right?

modules/shifted_boundary_method/src/auxkernels/DistanceAux.C
```diff
@@ -0,0 +1,49 @@
...
+Real
+DistanceAux::computeValue()
+{
+  // distance value calculation
+  const Point & pt = _current_elem->vertex_average();
+
+  const Point & closest_vec = _distance_to_surface->distanceVector(pt);
+
+  return closest_vec.norm();
+}
```
hugary1995: The purpose of this auxkernel is not clearly conveyed by its name. To me this auxkernel is more like `ElementCentroidToSurfaceDistanceAux`. In other words, there could be another auxkernel named `SurfaceDistanceAux`. The key difference is that the former always uses element centroids, and the latter directly queries at quadrature points. What you've implemented is the former, so I'd suggest a renaming.

hugary1995 (changes requested):

modules/shifted_boundary_method/doc/moosedocs.py
ChengHauYang: Fixed. Thanks!

modules/shifted_boundary_method/include/SBMBndElement/SBMBndEdge2.h
ChengHauYang: I removed this override. It is right now taking care of inside `LineSegment`.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndElementBase.h
ChengHauYang: Yes, I created a base class called `GeometryBase` and moved intersect into it.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndTri3.h
ChengHauYang: I removed this override. It is right now taking care of inside `Triangle`.

modules/shifted_boundary_method/include/SBMBndElement/SBMBndTri3.h
ChengHauYang: Yes. I moved it.

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
ChengHauYang: Fixed. Thanks!

modules/shifted_boundary_method/src/auxkernels/DistanceAux.C
ChengHauYang: Fixed the class name to be `ElementCentroidToSurfaceDistanceAux `. Thanks for the feedback.

hugary1995:
@kkiesling @shikhar413 @bwspenc This PR is partially related to what we discussed at the annual review meeting. We were thinking the CSG capabilities could be useful in some other modules such as XFEM, levelset, ray tracing etc. where some geometric primitives are needed. This PR is adding a new module named `shifted_boundary_method` which is yet another module that would make use of some geometric primitives.

It would be nice if CSG could support some baseline utilities for common operations on geometries. You can take a look at framework/include/geometries for the types of utilities we need. Do you think those are something CSG can extend?

hugary1995:
@loganharbour I think this PR is in a decent state now. Is there anything we should do before you incorporate this module's testing into the existing civet recipes?

loganharbour:
> @loganharbour I think this PR is in a decent state now. Is there anything we should do before you incorporate this module's testing into the existing civet recipes?

I'm out for a bit starting tomorrow, but I should have time at the very least to get the basic module setup that you guys can add as initial sanity checks. If you don't hear from me by monday on this, feel free to ping me.

kkiesling:
> It would be nice if CSG could support some baseline utilities for common operations on geometries. You can take a look at framework/include/geometries for the types of utilities we need. Do you think those are something CSG can extend?

Oh this definitely looks like something that could be supported in the CSGBase framework. We already have a `CSGSphere` class which is similar to the `Ball` primitive. We don't have an equivalent to a line segment, but it seems like the purpose of this line segment is essentially ray tracing. We don't support ray tracing either/yet but I think it would be reasonable to add support to CSGBase for your needs here for sure. I don't think it would necessarily be a heavy lift either.

Shikhar is on vacation right now, but I think when he gets back lets meet to discuss this in earnest detail. I'll reach out to setup a meeting in the next few weeks.

cc @eshemon

ChengHauYang:
Hi @kkiesling,

Thanks for the feedback!

At a minimum, what we need is the ability to determine whether a  line segment intersects with simple primitives such as a Ball/ Sphere. The motivation is to support a ray-casting algorithm (https://en.wikipedia.org/wiki/Point_in_polygon) for **inside-outside classification (Point-in-Polygon test)** in our SBM geometry handling, which would likely come as a follow-up PR.

Because of this, having baseline intersection queries in CSGBase (even just boolean intersection tests between a line segment and primitives like spheres/triangles/lines) would already cover most of our needs and integrate very naturally with our workflow.

Just sharing this for context!

hugary1995:
Not stale.

loganharbour:
FYI @ChengHauYang @hugary1995: the new module recipes have been added (see the failures)

framework/src/geometries/LineSegment.C
```diff
@@ -140,12 +140,32 @@ LineSegment::intersect(const LineSegment & l, Point & intersect_p) const
...
+  if (v.norm() < tol)
   {
-    // TODO: The lines are co-linear but more work is needed to determine and intersection point
-    //       it could be the case that the line segments don't lie on the same line or overlap only
-    //       a bit
+    // Parallel but not collinear: no intersection
+    if (c.cross(a).norm() >= tol)
+      return false;
+
+    // Collinear: segments overlap if any endpoint lies on the other segment
```
hugary1995, line 156: **Required.** This rewrite of the collinear/parallel branch (and the removal of the later `if (v.is_zero()) return false;`) changes the behavior of the two-argument `intersect(const LineSegment &, Point &)`, which is a framework primitive consumed outside SBM: `framework/src/utils/RayTracing.C:72` calls `line_segment.intersect(side_segment, intersection_point)`. Collinear overlap now returns a *defined* `intersect_p` (previously left unset), which is plausibly a fix but must be intentional. Please confirm the ray-tracing regression tests still pass and add a test that pins the new collinear contract. This change is also the most likely cause of the red **Framework 2** CI job - please check that log before merge. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/src/functions/UnsignedDistanceToSurfaceMesh.C
```diff
@@ -0,0 +1,96 @@
...
+  : Function(parameters)
+{
+}
+
+void
+UnsignedDistanceToSurfaceMesh::initialSetup()
+{
+  const auto builder = &getUserObject<SBMSurfaceMeshBuilder>("builder");
+
+  _kd_tree = const_cast<KDTree *>(&builder->getKDTree());
```
hugary1995: **Required (const-correctness).** `const_cast` to drop constness is disallowed by the Code Standard except in narrow, documented cases. The cast is only needed because `KDTree::neighborSearch` is non-const (`framework/src/utils/KDTree.C:37,46`) while `SBMSurfaceMeshBuilder::getKDTree() const` returns `const KDTree &`. Make `getKDTree()` return a non-const `KDTree &` and store `KDTree *` here without the cast. KDTree search is logically read-only and nanoflann's `knnSearch` is re-entrant, so cross-thread sharing is fine - state that reasoning in a comment. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
```diff
@@ -0,0 +1,82 @@
...
+public:
+  static InputParameters validParams();
+  SBMSurfaceMeshBuilder(const InputParameters & parameters);
+
+  virtual void initialSetup() override;
+  virtual void initialize() override {}
+  virtual void execute() override {}
+  virtual void finalize() override {}
+
+  const KDTree & getKDTree() const { return *_kd_tree; }
```
hugary1995: **Required.** Two issues here:
1. Returning `const KDTree &` forces the `const_cast` in `UnsignedDistanceToSurfaceMesh.C:39` (see that comment). Prefer returning a non-const `KDTree &`.
2. When the builder is configured with `build_kd_tree = false`, `_kd_tree` is a null `unique_ptr`, so `*_kd_tree` dereferences null. A `UnsignedDistanceToSurfaceMesh` pointed at such a builder then takes `&builder->getKDTree()` on a null object (the downstream `mooseAssert` is compiled out in opt). Please `mooseError` here (or in `UnsignedDistanceToSurfaceMesh::initialSetup`) with a clear message when the tree was not built. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/include/utils/SBMUtils.h
```diff
@@ -0,0 +1,62 @@
...
+                                      Real t);
+
+/// Scan all distance functions and return the corresponding normal vector.
+RealVectorValue closestTrueNormalVector(const std::vector<const Function *> & funcs,
+                                        const libMesh::Point & pt,
+                                        Real t);
+}
+
+/// Enum for different distance function types.
+enum class DistanceFunctionType
```
hugary1995: **Required (Simplicity First).** `DistanceFunctionType` is not referenced anywhere in this PR. Please remove it and reintroduce it in the follow-up PR that actually uses it - nothing speculative. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/include/SBMBndElement/SBMBndElementBase.h
```diff
@@ -0,0 +1,98 @@
...
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "MooseMesh.h"
+
+enum class IntersectionType
```
hugary1995: **Required (Simplicity First).** `IntersectionType` is declared but used nowhere in this PR. Please remove it (reintroduce with the in-out test that needs it). _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/src/base/ShiftedBoundaryMethodApp.C
```diff
@@ -0,0 +1,51 @@
+#include "ShiftedBoundaryMethodApp.h"
```
hugary1995, line 10: **Required.** This is the only source file in the PR missing the LGPL license header.

```suggestion
//* This file is part of the MOOSE framework
//* https://mooseframework.inl.gov
//*
//* All rights reserved, see COPYRIGHT for full restrictions
//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
//*
//* Licensed under LGPL 2.1, please see LICENSE for details
//* https://www.gnu.org/licenses/lgpl-2.1.html

#include "ShiftedBoundaryMethodApp.h"
``` _(generated by Claude Opus 4.8)_

File: modules/shifted_boundary_method/src/auxkernels/ElementCentroidToSurfaceDistanceAux.C [hunk 32]
```diff
@@ -0,0 +1,50 @@
...
+// Register the new AuxKernel class name
+registerMooseObject("ShiftedBoundaryMethodApp", ElementCentroidToSurfaceDistanceAux);
+
+InputParameters
+ElementCentroidToSurfaceDistanceAux::validParams()
+{
+  InputParameters params = AuxKernel::validParams();
+
+  params.addParam<UserObjectName>("distance_to_surface",
+                                  "UserObject that provides distance calculations.");
```
hugary1995: **Required.** `distance_to_surface` is declared optional but the constructor calls `getUserObject<ShortestDistanceToSurface>("distance_to_surface")` unconditionally, so omitting it yields a confusing lookup failure. Make it required (the test-only aux already does).

```suggestion
  params.addRequiredParam<UserObjectName>(
      "distance_to_surface", "UserObject that provides distance calculations.");
``` _(generated by Claude Opus 4.8)_

File: modules/shifted_boundary_method/src/userobjects/ShortestDistanceToSurface.C [hunk 33]
```diff
@@ -0,0 +1,74 @@
...
+registerMooseObject("ShiftedBoundaryMethodApp", ShortestDistanceToSurface);
+
+InputParameters
+ShortestDistanceToSurface::validParams()
+{
+  InputParameters params = ElementUserObject::validParams();
+  params.addRequiredParam<std::vector<FunctionName>>("surfaces",
+                                                     "Level-set function for distance.");
+
+  params.addClassDescription("Base distance user object.");
```
hugary1995: **Required (docs).** `addClassDescription("Base distance user object.")` is what renders in the syntax tables via `!syntax description`, and it contradicts the prose in `ShortestDistanceToSurface.md`. The `surfaces` help string is also singular/incomplete (it is a list and also accepts `UnsignedDistanceToSurfaceMesh`).

```suggestion
  params.addRequiredParam<std::vector<FunctionName>>(
      "surfaces",
      "Functions defining the surfaces to measure distance to; each must be a ParsedFunction "
      "level set or an UnsignedDistanceToSurfaceMesh.");

  params.addClassDescription(
      "Computes the shortest distance vector and true normal to one or more surfaces defined by "
      "parsed level-set functions and/or mesh-based distance functions.");
``` _(generated by Claude Opus 4.8)_

File: framework/include/geometries/Ball.h [hunk 34]
```diff
@@ -0,0 +1,53 @@
...
+
+#pragma once
+
+#include "GeometryBase.h"
+#include "libmesh/point.h"
+#include "libmesh/sphere.h"
+
+class LineSegment;
+
+using namespace libMesh;
```
hugary1995: **Required (standard) + reuse question.** `using namespace libMesh;` at namespace scope in a header leaks into every TU that includes `Ball.h` (now many, transitively via `LineSegment.h`). `Triangle.h` correctly qualifies `libMesh::Point` - do the same here and drop the directive.

Separately: `Ball` overlaps heavily with `libMesh::Sphere` (you already provide `toSphere()`). Please justify in the PR/design why a new framework primitive is needed rather than reusing/extending `libMesh::Sphere`. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/include/SBMBndElement/SBMBndElementBase.h
```diff
@@ -0,0 +1,98 @@
...
+  SBMBndElementBase(const Elem * elem);
+
+  /// Virtual destructor to ensure proper cleanup in derived classes
+  virtual ~SBMBndElementBase() = default;
+
+  /// Getter for the underlying element
+  const Elem & elem() const { return *_elem; }
+
+  /// Getter for the normal vector
+  const Point & normal();
```
hugary1995: **Required.** `normal()` mutates only `mutable` members and should be `const` (as `distanceFrom()` already is); that lets `UnsignedDistanceToSurfaceMesh::closestBoundaryElem` return `const &` too. More importantly, the boundary elements are owned once by the single builder and shared across threads, so the first concurrent `normal()` calls race on `_normal`/`_is_normal_initialized`. Since the elements are built once in `initialSetup`, compute the normal eagerly in the derived constructors (which already know their geometry) and store it `const` - this removes the `mutable`, the lazy init, and the race together. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/test/tests/distance_calc/tests
```diff
@@ -0,0 +1,102 @@
...
+  [distance_test]
+    requirement = 'The system shall compute '
+
+    [2D_distance_refine]
+      type = 'Exodiff'
+      input = '2D_distance_refine.i'
+      exodiff = '2D_distance_refine_out.e-s004'
+      mesh_mode = REPLICATED # SBMSurfaceMeshBuilder requires a serial mesh for its In-Out test
+      recover = false
+      detail = 'the distance field is computed correctly in 2D.'
```
hugary1995: **Required (tests).** With the stem `requirement = 'The system shall compute '`, this detail reads as "The system shall compute the distance field is computed correctly in 2D." Reword so it completes the stem cleanly. The `3D_distance_refine` detail just below has the same problem; please fix it the same way.

```suggestion
      detail = 'the distance field in 2D.'
``` _(generated by Claude Opus 4.8)_

**Review by hugary1995 — COMMENTED** [review]
## MOOSE PR Review (AI-generated)

Thanks for this - the overall architecture (KDTree-backed surface builder + distance Function/UserObject + visualization Aux, on a new `framework/geometries` primitive layer) is a reasonable foundation, docs/SQA scaffolding is complete, and valgrind is green. My recommendation is **changes-required** before merge: a framework primitive behavior change with cross-module blast radius, a const-correctness `const_cast`, some dead/speculative code, a few correctness/robustness gaps, and a scope question. Line-level findings are in the inline comments; cross-cutting items are below.

### Please resolve CI first
**Framework 2 = FAILURE** and **Coverage = ERROR** (all other jobs green). I could not read the failing test from the civet log. The `LineSegment::intersect` change (inline comment) is the most likely culprit since it alters a primitive used by ray tracing - please check.

### Reuse / reinvention (Code Standard: use existing functionality)
Beyond the `Ball` vs `libMesh::Sphere` point (inline), MOOSE already has a lot of nearest/distance/penetration machinery: `NearestNodeLocator`, `PenetrationLocator`/`PenetrationAux` (distance *and* normals between surfaces), `NearestNodeDistanceAux`, the `NearestPoint*` family, and libMesh `PointLocator`. The SBM case (distance from background-mesh points to a *standalone saved* surface mesh, not a sideset of the simulation mesh) is plausibly not served by those - but please state that justification explicitly in the PR/design so a reviewer doesn't have to infer it, and reuse/extend where any piece does fit.

### Tests & SQA
Good breadth (2D/3D single + multi-surface Exodiff, five `RunException` error paths, coverage omnibus, two gtest suites incl. the unsupported-geometry error branches); gold files present; `recover=false`/`mesh_mode=REPLICATED` justified inline. Fix: the requirement grammar and the oversized `bunny_2D.msh` (inline). The `coverage/tests` requirements are phrased as implementation branches ("exercising the zero-distance early-return branch in gradient()") rather than user-facing requirements - prefer requirement language since every test is a requirement.

### Documentation
Stub pages for all four registered objects with real Overview prose, plus the full SQA document set - nicely done. Fix the `ShortestDistanceToSurface` class description (inline) so the rendered description matches its page. A **newsletter entry is missing**: a whole new module warrants a short paragraph in `modules/doc/content/newsletter/2026/2026_06.md`.

### Correctness/security (deferred)
I did not duplicate deep correctness analysis. I recommend running `/code-review` focused on (a) the `LineSegment::intersect` collinear rewrite + its effect on `RayTracing.C`, and (b) the Moller-Trumbore `Triangle::intersect` tolerances. A dedicated security review is low-value here (no untrusted parsing beyond standard mesh I/O, RAII throughout, valgrind clean).

### Minor (non-blocking)
`_elem`/`_normal` are `protected` data members (Standard prefers non-public data; `_elem` can be `private`); `SBMSurfaceMeshBuilder.h` includes `SBMBndEdge2.h`/`SBMBndTri3.h`/`SBMUtils.h` that belong in the `.C`; `ShortestDistanceToSurface` is an `ElementUserObject` that never executes and ignores the current element (a `GeneralUserObject` may fit better).

_Review generated by Claude Opus 4.8 via Claude Code, at the request of @hugary1995._

File: modules/shifted_boundary_method/test/tests/distance_calc/tests [hunk 37]
```diff
@@ -0,0 +1,102 @@
...
+      mesh_mode = REPLICATED
+      recover = false
+      detail = 'if ShortestDistanceToSurface receives a function whose type is not ParsedFunction '
+               'or UnsignedDistanceToSurfaceMesh.'
+    []
+
+    [dim_mismatch]
+      type = 'RunException'
+      input = '3D_distance_refine.i'
+      cli_args = 'Mesh/boundary_mesh/file=bunny_2D.msh'
```
hugary1995: **Required (tests).** This `dim_mismatch` case loads `bunny_2D.msh` (~345k lines) purely to trigger the "does not match the expected mesh dimension" error, which fires before any geometry is consumed. Committing a 345k-line asset for a one-line error check bloats the repo - please swap in a trivially small 2D `.msh` (as `quad_surface.msh` does for the unsupported-element-type case), and confirm `bunny.msh` / `star.msh` are the minimum needed for the real distance tests. _(generated by Claude Opus 4.8)_

modules/modules.mk
```diff
@@ -250,10 +255,19 @@ ifeq ($(SOLID_PROPERTIES),yes)
...
+  APPLICATION_NAME   := shifted_boundary_method
+  SUFFIX             := sbm
+  include $(FRAMEWORK_DIR)/app.mk
+endif
+
 # Depended on by contact, fsi, misc, peridynamics, phase_field, porous_flow, xfem
 ifeq ($(SOLID_MECHANICS),yes)
   APPLICATION_DIR    := $(MOOSE_DIR)/modules/solid_mechanics
   APPLICATION_NAME   := solid_mechanics
+  DEPEND_MODULES     := shifted_boundary_method
```
hugary1995: **Required (scope).** This makes `solid_mechanics` (and everything downstream: contact, fsi, peridynamics, phase_field, porous_flow, xfem, combined) depend on the brand-new `shifted_boundary_method`, but no `solid_mechanics` source in this PR references SBM. Per Simplicity First, defer this dependency wiring to the follow-up PR that actually introduces the SCZM solid-mechanics coupling, or justify why it must land now. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/src/SBMBndElement/SBMBndElementBase.C
```diff
@@ -0,0 +1,166 @@
...
+
+bool
+SBMBndElementBase::intersect(const LineSegment & line_segment) const
+{
+  if (const auto * edge = dynamic_cast<const LineSegment *>(this))
+    return edge->intersect(line_segment);
+  if (const auto * tri = dynamic_cast<const Triangle *>(this))
+    return tri->intersect(line_segment);
+
+  mooseError("SBMBndElementBase::intersect: unsupported geometry type");
```
hugary1995, line 130: **Suggestion.** This per-type `dynamic_cast<const LineSegment*>` / `<const Triangle*>` ladder (here and in `computeBoundingBall` below) reimplements virtual dispatch that `GeometryBase` already provides: both `LineSegment` and `Triangle` derive from `GeometryBase` and override `intersect`/`computeBoundingBall`. A single `dynamic_cast<const GeometryBase*>(this)` would dispatch through the existing virtuals and extend automatically to future element types, removing the ladder and the `mooseError` tail. _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/src/functions/UnsignedDistanceToSurfaceMesh.C
```diff
@@ -0,0 +1,96 @@
...
+{
+  const RealVectorValue dv = distanceVectorToSurface(p);
+  const Real dist = dv.norm();
+
+  if (dist <= libMesh::TOLERANCE)
+    return RealGradient(0, 0, 0);
+
+  // dv points from the query point toward the nearest surface.
+  // The distance gradient must point in the direction of decreasing distance,
+  // hence the negative normalization: gradient = - (distance_vector) / |distance_vector|
```
hugary1995: **Suggestion (comment accuracy).** The code is correct (`-dv` points away from the surface), but the comment's justification is backwards: a distance gradient points toward *increasing* distance, not decreasing.

```suggestion
  // dv points from the query point toward the nearest surface, so -dv points away from it.
  // The gradient of a distance field points toward increasing distance (away from the surface),
  // matching the signed-distance gradient convention used by SBMUtils::distanceVectorFromFunction.
``` _(generated by Claude Opus 4.8)_

File: modules/shifted_boundary_method/include/userobjects/ShortestDistanceToSurface.h [hunk 41]
```diff
@@ -0,0 +1,58 @@
...
+
+  /// A true normal vector at the projected point of \p pt onto a specific surface specified
+  /// by index \p idx in the 'surfaces' parameter list.
+  RealVectorValue trueNormalByIndex(unsigned int idx, const Point & pt) const;
+
+  /// Compute distance vector using a specific distance function
+  const RealVectorValue distanceVectorByFunc(const Point & pt, Real t, const Function * func) const;
+
+  /// Compute true normal using a specific distance function
+  const RealVectorValue trueNormalByFunc(const Point & pt, Real t, const Function * func) const;
```
hugary1995: **Suggestion (scope).** These accessors (`trueNormal`, `*ByIndex`, `*ByFunc`) have no production callers; only the test-only `ShortestDistanceToSurfaceTestAux` exercises them, by its own description "without adding production callers." Consider trimming to what `ElementCentroidToSurfaceDistanceAux` actually uses (`distanceVector`) and reintroducing the rest with the SCZM kernels that consume them. If you keep them, note why in the PR so they don't read as dead code. (Also: `distanceVectorByFunc`/`trueNormalByFunc` return `const RealVectorValue` by value - the top-level `const` is a no-op; drop it.) _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
```diff
@@ -0,0 +1,82 @@
...
+  /// Get the SBM boundary elements
+  const std::vector<std::unique_ptr<SBMBndElementBase>> & getBoundaryElements() const;
+
+  /// Get the centroids of the boundary elements
+  const std::vector<Point> & getCentroids() const;
+
+  /**
+   * @brief Checks whether the boundary mesh is watertight.
+   *
+   * For a 3D surface mesh, watertightness means that every edge (defined by two nodes)
```
hugary1995: **Suggestion (doc accuracy).** This doxygen describes edge-sharing counts and node degree, but the implementation (`SBMUtils::checkWatertightnessFromRawElems`) simply checks that every side has a neighbor (`!el->neighbor_ptr(s)`). Please align the comment with the code (or strengthen the code if the richer check was intended). _(generated by Claude Opus 4.8)_

modules/shifted_boundary_method/include/userobjects/ShortestDistanceToSurface.h
ChengHauYang: Thanks! Kept these intentionally. 
The SCZM kernels that consume trueNormal / *ByIndex / *ByFunc are landing in the coming PR. The coverage test (ShortestDistanceToSurfaceTestAux + tests/coverage) is there to lock the API shape until those callers arrive.

Dropped the no-op top-level const on distanceVectorByFunc / trueNormalByFunc.

framework/src/geometries/LineSegment.C
ChengHauYang: 1. Confirm: Ray-tracing regression tests still pass
2. Added a test that pins the new collinear contract
3. Red Framework 2 CI job is because of `bcs/periodic.testperiodic_array` not related to this PR.

modules/shifted_boundary_method/src/functions/UnsignedDistanceToSurfaceMesh.C
ChengHauYang: Yes. We made `getKDTree()` return a non-const `KDTree &` and removed `const_cast`.

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
ChengHauYang: 1. Yes. Fixed.
2. We would `mooseError` out if the user calls `build_kd_tree = false` but the user also calls `getKDTree()`. We also do `mooseError` if the user wants to use `UnsignedDistanceToSurfaceMesh ` but sets `build_kd_tree = false` inside the `builder`.

modules/shifted_boundary_method/include/utils/SBMUtils.h
ChengHauYang: Removed. Thanks!

modules/shifted_boundary_method/include/SBMBndElement/SBMBndElementBase.h
ChengHauYang: Removed. Thanks!

modules/shifted_boundary_method/src/base/ShiftedBoundaryMethodApp.C
ChengHauYang: Added. Thanks!

modules/shifted_boundary_method/src/auxkernels/ElementCentroidToSurfaceDistanceAux.C
ChengHauYang: Sure. Fixed it to be a required parameter.

modules/shifted_boundary_method/src/userobjects/ShortestDistanceToSurface.C
ChengHauYang: I copied your suggestions to my code. Thanks!

framework/include/geometries/Ball.h
ChengHauYang: 1. Yes. I removed `using namespace libMesh; ` and used `libMesh::Point ` instead.
2. We added some justification to `Ball.h`:
```
 * Why not just use libMesh::Sphere?
 *  - libMesh::Sphere is 3D only; Ball supports both 2D (circle) and 3D (sphere) in a single type.
 *  - Ball provides the operations the framework geometries layer needs:
 *    intersection with a LineSegment and a tight bounding ball used for fast
 *    nearest-neighbor / spatial searches. libMesh::Sphere exposes a different
 *    set of surface queries (closest point, normal, above/below surface) that
 *    we do not need here.
 *  - Ball derives from GeometryBase so every framework geometry primitive can
 *    be used through one virtual interface.
```

modules/shifted_boundary_method/include/SBMBndElement/SBMBndElementBase.h
ChengHauYang: Sure. I followed your suggestions -- "compute the normal eagerly in the derived constructors and store it const".

modules/shifted_boundary_method/test/tests/distance_calc/tests
ChengHauYang: Fixed. Thanks!

modules/shifted_boundary_method/test/tests/distance_calc/tests
ChengHauYang: We simplify the tests that use `bunny_2D.msh ` and `star.msh ` to just use `quad_surface.msh`. And, `bunny.msh` (3D) to use `cube_surface.msh`.

modules/modules.mk
ChengHauYang: I just deleted this. I will land it later after we touched something in `solid_mechanics`.

modules/shifted_boundary_method/src/SBMBndElement/SBMBndElementBase.C
ChengHauYang: Sure. Now, I collapse the per-type ladder to a single dispatch.

modules/shifted_boundary_method/src/functions/UnsignedDistanceToSurfaceMesh.C
ChengHauYang: Sure. I copied your text. Thanks!

modules/shifted_boundary_method/include/userobjects/SBMSurfaceMeshBuilder.h
ChengHauYang: Thanks. I aligned the doxygen to what the code actually checks. The implementation walks every side and returns false on the first missing neighbor, which, on a manifold surface mesh, is equivalent to watertightness. The richer edge-count / node-degree check isn't needed here because the SBM workflow expects boundary meshes extracted from manifold sidesets; if a future use case requires validating non-manifold input (T-junctions, three faces sharing an edge), we can strengthen the check then. Updated the doc to call that out.

ChengHauYang:
Some feedback from me and Opus 4.7 helped me generate this reply:

Thanks for the thorough review — pretty much everything is addressed in the latest pushes. Mirroring your sections:

### CI
The Framework 2 failure is the pre-existing `testperiodic_array` flake on `next`, not from this PR; Coverage = ERROR was a CI dependency-resolution issue, also unrelated. Both should clear on the next rerun.

### Reuse / reinvention
- **`Ball` vs `libMesh::Sphere`**: added a class-level doxygen on `framework/include/geometries/Ball.h` spelling out the three reasons we don't reuse `libMesh::Sphere`: (1) it's 3D-only while `Ball` spans both 2D circles and 3D spheres, (2) it exposes surface-query operations (`closest_point`, `unit_normal`, `above_surface`/`below_surface`) rather than the intersection / bounding-ball operations the framework geometries layer needs, and (3) `Ball` derives from `GeometryBase` so all primitives dispatch through one virtual interface. `toSphere()` stays as an explicit adapter.
- **Broader nearest/distance machinery** (`NearestNodeLocator`, `PenetrationLocator`, `NearestPoint*`, `PointLocator`): they all operate on sidesets of the simulation mesh, while SBM needs distance from background-mesh points to a *standalone saved* surface mesh — the two cases don't overlap in inputs. Will add this justification paragraph to the PR description.

### Tests & SQA
- `coverage/tests` details rewritten in user-facing language (e.g. `gradient_on_surface` is now "return zero distance when the query point lies on the surface mesh" instead of "exercising the zero-distance early-return branch in `gradient()`"); same pass to the others.
- `distance_calc/tests` details tightened to your suggested compact form ("the distance field in 2D.", etc.).
- `bunny_2D.msh` removed; `bunny.msh` and `star.msh` also gone, replaced by trivial `cube_surface.msh` (12 TRI3, 387 B) and `square_boundary.msh` (4 EDGE2, 191 B). `dim_mismatch` now reuses `quad_surface.msh` under `2D_distance_refine.i`; `unsupported_element_type` continues to use `quad_surface.msh` under `3D_distance_refine.i` (same asset, two distinct error paths). Gold regenerated. Net: ~-16.5 MB / -513k lines.

### Documentation
- `ShortestDistanceToSurface` class description and `surfaces` help text fixed inline.
- Newsletter entry for `modules/doc/content/newsletter/2026/2026_06.md` will go in a follow-up once this lands.

### Correctness & robustness (inline threads)
- `LineSegment::intersect` collinear contract pinned by a new gtest (parallel-not-collinear, collinear disjoint, single-endpoint touch, partial overlap, full containment, identical, reversed, 3D diagonal).
- `getKDTree()` returns non-const, `mooseError`s when the tree wasn't built, and `UnsignedDistanceToSurfaceMesh::initialSetup` adds a friendly `hasKDTree()` precheck naming both objects; `const_cast` and opt-stripped `mooseAssert`s gone.
- `SBMBndElementBase::normal()` is now `const` and eager: derived constructors compute the unit normal once and pass it to the base, which stores it `const`. Removes the `mutable`/lazy-init pattern and the race on shared boundary elements.
- Per-type `dynamic_cast` ladder in `SBMBndElementBase.C` collapsed to a single `dynamic_cast<const GeometryBase*>(this)` that dispatches through the virtual `intersect` / `computeBoundingBall` — auto-extends to future primitives, and the `mooseError` tail is gone.
- Gradient comment in `UnsignedDistanceToSurfaceMesh::gradient` rewritten per your suggestion (distance gradient points toward *increasing* distance, away from the surface).

### Simplicity (dead-code / scope)
- Removed `DistanceFunctionType` (`SBMUtils.h`) and `IntersectionType` (`SBMBndElementBase.h`).
- Dropped `solid_mechanics → shifted_boundary_method` dependency from `modules.mk`.
- LGPL header added to `ShiftedBoundaryMethodApp.C`.
- `ElementCentroidToSurfaceDistanceAux::distance_to_surface` is now `addRequiredParam`.
- Keeping `trueNormal` / `*ByIndex` / `*ByFunc` accessors as-is — the SCZM interface-kernel PR that consumes them is going up shortly, so they'll have production callers in the very next PR rather than sitting unused.

### Minor
- `_elem` and `_normal` moved from `protected` to `private` in `SBMBndElementBase` (derived classes go through `elem()` / `normal()`).
- `SBMSurfaceMeshBuilder.h` includes pruned: `SBMBndEdge2.h`, `SBMBndTri3.h`, `SBMUtils.h` moved to the `.C`.
- `ShortestDistanceToSurface` now inherits from `GeneralUserObject` (was `ElementUserObject`) — only exposes query APIs, never used the current element.

### Framework geometries layer
- Dropped `using namespace libMesh;` from `Ball.h` (was leaking through every TU that transitively included `LineSegment.h`); qualified `libMesh::Point` / `libMesh::Real` in the header. The three `.C` files (`Ball.C`, `LineSegment.C`, `Triangle.C`) carry scoped `using libMesh::*;` declarations matching `Triangle.C`'s existing style.

Local `./run_tests --re='distance_calc|coverage'` is green (13/13) on the latest build.

ChengHauYang:
Hi @hugary1995,

I forced pushed the one without any unwanted files in the commit history. The CI failed case is related to `misc/signal_handler.test_signal`, which is not related to this PR.

This PR is good to merge.

Thanks!

hugary1995:
@loganharbour  This is ready from my perspective. The `misc/signal_handler.test_signal` intermittent failure is the only thing preventing the last check to turn green. Since this PR is related to some CIVET recipe, I'll let you control the merge button.

There's a pending review request to @bwspenc . I don't suppose he can get to this in a timely manner, so we can merge this without his approval. Note this is only part I of a series of stacked PR, so he will be able to chime in down the road. This module is tied to a milestone, so I'd like to avoid any potential delay.

hugary1995 (approved):

PR #29946, opened 2025-02-25
Implement parallel active learning capabilities

somu15:
### Motivation

Intelligently select under which inputs to run the computational model under a parallel active learning scheme, facilitated by a Gaussian process. Both saves computational cost and results in surrogates with better accuracy due to intelligent selection of inputs.  

### Design

The following new object type will be created: `ParallelAcquisitionFunction`

The following new sampler classes will be created: `GenericActiveLearningSampler`, `BayesianActiveLearningSampler`, `GPAffineInvariantDES`

The following new reporter classes will be created: `GenericActiveLearning`, `BayesianActiveLearner`, `GPAffineInvariantDifferentialDecision`

### Impact

No expected impact to existing objects.

closes #28196

PR #31943, opened 2025-11-19
Fix unitialized value in #29946

somu15:
Addressing a valgrind failure due to #29946


somu15:
@loganharbour Here is the fix.

loganharbour (approved):

PR #31945, opened 2025-11-20
Print range function in documentation for ranged-checked params

### LINKED ISSUE #31944: Add range for range-checked parameters in documentation
### Motivation

To close out a corrective action from the Grizzly/BlackBear NQA-1 assessment we need to include the valid range of parameters in the documentation. It's also just a good idea.

We have gone through and switched a lot of parameters to be range-checked, but the ranges are not yet shown in the documentation. We need some support for this in MooseDocs.

### Design

The easiest way to do this is to just print out the string defining the range-checking function that is supplied when range-checked parameters are defined in `addRangeCheckedParam()`. We could get fancier than that, but this will provide users all the needed information.

### Impact

Should only add new capability, with no impact on existing functionality.

bwspenc:
closes #31944

### DIFF
--- a/framework/doc/content/css/moose.css
+++ b/framework/doc/content/css/moose.css
@@ -608,6 +608,7 @@ li{
 .moose-parameter-description-default span,
 .moose-parameter-description-cpptype span,
 .moose-parameter-description-doc-unit span,
+.moose-parameter-description-doc-range span,
 .moose-parameter-description-options span,
 .moose-parameter-description-controllable span{
     padding-right:10px;
--- a/framework/src/outputs/formatters/YAMLFormatter.C
+++ b/framework/src/outputs/formatters/YAMLFormatter.C
@@ -108,6 +108,14 @@ YAMLFormatter::printParams(const std::string & prefix,
       oss << "'" << doc_unit << "'";
     oss << "\n";
 
+    oss << indent << "    doc_range: ";
+    std::string doc_range;
+    if (params.isRangeChecked(iter.first))
+      doc_range = params.rangeCheckedFunction(iter.first);
+    if (!doc_range.empty())
+      oss << "'" << doc_range << "'";
+    oss << "\n";
+
     if (params.have_parameter<MooseEnum>(name))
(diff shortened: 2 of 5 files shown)


cticenhour:
Thank you @bwspenc! Will look at this in the morning.

cticenhour (changes requested):
No concerns about the code itself. Since I can't add a suggestion to code that hasn't been touched yet, I will do so here. Regarding testing, it will need to be added in https://github.com/idaholab/moose/blob/ecf7a09933f550b625784f9d5a71d20985022261/python/MooseDocs/test/extensions/test_appsyntax.py#L44-L45

I would add a new `TEXT_RANGE` variable on the line below `TEXT`, using something from the framework:
```
TEXT_RANGE = "!syntax parameters /Functions/PeriodicFunction"
```

Then below `testMaterialize` within this class, I would add a new method, `testMaterializeRange`, that only tests this capability (the rest of the parameter functionality is tested using the former `TEXT` test). So we'd have:

```python
    def testMaterializeRange(self):
        _, res = self.execute(self.TEXT_RANGE, renderer=base.MaterializeRenderer())
        # Much of this is tested in the other methods in this class. This is meant to check
        # specifically for range checked param output in the documentation.
        self.assertSize(res, 7)
        self.assertHTMLTag(res(0), 'h2', id_='input-parameters', size=3)
        self.assertEqual(res(0).text(), 'Input Parameters')

        self.assertHTMLTag(res(4), 'ul', size=4, class_='collapsible')
        self.assertHTMLTag(res(4,0,0,0), 'span', class_='moose-parameter-name', string='period_time')
        self.assertHTMLTag(res(4,0,1,3), 'p', size=2, class_='moose-parameter-description-doc-range')
        self.assertEqual(res(4,0,1,3,0).text(), 'Range:')
        self.assertEqual(res(4,0,1,3,1)['content'], 'period_time>0')
```

cticenhour (approved):

cticenhour:
Tests with pending status have all passed successfully (https://civet.inl.gov/event/268597/). All changes to C++ are tested by these Python-based unit tests, so the Coverage failure is not a concern.

PR #31954, opened 2025-11-20
Catch when the server is shutting down

loganharbour:
## Reason
See https://civet.inl.gov/job/3397863/ for an example; we could catch the server just as it's shutting down. This should be considered the server not listening.

## Design
Catch the error "Control is no longer available" and consider it not listening.

## Impact
Catches an edge case. Will result in more consistent testing.


grmnptr (approved):
Looks good to me!

PR #31960, opened 2025-11-20
Add TH adjustments to facilitate fusion blanket modeling in SALAMANDER

### LINKED ISSUE #31958: Allow adjustment of the number of elements in each axial section by child classes in DiscreteLineSegmentInterface
### Motivation

When building child classes on top of DiscreteLineSegmentInterface, it can be useful to be able to set the number of elements in each axial section based on an external input (such as when pulling information from a file). Right now, this value is a `const` reference. Removing both the `const` and the reference would allow for this re-assignment. This value is already protected, so only child/inherited classes can make this adjustment.

This is an issue broken out of changes related to https://github.com/idaholab/moose/pull/30122 (which is being closed and moved to SALAMANDER).

### Design

Remove `const ... $` from `_n_elems` in `DiscreteLineSegmentInterface` and `const` from `_n_elem` (the total number of axial elements variable).

### Impact

More flexibility for child/inherited classes for this object.

cticenhour:
Closes #31958 

@joshuahansel Could you give this a review? ~Let me know if you want me to expand the new `Nu_scale` parameter to the non-AD version of the Dittus-Boelter material as well for consistency, or take some other approach.~ EDIT: This has been removed from the PR (see discussion below). This is mostly me pulling some Fande changes from the former "fusion module" PR and adding some testing/documentation.

Refs #30122

### DIFF
--- a/modules/thermal_hydraulics/include/interfaces/DiscreteLineSegmentInterface.h
+++ b/modules/thermal_hydraulics/include/interfaces/DiscreteLineSegmentInterface.h
@@ -102,9 +102,9 @@ protected:
   Real _length;
 
   /// Number of elements in each axial section
-  const std::vector<unsigned int> & _n_elems;
+  std::vector<unsigned int> _n_elems;
   /// Total number of axial elements
-  const unsigned int _n_elem;
+  unsigned int _n_elem;
 
   /// Number of axial sections
   const unsigned int _n_sections;


modules/thermal_hydraulics/include/interfaces/DiscreteLineSegmentInterface.h
```diff
@@ -102,9 +102,9 @@ class DiscreteLineSegmentInterface
   Real _length;
 
   /// Number of elements in each axial section
-  const std::vector<unsigned int> & _n_elems;
+  std::vector<unsigned int> _n_elems;
   /// Total number of axial elements
-  const unsigned int _n_elem;
+  unsigned int _n_elem;
```
joshuahansel, line 107: What is this about?

joshuahansel (changes requested):
Since I don't think that the scaling factor applies to Dittus Boelter alone, but to all HTCs: is there a way to accomplish this externally, like with a factor in your kernel(s) or another material property using ParsedMaterial for the enhanced HTC?

modules/thermal_hydraulics/include/interfaces/DiscreteLineSegmentInterface.h
joshuahansel: Nevermind, I see the reference issue now. This is fine.

cticenhour:
I could see this applying more broadly for sure (and it would lend itself to being applied at the level of SALAMANDER, as you said - using a Parsed Material or custom child class). The choice of Dittus-Boelter specifically here was made years ago at this point, and I don't know that reason. To me, a considerable part of this PR was to be faithful to some of the history and start a review conversation to get your input about what we actually wanted to do with it. 

I do not mind removing that change from this PR and closing the issue I opened for it. Are you then OK with the `DiscreteLineSegmentInterface` change remaining? I was looking for an alternate pathway without a larger refactor and I didn't see an obvious one at the moment. But I think that's a candidate for a future PR to clean up code now in a SALAMANDER branch.

joshuahansel:
Roger that, and yeah the DiscreteLineSegmentInterface change is fine, so this will be a boring PR 😄

joshuahansel (approved):

cticenhour:
Failures have been reviewed and are unrelated.

PR #31969, opened 2025-11-24
General improvements for MFEM auxkernels

nmnobre:
## Reason/design
- Generalises/refactors MFEM cross and inner product auxkernels to:
  - take any two vector coefficients (as opposed to just two vector variables)
  - allow scaling by an arbitrary scalar coefficient (as opposed to a scalar constant)
  - avoid pre-writing zeroes to the result gridfunction (for performance)
  - avoid creating coefficients we already have (for performance)
  - align naming with exiting inner product postprocessor and future complex auxkernel
- Removes map type input parameter (since, due to a bug, we were always using the default anyway)
- Fixes minor indentation inconsistencies
- Removes unnecessary header inclusions
- Removes summary section from all markdown documentation for MFEM classes
- Fixes FEC high order specifications
- Adds code to print MFEM version to the console output

## Impact
Simpler yet more general implementation.


framework/doc/content/source/mfem/auxkernels/MFEMInnerProductAux.md
```diff
@@ -0,0 +1,33 @@
+# MFEMInnerProductAux
+
+!if! function=hasCapability('mfem')
+
+## Summary
+
+!syntax description /AuxKernels/MFEMInnerProductAux
```
lindsayad: we generally omit this from the `.md` page since documentation such as in your `Overview` typically is a superset of the class description

framework/include/mfem/actions/AddMFEMComplexBCComponentAction.h
```diff
@@ -10,7 +10,6 @@
 #ifdef MOOSE_MFEM_ENABLED
 
 #pragma once
-
```
lindsayad, line 13: Having the newline here is typical in MOOSE

framework/include/mfem/auxkernels/MFEMCrossProductAux.h
```diff
@@ -9,16 +9,15 @@
 
 #ifdef MOOSE_MFEM_ENABLED
 #pragma once
-
 #include "MFEMAuxKernel.h"
-#include "mfem.hpp"
 
 /**
- * Project s(x) * (U x V) onto a vector MFEM auxvariable.
+ * Project s \vec u \times \vec v onto a vector MFEM auxvariable.
```
lindsayad: Is this going to be rendered correctly?

framework/include/mfem/actions/AddMFEMComplexBCComponentAction.h
lindsayad: but it's not a huge deal to me

framework/include/mfem/auxkernels/MFEMInnerProductAux.h
```diff
@@ -0,0 +1,39 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#ifdef MOOSE_MFEM_ENABLED
+#pragma once
+#include "MFEMAuxKernel.h"
+
+/**
+ * Project s \vec u \cdot \vec v onto a scalar MFEM auxvariable.
```
lindsayad: need math formatting?

framework/doc/content/source/mfem/auxkernels/MFEMInnerProductAux.md
nmnobre: Do you mean the `## Summary` heading itself? Because the syntax description is definitely a common feature and is what you get in the syntax index. As for the heading, it's missing in like 10 files, I can add it (I just checked, it was actually already on my todo list).

framework/include/mfem/actions/AddMFEMComplexBCComponentAction.h
nmnobre: Hehe, well, it's no longer "typical", but "universal". :P
I initially removed the line because it implied fewer changes for consistency within the MFEM subset, but now went full blast in the opposite direction.

framework/include/mfem/auxkernels/MFEMCrossProductAux.h
nmnobre: Good catch.

framework/include/mfem/auxkernels/MFEMInnerProductAux.h
nmnobre: Indeed.

framework/doc/content/source/mfem/auxkernels/MFEMInnerProductAux.md
lindsayad: The summary heading and the syntax description. Right, the description is used in the syntax index, but it doesn't need to be used on the class markdown page. I'm sure we have syntax descriptions on a very large number of our class markdown pages, perhaps even a majority, but I would say it's our best practice to not have it. (Clearly the best practice is often not followed)

lindsayad (approved):

framework/doc/content/source/mfem/auxkernels/MFEMInnerProductAux.md
nmnobre: Mind. Blown. 🤯

> the description is used in the syntax index, but it doesn't need to be used on the class markdown page

This is the point where I embarrassingly admit I thought the syntax index grabbed the first non-header line from the md to fill up its small description. 🙈

If that's not the case, I totally agree we should avoid the redundancy. I removed all such cases for the MFEM classes. EDIT: confirmed everything still looks fine on the newly generated website, thanks for the tip! 😊

lindsayad: No problem, thanks for the fixups!

framework/src/mfem/fespaces/MFEMScalarFESpace.C
```diff
@@ -55,10 +48,6 @@ MFEMScalarFESpace::getFECName() const
   // This is to get around an MFEM bug (to be removed in #31525)
   basis = (basis == "@G" || basis == "_T0") ? "" : basis;
 
-  if (_fec_map == "INTEGRAL")
-    return "L2Int" + basis + "_" + std::to_string(getProblemDim()) + "D_P" +
-           std::to_string(_fec_order);
-
```
alexanderianblair, line 61: Just to document here our offline discussion - I agree it makes more sense to add `L2Int` into the set of allowed `fec_types` rather than allowing the map to be separately specified - but that can be added in a follow-up PR, since it'd be best to add unit tests for that there

alexanderianblair (approved):

framework/src/mfem/fespaces/MFEMScalarFESpace.C
nmnobre: Addressed in #32021.

PR #31973, opened 2025-11-24
Fix uninitialized variable in hit; bumps wasp, neml2, moose-dev

### LINKED ISSUE #31972: Fix uninitialized variable in hit
### Motivation

The variable here: https://github.com/idaholab/moose/blob/5685ba093d40c41731669799c9692946167942b6/framework/contrib/hit/include/hit/parse.h#L132

does not have an initial value.

### Design

Set the initial value to nullptr.

### Impact

Enforces deterministic behavior.

loganharbour:
Closes https://github.com/idaholab/moose/issues/31972

Due to the insanity of how we use it, we need to update neml2. The neml2 hit header is used within moose builds when moose is built with neml2. Which is why this needs a neml2 update.

### DIFF
--- a/conda/moose-dev/conda_build_config.yaml
+++ b/conda/moose-dev/conda_build_config.yaml
@@ -7,7 +7,7 @@ moose_libmesh:
   - moose-libmesh 2025.10.15 openmpi_1
 
 moose_wasp:
-  - moose-wasp 2025.09.18
+  - moose-wasp 2025.09.19_02960f1
 
 moose_tools:
   - moose-tools 2025.10.21
--- a/conda/moose-dev/meta.yaml
+++ b/conda/moose-dev/meta.yaml
@@ -2,7 +2,7 @@
 # REMEMBER TO UPDATE the .yaml files for the following packages:
 #   moose/conda_build_config.yaml
 # As well as any directions pertaining to modifying those files.
-{% set version = "2025.10.27" %}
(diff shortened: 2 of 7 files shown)


conda/moose-dev/conda_build_config.yaml
```diff
@@ -7,7 +7,7 @@ moose_libmesh:
   - moose-libmesh 2025.10.15 openmpi_1
 
 moose_wasp:
-  - moose-wasp 2025.09.18
+  - moose-wasp 2025.09.19_02960f1
```
cticenhour, line 10: Why the extra hash?

cticenhour (commented):
Looks good at first glance, one question I had noted below.

conda/moose-dev/conda_build_config.yaml
loganharbour: Talked about this in person - but I made the executive decision to add in submodule commit hashes to these versions because dates don't uniquely describe a version. We do this already with the application releases

cticenhour: Right - the uniqueness fails during the situation when multiple version bumps occur in a single day. I am OK with this change.

cticenhour (approved):

loganharbour:
Failures (apptainer pronghorn and valgrind heavy) are unrelated. Due to a current breakage on devel. Patches are in progress.

PR #31906, opened 2025-11-13
Update MFEM, PETSc and libMesh

nmnobre:
## Reason/Impact
Unblocks #31525 and #31989. Lays some of the groundwork toward #31250.
Fixes memory deallocation bug in GPU builds of SuperLU_DIST. 

## Design
Update MFEM submodule, libMesh submodule, and PETSc submodule to version 3.24.2, to include fixes contributed upstream.

PR #31989, opened 2025-11-26
MFEM mumps solver

### LINKED ISSUE #31988: MFEM MUMPS Solver
### Motivation

We've been having some issues with the SuperLU solver.

I've done this, I just need a ref for the commits.

### Design

Wrap `mfem::MUMPSSolver`.

### Impact

Allows us to work around `SuperLU` solver problems while we get the necessary fixes upstreamed.

k-collie:
Closes #31988

Requires #31906, I will rebase once that is merged.

## Reason
Adds MUMPS solver

## Impact
Only adds to API

### DIFF
--- a/framework/doc/content/source/mfem/solvers/MFEMMUMPS.md
+++ b/framework/doc/content/source/mfem/solvers/MFEMMUMPS.md
@@ -1,0 +1,18 @@
+# MFEMMUMPS
+
+!if! function=hasCapability('mfem')
+
+## Overview
+
+Defines and builds an `mfem::MUMPSSolver` to use as a direct solver to solve the MFEM equation system.
+
+!syntax parameters /Solver/MFEMMUMPS
+
+!syntax inputs /Solver/MFEMMUMPS
+
+!syntax children /Solver/MFEMMUMPS
+
+!if-end!
+
+!else
+!include mfem/mfem_warning.md
--- a/framework/include/mfem/solvers/MFEMMUMPS.h
+++ b/framework/include/mfem/solvers/MFEMMUMPS.h
@@ -1,0 +1,33 @@
+//* This file is part of the MOOSE framework
+//* https://mooseframework.inl.gov
+//*
+//* All rights reserved, see COPYRIGHT for full restrictions
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#ifdef MOOSE_MFEM_ENABLED
+
+#pragma once
+
+#include "MFEMSolverBase.h"
+
+/**
+ * Wrapper for mfem::MUMPSSolver
+ */
+class MFEMMUMPS : public MFEMSolverBase
+{
+public:
+  static InputParameters validParams();
+
+  MFEMMUMPS(const InputParameters & parameters);
+
+protected:
+  void constructSolver(const InputParameters & parameters) override;
+
+  /// Updates the solver with the bilinear form in case LOR solve is required
+  void updateSolver(mfem::ParBilinearForm & a, mfem::Array<int> & tdofs) override;
+};
+
+#endif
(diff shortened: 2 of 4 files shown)


scripts/update_and_rebuild_mfem.sh
```diff
@@ -142,10 +143,12 @@ do
       -DHDF5_DIR="$HDF5_DIR" \
       -DHYPRE_DIR="$PETSC_DIR/$PETSC_ARCH" \
       -DMETIS_DIR="$PETSC_DIR/$PETSC_ARCH" \
+      -DMUMPS_DIR="$PETSC_DIR/$PETSC_ARCH" \
       -DNETCDF_DIR="$LIBMESH_DIR" \
       -DParMETIS_DIR="$PETSC_DIR/$PETSC_ARCH" \
       -DPETSC_ARCH="$PETSC_ARCH" \
       -DPETSC_DIR="$PETSC_DIR" \
+      -DScaLAPACK_DIR="$PETSC_DIR/$PETSC_ARCH/externalpackages/git.scalapack/petsc-build/" \
```
k-collie: I'm not certain that depending on this path is a good idea but I'm happy to hear thoughts

nmnobre: This won't work if we're using an installed version of PETSc (for which the source/build directories aren't present) as is the case for some CIVET recipes, but the good news is the following will:

```suggestion
      -DScaLAPACK_DIR="$PETSC_DIR/$PETSC_ARCH/lib/cmake/scalapack-2.2.2" \
```

or

```suggestion
      -DCMAKE_PREFIX_PATH="$PETSC_DIR/$PETSC_ARCH" \
```
I'm currently tempted to go with the latter (so it doesn't break if the version number ever changes). It also feels like `-DScaLAPACK_DIR="$PETSC_DIR/$PETSC_ARCH" \` should work (reading the CMake docs), but it's not working for me...

Either way, with your permission, I'm probably going to steal this one commit and move it to #31906 so we avoid the whole process with updating packages?

k-collie: Yes please steal away and modify as you see fit, thanks for saving me the extra work.

Avoiding hard-coding the scalapack version number was the reason I didn't use the former path.

`CMAKE_PREFIX_PATH` seems good since you can supply multiple paths if necessary. I think this is the best choice.

nmnobre: > It also feels like `-DScaLAPACK_DIR="$PETSC_DIR/$PETSC_ARCH" \` should work (reading the CMake docs), but it's not working for me...

Well, I misread the docs... `<PackageName>_DIR` only works if set as an environment variable NOT as a CMake variable, `<PackageName>_ROOT` on the other hand does work when set on the cmd line - this is what I'm going to do!

framework/include/mfem/solvers/MFEMMUMPS.h
```diff
@@ -0,0 +1,36 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#ifdef MOOSE_MFEM_ENABLED
+
+#pragma once
+#include "MFEMSolverBase.h"
+#include "libmesh/ignore_warnings.h"
+#include <mfem.hpp>
+#include "libmesh/restore_warnings.h"
+#include <memory>
```
nmnobre:
```suggestion

#include "MFEMSolverBase.h"
```

nmnobre (changes requested):
Excellent. Nice and short just as how I love it. :)

We're missing some docs in the form of `MFEMMUMPS.md` and perhaps an option to set the print level for consistency. I think we can postpone any other options to if/when we are pressed by any specific needs.

nmnobre (approved):
Pending a rebase, and assuming all tests pass, this is good to go.

framework/doc/content/source/mfem/solvers/MFEMMUMPS.md
```diff
@@ -0,0 +1,22 @@
+# MFEMMUMPS
+
+!if! function=hasCapability('mfem')
+
+## Summary
+
+!syntax description /Solver/MFEMMUMPS
```
nmnobre: Sorry, earlier I forgot we're dropping the summary section from the markdown docs.

```suggestion
```

k-collie:
Sorry for the pings, reviews got requested when I converted it from draft to ready to review.

nmnobre:
I've successfully tested this on the `MFEMDarcy` example test. The print level option is working as expected too.

PR #25040, opened 2023-07-26
Fix feature volume computation in FeatureVolumeVectorPostprocessor

Thomas-Ulrich:
closes #25035

PR #31990, opened 2025-11-26
Thomas/fix feature volume vector postprocessor

### LINKED ISSUE #25035: FeatureVolumeVectorPostprocessor should not integrate the variable used to detect the features
## Bug Description

FeatureVolumeVectorPostprocessor currently integrates the variable over the cell volume, while it should only compute the volume of the feature.
(note that If compute_var_to_feature_map=True, then the behavior is the one expected.)

A direct fix is the following:
https://github.com/Thomas-Ulrich/moose/commit/c19e3c71e64fbb7830dcf89394e1fd0a6892afa6
(but the solution of the unit tests would also need to be updated).

Thomas-Ulrich:
closes #25035
(there is already this PR https://github.com/idaholab/moose/pull/25040, which has been stalled, and it seems I cannot reopen it).

## Reason

to fix the issue identified.
## Design

sum only over the quadrature points exceeding the threshold.
(TODO:   double-check that compareValueWithThreshold returns the correct value with use_less_than_threshold_compariso=True).

## Impact

no

### DIFF
--- a/modules/phase_field/include/postprocessors/FeatureFloodCount.h
+++ b/modules/phase_field/include/postprocessors/FeatureFloodCount.h
@@ -326,6 +326,17 @@ public:
 
   /// Return a constant reference to the vector of all discovered features
   const std::vector<FeatureData> & getFeatures() const { return _feature_sets; }
+  /**
+   * Return the starting comparison threshold to use when inspecting an entity during the flood
+   * stage.
+   */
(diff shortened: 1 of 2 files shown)


GiudGiud:
@Thomas-Ulrich anybody at the US labs you work with we could ping for review?

PR #31998, opened 2025-11-26
Parameter application logic

dschwen:
## Reason
Allow common parameters to be specified at a higher level in the `tests` spec hierarchy.

## Design
Set parent block parameters as defaults for local parameters, cascading down.

## Impact
More compact tests specs.

### DIFF
--- a/python/FactorySystem/Parser.py
+++ b/python/FactorySystem/Parser.py
@@ -182,6 +182,14 @@ class Parser:
                 if key in params.keys():
                     if key == 'cli_args':
                         params[key].append(value)
+                    elif params.type(key) == list:
+                        if isinstance(value, str):
+                            value = value.replace('\n', ' ')
+                            params[key] = re.split(r'\s+', value)
+                        elif isinstance(value, (list, tuple)):
+                            params[key] = list(value)
+                        else:
+                            params[key] = [str(value)]
                     else:
                         params[key] = value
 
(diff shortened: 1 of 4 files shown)


python/FactorySystem/Parser.py
```diff
@@ -56,7 +56,19 @@ def parse(self, filename, default_values = None):
...
-        self._parseNode(filename, self.root, default_values)
+
+        # Convert any provided default values (from the testroot file) into a
+        # simple dictionary so that we can merge values without relying on
+        # pyhit.Node's type conversion rules.
+        if default_values is not None:
+            if isinstance(default_values, dict):
+                defaults = default_values
+            else:
+                defaults = {k: v for k, v in default_values.params()}
```
loganharbour, line 67: when would this be a dict and when would it not be?

python/FactorySystem/Parser.py
```diff
@@ -153,12 +178,20 @@ def _parseNode(self, filename, node, default_values):
...
-                      params[key].append(value)
+                        params[key].append(value)
+                    elif params.type(key) == list:
+                        if isinstance(value, str):
+                            value = value.replace('\n', ' ')
+                            params[key] = re.split(r'\s+', value)
+                        elif isinstance(value, (list, tuple)):
+                            params[key] = list(value)
+                        else:
+                            params[key] = [str(value)]
```
loganharbour, line 192: why is this now needed?

PR #32010, opened 2025-12-01
Update MooseServer formatting to not remove override syntax

brandonlangley:
refs #22766


brandonlangley:
@dschwen @loganharbour -

I saw a MOOSE Language Support discussion [report](https://github.com/idaholab/moose-language-support/discussions/32) that language server formatting was removing the override syntax.

So something like this:

```
param_01 := 4
param_02 :override= 5
```

gets formatted as this:

```
param_01 = 4
param_02 = 5
```

The document formatter was written before the override was added, and the formatter did not get updated to handle it.

This is just a two-line update to fix the issue along with a test, could one of you review this and merge it if you approve?

loganharbour (approved):

PR #32014, opened 2025-12-01
Add more thread safety to output in PerfGraphLive

### LINKED ISSUE #32012: Add more thread safety to `PerfGraphLive`
### Motivation

There is a race condition within `PerfGraphLive` for printing output as flushing the moose console isn't guarded.

### Design

Guard the moose console for use with `PerfGraphLive`

### Impact

Better live printing.

loganharbour:
Closes #32012

### DIFF
--- a/framework/src/utils/PerfGraphLivePrint.C
+++ b/framework/src/utils/PerfGraphLivePrint.C
@@ -22,9 +22,9 @@ PerfGraphLivePrint::PerfGraphLivePrint(PerfGraph & perf_graph, MooseApp & app)
     _mem_limit(perf_graph._live_print_mem_limit),
     _stack_level(0),
     _current_execution_list_end(0),
-    _current_execution_list_last(0),
     _last_execution_list_end(0),
-    _last_printed_increment(NULL),
+    _last_printed_increment(nullptr),
+    _last_printed_increment_finished(false),
     _last_num_printed(0),
     _console_num_printed(0),
     _stack_top_print_dots(true),
@@ -44,6 +44,7 @@ PerfGraphLivePrint::printLiveMessage(PerfGraph::SectionIncrement & section_incre
   {
     section_increment._state = PerfGraph::IncrementState::PRINTED;
     _last_printed_increment = &section_increment;
+    _last_printed_increment_finished = true;
     return;
   }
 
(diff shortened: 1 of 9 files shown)


loganharbour:
@lindsayad when you have a moment, please

framework/src/utils/PerfGraphLivePrint.C
```diff
@@ -354,20 +359,17 @@ PerfGraphLivePrint::start()
...
     // progress and exit
     if (_last_execution_list_end == _current_execution_list_end)
-      inSamePlace();
+      console_lock = inSamePlace(std::move(console_lock));
 
     // This means that new stuff has been added to the execution list.  We need to iterate through
     // it, modifying the stack and printing anything that needs printing
-    iterateThroughExecutionList();
-
-    _last_num_printed = _console.numPrinted();
```
loganharbour, line 370: Reminder for self; might need to add this back

lindsayad:
what exactly is the race condition? `libMesh::out` and `libMesh::err` should be thread safe these days after https://github.com/libMesh/libmesh/pull/4237

loganharbour:
> what exactly is the race condition? `libMesh::out` and `libMesh::err` should be thread safe these days after [libMesh/libmesh#4237](https://github.com/libMesh/libmesh/pull/4237)

When it is printing out what to be done, it relies on the state of the buffer (whether or not something else printed). If something else prints to the buffer while it is running, it'll end up clobbering the state of where the status printout is

lindsayad:
I don't understand yet. After the libMesh PR, each thread has its own buffer. Things like `std::flush` force sync'd output of the thread local buffers to the global owning buffer

lindsayad:
> When it is printing out what to be done, it relies on the state of the buffer (whether or not something else printed)

If you could point to this line(s), that would probably help me understand

loganharbour:
> > When it is printing out what to be done, it relies on the state of the buffer (whether or not something else printed)
> 
> If you could point to this line(s), that would probably help me understand

Unfortunately it's not that simple.

As the perf graph live thread wakes up - it records the state of the console (who printed to it last). How it prints things (like if it is printing a dot because the last thing was a section starting) depends on whether or not it was the last thing to print to the console or not. If something else is printed to console while the print thread wakes up and before it does its print, you get a garbled mess and the print thread could print in the middle of someone elses output.

lindsayad:
These changes just feel heavy. It feels disappointing to be introducing a `moose_console_mutex` in `Moose.h` after the work we've done at the libMesh level.

> If something else is printed to console while the print thread wakes up and before it does its print, you get a garbled mess and the print thread could print in the middle of someone elses output.

Garbled as in dots intermingling with other thread's output (which I agree is a problem)? The output should still be atomic, I hope, such that you would not get a race condition of the variety that TSAN reports.

Thinking about this makes me wonder why we have a separate thread printing dots anyway since we logically want the output to be sequential

lindsayad:
Unless people are intentionally printing console output from code within `Threads::parallel_reduce` but we would expect that to be garbled on its own. I just wonder if in serial code regions (outside of `Threads::paralle_reduce`) there should just be a single thread writing to the `_console`. This single thread would be notified when a task has finished so it could stop printing dots

loganharbour:
> Thinking about this makes me wonder why we have a separate thread printing dots anyway since we logically want the output to be sequential

I'm not aware of any way to do this without a thread.

lindsayad:
Do you have an example that (usually) demonstrates the issue?

PR #32017, opened 2025-12-02
MFEM complex variables transfer between MFEM applications

Giannis95:
closes #32008


nmnobre:
Tidied up to:
- unify real and complex code paths
- avoid nested `if` and `continue` statements
- verify, even in optimised builds, that the no. of dofs on the gridfunctions match (previously an MFEM assertion)
- remove unused `dim` input param from test files

I'm happy with this otherwise, and will merge tomorrow if people agree.

nmnobre (approved):

PR #31613, opened 2025-09-26
Add two MFEM auxkernels for L2-averaged inner and cross product of two source (par)gridfucntions projected onto an MFEM AuxVariable

Giannis95:
closes #31612

PR #32021, opened 2025-12-02
Allow L2 FE spaces w/ map type integral

nmnobre:
## Reason
Addresses concerns in #31969 which removed the ability to construct integral-mapped (as opposed to value-mapped) L2 FE spaces originally introduced in #31613.

## Design
Rather than specifying the map type, the user can just specify a different finite element collection type, eliminating the need for more complicated checks between the chosen space and Piola mapping.

## Impact
Allows L2 basis functions to be scaled by the volume integral of the spanned element, which can be useful when projecting between some spaces.


alexanderianblair (approved):
LGTM, pending remaining CI tests pass!

PR #32037, opened 2025-12-04
ExplicitMassDamping added.  Refs #32030

WilkAndy:
ExplicitMassDamping added.  Refs #32030


WilkAndy:
Hi @dschwen - the "restep" is failing.  After "restep", the MOOSE calculation at time step = 1 is identical to the correct result at time step = 2, and the MOOSE calculation at time step = 2 is identical to the correct result at time step = 3.  That is, the restep process doesn't get rid of the (disp, vel, accel) that were calculated prior to the restep, but instead retains the, and just keeps += onto them.   I'm pretty sure this is nothing to do with my PR: is it a "feature" of Explicit?

WilkAndy:
Hi @dschwen again.  When you review thhis, could you please inspect the input file paying particular attention to the `implicit = false` stuff.   Eg, i would put that in *all* Kernels (apart from MassMatrix), NodalKernels, BCs and Materials, but that's not the case in other solid_mechanics input files.  It's probably worth explaining to me the rationale behind including it, so i can document it.

WilkAndy:
Closing, as this addition will hopefully be combined with #30853

PR #32044, opened 2025-12-04
Track and limit CPU usage in TestHarness

### LINKED ISSUE #32043: Restrict CPU usage in test harness
### Motivation

We should track average cpu % for each job and then use it as a criteria to see if a job used too many resources.

### Design

Track CPU percentage in SubprocessRunner and fail the job if it uses too much CPU.

### Impact

New feature. May fail tests that are dirty.

loganharbour:
Closes #32043

### DIFF
--- a/test/tests/samplers/distribute/tests
+++ b/test/tests/samplers/distribute/tests
@@ -16,6 +16,8 @@
       installation_type = in_tree # see #26480
       hpc = false # see #26480
       capture_perf_graph = false
+      min_parallel = 4
+      max_parallel = 4
     []
     [plot]
       type = CheckFiles
--- a/python/TestHarness/runners/SubprocessRunner.py
+++ b/python/TestHarness/runners/SubprocessRunner.py
@@ -7,8 +7,10 @@
 #* Licensed under LGPL 2.1, please see LICENSE for details
 #* https://www.gnu.org/licenses/lgpl-2.1.html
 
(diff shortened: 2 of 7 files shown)


loganharbour:
I still need to work through the errors this causes. The limits might be a little too fine, I'm not sure. The meshgenerator failures are valid because netgen is dirty with threads.

test/tests/samplers/distribute/tests
```diff
@@ -16,6 +16,8 @@
       installation_type = in_tree # see #26480
       hpc = false # see #26480
       capture_perf_graph = false
+      min_parallel = 4
+      max_parallel = 4
```
lindsayad, line 20: why this?

loganharbour: This is a python script that manually runs moose with mpiexec -n 1 and mpiexec -n 4: https://github.com/idaholab/moose/blob/fb28c060e20cad83d698401c3d0f7b575de35980/test/tests/samplers/distribute/execute.py#L44

Without this, when the test harness is ran with `-p 1` through `-p 3`, slots are under allocated. And when ran with `-p 5` and greater, this will use more slots even though it isn't using that many ranks.

zachmprince (approved):
It's beautiful

loganharbour:
Superseded by https://github.com/idaholab/moose/pull/32424

PR #32060, opened 2025-12-08
Add aux-kernel update right before timestep_end output

GiudGiud:
closes #32059
early PR to see what fails


GiudGiud:
Tried with a couple test inputs and can't reproduce the error. They are checked in this branch. Closing for now.

PR #32061, opened 2025-12-09
Fix new MetaPhysicL 2 failures

lindsayad:
Refs #31906


nmnobre (approved):
Failures are due to missing container since next is stuck.
Merging would solve this chicken and egg problem.

loganharbour:
Merging despite failures; failures due to package update sitting on next.

loganharbour (approved):

PR #32067, opened 2025-12-09
Update MooseServer autocompletion lookups to use warehouses

brandonlangley:
refs #22766

### DIFF
--- a/framework/src/parser/Builder.C
+++ b/framework/src/parser/Builder.C
@@ -366,6 +366,7 @@ Builder::errorCheck(const Parallel::Communicator & comm, bool warn_unused, bool 
 
   std::vector<hit::ErrorMessage> messages;
 
+  if (_parser.queryCommandLineRoot())
   {
     UnusedWalker uw(_extracted_vars, *this);
     _parser.getCommandLineRoot().walk(&uw);
--- a/framework/include/base/MooseServer.h
+++ b/framework/include/base/MooseServer.h
@@ -39,6 +39,12 @@ public:
    */
   std::shared_ptr<wasp::lsp::Connection> getConnection() { return _connection; }
 
+  /**
+   * Public interface for writable check app reference with error checks.
+   * @return - writable reference to check app for current input document
+   */
+  MooseApp & getCheckApp();
+
 private:
   /**
    * SortedLocationNodes - type alias for set of nodes sorted by location
@@ -197,6 +203,14 @@ private:
   template <typename MooseEnumType>
   void getEnumsAndDocs(MooseEnumType & moose_enum_param,
                        std::map<std::string, std::string> & options_and_descs);
+
+  /**
+   * Supplement completion list with objects in warehouses if applicable.
+   * @param param_type - parameter type string to pick suitable warehouse
+   * @param options_and_descs - map to fill with options and descriptions
+   */
+  void addObjectsFromWarehouses(const std::string & param_type,
+                                std::map<std::string, std::string> & options_and_descs);
 
   /**
    * Gather definition locations - specific to this server implemention.
@@ -421,10 +435,6 @@ private:
    */
   const hit::Node * queryRoot() const;
 
-  /**
-   * @return The check app for the current document path, with error checking on if it exists
-   */
-  MooseApp & getCheckApp();
   /**
    * @return The root node from the check parser for the current document path, with error checking
    * on if it exists
--- a/framework/include/problems/FEProblemBase.h
+++ b/framework/include/problems/FEProblemBase.h
@@ -2118,6 +2118,11 @@ public:
   }
 
   /*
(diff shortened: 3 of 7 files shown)


framework/src/parser/Builder.C
```diff
@@ -366,6 +366,7 @@ Builder::errorCheck(const Parallel::Communicator & comm, bool warn_unused, bool
 
   std::vector<hit::ErrorMessage> messages;
 
+  if (_parser.queryCommandLineRoot())
```
loganharbour, line 369: Did you hit a case where this is needed?

brandonlangley:
@loganharbour @dschwen -

This updates the MooseServer autocompletion to use warehouses in support of action generated objects including:

- `MaterialName` types using the problem `MaterialWarehouse` objects
- `MaterialPropertyName` types using the `MaterialPropertyRegistry` names
- `FunctionName` types using the problem `FunctionWarehouse` objects
- `OutputName` types using the `OutputWarehouse` names and reserved words
- `UserObjectName` types using `TheWarehouse` parameterized attribute queries
- `NonlinearVariableName` types using the `NonlinearSystem` solver variables
- `AuxVariableName` types using the `AuxiliarySystem` equation variables
- `VariableName` types using `NonlinearSystem` and `AuxiliarySystem` names

The autocompletion is now using the `problem` from the `actionWarehouse` in order to access these various objects.

So I also updated the server logic so it continues to build the app if the parsing fails and run app if the building fails.

This lets the `problem` still be available for warehouse based autocompletion and plotting even if there was an error.

It previously did not build the app if parsing failed, and it deleted the app if building or running failed for any reason.

Without this change, warehouses could not be used to complete valueless parameters that cause a parse error like:

```
material =
```

Warehouses were also unavailable when something like duplicate parameters in an unrelated block caused an error.

I also added unit tests for all the above warehouse cases. Could one of you review this and merge it if you approve?

framework/src/parser/Builder.C
brandonlangley: @loganharbour -

Yes I did. I just [commented](https://github.com/idaholab/moose/pull/32067#issuecomment-3635465802) explaining that I updated the server to continue on to run the app if parsing fails.

So this means the `Builder::errorCheck()` method gets called now even if the `Parser::parse()` step fails.

This is fine except `Parser::parse()` throws certain errors, e.g., `DupParamWalker`, before `_cli_root` is set.

In these cases such as duplicate parameters where the parser did not set `_cli_root`, I would get this error:

```
Parser::getCommandLineRoot(): command line root is not set
```

This check just prevents that error as it does not need to check to the command line root if there is not one.

This is the same reason I removed `parser->queryCommandLineRoot()` from those two `mooseAssert` checks.

loganharbour:
> @loganharbour @dschwen -
> 
> This updates the MooseServer autocompletion to use warehouses in support of action generated objects including:
> 
>     * `MaterialName` types using the problem `MaterialWarehouse` objects
> 
>     * `MaterialPropertyName` types using the `MaterialPropertyRegistry` names
> 
>     * `FunctionName` types using the problem `FunctionWarehouse` objects
> 
>     * `OutputName` types using the `OutputWarehouse` names and reserved words
> 
>     * `UserObjectName` types using `TheWarehouse` parameterized attribute queries
> 
>     * `NonlinearVariableName` types using the `NonlinearSystem` solver variables
> 
>     * `AuxVariableName` types using the `AuxiliarySystem` equation variables
> 
>     * `VariableName` types using `NonlinearSystem` and `AuxiliarySystem` names
> 
> 
> The autocompletion is now using the `problem` from the `actionWarehouse` in order to access these various objects.
> 
> So I also updated the server logic so it continues to build the app if the parsing fails and run app if the building fails.
> 
> This lets the `problem` still be available for warehouse based autocompletion and plotting even if there was an error.
> 
> It previously did not build the app if parsing failed, and it deleted the app if building or running failed for any reason.
> 
> Without this change, warehouses could not be used to complete valueless parameters that cause a parse error like:
> 
> ```
> material =
> ```
> 
> Warehouses were also unavailable when something like duplicate parameters in an unrelated block caused an error.
> 
> I also added unit tests for all the above warehouse cases. Could one of you review this and merge it if you approve?

I already looked at it. There's a comment by me above your comment.

brandonlangley:
@loganharbour -

> I already looked at it. There's a comment by me above your comment.

I replied to that comment last night about why the `if (_parser.queryCommandLineRoot())` is needed. Do you see it?

framework/src/parser/Builder.C
brandonlangley: @loganharbour -

Circling back to make sure, did you see this reply to your comment last week? Does my response make sense?

loganharbour (approved):

PR #32068, opened 2025-12-09
Additional support for mixes of element dimensions

lindsayad:
- Remove flawed `isLowerD/hasLowerD` logic. The logic errors started when associating an `is_lower_d` attribute to subdomains. We (libmesh) do not require the same element type in a given subdomain, and we also do not require the same element dimension in a given subdomain. So all logic stemming from that assumed one-to-one mapping of subdomain to element dimension was also flawed.
- Add precise error checking in `ProjectionAux` where mixing dimensions is in fact logically flawed when using volume based weighting for element to node projections
- Add `GhostAllPointNeighbors` relationship manager which leverages the node-to-element maps we cache in `MooseMesh`
- Add identity weighting option when performing element to node projections

### DIFF
--- a/framework/include/auxkernels/AuxKernelBase.h
+++ b/framework/include/auxkernels/AuxKernelBase.h
@@ -99,10 +99,6 @@ protected:
    */
   const bool _check_boundary_restricted;
 
-  /// Whether we are computing for a lower dimensional variable using boundary restriction, e.g. a
-  /// variable whose block restriction is coincident with a higher-dimensional boundary face
-  std::optional<bool> _coincident_lower_d_calc;
-
   /// Subproblem this kernel is part of
   SubProblem & _subproblem;
   /// System this kernel is part of
--- a/framework/include/auxkernels/AuxKernelBase.h
+++ b/framework/include/auxkernels/AuxKernelBase.h
@@ -80,6 +80,8 @@ public:
 
   virtual const std::set<std::string> & getSuppliedItems() override;
 
+  virtual void initialSetup() override;
+
 protected:
   /// Base MooseVariable
   MooseVariableFieldBase & _var;
--- a/framework/include/auxkernels/AuxKernelBase.h
+++ b/framework/include/auxkernels/AuxKernelBase.h
@@ -99,7 +99,7 @@ protected:
 
   /// Whether we are computing for a lower dimensional variable using boundary restriction, e.g. a
   /// variable whose block restriction is coincident with a higher-dimensional boundary face
-  const bool _coincident_lower_d_calc;
+  std::optional<bool> _coincident_lower_d_calc;
 
   /// Subproblem this kernel is part of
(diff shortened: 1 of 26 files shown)


framework/include/auxkernels/AuxKernelBase.h
```diff
@@ -99,7 +99,7 @@ class AuxKernelBase : public MooseObject,
 
   /// Whether we are computing for a lower dimensional variable using boundary restriction, e.g. a
   /// variable whose block restriction is coincident with a higher-dimensional boundary face
-  const bool _coincident_lower_d_calc;
+  std::optional<bool> _coincident_lower_d_calc;
```
loganharbour: > `std::optional`

❤️ ❤️ ❤️

framework/src/auxkernels/AuxKernel.C
```diff
@@ -124,7 +124,7 @@ template <typename ComputeValueType>
 void
 AuxKernelTempl<ComputeValueType>::insert()
 {
-  if (_coincident_lower_d_calc)
+  if (_coincident_lower_d_calc.value())
```
loganharbour, line 133: I think you probably want to assert that it has a value or use `value_or()`?

lindsayad: It throws if it doesn't have a value. I guess an assertion could be used to add a more friendly error message in devel/dbg

lindsayad:
@roystgnr would appreciate your review on this one

GiudGiud (approved):

framework/include/relationshipmanagers/GhostAllPointNeighbors.h
```diff
@@ -0,0 +1,37 @@
...
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "RelationshipManager.h"
+
+/**
+ * Ghosts all point neighbors (not including peridioc neighbors) regardless of the dimensionality of
```
roystgnr: periodic

framework/src/auxkernels/AuxKernel.C
```diff
@@ -328,6 +321,27 @@ AuxKernelTempl<ComputeValueType>::isMortar()
...
 
+template <typename ComputeValueType>
+void
+AuxKernelTempl<ComputeValueType>::determineWhetherCoincidentLowerDCalc()
+{
+  mooseAssert(_bnd && !isNodal(),
+              "We can never be a lower-dimensional calculation if we are not boundary restricted "
+              "or if we are a nodal auxiliary kernel");
+
+  _coincident_lower_d_calc = !_var.numberOfDofs();
```
roystgnr, line 340: A comment here would be good; I'm not following the reasoning.

roystgnr (approved):

PR #32073, opened 2025-12-10
Xfem meshcut 3d

lynnmunday:
closes #32072


lynnmunday:
With this change, it matches the 2D case and the Belytschko results pretty well for the double cantilever beam.  It does crash after 15 steps due to a different issue.  Compare the black dots from the Belytschko with the blue dots which would be from the center of the beam where the crack would be under plane strain conditions.  The red dots are for the crack path near edges where the crack has a free surface.

crack_close

modules/xfem/src/userobjects/CrackMeshCut3DUserObject.C
```diff
@@ -765,10 +765,39 @@ CrackMeshCut3DUserObject::findActiveBoundaryDirection()
...
+        // check that kii is not zero
+        Real theta_m;
+        Real theta_p;
+        if (std::abs(kii) < libMesh::TOLERANCE)
+          theta_m = theta_p = 0;
+        else
+        {
+          theta_m = 2 * std::atan((ki - sqrt_k) / (4 * kii));
+          theta_p = 2 * std::atan((ki + sqrt_k) / (4 * kii));
+        }
```
bwspenc:
```suggestion
        Real theta_m = 0;
        Real theta_p = 0;
        if (std::abs(kii) > libMesh::TOLERANCE)
        {
          theta_m = 2 * std::atan((ki - sqrt_k) / (4 * kii));
          theta_p = 2 * std::atan((ki + sqrt_k) / (4 * kii));
        }
```

modules/xfem/test/tests/solid_mechanics_basic/edge_crack_3d_mhs.i
```diff
@@ -40,7 +40,7 @@
   []
   [top_trac_y]
     type = ConstantFunction
-    value = 0
+    value = 100
```
bwspenc, line 43: Just curious what this was for. Was it to induce mixed-mode behavior?

bwspenc (changes requested):

modules/xfem/test/tests/solid_mechanics_basic/edge_crack_3d_mhs.i
lynnmunday: the y-axis is normal to the crack face so before this the cracks were mostly loaded in only mode II.  With this change to computing the crack direction it really curved down and the solve failed.  So I added more mode I loading to straighten out the crack.

bwspenc (approved):

PR #32076, opened 2025-12-11
Periodic mfem mesh

### LINKED ISSUE #32075: Add support for periodic MFEM meshes
### Motivation

Libmesh meshes can already support enforcing boundary conditions via vector. The goal here is to do the same with MFEM meshes. Periodic gmesh meshes are already natively supported within MFEM, and here our aim is to leverage mfem's ability to enforce [periodic boundary conditions supplied by a vector.](https://mfem.org/howto/periodic-boundaries/)

### Design

We add a new action, `AddMFEMPeriodicBCs`, which picks up n different n-vectors (n being the space dimension of the mesh) which specifies translation vectors for vertices on the boundary of the mesh. Any vertices which are determined to be coincident under this translation will be considered topologically identical.

Since MFEM enforces periodic boundary conditions at the mesh level, we cannot re-use the existing periodic BC system.

### Impact

This will allow us to enforce periodic boundary conditions on a wider array of problems.

sean-baccas:
## Reason

Libmesh meshes can already support enforcing boundary conditions via vector. The goal here is to do the same with MFEM meshes. Periodic gmesh meshes are already natively supported within MFEM, and here our aim is to leverage mfem's ability to enforce [periodic boundary conditions supplied by a vector.](https://mfem.org/howto/periodic-boundaries/)

## Design

We add a new action, `AddMFEMPeriodicBCs`, which picks up n different n-vectors (n being the space dimension of the mesh) which specifies translation vectors for vertices on the boundary of the mesh. Any vertices which are determined to be coincident under this translation will be considered topologically identical.

## Impact
This will allow us to enforce periodic boundary conditions on a wider array of problems.


framework/doc/content/syntax/MFEMPeriodic/index.md
```diff
@@ -0,0 +1,9 @@
+# MFEMPeriodic
+
+The `MFEMPeriodic` syntax is used to add periodic boundary conditions to an MFEM mesh.
+The user must specify some translation vectors: either 2 vectors of dimension 2 for a 2D
+problem, or 3 vectors of dimension 3 for a 3D problem. These are passed to the MFEM mesh,
```
nmnobre:
```suggestion
The user can specify up to n translation vectors of dimension n for an n-dimensional problem. These are passed to the MFEM mesh,
```

framework/doc/content/source/mfem/actions/AddMFEMPeriodicBCs.md
```diff
@@ -0,0 +1,19 @@
...
+!if! function=hasCapability('mfem')
+
+## Overview
+
+Action called to specify a periodic boundary condition on an MFEM mesh, where one isn't there already.
+It parses the `MFEMPeriodic` block in the user input.
+
+## Example Input File Syntax
+
+!listing test/tests/mfem/periodic/periodic_cube_vector.i block=MFEMPeriodic
```
nmnobre, line 12: How difficult would it be to move this under the usual `[BCs]` block?

framework/include/mfem/mesh/MFEMPeriodicBCs.h
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include "MooseObject.h"
+
+class MFEMPeriodicByVector : public MooseObject
+{
+public:
+  static InputParameters validParams();
+  MFEMPeriodicByVector(const InputParameters & parameters);
+  mfem::Vector GetPeriodicBc(int dim);
```
nmnobre:
```suggestion
  mfem::Vector GetPeriodicBC(int dim);
```

test/tests/mfem/kernels/diffusion_cube.i
```diff
@@ -0,0 +1,105 @@
+[Mesh]
```
nmnobre, line 1: Looks like a forgotten leftover file?

test/tests/mfem/periodic/tests
```diff
@@ -0,0 +1,36 @@
+[Tests]
```
nmnobre, line 1: Let's switch all three tests to CSVDiff.
Also, we're missing `design = 'some-doc.md another-doc.md'`, that's why the tests are failing.

framework/src/mfem/mesh/MFEMPeriodicBCs.C
```diff
@@ -0,0 +1,70 @@
...
+                                             "Vector specifying translation in y direction.");
+  params.addParam<std::vector<Real>>("translation_z",
+                                     "Vector specifying translation in z direction.");
+  return params;
+}
+
+MFEMPeriodicByVector::MFEMPeriodicByVector(const InputParameters & parameters)
+  : MooseObject(parameters)
+{
+  _translation_x = getParam<std::vector<Real>>("translation_x");
```
nmnobre, line 33: I think all three parameters should be optional and there should be no special treatment for the 3d case. Ultimately, we need to build an `std::vector<mfem::Vector>` (`GetPeriodicBC` could probably return that directly?), and given the mesh spatial dimension (we need to know the variable we're applying this BC to, so we can get the appropriate (sub)mesh - which would be very natural if this is moved to the `[BCs]` input syntax block), we can provide suitable defaults (zero-vectors) for the vectors the user doesn't provide.

nmnobre (changes requested):

framework/doc/content/source/mfem/actions/AddMFEMPeriodicBCs.md
sean-baccas: I definitely tried this, but it was quite a while ago now. I remember finding it tricky and giving up

test/tests/mfem/mesh/cube_hex10.e
nmnobre, line 1: Rename to `cube_hex8.e`.

framework/include/mfem/mesh/MFEMPeriodicBCs.h
```diff
@@ -0,0 +1,29 @@
...
+
+class MFEMPeriodicByVector : public MooseObject
+{
+public:
+  static InputParameters validParams();
+  MFEMPeriodicByVector(const InputParameters & parameters);
+  std::vector<mfem::Vector> GetPeriodicBCs();
+
+private:
+  std::vector<Real> _translation_x;
```
nmnobre, line 24:
```suggestion
  std::vector<mfem::real_t> _translation_x;
```

Likewise for all other occurrences of `Real` in this PR.

nmnobre (changes requested):
What if `MFEMPeriodicBCs` becomes an interface like, e.g., `MFEMBlockRestrictable`, that both `MFEMMesh` and `MFEMSubMesh` inherit from? Maybe we can call it `MFEMPeriodical`? That interface defines the input parameters, i.e. the three (optional) translation vectors, perhaps named `periodic_translation_{x,y,z}`, defaulting to '0. 0. 0.' (because we know the mesh dim, we can trim this default if needed a-posteriori*). The condition to determine periodicity is then iff any of the three vectors verifies `isParamSetByUser()` (no need for an extra input parameter). The new action and syntax block are gone. I think you might've already started moving in this direction judging by the input parameters in `MFEMMesh`?

\* To elaborate on that, we could move `applyPeriodicBoundaryByTranslation(mfem::Mesh & input)` to `MFEMPeriodical` and if `isParamSetByUser()`, we complain if the dimensions don't match, but if it's not, we just trim the default to fit the mesh dim.

framework/doc/content/source/mfem/actions/AddMFEMPeriodicBCs.md
nmnobre: I'll mark this resolved as it's redundant if we move the input parameters to the Mesh/SubMesh block anyway.

framework/src/mfem/mesh/MFEMPeriodicBCs.C
```diff
@@ -0,0 +1,75 @@
...
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#ifdef MOOSE_MFEM_ENABLED
+
+#include "MFEMPeriodicBCs.h"
+
+registerMooseObject("MooseApp", MFEMPeriodicByVector);
+
+InputParameters
+MFEMPeriodicByVector::validParams()
```
alexanderianblair, line 17: Can we rename this? `MFEMPeriodicByVector` doesn't seem to clearly capture that it's mapping periodic copies by translation vector IMO

framework/src/mfem/mesh/MFEMPeriodicBCs.C
```diff
@@ -0,0 +1,75 @@
...
+{
+  _translation_x = getParam<std::vector<Real>>("translation_x");
+  _translation_y = getParam<std::vector<Real>>("translation_y");
+  if (isParamSetByUser("translation_z"))
+  {
+    _translation_z = getParam<std::vector<Real>>("translation_z");
+  }
+}
+
+std::vector<mfem::Vector>
```
alexanderianblair, line 41: This could be made more general if the `MFEMPeriodicBC` derived class returned the v2v map between vertices on the primary and replica boundaries rather than translation vectors, since this is what is needed for more generalised transforms

Proof-of-principle for rotational transforms over at https://github.com/alexanderianblair/platypus-moose/tree/alexanderianblair/mfem-periodic-wedge-sectors effectively just creates a custom `CreatePeriodicVertexMapping` for a rotational transform instead; there's no reason we could do this consistently across all mapping types, since they're passed to `mfem::Mesh::MakePeriodic` either way

alexanderianblair:
> What if `MFEMPeriodicBCs` becomes an interface like, e.g., `MFEMBlockRestrictable`, that both `MFEMMesh` and `MFEMSubMesh` inherit from? Maybe we can call it `MFEMPeriodical`? That interface defines the input parameters, i.e. the three (optional) translation vectors, perhaps named `periodic_translation_{x,y,z}`, defaulting to '0. 0. 0.' (because we know the mesh dim, we can trim this default if needed a-posteriori*). The condition to determine periodicity is then iff any of the three vectors verifies `isParamSetByUser()` (no need for an extra input parameter). The new action and syntax block are gone. I think you might've already started moving in this direction judging by the input parameters in `MFEMMesh`?
> 
> * To elaborate on that, we could move `applyPeriodicBoundaryByTranslation(mfem::Mesh & input)` to `MFEMPeriodical` and if `isParamSetByUser()`, we complain if the dimensions don't match, but if it's not, we just trim the default to fit the mesh dim.

I worry that there's a loss of generality here by restricting the options for forming the vertex-to-periodic-vertex map to just translations. That said, translations + rotations give pretty broad coverage for common use-cases, and certainly would catch the ones on our immediate roadmap

alexanderianblair:
Just to note here, following offline conversations with @sean-baccas - I'll be picking up this PR to free up some of his bandwidth, as the periodic wedge BCs have become a particular priority on our side - I will add an associated test here, and address the remaining requested changes

alexanderianblair:
Moving development over to #33337 given a) the change in ownership of the branch and b) the redesign to use an interface class to allow periodic transformations to be specified in the [Mesh] block rather than in a new action.

Currently requested changes will be addressed their prior to fresh review

PR #32080, opened 2025-12-11
Limit threads in external TestHarness testers and external plotting scripts

### LINKED ISSUE #32079: Better restrict threads and slots in command tests
### Motivation

TestHarness tests that run external tests (AnalyzeJacobian, PythonUnitTest, tests with `command` or `command_proxy`) should have `OMP_NUM_THREADS` set in order to restrict the number of threads within the tests they run. Packages like pandas, for example, try to run with lots of threads. This is a waste and is also an abuse of resources.

### Design

Restrict `OMP_NUM_THREADS` in external tests.

### Impact

Improves resource contention. Could possibly make some tests a little slower, but it should be negligible.

loganharbour:
Closes https://github.com/idaholab/moose/issues/32079

### DIFF
--- a/python/MooseDocs/extensions/media.py
+++ b/python/MooseDocs/extensions/media.py
@@ -146,6 +146,8 @@ class ScriptCommand(ImageCommand):
         python_dir = os.path.abspath(os.path.join(this_dir, '..', '..'))
         run_env = os.environ.copy()
         run_env['PYTHONPATH'] = f'{python_dir}:' + os.environ.get('PYTHONPATH', '')
+        # Don't allow threads in python scripts
+        run_env['OMP_NUM_THREADS'] = '1'
 
         # Generate the plot
         LOG.info("Executing plot script %s", script_path)
--- a/python/TestHarness/runners/SubprocessRunner.py
+++ b/python/TestHarness/runners/SubprocessRunner.py
@@ -69,16 +69,21 @@ class SubprocessRunner(Runner):
         else:
             process_kwargs['preexec_fn'] = os.setsid
 
+        # Augment the environment if needed
+        process_env = tester.augmentEnvironment(self.options)
+
         # Special logic for openmpi runs
         if tester.hasOpenMPI():
-            process_env = os.environ.copy()
-
             # Don't clobber state
             process_env['OMPI_MCA_orte_tmpdir_base'] = self.job.getTempDirectory().name
             # Allow oversubscription for hosts that don't have a hostfile
             process_env['PRTE_MCA_rmaps_default_mapping_policy'] = ':oversubscribe'
 
-            process_kwargs['env'] = process_env
+        # Add to environment if requested
+        if process_env:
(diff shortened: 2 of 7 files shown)


loganharbour:
@roystgnr @lindsayad @hugary1995 I need some insight on this.

In https://github.com/idaholab/moose/pull/32044, I'm working on adding in CPU timing polling in an effort to find tests that are oversubscribing resources. This led me to working on some simple restrictions that take care of a lot of the tests that are oversubscribing.

The piece that I need insight from all of this - setting `OMP_NUM_THREADS` for all tests. As it stands right now, I set `OMP_NUM_THREADS` equal to whatever we set `--n-threads=` to in the command line arguments. This does a few things:

- For python scripts, some libs (numpy, pandas, etc) default the number of the threads to the number of CPUs. This limits those. In a lot of cases I realized we're getting a CPU spike as a python script stupidly instantiates 64 or 128 threads and then does little work. 
- For MOOSE? I'm not sure. Do we have any libraries for which we're _not_ properly setting the number of threads (or they stupidly default to dome value)? Probably. 

The other option is to completely unset `OMP_NUM_THREADS` for moose runs, but I'm not sure I like that. We have a lot of cases where we have a mixed run of a python script and MOOSE.

roystgnr:
> Do we have any libraries for which we're not properly setting the number of threads (or they stupidly default to dome value)? Probably.

We have at least one.  The tetrahedralization code in Netgen spawns 4 threads.  If you set OMP_NUM_THREADS to something different, it spawns 4 threads, and if you set their NGS_NUM_THREADS to something different, it spawns 4 threads.  ;-)

I wouldn't be too surprised if we had others, maybe some obscure PETSc preconditioner options.  We call `omp_set_num_threads(_n_threads)` with whatever we got from `--n_threads`, but if a library doesn't use OpenMP or if it overrides the OpenMP setting or even if it just queries the environment variable directly rather than going through `omp_get_num_threads()` then it won't be controlled by our call.

loganharbour:
>  but if a library doesn't use OpenMP or if it overrides the OpenMP setting or even if it just queries the environment variable directly rather than going through `omp_get_num_threads()` then it won't be controlled by our call.

This is my fear. We have a balance of two things:

- Don't oversubscribe build machines (which they definitely are at the moment and it's very hard to track
- Let things behave as they would with default options

However, if we can't easily tell if something is using extra threads anyway I see less harm in just setting `OMP_NUM_THREADS`

lindsayad:
I think we should definitely set `OMP_NUM_THREADS` and it makes the most sense to set it to `--n-threads` IMO. We do this in PETSc CI and `make check` otherwise 3rd party libs downloaded by PETSc also will toss out num-cpu threads

zachmprince (approved):
Beautiful

loganharbour:
With a general consensus from Roy and Alex, I'm happy with this one.

Merging despite test success as I just added additional recipes to get a quick sanity check for the rest. The running recipes aren't required on PRs.

Issue #22141, opened 2022-09-19
Enable boundary nodes to move in the `SmoothMeshGenerator`


dschwen:
## Reason
https://github.com/libMesh/libmesh/pull/3399 improved the Laplace mesh smoother to permit boundary nodes to move along flat parts of the mesh boundary (and internal subdomain boundaries). This can significantly improve the mesh quality.

### Current smoothing result
![image](https://user-images.githubusercontent.com/202302/191128716-571edbe3-3c5f-44f6-a333-5caff92b5635.png)

### New smoothing result with movable boundary nodes

![image](https://user-images.githubusercontent.com/202302/191128789-b0a8bbd8-c8ef-4845-b5d7-351f15c1c127.png)

## Design
Add a new parameter to `SmoothMeshGenerator` to enable the libmesh functionality. This parameter will default to the old behavior where boundary nodes are immovable.

## Impact
Added capability.

dschwen:
(ignore the aspect ratio change in the example, those are two slightly different starting meshes)

PR #32085, opened 2025-12-11
Add Variational Mesh Smoother API

### LINKED ISSUE #28424: Variational Smoother

pbehne:
This PR adds an API and supporting documentation to use libMesh's `VariationalMeshSmoother` as developed in https://github.com/libMesh/libmesh/issues/4082.

Closes #28424, #22141.


test/tests/meshgenerators/smooth_mesh_generator/tests
```diff
@@ -1,13 +1,48 @@
...
-    exodiff = 'mesh_smoother_generator_in.e'
-    requirement = 'The system shall be able to use libmesh mesh modification tools.'
-    design = 'meshgenerators/SmoothMeshGenerator.md'
-    issues = '#11640'
-    mesh_mode = 'REPLICATED'
-    recover = false
-  [../]
+  design = 'meshgenerators/SmoothMeshGenerator.md'
+  [algorithms]
+    requirement = 'The system shall be able to use libmesh mesh modification tools:'
```
GiudGiud:
```suggestion
    requirement = 'The system shall be able to smooth a mesh using a:'
```

test/tests/meshgenerators/smooth_mesh_generator/tests
```diff
@@ -1,13 +1,48 @@
...
+    issue = '#11640 #28424 #22141'
+    [laplace_smooth_mesh_generator]
+      type = 'Exodiff'
+      input = 'laplace_mesh_smoother_generator.i'
+      cli_args = '--mesh-only'
+      exodiff = 'laplace_mesh_smoother_generator_in.e'
+      mesh_mode = 'REPLICATED'
+      recover = false
+      detail = 'lapalce smoothing algorithm'
```
GiudGiud:
```suggestion
      detail = 'Laplace smoothing algorithm'
```

test/tests/meshgenerators/smooth_mesh_generator/tests
```diff
@@ -1,13 +1,48 @@
...
+      input = 'variational_mesh_smoother_generator.i'
+      cli_args = '--mesh-only'
+      exodiff = 'variational_mesh_smoother_generator_in.e'
+      recover = false
+      detail = 'variational smoothing algorithm'
+    []
+  []
+
+  [warnings]
+    requirement = 'The SmoothMeshGenerator shall warn the user when:'
```
GiudGiud:
```suggestion
    requirement = 'The system shall emit an error if'
```

GiudGiud: warnings would be missed in mesh generation output. it can be quite lengthy on the console

test/tests/meshgenerators/smooth_mesh_generator/tests
```diff
@@ -1,13 +1,48 @@
...
+    issues = '#28424'
+    [unused_variational_params]
+       type = RunException
+       input = 'laplace_mesh_smoother_generator.i'
+       cli_args = 'Mesh/smooth/preserve_subdomain_boundaries=false --mesh-only'
+       mesh_mode = 'REPLICATED'
+       recover = false
+       detail = 'they set variational params for a laplace smooth'
```
GiudGiud:
```suggestion
       detail = 'variational smoothing parameters are passed for the laplace smoothing algorithm,'
```

test/tests/meshgenerators/smooth_mesh_generator/tests
```diff
@@ -1,13 +1,48 @@
...
+       expect_err = "SmoothMeshGenerator param 'preserve_subdomain_boundaries' applies to algorithm='variational' only and has no effect on the currently selected algorithm='laplace'."
+    []
+
+    [unused_laplace_params]
+       type = RunException
+       input = 'variational_mesh_smoother_generator.i'
+       cli_args = 'Mesh/smooth/iterations=10 --mesh-only'
+       recover = false
+       detail = 'they set laplace params for a variational smooth'
```
GiudGiud: same

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
 
-As an example here is an original mesh going through 12 iterations of this smoother:
+The Laplace smoothing algorithm utilizes a classical Laplacian smoother that iteratively relocates interior nodes to the average of their neighbors.
+Boundary and subdomain boundary nodes are not moved.
+This implementation currently requires a replicated mesh.
+
+## Variational Algorithm
+
+The variational smoothing algorithm is a Newton-based variational optimizer that minimizes a mixed *distortion–dilation* energy.
+It supports 1D/2D/3D meshes, preserves exterior boundaries by constraining nodes, can optionally preserve subdomain (block) boundaries, and can *untangle some* tangled meshes before smoothing.
```
GiudGiud, line 17: I think laplace smoother does too for all but last item. let s add those details above too

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+Internally it builds a libMesh system that assembles the gradient (residual) and analytic Hessian (jacobian) of the energy and solves to a stationary point.
+
+#### Mathematical Formulation
```
GiudGiud:
```suggestion
### Mathematical Formulation
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+| $n$ | spatial dimension of the mesh (1, 2, or 3) |
+| $\det S$ | jacobian determinant of the target-to-physical element mapping |
+| $\mathrm{tr}(S^{\mathsf T}S)$ | trace of the specified matrix |
+| $v$ | reference (i.e., desired) $\det S$, or "volume" for each element |
+| $\chi_\varepsilon(x) = \tfrac{1}{2}(x + \sqrt{\varepsilon^2 + x^2})$ | smooth barrier ensuring well-defined metrics for degenerate or folded elements (i.e., $\det S < 0$) |
+| $\varepsilon$ | small nonzero regularization constant |
+
+
+#### Notes & Limitations
+
```
GiudGiud:
```suggestion
### Notes & Limitations

```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+##### Target Elements
```
GiudGiud:
```suggestion
#### Target Elements
```

GiudGiud: same for the others

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
 
 !syntax description /Mesh/SmoothMeshGenerator
 
-## Example
+The `SmoothMeshGenerator` exposes two mesh smoothing algorithms, which are described below.
```
GiudGiud:
```suggestion
The `SmoothMeshGenerator` supports two mesh-smoothing algorithms, which are described below.
```
exposes is seeing it under a libmesh eye

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+
+The variational smoothing algorithm is a Newton-based variational optimizer that minimizes a mixed *distortion–dilation* energy.
+It supports 1D/2D/3D meshes, preserves exterior boundaries by constraining nodes, can optionally preserve subdomain (block) boundaries, and can *untangle some* tangled meshes before smoothing.
+Internally it builds a libMesh system that assembles the gradient (residual) and analytic Hessian (jacobian) of the energy and solves to a stationary point.
+
+#### Mathematical Formulation
+
+The Variational Mesh Smoother minimizes a *distortion–dilation energy functional*, $I_h(R)$, originally proposed by *Branets (2005)*:
+
+$I_h(R)=\sum_{c=1}^{N_{\text{elem}}} \int_{\hat{\Omega}_c} E_\theta(S_c(R))\, d\hat{\xi}\,,$
```
GiudGiud:
```suggestion
!equation
I_h(R)=\sum_{c=1}^{N_{\text{elem}}} \int_{\hat{\Omega}_c} E_\theta(S_c(R))\, d\hat{\xi}\,
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+- $\hat{\Omega}_c$ is the target element corresponding to physical element $c$
+- $\hat{\xi}$ is the target space coordinate
+
+The relationship between the target and reference elements is discussed below in the Notes & Limitations section.
+Note that for each element, integration is over the target space, not the physical space.
+Minimization is achieved using a damped Newton method with analytically derived gradient and Hessian of $E_\theta(S)$.
+
+For each mesh element, the local energy density is
+
+$E_\theta(S) = \theta\,\mu(S) + (1 - \theta)\,\beta(S) \,,$
```
GiudGiud:
```suggestion
!equation
E_\theta(S) = \theta\,\mu(S) + (1 - \theta)\,\beta(S)
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+
+where:
+
+- $0 \le \theta \le 1$ is the *dilation weight* (`dilation_weight` in input)
+- $\mu(S)$, penalizes *volumetric dilation* from the target element
+- $\beta(S)$, penalizes *shape distortion* from the target element
+
+The *distortion metric* is
+
+$\beta(S)=\dfrac{\left(\tfrac{1}{n}\,\mathrm{tr}(S^{\mathsf T}S)\right)^{n/2}}{\chi_\varepsilon(\det S)}$,
```
GiudGiud:
```suggestion
!equation
\beta(S)=\dfrac{\left(\tfrac{1}{n}\,\mathrm{tr}(S^{\mathsf T}S)\right)^{n/2}}{\chi_\varepsilon(\det S)}
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+
+and the *dilation metric* is
+
+$\mu(S)=\dfrac{1}{2\,\chi_\varepsilon(\det S)}\left(v + \dfrac{(\det S)^2}{v}\right)$.
```
GiudGiud:
```suggestion
!equation
\mu(S)=\dfrac{1}{2\,\chi_\varepsilon(\det S)}\left(v + \dfrac{(\det S)^2}{v}\right)
```

framework/src/meshgenerators/SmoothMeshGenerator.C
```diff
@@ -24,34 +25,85 @@ SmoothMeshGenerator::validParams()
...
+    if (!old_mesh->is_replicated())
+      mooseError(
+          "SmoothMeshGenerator with algorithm='laplace' is not implemented for distributed meshes");
 
-  libMesh::LaplaceMeshSmoother lms(static_cast<UnstructuredMesh &>(*mesh), _iterations);
+    mesh = dynamic_pointer_cast<ReplicatedMesh>(old_mesh);
 
-  lms.smooth();
+    smoother = std::make_unique<libMesh::LaplaceMeshSmoother>(
+        static_cast<UnstructuredMesh &>(*mesh), _iterations);
```
roystgnr: As long as we're futzing with this code anyway - can we get rid of this static_cast?  I'd bet it was originally a cast from MeshBase, which would have been necessary, but every ReplicatedMesh casts implicitly to UnstructuredMesh.

For that matter, instead of testing `old_mesh->is_replicated()`, we could just move the `mesh` `dynamic_pointer_cast` up and test to see if the pointer is empty afterwards (every `ReplicatedMesh` `is_replicated()`) for simplicity.

test/tests/meshgenerators/smooth_mesh_generator/tests
```diff
@@ -1,24 +1,49 @@
...
+    requirement = 'The system shall be able to use libmesh mesh modification tools:'
+    [laplace_smooth_mesh_generator]
+      type = 'Exodiff'
+      input = 'laplace_mesh_smoother_generator.i'
+      cli_args = '--mesh-only'
+      exodiff = 'laplace_mesh_smoother_generator_in.e'
+      issues = '#11640'
+      mesh_mode = 'REPLICATED'
+      recover = false
+      detail = 'lapalce smoothing algorithm'
```
roystgnr: laplace

roystgnr (changes requested):

roystgnr:
One more subtle concern: have you timed the new tests here?  The variational smoother tests are currently some of the most heavyweight unit tests in libMesh.

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+| Element Type | Reference Element | Target Element |
+|:--|:--|:--|
+| EDGE | line | line |
+| TRI | right triangle | equilateral triangle |
+| QUAD | square | square |
+| PRISM | right triangular base | equilateral triangular base, equal face areas |
+| HEX | cube | cube |
+| PYRAMID | square base, isosceles triangular sides  | square base, equilateral triangular sides |
+| TET | right tet | regular tet |
+
```
GiudGiud, line 93: what about polys?

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+The term reference "volume" is a bit misleading.
+A more descriptive, albeit longer, term is "reference target-to-physical jacobian determinant".
+Basically, $v$ is the jacobian determinant that we want the smoother to push our mesh elements towards.
+The term "volume" is used here because the jacobian determinant is proportional to the volume of an element.
+When a given mesh element has $\det S \neq v\,,$ the element is considered to be "dilated" from an element with $\det S = v\,.$
+
+The reference volume is automatically calculated as the volumetric average of $\det S$ for each element, averaged over the number of elements in the mesh:
+
+$v = \dfrac{1}{N_\text{elem}} \sum_{c=1}^{N_\text{elem}} \dfrac{\int_{\hat{\Omega}_c} \det S_c d\hat{\xi}}{\int_{\hat{\Omega}_c} d\hat{\xi}} \,.$
```
GiudGiud:
```suggestion
!equation
v = \dfrac{1}{N_\text{elem}} \sum_{c=1}^{N_\text{elem}} \dfrac{\int_{\hat{\Omega}_c} \det S_c d\hat{\xi}}{\int_{\hat{\Omega}_c} d\hat{\xi}}
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+The barrier term $\chi_\varepsilon(x)$ behaves like $x$ for $x \gg 0$, is nonzero at $x = 0$, and asymptotes to 0 as $x$ approaches $-\infty$.
+The use of this term in place of $\det S$ in denominators allows the smoother to untangle meshes with degenerate/folded elements.
+
+Once the mesh has been untangled, $\varepsilon$ is set to zero to prevent re-tangling during the smoothing solve.
+
+##### Optimal Solutions
+
+It can be shown that, for the smoothing solve, when $\varepsilon = 0$ and $\chi_\varepsilon(\det S) = \det S$, the distortion metric $\beta(S)$ is minimized by any constant-scaled special orthogonal matrix such that
+
+$S^TS = cI \,,\qquad \det S = c^n \,,\qquad \forall c > 0 \,.$
```
GiudGiud:
```suggestion
!equation
S^TS = cI \,,\qquad \det S = c^n \,,\qquad \forall c > 0
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+
+The implication of this is that the distortion metric is minimized by a mesh element that is a scaled rotation of the target element.
+This makes sense because the target element is our definition of the "ideal" smooth element shape.
+
+For untangled meshes, the dilation metric simplifies to
+
+$\mu(S)=\dfrac{1}{2}\left(\dfrac{v}{\det S} + \dfrac{\det S}{v}\right) \,.$
```
GiudGiud:
```suggestion
!equation
\mu(S)=\dfrac{1}{2}\left(\dfrac{v}{\det S} + \dfrac{\det S}{v}\right)
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
 !media media/mesh/smooth.gif
-       id=inl-logo
+       id=laplace-smoother
        caption=12 iterations of Laplacian smoothing.  Coloring is by element quality (higher is better).
        style=width:50%;padding:20px;
 
+### Example 2 — Variational smoothing (fixed boundary nodes)
+
+Here is an example of the variational smoother applied to a mesh with 3 subdomains and nonlinear boundaries.
+Note that the locations of the (subdomain)boundary nodes do not change.
```
GiudGiud:
```suggestion
Note that the locations of the (subdomain) boundary nodes do not change.
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+
+!media media/mesh/variational_smoother_ex1.png
+       id=variational-smoother-1
+       caption=Variational smoother applied to mesh with nonlinear boundary. The original mesh is shown on the left and the smoothed mesh is shown on the right.
+       style=width:50%;padding:20px;
+
+### Example 3 — Variational smoothing (sliding boundary nodes)
+
+Here is an example of the variational smoother applied to a mesh with 2 subdomains and linear boundaries.
+Note that while the locations of the (subdomain)boundary nodes change, these boundaries are still preserved.
```
GiudGiud:
```suggestion
Note that while the locations of the (subdomain) boundary nodes change, these boundaries are still preserved.
```

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
```diff
@@ -2,20 +2,189 @@
...
+
+If the surfaces are not linear, the boundary/subdomain nodes are constrained (i.e., fixed) to their original location.
+It has been observed in some cases that fixing boundary nodes on nonlinear boundaries significantly restricts the space of optimally smooth solutions.
+As a result, the original mesh may not change significantly.
+This is because fixing boundary nodes reduces the space of possible smooth boundary elements.
+These restrictions then propagate into the mesh interior.
+In the future, it will be useful to allow boundary nodes to slide along nonlinear boundaries.
+This is not currently possible because libMesh does not yet support nonlinear constraints.
+
+For further theoretical background, see: Larisa V. Branets, *A Variational Grid Optimization Method Based on a Local Cell Quality Metric*, Ph.D. Dissertation, University of Texas at Austin, 2005.
```
GiudGiud: let's add an entry to the bib instead

framework/src/meshgenerators/SmoothMeshGenerator.C
```diff
@@ -24,34 +25,120 @@ SmoothMeshGenerator::validParams()
   InputParameters params = MeshGenerator::validParams();
 
   params.addRequiredParam<MeshGeneratorName>("input", "The mesh we want to smooth.");
-  params.addClassDescription("Utilizes a simple Laplacian based smoother to attempt to improve "
-                             "mesh quality.  Will not move boundary nodes or nodes along "
-                             "block/subdomain boundaries");
+  params.addClassDescription(
+      "Utilizes the specified smoothing algorithm to attempt to improve "
+      "mesh quality. The Laplace algorithm does not move boundary nodes or nodes along "
+      "block/subdomain boundaries.");
```
GiudGiud:
```suggestion
      "mesh quality.");
```

framework/src/meshgenerators/SmoothMeshGenerator.C
```diff
@@ -24,34 +25,120 @@ SmoothMeshGenerator::validParams()
...
 
-  params.addParam<unsigned int>("iterations", 1, "The number of smoothing iterations to do.");
+  MooseEnum SmootherAlgorithm("variational laplace", "variational");
+  params.addParam<MooseEnum>("algorithm", SmootherAlgorithm, "The smoothing algorithm to use.");
+  params.addParam<unsigned int>(
+      "iterations", 1, "Laplace algorithm only: the number of smoothing iterations to do.");
```
GiudGiud, line 34: if we are using newton's method for variational, why can't we set the (max) number of iterations?

framework/src/meshgenerators/SmoothMeshGenerator.C
```diff
@@ -24,34 +25,120 @@ SmoothMeshGenerator::validParams()
...
+
+    else // _algorithm == "variational"
+    {
+      other_algorithm = "laplace";
+      check_params = {"iterations"};
+    }
+
+    for (const auto & param_name : check_params)
+      if (isParamSetByUser(param_name))
+        mooseWarning(type(),
```
GiudGiud:
```suggestion
        mooseError(
```
type is already output

GiudGiud (commented):
very nice I am excited to use this on the MCRE and the blanket sector meshes!

GiudGiud:
> One more subtle concern: have you timed the new tests here? The variational smoother tests are currently some of the most heavyweight unit tests in libMesh.

See latest CIVET post. Looks very reasonable to me!

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
roystgnr: I don't believe Laplace can preserve subdomain boundaries.

There's also a difference between the "constraining nodes" Laplace does (fix nodes in place) and the "constraining nodes" that only Variational can do (fixing nodes to only slide along a line or a plane, to allow the node to move without allowing the boundary to move), though the doc doesn't clarify the distinction until later.  Maybe it would be clear enough to say that Laplace "tries to preserve exterior boundaries by fixing boundary nodes in place" and Variational "preserves exterior boundaries by only allowing boundary node movement that leaves the domain unchanged"?

I write "tries to preserve", because Laplace mesh "smoothing" can actually tangle a mesh and send interior nodes outside the mesh boundary when it's faced with a reentrant corner.  It's just not a good algorithm.  We only have the "iterations" control because sometimes a couple iterations can be a big improvement where iterating to convergence would give garbage.

I'm not saying we should put "This sucks, don't use it" in the Laplace option docs, but now that there's finally a much better alternative I would like users to be aware that it does suck and they shouldn't use it.  ;-)

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
pbehne: Good catch. We forgot to test the smoother on polygons and polyhedra. From the description of the [C0Polygon](https://mooseframework.inl.gov/docs/doxygen/libmesh/classlibMesh_1_1C0Polygon.html#details), it seems like the current reference (master) element is already the idea shape (i.e., equal to the target element). Because of this, no new code should be necessary in the `VariationalMeshSmoother`, but adding a test to libmesh would probably be wise.

From the desription of the [C0Polyhedron](https://mooseframework.inl.gov/docs/doxygen/libmesh/classlibMesh_1_1C0Polyhedron.html#details), it seems like Roy gave up:
> We've practically got to have a definition distinct for every physical element anyway, and we've got to give FE and QBase objects access to the physical element, so we'll just do all our FE and Quadrature calculations there.

@roystgnr, how would this affect the logic in the `VariationalMeshSmoother`?

roystgnr: Yup, the master element should work as the target element for `C0Polygon`.

For `C0Polyhedron` ... I wouldn't dare risk smoothing them in general without a major upgrade.  They currently assume planar faces around a convex volume, for instance, and our smoother doesn't keep planar faces planar if they've got more than 3 nodes ... nor would it guarantee convexity; it would prevent individual internal tets from inverting but it wouldn't stop them from bending enough to give the total polyhedron a reentrant edge.

framework/doc/content/source/meshgenerators/SmoothMeshGenerator.md
pbehne: > tries to preserve exterior boundaries by fixing boundary nodes in place

Just to be clear, does Laplace preserve exterior boundary nodes, or does it just "try" to preserve them, like it tries to preserve subdomain boundary nodes?

GiudGiud: i think it preserves exterior boundary nodes. Never observed the contrary

roystgnr: Laplace does preserve exterior boundary nodes.  It does not necessarily preserve exterior boundaries.  It does not even try to preserve subdomain boundary nodes.

framework/src/meshgenerators/SmoothMeshGenerator.C
pbehne: As in the number of nonlinear iterations? We could in the future after adding a setter method to the `VariationalMeshSmoother` class and updating the libMesh submodule. I think for now, the ability to set the absolute and relative residual tolerances should give the user sufficient flexibility.

framework/src/meshgenerators/SmoothMeshGenerator.C
```diff
@@ -109,27 +109,24 @@ SmoothMeshGenerator::SmoothMeshGenerator(const InputParameters & parameters)
 std::unique_ptr<MeshBase>
 SmoothMeshGenerator::generate()
 {
-  std::unique_ptr<MeshBase> mesh = nullptr;
+  std::unique_ptr<MeshBase> input = std::move(_input);
+  std::unique_ptr<UnstructuredMesh> mesh = dynamic_pointer_cast<UnstructuredMesh>(input);
```
roystgnr: Could these two lines just be replaced with `std::unique_ptr<UnstructuredMesh> mesh = dynamic_pointer_cast<UnstructuredMesh>(_input);` ?  `dynamic_pointer_cast` also effectively moves a `unique_ptr` argument, doesn't it?

framework/src/meshgenerators/SmoothMeshGenerator.C
```diff
@@ -109,27 +109,24 @@ SmoothMeshGenerator::SmoothMeshGenerator(const InputParameters & parameters)
...
   std::unique_ptr<libMesh::MeshSmoother> smoother = nullptr;
 
   if (_algorithm == "laplace")
   {
-    std::unique_ptr<MeshBase> old_mesh = std::move(_input);
-    if (!old_mesh->is_replicated())
+    if (!mesh->is_replicated())
```
roystgnr: Now that I think about it - could we weaken this to `if (!mesh->is_serial())`?  That's the actual requirement that we need for the libMesh Laplace smoother - we don't care if the mesh could be distributed later, just as long as it's serialized at the moment.

roystgnr (approved):
I'm happy with this as-is, but let's wait for @GiudGiud to get another look too, and if he still sees real issues I have some trivia that could be updated alongside them.

framework/src/meshgenerators/SmoothMeshGenerator.C
pbehne: > dynamic_pointer_cast also effectively moves a unique_ptr argument, doesn't it?

Seems like it:

```
template <typename T_DEST, typename T_SRC>
std::unique_ptr<T_DEST>
dynamic_pointer_cast(std::unique_ptr<T_SRC> & src)
{
  if (!src)
    return std::unique_ptr<T_DEST>(nullptr);

  T_DEST * dest_ptr = dynamic_cast<T_DEST *>(src.get());
  if (!dest_ptr)
    return std::unique_ptr<T_DEST>(nullptr);

  std::unique_ptr<T_DEST> dest_temp(dest_ptr);

  src.release();

  return dest_temp;
}
```

roystgnr:
Wait a minute, Guillaume just started his vacation, didn't he?  On a plane today, and hopefully having fun rather than reviewing smoother code during the next three weeks?  I'll say you addressed his review well enough and take the flak if we're wrong.  ;-)

framework/src/meshgenerators/SmoothMeshGenerator.C
roystgnr: Agreed.  "Stop when your work is good enough" is a strict improvement over "Stop when you've worked long enough".  ;-)

pbehne:
> Guillaume just started his vacation, didn't he?

Is Guillaume ever REALLY on vacation?

PR #31985, opened 2025-11-26
Add memory tracking to the test harness

loganharbour:
Closes https://github.com/idaholab/moose/issues/27479, recreation of https://github.com/idaholab/moose/pull/27480

Issue #27479, opened 2024-04-26
Track approximate max memory in the test harness


loganharbour:
### Motivation

Now that we're running larger tests with the test harness, we need to be equitable in terms of memory usage. That is, we want to make sure that we don't oversubscribe CI with our tests in terms of memory.

### Design

Track memory with the TestHarness and make it an option for output. Add a `--max-memory` option or something of the sort to fail tests if they use too much memory.

### Impact

CI/CD equitability.

PR #32090, opened 2025-12-12
Remove type hint as psutil may not be available

loganharbour:
refs #31985, #27479, #32077


roystgnr (approved):
This fixes the regression in https://github.com/idaholab/moose/pull/31985#issuecomment-3648239252 for me.

PR #32097, opened 2025-12-15
MFEM mixed forms fix

### LINKED ISSUE #32096: MFEM mixed bilinear forms test and trial space ordering issues
### Bug Description

When using MFEM to solve a mixed system (i. e. different test and trial spaces), the `FormSystemMatrix` method in the `EquationSystem` class calls 

`mblf->FormRectangularLinearSystem(_ess_tdof_lists.at(j),
                                          _ess_tdof_lists.at(i),
                                          *(_var_ess_constraints.at(j)),
                                          aux_lf,
                                          *aux_a,
                                          aux_x,
                                          aux_rhs);`

where `i` represents the index of the test finite element space, and `j` represents the trial space. In particular, we wish to use the trial variable within the `j`th entry in `_var_ess_constraints`. However, as it currently stands, `_var_ess_constraints` has been created with the ordering of test spaces, not trial spaces. In square systems, with no mixed kernels, the `_test_var_names` and `_trial_var_names` are created with the same ordering of spaces, so this isn't an issue. However, in mixed systems, depending on how the `Kernels` block in the MOOSE script was arranged, the orderings might not be the same and the simulation will crash.

A secondary issue is that when `SetSize()` is called on `_h_blocks`, it only allocates space for the blocks but it doesn't clear what might be in that space, so in some cases it might cause segfaults if the blocks in `h_blocks` are not initialised as `nullptr`s.

### Steps to Reproduce

With the current version of MOOSE, run the following script:

```
[Mesh]
  type = MFEMMesh
  file = ../mesh/square.e
  dim = 2
[]

[Problem]
  type = MFEMProblem
[]

[FESpaces]
  [H1FESpace]
    type = MFEMScalarFESpace
    fec_type = H1
    fec_order = FIRST
  []
  [HCurlFESpace]
    type = MFEMVectorFESpace
    fec_type = ND
    fec_order = FIRST
  []
[]

[Variables]
  [E]
    type = MFEMVariable
    fespace = HCurlFESpace
  []
  [p]
    type = MFEMVariable
    fespace = H1FESpace
  []
[]

[Kernels]
  [scalarcurl_E]
    type = MFEMMixedScalarCurlKernel
    trial_variable = E
    variable = p
    coefficient = 1.0
  []
  [mass_p]
    type = MFEMMassKernel
    variable = p
    coefficient = 1.0
  []
  [mass_E]
    type = MFEMVectorFEMassKernel
    variable = E
    coefficient = 1.0
  []
[]

[Solver]
  type = MFEMSuperLU
[]

[Executioner]
  type = MFEMSteady
  assembly_level = legacy
[]
```

It will likely segfault, due to the secondary issue described above. The solution to the segfault is to initialise all blocks in `_h_blocks` as `nullptr`s before forming the linear system. This is done in the `ClearAllBlocks()` method in the PR that will be opened to fix this issue. However, even after fixing the segfault, the code will still crash due to the ordering of the FE spaces. Note, however, that if you change the order of the kernels in this script by putting the `scalarcurl_E` last in the `Kernels` block, the code should run normally. That is because in this case the `_test_var_names` and `_trial_var_names` end up with the same ordering.

### Impact

Depending on the particularities of the mixed system of equations being solved for, this issue might completely preclude the user from solving a perfectly valid and well-defined problem.

Heinrich-BR:
Closes #32096

### DIFF
--- a/framework/src/mfem/equation_systems/EquationSystem.C
+++ b/framework/src/mfem/equation_systems/EquationSystem.C
@@ -69,6 +69,25 @@ EquationSystem::SetTrialVariableNames()
   }
 }
 
+void EquationSystem::SortTestAndTrialNames()
+{
+  std::vector<std::string> common_vars;
+  for (const auto & name : _test_var_names)
+  {
+    if (VectorContainsName(_trial_var_names, name))
+      common_vars.push_back(name);      
+  }
+
+  // Move common variables to beginning of both vectors, and in the same order
+  for (const auto & name : common_vars)
+  {
+    _test_var_names.erase(std::find(_test_var_names.begin(), _test_var_names.end(), name));
+    _test_var_names.insert(_test_var_names.begin(), name);
+    _trial_var_names.erase(std::find(_trial_var_names.begin(), _trial_var_names.end(), name));
+    _trial_var_names.insert(_trial_var_names.begin(), name);
+  }
+}
+
 void
 EquationSystem::AddKernel(std::shared_ptr<MFEMKernel> kernel)
 {
@@ -138,6 +157,12 @@ EquationSystem::Init(Moose::MFEM::GridFunctions & gridfunctions,
     mooseError("Complex variables have been created but the executioner numeric type has not been "
                "set to complex. Please set Executioner/numeric_type = complex.");
 
+  // Extract which coupled variables are to be trivially eliminated and which are trial variables
+  SetTrialVariableNames();
+
+  // Now that all test and trial variables are known, order them
+  SortTestAndTrialNames();
+
   for (auto & test_var_name : _test_var_names)
   {
     if (!gridfunctions.Has(test_var_name))
@@ -149,13 +174,22 @@ EquationSystem::Init(Moose::MFEM::GridFunctions & gridfunctions,
     }
     // Store pointers to test FESpaces
     _test_pfespaces.push_back(gridfunctions.Get(test_var_name)->ParFESpace());
-    // Create auxiliary gridfunctions for storing essential constraints from Dirichlet conditions
-    _var_ess_constraints.emplace_back(
-        std::make_unique<mfem::ParGridFunction>(gridfunctions.Get(test_var_name)->ParFESpace()));
   }
 
-  // Extract which coupled variables are to be trivially eliminated and which are trial variables
-  SetTrialVariableNames();
+  for (auto & trial_var_name : _trial_var_names)
+  {
+    if (!gridfunctions.Has(trial_var_name))
+    {
+      mooseError("MFEM variable ",
+                 trial_var_name,
+                 " requested by equation system during initialisation was "
+                 "not found in gridfunctions");
+    }
+    // Create auxiliary gridfunctions for storing essential constraints from Dirichlet conditions
+    _var_ess_constraints.emplace_back(
+        std::make_unique<mfem::ParGridFunction>(gridfunctions.Get(trial_var_name)->ParFESpace()));
+  }
+
 
   // Store pointers to FESpaces of all coupled variables
   for (auto & coupled_var_name : _coupled_var_names)
@@ -192,16 +226,16 @@ EquationSystem::ApplyEssentialBC(const std::string & var_name,
 void
 EquationSystem::ApplyEssentialBCs()
 {
-  _ess_tdof_lists.resize(_test_var_names.size());
-  for (const auto i : index_range(_test_var_names))
+  _ess_tdof_lists.resize(_trial_var_names.size());
+  for (const auto i : index_range(_trial_var_names))
   {
-    const auto & test_var_name = _test_var_names.at(i);
+    const auto & trial_var_name = _trial_var_names.at(i);
     mfem::ParGridFunction & trial_gf = *(_var_ess_constraints.at(i));
     mfem::Array<int> global_ess_markers(trial_gf.ParFESpace()->GetParMesh()->bdr_attributes.Max());
     global_ess_markers = 0;
     // Set strongly constrained DoFs of trial_gf on essential boundaries and add markers for all
     // essential boundaries to the global_ess_markers array
-    ApplyEssentialBC(test_var_name, trial_gf, global_ess_markers);
+    ApplyEssentialBC(trial_var_name, trial_gf, global_ess_markers);
     trial_gf.ParFESpace()->GetEssentialTrueDofs(global_ess_markers, _ess_tdof_lists.at(i));
   }
 }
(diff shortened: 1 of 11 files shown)


framework/src/mfem/equation_systems/EquationSystem.C
```diff
@@ -272,6 +313,8 @@ EquationSystem::FormSystemMatrix(mfem::OperatorHandle & op,
   // Allocate block operator
   DeleteAllBlocks();
   _h_blocks.SetSize(_test_var_names.size(), _test_var_names.size());
+  ClearAllBlocks();
+
```
nmnobre:
```suggestion
  _h_blocks = nullptr;
```

framework/src/mfem/equation_systems/EquationSystem.C
```diff
@@ -142,14 +146,22 @@ EquationSystem::Init(Moose::MFEM::GridFunctions & gridfunctions,
...
     _test_pfespaces.push_back(gridfunctions.Get(test_var_name)->ParFESpace());
+  }
+
+  for (auto & trial_var_name : _trial_var_names)
+  {
+    if (!gridfunctions.Has(trial_var_name))
+    {
+      mooseError("MFEM variable ",
+                 trial_var_name,
+                 " requested by equation system during initialisation was "
```
nmnobre: initialisation -> initialization (change all three appearances)

framework/src/mfem/problem_operators/ComplexEquationSystemProblemOperator.C
```diff
@@ -17,28 +17,42 @@ void
...
   }
-  _block_true_offsets.PartialSum();
+  _block_true_offsets_trial.PartialSum();
 
-  _true_x.Update(_block_true_offsets);
-  _true_rhs.Update(_block_true_offsets);
+  // Set operator size and block structure for test spaces
+  _block_true_offsets_test.SetSize(_cmplx_test_variables.size() + 1);
+  _block_true_offsets_test[0] = 0;
+  for (unsigned int ind = 0; ind < _cmplx_test_variables.size(); ++ind)
```
nmnobre: change all instances to use `index_range()` instead

nmnobre (changes requested):
Thanks Henrique, I love the conceptual distinction between trial and test variables that this PR is pushing for :)

I've taken the liberty to push a commit that guarantees test and trial variables are in the same order. We probably don't want to place diagonal blocks (i.e. those with the same trial and test variable) outside the diagonal or off-diagonal blocks (i.e. different trial and test variables) on the diagonal; this would be a headache for conditioning, and preconditioning in particular - I don't doubt we could work out the algebra, but I fear the amount of technical debt we'd be buying into isn't justifiable.

So let's also assert that:
a) the no. of trial and test variables is the same (this will force future us to intentionally remove the assertion when we're ready to support rectangular, i.e. underconstrained, systems).
b) `i == j` when `test_var_name == trial_var_name` in `FormSystemMatrix()`

Inside that `if (test_var_name == trial_var_name)`, since we use `test_var_name`, let's also use `i` and not `j` (now we're asserting they're equal anyway).

Heinrich-BR:
Thank you for the comments @nmnobre ! I've addressed them now. My only question is about this:

> Inside that `if (test_var_name == trial_var_name)`, since we use `test_var_name`, let's also use `i` and not `j` (now we're asserting they're equal anyway).

I understand that they're equal anyway, but isn't it cleaner if we keep the conceptually correct index?

framework/src/mfem/problem_operators/ComplexEquationSystemProblemOperator.C
```diff
@@ -17,28 +17,42 @@ void
...
   }
-  _block_true_offsets.PartialSum();
+  _block_true_offsets_trial.PartialSum();
 
-  _true_x.Update(_block_true_offsets);
-  _true_rhs.Update(_block_true_offsets);
+  // Set operator size and block structure for test spaces
+  _block_true_offsets_test.SetSize(_cmplx_test_variables.size() + 1);
+  _block_true_offsets_test[0] = 0;
+  for (const auto ind : index_range(_cmplx_trial_variables))
```
nmnobre:
```suggestion
  for (const auto ind : index_range(_cmplx_test_variables))
```

framework/src/mfem/problem_operators/ProblemOperatorBase.C
```diff
@@ -25,22 +25,30 @@ void
...
   _trial_variables = _problem_data.gridfunctions.Get(_trial_var_names);
+  _test_variables = _problem_data.gridfunctions.Get(_test_var_names);
 
-  // Set operator size and block structure
-  _block_true_offsets.SetSize(_trial_variables.size() + 1);
-  _block_true_offsets[0] = 0;
+  // Set operator size and block structure for trial spaces
+  _block_true_offsets_trial.SetSize(_trial_variables.size() + 1);
+  _block_true_offsets_trial[0] = 0;
   for (unsigned int ind = 0; ind < _trial_variables.size(); ++ind)
```
nmnobre:
```suggestion
  for (const auto ind : index_range(_trial_variables))
```

framework/src/mfem/problem_operators/ProblemOperatorBase.C
```diff
@@ -25,22 +25,30 @@ void
...
-  _block_true_offsets.PartialSum();
+    _block_true_offsets_trial[ind + 1] = _trial_variables.at(ind)->ParFESpace()->TrueVSize();
+  _block_true_offsets_trial.PartialSum();
 
-  _true_x.Update(_block_true_offsets);
-  _true_rhs.Update(_block_true_offsets);
+  // Set operator size and block structure for test spaces
+  _block_true_offsets_test.SetSize(_test_variables.size() + 1);
+  _block_true_offsets_test[0] = 0;
+  for (unsigned int ind = 0; ind < _test_variables.size(); ++ind)
```
nmnobre:
```suggestion
  for (const auto ind : index_range(_test_variables))
```

framework/src/mfem/problem_operators/ProblemOperatorBase.C
nmnobre:
```suggestion
  for (const auto ind : index_range(_trial_variables))
```

nmnobre (changes requested):

nmnobre:
> but isn't it cleaner if we keep the conceptually correct index?

Fair, agreed.

nmnobre (approved):

nmnobre:
Test failures unrelated.

PR #32099, opened 2025-12-15
Improve MFEMRWTE10IntegratedBC's implementation and documentation

nmnobre:
## Reason and Design
Initially the idea was simply to fix the typos/broken links in the docs, but then I thought I could better align the code with the docs for someone reading the implementation with fresh eyes, so that's what I did.

## Impact
Improved readability. Documentation correctness.

### DIFF
--- a/framework/doc/content/source/mfem/auxkernels/MFEMGradAux.md
+++ b/framework/doc/content/source/mfem/auxkernels/MFEMGradAux.md
@@ -5,7 +5,7 @@
 ## Overview
 
 AuxKernel for calculating the gradient of a scalar $H^1$ conforming source variable and storing it in
-a scalar elemental result variable defined on an $H(\mathrm{curl})$ conforming Nédélec finite element space.
+a result variable defined on an $H(\mathrm{curl})$ conforming Nédélec finite element space.
 
 The result may be scaled by an optional (global) scalar factor.
 
--- a/framework/doc/content/source/mfem/bcs/MFEMRWTE10IntegratedBC.md
+++ b/framework/doc/content/source/mfem/bcs/MFEMRWTE10IntegratedBC.md
@@ -4,31 +4,31 @@
 
 ## Overview
 
-Adds the frequency space integrator corresponding to the value upon a cross-sectional boundary of the TE10 mode of an electromagnetic wave traveling through an infinite rectangular waveguide. The physical system in question is explained in more detail [here](https://phys.libretexts.org/Bookshelves/Electricity_and_Magnetism/Electromagnetics_II_(Ellingson)/06%3A_Waveguides/6.09%3A_Rectangular_Waveguide-_TE_Modes). 
+Adds the frequency space integrator corresponding to the value upon a cross-sectional boundary of the TE10 mode of an electromagnetic wave traveling through an infinite rectangular waveguide. The physical system in question is explained in more detail [here](https://phys.libretexts.org/Bookshelves/Electricity_and_Magnetism/Electromagnetics_II_(Ellingson%29/06%3A_Waveguides/6.09%3A_Rectangular_Waveguide-_TE_Modes).
 
 If the parameter [!param](/BCs/MFEMRWTE10IntegratedBC/input_port) is set to `false`, corresponding to the output boundary of the waveguide, the following boundary integrator is applied:
 
 !equation
-iIm[k]\mu_0^{-1}(\vec u, \vec v)_{\partial\Omega} \,\,\, \forall v \in V
+i\mathrm{Im}[k]\mu^{-1}(\vec u, \vec v)_{\partial\Omega} \,\,\, \forall \vec v \in V
 
-where $\vec v \in H(\mathrm{curl})$ or $H(\mathrm{div})$, $k$ is a complex scalar coefficient defined as follows:
+where $\vec u, \vec v \in H(\mathrm{curl})$ or $H(\mathrm{div})$, $k$ is a complex scalar coefficient defined as follows:
 
 !equation
-k = i\sqrt{(2\pi\nu)^2\mu_0\epsilon_0-\frac{\pi}{|\vec a_1|}}
+k = i\sqrt{(2\pi\nu)^2\mu\epsilon-\frac{\pi^2}{|\vec a_1|^2}}
 
-where $\nu$ is the [!param](/BCs/MFEMRWTE10IntegratedBC/frequency) parameter, $\vec a_1$ is the port length vector, and $\mu_0$ and $\epsilon_0$ are the magnetic permeability and electric permittivity of the material in which the wave is propagating. By default, these are set to be the permeability and permittivity of vacuum.
+where $\nu$ is the [!param](/BCs/MFEMRWTE10IntegratedBC/frequency) parameter, $\vec a_1$ is the port length vector, and $\mu$ and $\epsilon$ are the magnetic permeability and electric permittivity of the material in which the wave is propagating.
 
 If the parameter [!param](/BCs/MFEMRWTE10IntegratedBC/input_port) is set to `true`, corresponding to the input boundary of the waveguide, the following boundary integrator is applied:
 
 !equation
-iIm[k]\mu_0^{-1}(\vec u, \vec v)_{\partial\Omega} + 2Im[k]\mu_0^{-1}(\vec n \times (Im[\vec f]-iRe[\vec f]), \vec v)_{\partial\Omega}  \,\,\, \forall v \in V
+i\mathrm{Im}[k]\mu^{-1}(\vec u, \vec v)_{\partial\Omega} + 2\mathrm{Im}[k]\mu^{-1}(\vec n \times (\mathrm{Im}[\vec f]-i\mathrm{Re}[\vec f]), \vec v)_{\partial\Omega}  \,\,\, \forall \vec v \in V
 
 where $\vec f$ is defined by 
 
 !equation
-\vec f=\sqrt{\frac{2\omega\mu_0}{|\vec a_1||\vec a_2|Im[k]}} \sin(\vec k_a \cdot \vec x)e^{-i\vec k_c \cdot x}\hat e
+\vec f=\sqrt{\frac{2\omega\mu}{|\vec a_1||\vec a_2|\mathrm{Im}[k]}} \sin(\vec k_a \cdot \vec x)e^{-i\vec k_c \cdot \vec x}\hat e
 
-with $\vec a_2$ being the port width vector, $\vec k_a=\vec a_2 \times \vec k_c$ and $\vec k_c=\vec a_1 \times \vec a_2$. Lastly, $\hat e$ is a unit vector in the direction of $\vec k_c \times \vec k_a$.
+with $\vec a_2$ being the port width vector, $\vec k_a=\frac{\pi}{|\vec a_1|} \frac{\vec a_2 \times \vec k_c}{|\vec a_2 \times \vec k_c|}$ and $\vec k_c=k \frac{\vec a_1 \times \vec a_2}{|\vec a_1 \times \vec a_2|}$. Lastly, $\hat e$ is a unit vector in the direction of $\vec k_c \times \vec k_a$.
 
 
 ## Example Input File Syntax
--- a/framework/doc/content/source/mfem/bcs/MFEMRWTE10IntegratedBC.md
+++ b/framework/doc/content/source/mfem/bcs/MFEMRWTE10IntegratedBC.md
@@ -23,7 +23,7 @@ If the parameter [!param](/BCs/MFEMRWTE10IntegratedBC/input_port) is set to `tru
 !equation
 i\mathrm{Im}[k]\mu^{-1}(\vec u, \vec v)_{\partial\Omega} + 2\mathrm{Im}[k]\mu^{-1}(\vec n \times (\mathrm{Im}[\vec f]-i\mathrm{Re}[\vec f]), \vec v)_{\partial\Omega}  \,\,\, \forall \vec v \in V
 
-where $\vec f$ is defined by 
+where $\vec f$ is defined by
 
 !equation
 \vec f=\sqrt{\frac{2\omega\mu}{|\vec a_1||\vec a_2|\mathrm{Im}[k]}} \sin(\vec k_a \cdot \vec x)e^{-i\vec k_c \cdot \vec x}\hat e
--- a/framework/include/mfem/bcs/MFEMRWTE10IntegratedBC.h
+++ b/framework/include/mfem/bcs/MFEMRWTE10IntegratedBC.h
@@ -40,56 +40,28 @@ public:
     return new mfem::VectorFEMassIntegrator(_robin_coef_im.get());
   }
 
-  // Utilities
-  mfem::Vector CrossProduct(const mfem::Vector & va, const mfem::Vector & vb)
+  /// Computes the unit vector in the direction of the cross product of two 3d vectors
+  mfem::Vector normalizedCrossProduct(const mfem::Vector & va, const mfem::Vector & vb)
   {
     mfem::Vector vec;
-    vec.SetSize(3);
-    vec[0] = va[1] * vb[2] - va[2] * vb[1];
-    vec[1] = va[2] * vb[0] - va[0] * vb[2];
-    vec[2] = va[0] * vb[1] - va[1] * vb[0];
-    return vec;
-  }
-
-  // Utilities
-  mfem::real_t InnerProduct(const mfem::Vector & va, const mfem::Vector & vb)
-  {
-    if (va.Size() != vb.Size())
-      mooseError("InnerProduct: vectors must have the same size");
-
-    mfem::real_t res = 0.0;
-    for (int i = 0; i < va.Size(); ++i)
-      res += va[i] * vb[i];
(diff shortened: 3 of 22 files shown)


framework/doc/content/source/mfem/auxkernels/MFEMGradAux.md
```diff
@@ -5,7 +5,7 @@
 ## Overview
 
 AuxKernel for calculating the gradient of a scalar $H^1$ conforming source variable and storing it in
-a scalar elemental result variable defined on an $H(\mathrm{curl})$ conforming Nédélec finite element space.
+a result variable defined on an $H(\mathrm{curl})$ conforming Nédélec finite element space.
```
lindsayad, line 8: why do we map to a nedelec space instead of a vector L2?

framework/doc/content/source/mfem/bcs/MFEMRWTE10IntegratedBC.md
```diff
@@ -4,31 +4,31 @@
...
 
-Adds the frequency space integrator corresponding to the value upon a cross-sectional boundary of the TE10 mode of an electromagnetic wave traveling through an infinite rectangular waveguide. The physical system in question is explained in more detail [here](https://phys.libretexts.org/Bookshelves/Electricity_and_Magnetism/Electromagnetics_II_(Ellingson)/06%3A_Waveguides/6.09%3A_Rectangular_Waveguide-_TE_Modes). 
+Adds the frequency space integrator corresponding to the value upon a cross-sectional boundary of the TE10 mode of an electromagnetic wave traveling through an infinite rectangular waveguide. The physical system in question is explained in more detail [here](https://phys.libretexts.org/Bookshelves/Electricity_and_Magnetism/Electromagnetics_II_(Ellingson%29/06%3A_Waveguides/6.09%3A_Rectangular_Waveguide-_TE_Modes).
 
 If the parameter [!param](/BCs/MFEMRWTE10IntegratedBC/input_port) is set to `false`, corresponding to the output boundary of the waveguide, the following boundary integrator is applied:
 
 !equation
-iIm[k]\mu_0^{-1}(\vec u, \vec v)_{\partial\Omega} \,\,\, \forall v \in V
+iIm[k]\mu^{-1}(\vec u, \vec v)_{\partial\Omega} \,\,\, \forall \vec v \in V
```
lindsayad: What is $I$ and what do the brackets in the $[k]$ notation mean?

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
```diff
@@ -41,21 +41,14 @@ MFEMRWTE10IntegratedBC::MFEMRWTE10IntegratedBC(const InputParameters & parameter
...
     _omega(2 * M_PI * getParam<Real>("frequency")),
-    _a1_vec(to3DMFEMVector(getParam<RealVectorValue>("port_length_vector"))),
-    _a2_vec(to3DMFEMVector(getParam<RealVectorValue>("port_width_vector"))),
-    _a3_vec(CrossProduct(_a1_vec, _a2_vec)),
-    _a2xa3(CrossProduct(_a2_vec, _a3_vec)),
-    _a3xa1(CrossProduct(_a3_vec, _a1_vec)),
-    _v(mfem::InnerProduct(_a1_vec, _a2xa3)),
-    _kc(M_PI / _a1_vec.Norml2()),
+    _a1_vec(const_cast<mfem::real_t *>(&getParam<RealVectorValue>("port_length_vector")(0)), 3),
+    _a2_vec(const_cast<mfem::real_t *>(&getParam<RealVectorValue>("port_width_vector")(0)), 3),
```
lindsayad:
```suggestion
    _a1_vec(const_cast<mfem::real_t *>(&getParam<RealVectorValue>("port_length_vector")(0)), Moose::dim),
    _a2_vec(const_cast<mfem::real_t *>(&getParam<RealVectorValue>("port_width_vector")(0)), Moose::dim),
```

framework/include/mfem/bcs/MFEMRWTE10IntegratedBC.h
```diff
@@ -41,55 +41,27 @@ class MFEMRWTE10IntegratedBC : public MFEMComplexIntegratedBC
   }
 
   // Utilities
-  mfem::Vector CrossProduct(const mfem::Vector & va, const mfem::Vector & vb)
+  mfem::Vector NormalizedCrossProduct(const mfem::Vector & va, const mfem::Vector & vb)
```
lindsayad:
```suggestion
  mfem::Vector normalizedCrossProduct(const mfem::Vector & va, const mfem::Vector & vb)
```

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
```diff
@@ -41,21 +41,14 @@ MFEMRWTE10IntegratedBC::MFEMRWTE10IntegratedBC(const InputParameters & parameter
...
     _k0(_omega * sqrt(_mu * _epsilon)),
-    _k(std::complex<double>(0., sqrt(_k0 * _k0 - _kc * _kc))),
-    _k_a(_a2xa3),
-    _k_c(_a3_vec)
+    _kc(M_PI / _a1_vec.Norml2()),
+    _k(std::complex<mfem::real_t>(0., sqrt(_k0 * _k0 - _kc * _kc))),
+    _k_c(NormalizedCrossProduct(_a1_vec, _a2_vec) *= _k.imag()),
+    _k_a(NormalizedCrossProduct(_a2_vec, _k_c) *= _kc)
```
lindsayad:
```suggestion
    _k_c(normalizedCrossProduct(_a1_vec, _a2_vec) *= _k.imag()),
    _k_a(normalizedCrossProduct(_a2_vec, _k_c) *= _kc)
```

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
```diff
@@ -69,15 +62,15 @@ MFEMRWTE10IntegratedBC::MFEMRWTE10IntegratedBC(const InputParameters & parameter
...
 
 void
-MFEMRWTE10IntegratedBC::RWTE10(const mfem::Vector & x, std::vector<std::complex<double>> & E)
+MFEMRWTE10IntegratedBC::RWTE10(const mfem::Vector & x, std::vector<std::complex<mfem::real_t>> & E)
 {
 
-  mfem::Vector e_hat(CrossProduct(_k_c, _k_a));
-  e_hat *= 1.0 / e_hat.Norml2();
-  std::complex<double> zi(0., 1.);
+  mfem::Vector e_hat(NormalizedCrossProduct(_k_c, _k_a));
```
lindsayad:
```suggestion
  mfem::Vector e_hat(normalizedCrossProduct(_k_c, _k_a));
```

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
```diff
@@ -41,21 +41,14 @@ MFEMRWTE10IntegratedBC::MFEMRWTE10IntegratedBC(const InputParameters & parameter
...
-    _a1_vec(to3DMFEMVector(getParam<RealVectorValue>("port_length_vector"))),
-    _a2_vec(to3DMFEMVector(getParam<RealVectorValue>("port_width_vector"))),
-    _a3_vec(CrossProduct(_a1_vec, _a2_vec)),
-    _a2xa3(CrossProduct(_a2_vec, _a3_vec)),
-    _a3xa1(CrossProduct(_a3_vec, _a1_vec)),
-    _v(mfem::InnerProduct(_a1_vec, _a2xa3)),
-    _kc(M_PI / _a1_vec.Norml2()),
+    _a1_vec(const_cast<mfem::real_t *>(&getParam<RealVectorValue>("port_length_vector")(0)), 3),
+    _a2_vec(const_cast<mfem::real_t *>(&getParam<RealVectorValue>("port_width_vector")(0)), 3),
     _k0(_omega * sqrt(_mu * _epsilon)),
```
lindsayad: Do we have a `using namespace std` somewhere or are there `sqrt` overloads defined in namespace `mfem`?

modules/navier_stokes/doc/content/source/bcs/INSADMomentumNoBCBC.md
```diff
@@ -2,7 +2,7 @@
 
 The `INSADMomentumNoBCBC` class is used to implement the "No Boundary Condition"
 developed by
-[Griffiths](https://onlinelibrary.wiley.com/doi/abs/10.1002/(SICI)1097-0363(19970228)24:4%3C393::AID-FLD505%3E3.0.CO;2-O). According
+[Griffiths](https://onlinelibrary.wiley.com/doi/abs/10.1002/(SICI%291097-0363(19970228%2924:4%3C393::AID-FLD505%3E3.0.CO;2-O). According
```
lindsayad, line 5: lol how do you find this stuff?

framework/doc/content/source/mfem/auxkernels/MFEMGradAux.md
nmnobre: Because that's where the gradients [naturally belong](https://defelement.org/families/index.html).

For a motivating example, picture a 2 x 2 quad mesh, and a first-order Lagrange space, that's 9 DoFs. Let's assume we have a function $f(x,y) = xy$ defined over that domain, its gradient is $\nabla f = (y,x)$. To represent that you need a first-order, vector discontinuous Lagrange space with 32 DoFs. That same gradient can be represented in a first-order Nédélec space with 12 DoFs - and it actually properly encodes the tangential continuity of the gradient by virtue of the shared DoFs.

framework/doc/content/source/mfem/bcs/MFEMRWTE10IntegratedBC.md
nmnobre: $\mathrm{Im}[k]$ denotes the imaginary part of complex number $k$. Should be in roman type, I'll make that change.

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
nmnobre: Neither, they are just available in the global namespace:

"In the C++ standard library, however, the declarations (except for names which are defined as macros in C) are within namespace scope (3.3.6) of the namespace std. It is unspecified whether these names are first declared within the global namespace scope and are then injected into namespace std by explicit using-declarations (7.3.3)."

I've changed our implementation to use `std::sqrt()`.

modules/navier_stokes/doc/content/source/bcs/INSADMomentumNoBCBC.md
nmnobre: We have one such case in the docs for the BC I'm editing and I couldn't figure out how to fix it. So then I regexed the entire repo looking for similar cases and found a few. Most broken (which I've now fixed), except for one (which I've used as inspiration):

https://github.com/idaholab/moose/blob/fb28c060e20cad83d698401c3d0f7b575de35980/modules/doc/content/getting_started/examples_and_tutorials/tutorial01_app_development/step07_parallel.md?plain=1#L11

There are other cases in the docs where people have used the ASCII encoding for both "(" [%28] and ")" [%29].

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
nmnobre: Done. Though this is just a longer way of writing "3". xD

framework/include/mfem/bcs/MFEMRWTE10IntegratedBC.h
nmnobre: Done. Good catch, thank you.

framework/doc/content/source/mfem/auxkernels/MFEMGradAux.md
lindsayad: What do you do in 1D?

framework/doc/content/source/mfem/bcs/MFEMRWTE10IntegratedBC.md
lindsayad: Duh. 🤦 That's embarrassing for me haha

lindsayad: I think the Roman will help there thanks

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
lindsayad: In theory someone could configure with a different `LIBMESH_DIM`

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
lindsayad: Glad you changed. Because this would have failed if `mfem::real_t` was not `double`

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
nmnobre: In theory :P

https://github.com/idaholab/moose/blob/fb28c060e20cad83d698401c3d0f7b575de35980/framework/include/base/Moose.h#L152-L155

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
nmnobre: How come? Does this have anything to do with the MetaPhysicL 2 changes I've seen you do? I still don't quite understand the changes from qualified names such as `std::sqrt()` to `using std::sqrt`?

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
lindsayad: If we removed that static assertion, I'm assuming you'd like your code to still work, slash, I'd like to minimize the number of things I'd have to fix in the MOOSE herd

nmnobre: Of course, I agree, I did make the change :)

framework/doc/content/source/mfem/auxkernels/MFEMGradAux.md
nmnobre: The Nédélec space degenerates into a discontinuous Lagrange space.

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
lindsayad: Your code might compile, but clang at least would will give you a warning for some cmath functions. You could try this test program
```
#include <cmath>

int
main()
{
  long double foo = 5;
  fabs(foo);
  std::fabs(foo);
}
```
Clang at least will give you the warning
```
calculate-this.cpp:7:3: warning: absolute value function 'fabs' given an argument of type 'long double' but has parameter of type 'double' which may cause truncation of value [-Wabsolute-value]
    7 |   fabs(foo);
      |   ^
calculate-this.cpp:7:3: note: use function 'std::abs' instead
    7 |   fabs(foo);
```
This is because without `std::` qualification or pulling `std` functions into overload consideration via `using` you are relying on the global c functions. And c functions do not support overloading. So if you were to use c functions in my example you'd have to use `fabsf` for floats and `fabsl` for long doubles in order to prevent implicit conversions, which would narrow in the case of long double.

lindsayad: > "In the C++ standard library, however, the declarations (except for names which are defined as macros in C) are within namespace scope (3.3.6) of the namespace std. It is unspecified whether these names are first declared within the global namespace scope and are then injected into namespace std by explicit using-declarations (7.3.3)."

How I read this passage is that `std` math functions are guaranteed to be declared in namespace `std` and that it's unspecified whether those declarations are introduced by `using` functions from the global namespace

framework/doc/content/source/mfem/auxkernels/MFEMGradAux.md
lindsayad: I understand now that while obviously you do not do a good job of representing the finite element solution gradient $\nabla u_h$ by projecting into the Nedelec space, you do a *better* job of representing the true solution gradient by projecting into the Nedelec space 👍 

> What do you do in 1D?

Oh I see that MFEM supports Nedelec on segments. Nice

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
lindsayad: (And the only functions you could pull into namespace `std` from the global namespace would be the c functions that take double arguments)

lindsayad (approved):

framework/src/mfem/bcs/MFEMRWTE10IntegratedBC.C
nmnobre: Yes, and I agree with all this; but `mfem::real_t` can only be `float` or `double` so the C versions would work fine. (For now, and if that were to change, we of course would like our code to be robust, thus the change anyway)

nmnobre:
Failure unrelated. See https://github.com/idaholab/virtual_test_bed/pull/727.

PR #32103, opened 2025-12-16
Add quadrature option to ContactAction

lindsayad:
Replaces https://github.com/idaholab/moose/pull/30198 to get automated testing

Closes #32095


lindsayad:
@bwspenc could you review this when you have a chance?

framework/src/constraints/MortarConstraintBase.C
```diff
@@ -80,7 +80,8 @@ MortarConstraintBase::validParams()
       "For 3D mortar, QUAD meshes are integrated using triangle mortar segments. "
       "While DEFAULT quadrature order is typically sufficiently accurate, exact integration of "
       "QUAD mortar faces requires SECOND order quadrature for FIRST variables and FOURTH order "
-      "quadrature for SECOND order variables.");
+      "quadrature for SECOND order variables. Note that the actual order supplied to the libMesh "
+      "Quadrature class will be two times this value plus one.");
```
bwspenc: Why is there a difference between this value and the one passed to libMesh?
Is this referring to the relationship between the order of the polynomial exactly integrated (p) and the number of quadrature points in one direction (n):
p=2n-1
The libmesh quadrature orders are specified in terms of p. Is this specifying n instead? If so, we should explain that, and also, shouldn't it be minus one rather than plus one?

bwspenc: If what I said above is correct, it seems like the name of this parameter should be somethin that's different from `quadrature` to avoid confusion. Maybe `quadrature_points`? I realize that might be unclear for multidimensional cases, though.

lindsayad: In @roystgnr's words

> We default to 2n+1, based on the idea that we want to capture "a little extra" when a mass matrix (normally 2n) has a nonlinear or a spatially-dependent constitutive coefficient.

So the enumerator values are indeed all about the polynomial basis order of the variable. It's just that in a mass matrix bilinear form you'll be integrating a polynomial of at least degree 2x the enumerator value and then libMesh tosses that extra 1 on top for the reasons Roy mentions. Additionally, also in Roy's words ([from slack](https://moosedevelopers.slack.com/archives/C010M4X8ZK2/p1741631282369729))

> For tensor product elements a Gaussian rule for 2n also gives you 2n+1 for free, so it was a relatively cheap thing to request by default.

Does that make sense?

lindsayad: If it does make sense, perhaps I should add some of that explanation to the doc string? Or perhaps it would be better for me to remove that last sentence from the doc string entirely and perhaps instead add it as an inline comment? I think it's good to document for a developer but might just add confusion for an input file user

roystgnr: The comment as it's phrased here is confusing to me.  "requires SECOND order quadrature for FIRST variables" makes me think that if I'm using p=1 for variables then I could replicate the default here by specifying p=2 for quadrature, but then "Note that the actual order supplied to the libMesh Quadrature class will be two times this value plus one." makes me think that if I specify p=2 then libMesh will be told to use p=5.

lindsayad: Ok what do you guys think of the new (long) parameter name with revised doc string?

lindsayad: Ping @roystgnr @bwspenc on the name

roystgnr: Both the name and the new doc string look reasonable to me.  Explaining the bilinear->quadratic issue really helps!

lindsayad:
This should be ready for more review/merge

framework/src/constraints/MortarConstraintBase.C
```diff
@@ -74,13 +74,17 @@ MortarConstraintBase::validParams()
   params.addParam<bool>(
       "compute_lm_residuals", true, "Whether to compute Lagrange Multiplier residuals");
   params.addParam<MooseEnum>(
-      "quadrature",
+      "polynomial_basis_order_for_integration",
```
bwspenc: Per our discussion, let's deprecate `quadrature` and add a new parameter called `segment_quadrature` that specifies actual quadrature order on the face, and support up to NINTH order. Have `segment_quadrature` also exposed in the action.

Doc string: Say something about what DEFAULT order is (the base elements).

bwspenc (changes requested):

bwspenc (approved):

PR #32104, opened 2025-12-16
Add new AuxKernels and AuxScalarKernels for calculation of Henry and mass transfer coefficient values

bobsalko:
## Reason

The mass transfer modeling work that supports modeling of molten salt reactors and fission gas transfer requires calculation of Henry coefficients and mass transfer coefficients.  Models have been developed for that calculations as part of MSR modeling efforts in NEAMS.  This MR will merge these new capabilities into MOOSE.

## Design

A set of AuxKernels and AuxScalarKernels have been added for the Henry coefficient calculation as well as the mass transfer coefficient calculation.  The Henry coefficient kernels include models for both FLIBE and FLINAK salts as well as an option for the user to enter a custom model.  The mass transfer coefficient model includes options using the Wilke-Chang and Stokes-Einstein models.  Tests have been added for all kernels to ensure the values are being correctly calculated.

## Impact

This will add new AuxKernels and AuxScalarKernels that other MOOSE applications can utilize for modeling of mass transfer.


bobsalko:
added issue to track: #32106

GiudGiud:
@bobsalko who are you working with on this? To see if there is a preferred reviewer

wgurecky:
> @bobsalko who are you working with on this? To see if there is a preferred reviewer

This MR is to satisfy a milestone for a WP in SM&C. @tanoret

wgurecky:
One question: should these aux kernels live in the chemical_reactions or scalar_transport module?

GiudGiud:
> One question: should these aux kernels live in the chemical_reactions or scalar_transport module?

I'm thinking chemical reactions since that's where the rest of the MSR chemistry work is.
Henry coefficients have been in the fluid properties historically in MOOSE. Would it be feasible to put them there? Which fluid properties class do you guys use in SAM/Pronghorn for the salt?

bobsalko:
SAM has internal properties for salt.  MOOSE also provides some salt fluid properties.  Then we have the Saline interface to MSTDB-TP.  So there are three different locations to consider.  The Saline interface would be the default one, I would say.  But that would pretty much require putting this into MSTDB-TP instead and having Saline expose it.

GiudGiud:
That's a good point. It would need to be implemented in the MoltenSaltSalineFluidProperties most likely, that's the most general. 

Also henry's constant depends on the gas as well not just the fluid? So you would need more than one routine per fluid property?

bobsalko:
Yes, we could implement it into the `SalineMoltenSaltFluidProperties` class instead.  We should probably loop SM&C in on this to see what they'd prefer.  Either way, we need a wrapper in `SalineMoltenSaltFluidProperties`.  We could put these AuxKernels in the wrapper directly for now.  Maybe SM&C would want to have a longer term plan to put it into Saline proper.  

I'm no expert on Henry's Law.  I think this model is applicable to noble gases.  The paper shows results for helium, argon, xenon, and krypton, specifically.  Maybe we add a specifier to the procedure name to clarify this (i.e., henryConstantNobleGas).

GiudGiud:
Might not be for all noble gases at once, seeing one coefficient per noble gas on this page for water
https://en.wikipedia.org/wiki/Henry%27s_law

SM&C as in INL's SMC facility?

bobsalko:
By "all at once" you mean a mixture of multiple noble gases?  This is the paper the model came from: https://www.nature.com/articles/s41598-024-60006-9.  By SM&C, I mean the structural materials and chemistry technical area in NEAMS.  They are driving this current effort.

GiudGiud:
No I meant individually, it seems each noble gas has a different Henry's constant for equilibrium with water

bobsalko:
Each gas ends up with a different Henry coefficient using this model also.  It is a function of the van der Waals radius, which is provided by user, and this leads to different coefficients for different gases.

GiudGiud:
ok so if we went through the FluidProperties classes we would need something like henryCoefficient(p, T, gas type), which is not an API which exists currently, so it does not seem like the way forward. If there are additional requirements from SM&C let's have them pitch in and see if we keep the current design.

tanoret:
Nice contribution!

I would say that the transfer kernel `GasLiquidMassTransferAux` actually belongs to scalar_transport or navier_stokes, as @wgurecky previously mentioned. The mass transfer by partial pressures is a physical process, without chemical reactions, i.e. electrons transferred involved. So, chemical_reactions could be a bit confusing.

As for `HenryGasConstantAux`, I agree with you guys that `FluidProperties` may be a better choice.

What do you think?

modules/scalar_transport/include/userobjects/GasLiquidMassTransfer.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
```
GiudGiud, line 10: missing copyright

GiudGiud:
```suggestion
//* This file is part of the MOOSE framework
//* https://mooseframework.inl.gov
//*
//* All rights reserved, see COPYRIGHT for full restrictions
//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
//*
//* Licensed under LGPL 2.1, please see LICENSE for details
//* https://www.gnu.org/licenses/lgpl-2.1.html

#pragma once
```

modules/chemical_reactions/include/auxkernels/GasLiquidMassTransferAux.h
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include "AuxKernel.h"
+#include "Function.h"
+
+// Forward Declarations
+class SinglePhaseFluidProperties;
+
+class GasLiquidMassTransferAux : public AuxKernel
```
GiudGiud:
```suggestion
/**
  * Computes the ...
  */
class GasLiquidMassTransferAux : public AuxKernel
```

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
```diff
@@ -0,0 +1,56 @@
...
+  static constexpr Real _Rgas = 8.31446261815324;
+
+  /// Model constants obtained from
+  /// K. Lee, et al., "Semi-empirical model for Henry's law constant of noble gases in molten salt",
+  /// Scientific Reports (2024) 14:12847, https://doi.org/10.1038/s41598-024-60006-9.
+
+  /// Model constants for FLiBe
+  static constexpr Real _alpha_FLiBe = -3.3584;
+  static constexpr Real _beta_FLiBe = -0.0215;
+  static constexpr Real _KH0_FLiBe = 7.8622e-1 / _atm_to_Pa;
```
GiudGiud, line 46: paper says 7.8622e-7 ? this would be e-6 ?

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
```diff
@@ -0,0 +1,56 @@
...
+  static constexpr Real _gamma_0_FLiBe = 235.5;
+  static constexpr Real _dgamma_dT_FLiBe = -0.09;
+
+  /// Model constants for FLiNaK
+  static constexpr Real _alpha_FLiNaK = -4.6541;
+  static constexpr Real _beta_FLiNaK = 0.0105;
+  static constexpr Real _KH0_FLiNaK = 1.4246 / _atm_to_Pa;
```
GiudGiud, line 53: same

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
```diff
@@ -0,0 +1,99 @@
+#include "GasLiquidMassTransferAux.h"
```
GiudGiud: same comment for copyright in source

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
```diff
@@ -0,0 +1,99 @@
...
+#include "MathUtils.h"
+#include "SinglePhaseFluidProperties.h"
+
+registerMooseObject("MooseApp", GasLiquidMassTransferAux);
```
GiudGiud: register this to the ScalarTransportApp if in scalar_transport module, FLuidPropertiesApp if in Fluid_properties etc
otherwise doco check will fail

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
```diff
@@ -0,0 +1,99 @@
...
+
+InputParameters
+GasLiquidMassTransferAux::validParams()
+{
+  InputParameters params = AuxKernel::validParams();
+  // Required params
+  params.addRequiredCoupledVar("p", "Pressure [Pa]");
+  params.addRequiredCoupledVar("T", "Temperature [K]");
+  params.addRequiredCoupledVar("fluid_velocity", "Velocity vector [m/s]");
```
GiudGiud: Says vector here but it is being stored in a single-component variable value ?

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
```diff
@@ -0,0 +1,99 @@
...
+    _phi(getParam<Real>("phi")),
+    _wc(getParam<Real>("wc")),
+    _db(getParam<Real>("db"))
+{
+  switch (_equation_list)
+  {
+    case Equationlist::WILKECHANG:
+    {
+      if (!isParamSetByUser("molar_weight"))
+        mooseError("Must set the molecular weight of the gas when using WilkeChang");
```
GiudGiud:
```suggestion
        paramError("molar_weight", "Must set the molecular weight of the gas when using WilkeChang");
```
same just below please

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
```diff
@@ -0,0 +1,99 @@
...
+      Diffusivity = _kB * _temperature[_qp] / (6.0 * libMesh::pi * mu * _radius);
+      break;
+
+    case Equationlist::WILKECHANG:
+      // Equation 5 of:
+      // C.R. Wilke, P. Chang, Correlation of diffusion coefficients in dilute solutions, AICHE J.,
+      // 1955, 1(2) 264-270 Units of Wilke Chang are g-cm-s
+      Real kg_to_g = 1000.0;
+      Real m_to_cm = 100.0;
+      Real cm_to_m = 1 / m_to_cm;
```
GiudGiud:
```suggestion
      Real cm_to_m = 1. / m_to_cm;
```

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
```diff
@@ -0,0 +1,99 @@
...
+      Real mu_cgs = mu * kg_to_g / m_to_cm; // g/cm/s = Poise
+      Real poise_to_centipoise = 100;
+      mu_cgs = mu_cgs * poise_to_centipoise;           // cP
+      Real molar_volume = _mw / rho * pow(m_to_cm, 3); // cm3/mol
```
GiudGiud:
```suggestion
      Real molar_volume = _mw / rho * Utility::pow<3>(m_to_cm); // cm3/mol
```
only for integer power

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
```diff
@@ -0,0 +1,99 @@
...
+      Real molar_weight = _mw * kg_to_g;               // g/mol
+      Diffusivity = _wc * _temperature[_qp] * std::sqrt(_phi * molar_weight) /
+                    (mu_cgs * std::pow(molar_volume, 0.6));
+      Diffusivity = Diffusivity * pow(cm_to_m, 2);
```
GiudGiud:
```suggestion
      Diffusivity = Diffusivity * Utility::pow<2>(cm_to_m);
```

modules/chemical_reactions/src/auxkernels/HenryGasConstantAux.C
```diff
@@ -0,0 +1,88 @@
+#include "HenryGasConstantAux.h"
```
GiudGiud: same comments

modules/chemical_reactions/src/auxkernels/HenryGasConstantAux.C
```diff
@@ -0,0 +1,88 @@
...
+#include "MathUtils.h"
+
+// This AuxKernel performs a calculation of the Henry coefficient for noble gases using the model
+// by K. Lee, et al., "Semi-empirical model for Henry's law constant of noble gases in molten salt",
+// Scientific Reports (2024) 14:12847, https://doi.org/10.1038/s41598-024-60006-9.
+
```
GiudGiud: let s move this to the header instead, can be the docstring

modules/chemical_reactions/src/auxkernels/HenryGasConstantAux.C
```diff
@@ -0,0 +1,88 @@
...
+HenryGasConstantAux::computeValue()
+{
+
+  const Real m_to_ang = 1e10;
+
+  // Equations result in units of mol/cm^3/atm, so convert to mol/m^3/atm
+  Real value =
+      _alpha * 4 * libMesh::pi * std::pow(_radius * m_to_ang, 2) *
+          (_gamma_0 + _dgamma_dT * (_temperature[_qp] - 273.15)) +
+      _beta * 4 / 3 * libMesh::pi * std::pow(_radius * m_to_ang, 3) * _Rgas * _temperature[_qp];
```
GiudGiud:
```suggestion
      _beta * 4. / 3. * libMesh::pi * std::pow(_radius * m_to_ang, 3) * _Rgas * _temperature[_qp];
```
careful with those. 4 / 3 in C++ is integer division

modules/chemical_reactions/src/auxkernels/HenryGasConstantAux.C
```diff
@@ -0,0 +1,88 @@
...
+
+Real
+HenryGasConstantAux::computeValue()
+{
+
+  const Real m_to_ang = 1e10;
+
+  // Equations result in units of mol/cm^3/atm, so convert to mol/m^3/atm
+  Real value =
+      _alpha * 4 * libMesh::pi * std::pow(_radius * m_to_ang, 2) *
```
GiudGiud:
```suggestion
      _alpha * 4 * libMesh::pi * Utility::pow<2>(_radius * m_to_ang) *
```

modules/chemical_reactions/src/auxscalarkernels/HenryGasConstantScalarAux.C
```diff
@@ -0,0 +1,88 @@
...
+      _beta = HenryGasConstantAux::_beta_FLiNaK;
+      _KH0 = HenryGasConstantAux::_KH0_FLiNaK;
+      _gamma_0 = HenryGasConstantAux::_gamma_0_FLiNaK;
+      _dgamma_dT = HenryGasConstantAux::_dgamma_dT_FLiNaK;
+      break;
+
+    case Saltlist::CUSTOM:
+      // Ensure that all parameters were set by user
+      if (!isParamSetByUser("alpha"))
+        mooseError("Must include alpha parameter when using custom salt Henry gas model");
```
GiudGiud:
```suggestion
        paramError("alpha", "Must include alpha parameter when using custom salt Henry gas model");
```

GiudGiud (changes requested):
which code is going to use this? An auxkernel does not allow the propagation of derivative terms, so it prevents getting a perfect Jacobian when the problem is solved in a fully coupled approach. Depending on the code you want to use this with, I might recommend material properties, functor properties, or auxkernels if (linearFV, no Jacobian there)

If the equations are coupled only 1 way (p,T,v influence transport, but transport does not influence them), we might be looking at a segregated multi-system solve in which case auxkernels would be fine too

bobsalko:
> which code is going to use this? An auxkernel does not allow the propagation of derivative terms, so it prevents getting a perfect Jacobian when the problem is solved in a fully coupled approach. Depending on the code you want to use this with, I might recommend material properties, functor properties, or auxkernels if (linearFV, no Jacobian there)

That is a good point about the derivative term.  If we put it into the EOS class (i.e., `SalineMoltenSaltFluidProperties`, and maybe also `FlibeFluidProperties` and `FlinakFluidProperties`), then the derivative terms can be retrieved.  I'm not sure about your comment about making it material properties or functor properties.  What would be the benefit there over EOS functions?

Also, I think the Henry coefficient makes sense for EOS potentially, but what about the mass transfer coefficient calculation?  Would that be more suitable for `AuxKernels`?

GiudGiud:
> I'm not sure about your comment about making it material properties or functor properties. What would be the benefit there over EOS functions?

The problem with EOS (FluidProperties) is that the routines are not very general. The calculation of the coefficients is only for a few fluids (not all the single phase ones) and is one model among others, so the parameter list would need to grow every time we add a new model.
I could see it done within the EOS, but with a slightly different design. Create a helper class that adds the parameters and the coefficient routines, and add that helper to the 3 fluids of interest. Then when using it, you will first cast the fluid properties to the helper class. 

On the other hand, a material is better self-contained. It's very close to the auxkernel design, it's just more efficient because it's computed on the fly rather than storing a whole field for each coefficient.

bobsalko:
Ok, that makes sense to me.  I suppose there is still a way for a user to get a printout of the Henry coefficient on the mesh if it were implemented as a material instead of an AuxKernel?

GiudGiud:
Yes you can output material properties as auxiliary variables using a single "outputs" parameter where you specify usually "exodus"

There are more parameters to restrict the output to a few properties

bobsalko:
@GiudGiud , switching to material properties seems like a reasonable approach to me.  @tanoret, would that be reasonable for Pronghorn integration in the future?

GiudGiud:
The linearFV discretization would prefer AuxKernels as you have now, or functor material properties (a special type of material, which is mostly used by FV simulations, but can be used by FE). It depends on the target application here

wgurecky:
@GiudGiud  and @tanoret Thanks for the feedback and discussion!  I'm in agreement that migrating to the Henry's coefficient callculation for high volatile, noble gases in molten salts over to the MOOSE Material system is a good strategy.  The Henry's coefficient is indeed depedant on temperature and, weakly, on pressure (through mu and rho).

Should the present auxkernels/GasLiquidMassTransferAux   likewise be migrated over to the Material system as well?  The mass transfer coefficients are a function of the local fluid velocity, and temperature, so I could also see how accounting jacobian contributions from these could help in some cases.  

I think SAM and potentially Pronghorn would be the main use case that I am aware of for these new routines.  I think SAM can easily shift over from near-identical kernels that it currently uses.

GiudGiud:
> I think SAM and potentially Pronghorn would be the main use case that I am aware of for these new routines. I think SAM can easily shift over from near-identical kernels that it currently uses.

Ok for SAM regular material properties, with additional properties for the derivatives, would work for sure. However, SAM has started to use multi-system segregated solves as well for passive scalar (for example neutron precursor) advection, in which case the derivatives are not needed, and auxkernels would work just fine imo. "passive" here means that they do not influence the flow conditions (as opposed to active, like a gaseous phase influencing density)
For Pgh, the auxkernels are better at this time for the linearFV discretization (in fact you could not use material properties with that discretization).

modules/chemical_reactions/src/userobjects/HenryGasConstant.C
```diff
@@ -0,0 +1,88 @@
...
+    _salt_list(getParam<MooseEnum>("salt").getEnum<Saltlist>()),
+    _alpha(isParamSetByUser("alpha") ? getParam<Real>("alpha") : 0.0),
+    _beta(isParamSetByUser("beta") ? getParam<Real>("beta") : 0.0),
+    _gamma_0(isParamSetByUser("gamma_0") ? getParam<Real>("gamma_0") : 0.0),
+    _dgamma_dT(isParamSetByUser("dgamma_dT") ? getParam<Real>("dgamma_dT") : 0.0),
+    _KH0(isParamSetByUser("KH0") ? getParam<Real>("KH0") : 0.0)
+{
+  switch (_salt_list)
+  {
+    case Saltlist::FLIBE:
```
bwspenc, line 46: At the MOOSE level it might be best to keep this more generic and not have options for specific systems. We can discuss this.

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
bobsalko: Units in the paper are mol/cm^3/atm.  We want to be in mol/m^3/atm, so I multiplied by 1e6... then divide by atm to Pa conversion factor to get to units of mol/m^3/Pa.

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
bobsalko: See above explanation.

modules/chemical_reactions/src/auxkernels/GasLiquidMassTransferAux.C
bobsalko: added

modules/chemical_reactions/src/auxkernels/HenryGasConstantAux.C
bobsalko: Moved into the header.  Note that HenryGasConstant is now a UserObject

modules/chemical_reactions/src/auxkernels/HenryGasConstantAux.C
bobsalko: notice was added in new header file

bobsalko:
The GasLiquidMassTransfer class was moved to ScalarTransport.  The HenryGasConstant stays in ChemicalReactions.  Both tools have been switched to UserObjects.  The regression tests have been removed in favor of unit tests on the new objects.

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
```diff
@@ -0,0 +1,68 @@
...
+//* https://github.com/idaholab/moose/blob/master/COPYRIGHT
+//*
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "GeneralUserObject.h"
+
+// This AuxKernel performs a calculation of the Henry coefficient for noble gases using the model
```
GiudGiud:
```suggestion
/**
 * This UserObject performs a calculation of the Henry coefficient for noble gases using the model
```

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
```diff
@@ -0,0 +1,68 @@
...
+// by K. Lee, et al., "Semi-empirical model for Henry's law constant of noble gases in molten salt",
+// Scientific Reports (2024) 14:12847, https://doi.org/10.1038/s41598-024-60006-9.
+
```
GiudGiud:
```suggestion
```

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
```diff
@@ -0,0 +1,68 @@
...
+//* Licensed under LGPL 2.1, please see LICENSE for details
+//* https://www.gnu.org/licenses/lgpl-2.1.html
+
+#pragma once
+
+#include "GeneralUserObject.h"
+
+// This AuxKernel performs a calculation of the Henry coefficient for noble gases using the model
+// by K. Lee, et al., "Semi-empirical model for Henry's law constant of noble gases in molten salt",
+// Scientific Reports (2024) 14:12847, https://doi.org/10.1038/s41598-024-60006-9.
```
GiudGiud:
```suggestion
* by K. Lee, et al., "Semi-empirical model for Henry's law constant of noble gases in molten salt",
* Scientific Reports (2024) 14:12847, https://doi.org/10.1038/s41598-024-60006-9.
*/
```

modules/chemical_reactions/src/base/ChemicalReactionsApp.C
```diff
@@ -65,6 +67,8 @@ ChemicalReactionsApp::registerApps()
 #endif
 
   registerApp(ChemicalReactionsApp);
+
+  FluidPropertiesApp::registerApps();
```
GiudGiud: do you still user fluid properties? if not let's undo this
same for Makefile changes

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+  static InputParameters validParams();
+  HenryGasConstant(const InputParameters & parameters);
+  virtual ~HenryGasConstant() {}
+
+  virtual void initialSetup() override;
+  virtual void initialize() override;
+  virtual void finalize() override;
+  virtual void execute() override;
+  virtual Real henry(Real temperature) const;
```
GiudGiud, line 31:
```suggestion
  /// Returns the henry constant at the specified temperature
  virtual Real henry(Real temperature) const;
```

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
```diff
@@ -0,0 +1,68 @@
...
+  static constexpr Real _atm_to_Pa = 101325;
+
+  /// Universal gas constant [J/mol/K]
+  static constexpr Real _Rgas = 8.31446261815324;
+
+  /// Model constants obtained from
+  /// K. Lee, et al., "Semi-empirical model for Henry's law constant of noble gases in molten salt",
+  /// Scientific Reports (2024) 14:12847, https://doi.org/10.1038/s41598-024-60006-9.
+
+  /// Model constants for FLiBe
```
GiudGiud, line 43: I think I'd prefer derived class. "FliBeHenryGasConstant" and "FLiNakHenriGasConstant" over a single class with an enum

GiudGiud: let's change that on the next PR though, if more models / salts are added

modules/chemical_reactions/src/userobjects/HenryGasConstant.C
GiudGiud: @bwspenc we could do this with derived classes at this point. Is that needed now?

modules/chemical_reactions/src/userobjects/HenryGasConstant.C
```diff
@@ -0,0 +1,111 @@
...
+}
+
+void
+HenryGasConstant::finalize()
+{
+}
+
+void
+HenryGasConstant::execute()
+{
```
GiudGiud: move these as {} in the header please

modules/chemical_reactions/unit/include/HenryGasConstantTest.h
```diff
@@ -0,0 +1,61 @@
...
+    }
+    // Set up object for helium and FLiNaK
+    {
+      InputParameters params = _factory.getValidParams("HenryGasConstant");
+      params.set<MooseEnum>("salt") = "FLINAK";
+      params.set<Real>("radius") = 1.4e-10;
+      _fe_problem->addUserObject("HenryGasConstant", "henry_helium_flinak", params);
+      _henry_helium_flinak = &_fe_problem->getUserObject<HenryGasConstant>("henry_helium_flinak");
+    }
+    // Set up object for argon and FLiNaK
```
GiudGiud, line 47: could you set it up the custom one instead? That way you can verify that custom matches fliank when the constants passed are the same

modules/chemical_reactions/unit/src/HenryGasConstantTest.C
```diff
@@ -0,0 +1,60 @@
...
+
+#include "gtest/gtest.h"
+
+#include "HenryGasConstantTest.h"
+
+#include <cmath>
+
+// Leave relative tolerance loose because the gold values were digitized from
+// a publication
+const double tol = 5.0e-2;
```
GiudGiud, line 18: ok this is a good start but we should be testing it accurately too.
I'd double the tests, by outputting the value with 
```
std::cout << std::setprecision(12) << _henry_helium_flibe->henry(750.0) << std::endl;
```
with 
```
  Real saved_value_tol = 1e-10;
  EXPECT_LT(std::abs(_henry_helium_flibe->henry(750.0) - saved_value), saved_value_tol);
```

modules/scalar_transport/src/userobjects/GasLiquidMassTransfer.C
```diff
@@ -0,0 +1,123 @@
...
+    case Equationlist::STOKESEINSTEIN:
+    {
+      if (!isParamSetByUser("radius"))
+        paramError("radius", "Must set particle radius when using StokesEinstein");
+      break;
+    }
+  }
+}
+
+void
```
GiudGiud: same let s move those to the header

modules/modules.mk
```diff
@@ -126,10 +126,15 @@ ifeq ($(XFEM),yes)
         SOLID_MECHANICS             := yes
 endif
 
+ifeq ($(CHEMICAL_REACTIONS),yes)
+        FLUID_PROPERTIES            := yes
+endif
```
GiudGiud: let's undo that now

modules/scalar_transport/unit/src/GasLiquidMassTransferTest.C
```diff
@@ -0,0 +1,51 @@
...
+
+#include "gtest/gtest.h"
+
+#include "GasLiquidMassTransferTest.h"
+
+#include <cmath>
+
+const double tol = 1.0e-6;
+
+const double p = 1e5;
```
GiudGiud:
```suggestion
const Real tol = libMesh::TOLERANCE;

const Real p = 1e5;
```
etc

that way all the tests scale with the floating point precision MOOSE is compiled with

GiudGiud (changes requested):

GiudGiud:
I rebased to get the precheck to pass. You'll want to pull down the online version locally to build on it

modules/chemical_reactions/src/base/ChemicalReactionsApp.C
bobsalko: No, we don't need FluidProperties anymore.  Fixed this and the makefile.

modules/chemical_reactions/src/userobjects/HenryGasConstant.C
bobsalko: Ok.  I assume you want all of these in the .h.

modules/chemical_reactions/unit/include/HenryGasConstantTest.h
bobsalko: Good idea.  I set up a custom one in addition to the flinak one and ensure they match.

modules/chemical_reactions/unit/src/HenryGasConstantTest.C
bobsalko: I don't understand, what is "saved_value" here?

modules/scalar_transport/src/userobjects/GasLiquidMassTransfer.C
bobsalko: Done

modules/modules.mk
bobsalko: Done

bobsalko (commented):
Most comments addressed.  One question on my end.

modules/chemical_reactions/unit/src/HenryGasConstantTest.C
GiudGiud: it's a value you output to console and save manually in the unit test file

modules/chemical_reactions/include/userobjects/HenryGasConstant.h
GiudGiud: nevermind, I changed my mind.

I prefer the enum since the model is shared between all three

modules/chemical_reactions/unit/src/HenryGasConstantTest.C
bobsalko: got it.  I expanded these tests.

GiudGiud:
https://mooseframework.inl.gov/help/development/VSCode.html#d3535156-89d0-44a4-a1fe-5df9ea3c0556
this page has tips for not doing the formatting manually btw

GiudGiud:
Nevermind this was just "std::cout" being banned from the code so that fails precheck too

GiudGiud (dismissed):

GiudGiud:
my contributions were minor (commit history, formatting / precheck)

GiudGiud:
@bobsalko looks like we are missing a GasLiquidMassTransfer markdown doco page? Did I accidentally delete it while editing the commits?

bobsalko:
I'm not sure if I added anything.  Is that something that is added directly to the repo?  I don't see anything for HenryGasConstant either.

GiudGiud:
yes in the modules / ... / doc/content/source/userobject folder

you're right it's missing there too.
There are examples in the folder. it s the user facing documentation

bobsalko:
Ok, I've added two md files.  Let me know if that is what you are looking for.

loganharbour:
@bobsalko does this close https://github.com/idaholab/moose/issues/32106? If so, add `Closes #32106` to the initial comment and it'll get auto closed once this is merged

GiudGiud (approved):

GiudGiud:
Thanks for the contribution!

PR #32108, opened 2025-12-16
Remove usage of -ksp_snes_ew petsc option

bwspenc:
This may have been a valid option at one time, but I keep seeing models that give errors because this is not recognized as a valid PETSc option. Replace all instances of its use with -snes_ksp_ew.

## Reason
I keep running into models that use the invalid `-ksp_snes_ew` PETSc option. Getting rid of all usage of this in MOOSE will help avoid this (and get rid of some warnings in our tests).

## Design
Switch usages of `-ksp_snes_ew` to `-snes_ksp_ew` in all MOOSE tests.

## Impact
This will only update the test suite. No impact on code behavior.

### DIFF
--- a/framework/src/utils/PetscSupport.C
+++ b/framework/src/utils/PetscSupport.C
@@ -924,7 +924,7 @@ MultiMooseEnum
 getCommonSNESFlags()
 {
   return MultiMooseEnum(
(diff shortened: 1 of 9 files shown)


lindsayad (approved):
thanks Ben!

PR #32111, opened 2025-12-16
Mesh preparedness fixes

roystgnr:
## Reason
Some of the additional error checking in https://github.com/libMesh/libmesh/pull/3759 has exposed issues in MOOSE code; those need to be fixed before we can be confident the libMesh PR is ready to merge.

## Design
1. In places where MOOSE is relying on cached data from an unprepared mesh, test `is_prepared()` first and prepare if necessary.
2. When MOOSE is doing mesh splitting, keep repartitioning disabled until *right before* splitting the mesh; otherwise when libMesh starts correctly noting that unpartitioned meshes are unprepared, earlier MOOSE prepare_for_use() calls can trigger partitioning before the split, when a split-detail-dependent partitioner like a manual GridPartitioner setting may be unable to handle it.

## Impact
In theory no impact.  When we merge the next libMesh submodule there may be other similar errors exposed downstream.

### DIFF
--- a/framework/src/actions/SplitMeshAction.C
+++ b/framework/src/actions/SplitMeshAction.C
@@ -32,7 +32,7 @@ SplitMeshAction::act()
 {
   auto mesh = _app.actionWarehouse().mesh();
 
-  // Keep the mesh unpartitioned until right before the split 
+  // Keep the mesh unpartitioned until right before the split
   mesh->getMesh().skip_partitioning(false);
 
   const std::string split_file_arg = _app.parameters().isParamSetByUser("split_file")
--- a/framework/src/actions/SplitMeshAction.C
+++ b/framework/src/actions/SplitMeshAction.C
@@ -32,6 +32,7 @@ SplitMeshAction::act()
 {
   auto mesh = _app.actionWarehouse().mesh();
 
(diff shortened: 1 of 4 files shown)


framework/src/actions/SplitMeshAction.C
```diff
@@ -31,6 +31,9 @@ void
 SplitMeshAction::act()
 {
   auto mesh = _app.actionWarehouse().mesh();
+
+  mesh->getMesh().skip_partitioning(false);
```
GiudGiud, line 36:
```suggestion
  // Keep the mesh unpartitioned until right before the split 
  mesh->getMesh().skip_partitioning(false);
```

GiudGiud (approved):

framework/src/actions/SplitMeshAction.C
```diff
@@ -31,6 +31,10 @@ void
 SplitMeshAction::act()
 {
   auto mesh = _app.actionWarehouse().mesh();
+
+  // Keep the mesh unpartitioned until right before the split
```
loganharbour:
```suggestion
  // Keep the mesh unpartitioned until right before the split
```

loganharbour (dismissed):

PR #32120, opened 2025-12-17
Add Gap Mesh Generator

miaoyinb:
closes #32119


miaoyinb:
I guess the failed test is not related.

framework/include/utils/MooseMeshUtils.h
```diff
@@ -470,4 +470,41 @@ void copyIntoMesh(MeshGenerator & mg,
...
+ * @param end_boundary The boundary name to assign to the end of the polyline (if
+ * not a loop)
+ * @param max_elem_size The maximum element size for the mesh
+ */
+void buildPolyLineMesh(MeshBase & mesh,
+                       const std::vector<Point> & points,
+                       const bool loop,
+                       const BoundaryName & start_boundary,
+                       const BoundaryName & end_boundary,
+                       const Real & max_elem_size);
```
GiudGiud:
```suggestion
                       const Real max_elem_size);
```

modules/reactor/doc/content/source/meshgenerators/GapMeshGenerator.md
```diff
@@ -0,0 +1,40 @@
+# GapMeshGenerator
+
+!syntax description /Mesh/GapMeshGenerator
+
+## Overview
+
+In the mesh generation for reactor geometries, it is often necessary to create gaps that are conformal to the external boundaries of components. The gap is meant to have a uniform thickness throughout the geometry. This `GapMeshGenerator` class provides functionality to generate a polyline mesh (similar to the meshes generated by [`PolyLineMeshGenerator`](/PolyLineMeshGenerator.md)) with a specified gap thickness based on an 2D input mesh residing in the XY plane. The input mesh must be a connected mesh with only one outer boundary manifold. The generated polyline mesh can then be used for further mesh generation processes, such as [`XYDelaunayGenerator`](/XYDelaunayGenerator.md).
```
GiudGiud:
```suggestion
In the mesh generation for reactor geometries, it is often necessary to create unmeshed gaps along the external boundaries of components. The gap is meant to have a uniform thickness all around the boundary. This `GapMeshGenerator` class provides functionality to generate a polyline mesh (similar to the meshes generated by [`PolyLineMeshGenerator`](/PolyLineMeshGenerator.md)) with a specified gap thickness based on an 2D input mesh residing in the XY plane. The input mesh must be a connected mesh with only one outer boundary manifold. The generated polyline mesh can then be used for further mesh generation processes, such as [`XYDelaunayGenerator`](/XYDelaunayGenerator.md).
```

modules/reactor/doc/content/source/meshgenerators/GapMeshGenerator.md
```diff
@@ -0,0 +1,40 @@
...
+By default, this mesh generator only captures the vertices of the input mesh that are on the external polygon boundary and generate only one `EDGE2` element for each side of the polygon. In that case, users need to rely on the refinement capabilities in [`XYDelaunayGenerator`](/XYDelaunayGenerator.md) to achieve the desired mesh density. Optionally, users can specify a maximum element size ([!param](/Mesh/GapMeshGenerator/max_elem_size)) to have this mesh generator automatically insert additional nodes along each side of the polygon.
+
+## Examples
+
+A simple example shown here is based on a set of patterned hexagons. As shown in [simple_gap], the generated polyline mesh without refinement only contains one `EDGE2` element for each side of the hexagons with a gap from the original hexagons.
+
+!media meshgenerators/simple_gap.png
+      style=display: block;margin-left:auto;margin-right:auto;width:40%;
+      id=simple_gap
+      caption=A simple non-refined polyline mesh that is generated based on a set of patterned hexagons with a gap.
```
GiudGiud:
```suggestion
      caption=A simple non-refined polyline mesh that is generated around set of patterned hexagons, distanced with a gap.
```

modules/reactor/doc/content/source/meshgenerators/GapMeshGenerator.md
```diff
@@ -0,0 +1,40 @@
...
+
+After specifying a maximum element size, additional nodes are inserted along each side of the hexagons to create smaller `EDGE2` elements, as shown in [refined_gap]. Finally, using the refined polyline mesh as input, a triangular mesh with a circular outer boundary is generated by [`XYDelaunayGenerator`](/XYDelaunayGenerator.md) as shown in [gap_mesh].
+
+!media meshgenerators/refined_gap.png
+      style=display: block;margin-left:auto;margin-right:auto;width:40%;
+      id=refined_gap
+      caption=A refined polyline mesh that is generated based on a set of patterned hexagons with a gap.
```
GiudGiud:
```suggestion
      caption=A refined polyline mesh that is generated around a set of patterned hexagons with a gap.
```

modules/reactor/include/meshgenerators/GapMeshGenerator.h
```diff
@@ -0,0 +1,55 @@
...
+
+#pragma once
+
+#include "PolygonMeshGeneratorBase.h"
+
+/**
+ * Generate a polyline mesh that is based on an input 2D-XY mesh. The 2D-XY mesh needs to be a
+ * connected mesh with only one outer boundary manifold. The polyline mesh generated along with the
+ * boundary of the input mesh form a gap with a specified thickness. The mesh can further be used by
+ * XYDelaunayGenerator to generate a triangulation mesh that takes the gap into account.
```
GiudGiud:
```suggestion
 * boundary of the input mesh form a gap with a specified thickness.
```

modules/reactor/include/meshgenerators/GapMeshGenerator.h
```diff
@@ -0,0 +1,55 @@
...
+  GapMeshGenerator(const InputParameters & parameters);
+
+  std::unique_ptr<MeshBase> generate() override;
+
+protected:
+  /// Input mesh defining the boundary
+  std::unique_ptr<MeshBase> & _input;
+
+  /// The thickness of the gap to be created
+  const Real & _thickness;
```
GiudGiud:
```suggestion
  const Real _thickness;
```
not controlled, does not need to be a reference

modules/reactor/include/meshgenerators/GapMeshGenerator.h
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Check if three points are collinear.
+   * @param p1 First point
+   * @param p2 Second point
+   * @param p3 Third point
+   * @return true if the three points are collinear, false otherwise
+   */
+  bool isPointsColinear(const Point & p1, const Point & p2, const Point & p3) const;
```
GiudGiud: this should go in a utils if it s not there already

modules/reactor/include/meshgenerators/GapMeshGenerator.h
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Check if the line segment p1-p2 intersects with line segment p3-p4
+   * @param p1 First point of first line segment
+   * @param p2 Second point of first line segment
+   * @param p3 First point of second line segment
+   * @param p4 Second point of second line segment
+   * @return true if the two line segments intersect, false otherwise
+   */
+  bool
+  fourPointOverlap(const Point & p1, const Point & p2, const Point & p3, const Point & p4) const;
```
GiudGiud:
```suggestion
  segmentsIntersect(const Point & p1, const Point & p2, const Point & p3, const Point & p4) const;
```
same could go in a utils

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+  params.addRequiredParam<Real>("thickness", "The thickness of the gap to be created.");
+
+  params.addParam<Real>("max_elem_size", "The maximum element size for the generated gap mesh.");
+
+  params.addClassDescription(
+      "Generate a polyline mesh that is based on an input 2D-XY mesh. The 2D-XY mesh needs to be a "
+      "connected mesh with only one outer boundary manifold. The polyline mesh generated along "
+      "with the boundary of the input mesh form a gap with a specified thickness. The mesh can "
+      "further be used by XYDelaunayGenerator to generate a triangulation mesh that takes the gap "
+      "into account.");
```
GiudGiud:
```suggestion
      "with the boundary of the input mesh form an unmeshed gap with a specified thickness.");
```
the xy delaunay comment is already in the md file

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+  poly2tri.set_refine_boundary_allowed(false);
+  poly2tri.set_verify_hole_boundaries(false);
+  poly2tri.desired_area() = 0;
+  poly2tri.minimum_angle() = 0; // Not yet supported
+  poly2tri.smooth_after_generating() = false;
+  poly2tri.triangulate();
+
+  // We need to serialize the mesh for next steps
+  libMesh::MeshSerializer serial(*ply_mesh_u);
+  // The mesh now only contain one side set that corresponds to the outer boundary with an ID of 0
```
GiudGiud:
```suggestion
  // The mesh now only contains one side set that corresponds to the outer boundary with an ID of 0
```

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+#include "libmesh/elem.h"
+#include "libmesh/enum_to_string.h"
+#include "libmesh/int_range.h"
+#include "libmesh/mesh_modification.h"
+#include "libmesh/mesh_serializer.h"
+#include "libmesh/mesh_triangle_holes.h"
+#include "libmesh/parsed_function.h"
+#include "libmesh/poly2tri_triangulator.h"
+#include "libmesh/unstructured_mesh.h"
+#include "DelimitedFileReader.h"
```
GiudGiud:
```suggestion
#include "MooseUtils.h"
#include "DelimitedFileReader.h"

#include "libmesh/elem.h"
#include "libmesh/enum_to_string.h"
#include "libmesh/int_range.h"
#include "libmesh/mesh_modification.h"
#include "libmesh/mesh_serializer.h"
#include "libmesh/mesh_triangle_holes.h"
#include "libmesh/parsed_function.h"
#include "libmesh/poly2tri_triangulator.h"
#include "libmesh/unstructured_mesh.h"
```

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+
```
GiudGiud, line 19:
```suggestion
```

modules/reactor/include/meshgenerators/GapMeshGenerator.h
GiudGiud: there are more similar routines in PolygonMeshGeneratorBase that could go too

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+      dynamic_pointer_cast<UnstructuredMesh>(std::move(_input));
+
+  // MeshedHole is a good tool to extract and sort boundary points
+  TriangulatorInterface::MeshedHole bdry_mh(*mesh);
+
+  // Reduce the point list to only contain vertices
+  std::vector<Point> reduced_pts_list;
+  for (const auto i : make_range(bdry_mh.n_points()))
+  {
+    if (!isPointsColinear(bdry_mh.point((i - 1 + bdry_mh.n_points()) % bdry_mh.n_points()),
```
GiudGiud: should we look at the 0 - 1 point when the boundary is not a loop?

GiudGiud: I guess it is assumed to always be a loop

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+                          bdry_mh.point(i),
+                          bdry_mh.point((i + 1) % bdry_mh.n_points())))
+      reduced_pts_list.push_back(bdry_mh.point(i));
+  }
+  // Here we need a method to generate the outward normals of each external side
+  auto ply_mesh = buildMeshBaseObject();
+
+  MooseMeshUtils::buildPolyLineMesh(
+      *ply_mesh, reduced_pts_list, true, "dummy", "dummy", std::vector<unsigned int>({1}));
```
GiudGiud:
```suggestion
      *ply_mesh, reduced_pts_list, /*loop*/true, "dummy", "dummy", std::vector<unsigned int>({1}));
```

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+    mooseAssert(std::count(reduced_pts_list.begin(), reduced_pts_list.end(), original_pt) == 1,
+                "The original point should be found exactly once in the reduced points list.");
+    mod_reduced_pts_list[std::distance(
+        reduced_pts_list.begin(),
+        std::find(reduced_pts_list.begin(), reduced_pts_list.end(), original_pt))] =
+        original_pt + move_dir * mov_dist;
+  }
+
+  // To ensure no overlapping, we need to check set of four points
+  // p1 and p2 should be the pair of points before and after shifting
```
GiudGiud:
```suggestion
  // p1 and p2 should be the pair of points before shifting the side
```
?

modules/reactor/src/meshgenerators/GapMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+    for (const auto & i_node_2 : make_range(mod_reduced_pts_list.size()))
+    {
+      if (i_node_2 == i_node_1 || (i_node_2 + 1) % mod_reduced_pts_list.size() == i_node_1)
+        continue;
+      const Point & p3 = mod_reduced_pts_list[i_node_2];
+      const Point & p4 = mod_reduced_pts_list[(i_node_2 + 1) % mod_reduced_pts_list.size()];
+      if (fourPointOverlap(p1, p2, p3, p4))
+        paramError("thickness",
+                   "The specified thickness creates overlapping in the gap mesh. Please reduce the "
+                   "thickness value.");
```
GiudGiud: let's give more details on what is happening. 
The thickness is so large that the mesh is tangled because the offset nodes are no longer in the same order when following the original boundary

modules/reactor/src/meshgenerators/GapLineMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  std::vector<Point> mod_reduced_pts_list(reduced_pts_list);
+  for (const auto & [node_id, normal_vecs] : node_normal_map)
+  {
+    mooseAssert(normal_vecs.size() == 2,
+                "Each vertex should be connected to exactly two sides in a polygon.");
+
+    const Point original_pt = *(ply_mesh_u->node_ptr(node_id));
+    const Point move_dir = (normal_vecs.front() + normal_vecs.back()).unit();
```
GiudGiud:
```suggestion
    // Form an average normal at the vertex from the two connected sides' normals
    const Point move_dir = (normal_vecs.front() + normal_vecs.back()).unit();
```

modules/reactor/src/meshgenerators/GapLineMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+    const Real mov_dist =
+        _thickness /
+        std::sqrt((1.0 + (normal_vecs.front() * normal_vecs.back()) /
+                             (normal_vecs.front().norm() * normal_vecs.back().norm())) /
+                  2.0);
```
GiudGiud, line 161: can you add a comment on the scaling factor here? not sure I understand it

modules/reactor/src/meshgenerators/GapLineMeshGenerator.C
```diff
@@ -0,0 +1,219 @@
...
+  poly2tri.triangulate();
+
+  // We need to serialize the mesh for next steps
+  libMesh::MeshSerializer serial(*ply_mesh_u);
+  // The mesh now only contain one side set that corresponds to the outer boundary with an ID of 0
+  auto bdry_list(ply_mesh_u->get_boundary_info().build_side_list());
+
+  // For each vertex, the shifting direction to form the gap is defined by the normal vectors of the
+  // two sides that contain the vertex
+  std::map<dof_id_type, std::vector<Point>> node_normal_map;
```
GiudGiud, line 118:
```suggestion
  // We gather the normals for each node on the boundary here, which are all vertices because of the pre-selection of nodes
  std::map<dof_id_type, std::vector<Point>> node_normal_map;
```

GiudGiud (changes requested):
GapMeshGenerator could be meshing the gap instead of circling it on one side.
Maybe GapLineMeshGenerator ? Or BoundaryOffsetLineGenerator ? WallOffsetLineGenerator?

Does this generator work to mesh a gap inside rather than outside? Like inside the clad rather than outside the fuel?

miaoyinb:
> GapMeshGenerator could be meshing the gap instead of circling it on one side. Maybe GapLineMeshGenerator ? Or BoundaryOffsetLineGenerator ? WallOffsetLineGenerator?
> 
> Does this generator work to mesh a gap inside rather than outside? Like inside the clad rather than outside the fuel?

`GapLineMeshGenerator` sound good.

Although the offset algorithm works for "inside" gap, the current approach to identify the external boundary would not work for an internal boundary. I think this can be developed later.

modules/reactor/src/meshgenerators/GapMeshGenerator.C
miaoyinb: Yes, it is alway a loop as its from the MeshedHole class.

miaoyinb:
> > GapMeshGenerator could be meshing the gap instead of circling it on one side. Maybe GapLineMeshGenerator ? Or BoundaryOffsetLineGenerator ? WallOffsetLineGenerator?
> > Does this generator work to mesh a gap inside rather than outside? Like inside the clad rather than outside the fuel?
> 
> `GapLineMeshGenerator` sound good.
> 
> Although the offset algorithm works for "inside" gap, the current approach to identify the external boundary would not work for an internal boundary. I think this can be developed later.

It seems MeshedHole may be able to do this using a given boundary id. Let me look into this.

framework/include/utils/MooseMeshUtils.h
```diff
@@ -470,4 +470,60 @@ void copyIntoMesh(MeshGenerator & mg,
...
+ * @param max_elem_size The maximum element size for the mesh
+ */
+void buildPolyLineMesh(MeshBase & mesh,
+                       const std::vector<Point> & points,
+                       const bool loop,
+                       const BoundaryName & start_boundary,
+                       const BoundaryName & end_boundary,
+                       const Real max_elem_size);
+
+/**
```
GiudGiud: I think the two below fit better under GeometryUtils

miaoyinb:
@GiudGiud ready for another round of review. Thanks.

framework/include/utils/GeometryUtils.h
```diff
@@ -167,4 +167,23 @@ bool pointOnEdge(const libMesh::Point & point,
  */
 std::vector<libMesh::Point> boxCorners(const libMesh::BoundingBox & box,
                                        const libMesh::Real factor);
+
+/**
+ * Check if three points are collinear.
```
GiudGiud:
```suggestion
 * Check if three points are colinear.
```
american english spelling

framework/include/utils/GeometryUtils.h
```diff
@@ -167,4 +167,23 @@ bool pointOnEdge(const libMesh::Point & point,
...
+ * @param p1 First point
+ * @param p2 Second point
+ * @param p3 Third point
+ * @return true if the three points are collinear, false otherwise
```
GiudGiud:
```suggestion
 * @return true if the three points are colinear, false otherwise
```

framework/src/utils/GeometryUtils.C
```diff
@@ -277,4 +277,45 @@ boxCorners(const libMesh::BoundingBox & box, const Real factor)
...
+  const Real b1 = p1(0) - p2(0);
+  const Real c1 = p2(0) * p1(1) - p1(0) * p2(1);
+
+  const Real a2 = p4(1) - p3(1);
+  const Real b2 = p3(0) - p4(0);
+  const Real c2 = p4(0) * p3(1) - p3(0) * p4(1);
+
+  const Real denom = a1 * b2 - a2 * b1;
+  // We should not worry about the parallel case here
+  // If there is an overlap issue, it will be captured by other line segments
```
GiudGiud:
```suggestion
```

framework/include/utils/GeometryUtils.h
```diff
@@ -167,4 +167,23 @@ bool pointOnEdge(const libMesh::Point & point,
...
+bool isPointsColinear(const Point & p1, const Point & p2, const Point & p3);
+
+/**
+ * Check if the line segment p1-p2 intersects with line segment p3-p4
+ * @param p1 First point of first line segment
+ * @param p2 Second point of first line segment
+ * @param p3 First point of second line segment
+ * @param p4 Second point of second line segment
+ * @return true if the two line segments intersect, false otherwise
+ */
```
GiudGiud, line 188:
```suggestion
 * Parallel overlapping segments are considered not intersecting.
 */
```

GiudGiud: or add some code to detect that

framework/src/utils/MooseMeshUtils.C
```diff
@@ -1019,4 +1019,112 @@ copyIntoMesh(MeshGenerator & mg,
...
+    const auto & num_edges_between_points = (nums_edges_between_points.size() == 1)
+                                                ? nums_edges_between_points[0]
+                                                : nums_edges_between_points[i];
+
+    Point p = points[i];
+    mesh.add_point(p,
+                   nums_edges_between_points.size() == 1
+                       ? (i * num_edges_between_points)
+                       : (std::accumulate(nums_edges_between_points.begin(),
+                                          nums_edges_between_points.begin() + i,
```
GiudGiud: this wont scale but no need to change

framework/src/utils/MooseMeshUtils.C
```diff
@@ -1019,4 +1019,112 @@ copyIntoMesh(MeshGenerator & mg,
...
+
+      for (auto j : make_range(1u, num_edges_between_points))
+      {
+        p += pvec;
+        mesh.add_point(p,
+                       (nums_edges_between_points.size() == 1
+                            ? (i * num_edges_between_points)
+                            : (std::accumulate(nums_edges_between_points.begin(),
+                                               nums_edges_between_points.begin() + i,
+                                               0))) +
```
GiudGiud: could store this quantity from previous computation right above

framework/src/utils/MooseMeshUtils.C
```diff
@@ -1019,4 +1019,112 @@ copyIntoMesh(MeshGenerator & mg,
...
+  }
+
+  if (!loop)
+  {
+    BoundaryInfo & bi = mesh.get_boundary_info();
+    std::vector<BoundaryName> bdy_names{start_boundary, end_boundary};
+    std::vector<boundary_id_type> ids = MooseMeshUtils::getBoundaryIDs(mesh, bdy_names, true);
+    bi.add_side(mesh.elem_ptr(0), 0, ids[0]);
+    bi.add_side(mesh.elem_ptr(n_elem - 1), 1, ids[1]);
+  }
```
GiudGiud, line 1096: else mooseAssert start_bdy and end_bdy are ""

modules/reactor/test/tests/meshgenerators/gap_line_mesh_generator/tests
```diff
@@ -0,0 +1,52 @@
...
+  []
+  [error]
+    requirement = "The system shall throw an error "
+    [too_thick]
+      type = RunException
+      input = simple_gap.i
+      cli_args = "--mesh-only err.e
+                  Mesh/gap/thickness=1.0"
+      expect_err = "mesh is tangled"
+      detail = "when the gap thickness exceeds the available space and leads to overlapping."
```
GiudGiud:
```suggestion
      detail = "when the gap thickness exceeds the available space and leads to overlapping elements."
```

modules/reactor/doc/content/source/meshgenerators/GapLineMeshGenerator.md
```diff
@@ -0,0 +1,42 @@
+# GapLineMeshGenerator
+
+!syntax description /Mesh/GapLineMeshGenerator
+
+## Overview
+
+In the mesh generation for reactor geometries, it is often necessary to create unmeshed gaps along the external boundaries of components. The gap is meant to have a uniform thickness all around the boundary. This `GapLineMeshGenerator` class provides functionality to generate a polyline mesh (similar to the meshes generated by [`PolyLineMeshGenerator`](/PolyLineMeshGenerator.md)) with a specified gap thickness based on an 2D input mesh residing in the XY plane. If no specific boundary IDs are provided ([!param](/Mesh/GapLineMeshGenerator/boundary_ids)), the mesh generator will extract the external boundary of the input mesh to create the gap. In that case, the input mesh must be a connected mesh with only one outer boundary manifold. Otherwise, only the specified boundary IDs will be used to create the gap. The generated polyline mesh can then be used for further mesh generation processes, such as [`XYDelaunayGenerator`](/XYDelaunayGenerator.md).
+
```
GiudGiud, line 8: I think there needs to be a caveat on the list of boundary ids

can you do non-contiguous groups of boundary IDs? Does it detect it when the boundaries are contiguous?

modules/reactor/doc/content/source/meshgenerators/GapLineMeshGenerator.md
```diff
@@ -0,0 +1,42 @@
...
+## Methods
+
+This mesh generator first utilizes the `libMesh` utility used by [`XYDelaunayGenerator`](/XYDelaunayGenerator.md) to extract the external boundary or the specified boundary of the input mesh ([!param](/Mesh/GapLineMeshGenerator/input)). The boundary vertices are then shifted outward along the averaged normal directions of the connected edges to form the designated gap thickness ([!param](/Mesh/GapLineMeshGenerator/thickness)). The shifted vertices are then connected to form a new polyline mesh representing the gap using the meshing utilities of [`PolyLineMeshGenerator`](/PolyLineMeshGenerator.md).
+
+The gap can be created either outward or inward with respect to the boundary of the input mesh by specifying the gap direction ([!param](/Mesh/GapLineMeshGenerator/gap_direction)).
```
GiudGiud:
```suggestion
The gap can be created either outward or inward with respect to the boundary of the input mesh by specifying the gap direction ([!param](/Mesh/GapLineMeshGenerator/gap_direction)). This direction is relative to the orientation of the sideset used.
```

modules/reactor/doc/content/source/meshgenerators/GapLineMeshGenerator.md
```diff
@@ -0,0 +1,42 @@
...
+
+By default, this mesh generator only captures the vertices of the input mesh that are on the polygon boundary and generate only one `EDGE2` element for each side of the polygon. In that case, users need to rely on the refinement capabilities in [`XYDelaunayGenerator`](/XYDelaunayGenerator.md) to achieve the desired mesh density. Optionally, users can specify a maximum element size ([!param](/Mesh/GapLineMeshGenerator/max_elem_size)) to have this mesh generator automatically insert additional nodes along each side of the polygon.
```
GiudGiud:
```suggestion
By default, this mesh generator only captures the vertices of the input mesh that are on the polygon boundary and generate only one `EDGE2` element for each side of the polygon. In that case, users need to rely on the refinement capabilities in [`XYDelaunayGenerator`](/XYDelaunayGenerator.md) to achieve the desired mesh density. Alternatively, users can specify a maximum element size ([!param](/Mesh/GapLineMeshGenerator/max_elem_size)) to have this mesh generator automatically insert additional nodes along each side of the polygon.
```

GiudGiud: while we can combine both, seems like this second method is easier and thus is an alternartive

modules/reactor/doc/content/source/meshgenerators/GapLineMeshGenerator.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Examples
+
+A simple example shown here is based on a set of patterned hexagons. As shown in [simple_gap], the generated polyline mesh without refinement only contains one `EDGE2` element for each side of the hexagons with a gap from the original hexagons.
```
GiudGiud:
```suggestion
A simple example shown here is based on a set of patterned hexagons. As shown in [simple_gap], the generated polyline mesh without refinement only contains one `EDGE2` element for each side of the hexagons with a gap from the original hexagons' sides.
```

modules/reactor/doc/content/source/meshgenerators/GapLineMeshGenerator.md
GiudGiud:
```suggestion
The gap can be created either outward or inward with respect to the boundary of the input mesh by specifying the gap direction ([!param](/Mesh/GapLineMeshGenerator/gap_direction)). This direction is defined with regards to the volume enclosed by the boundaries specified (or the entire input mesh if unspecified).
```

modules/reactor/src/meshgenerators/GapLineMeshGenerator.C
```diff
@@ -0,0 +1,197 @@
...
+  params.addRequiredParam<MeshGeneratorName>("input", "The input mesh to create the gap based on.");
+
+  params.addRequiredParam<Real>("thickness", "The thickness of the gap to be created.");
+
+  MooseEnum gap_direction("OUTWARD INWARD", "OUTWARD");
+
+  params.addParam<MooseEnum>(
+      "gap_direction",
+      gap_direction,
+      "In which direction the gap is created with respect to the boundary of the input mesh.");
```
GiudGiud:
```suggestion
      "In which direction the gap is created with respect to the side normal of the elements along the boundary of the input mesh.");
```

modules/reactor/src/meshgenerators/GapLineMeshGenerator.C
```diff
@@ -0,0 +1,197 @@
...
+
+  params.addParam<std::vector<boundary_id_type>>(
+      "boundary_ids",
+      std::vector<boundary_id_type>(),
+      "The boundary IDs around which the gap will be created.");
+
+  params.addParam<Real>("max_elem_size", "The maximum element size for the generated gap mesh.");
+
+  params.addClassDescription(
+      "Generate a polyline mesh that is based on an input 2D-XY mesh. The 2D-XY mesh needs to be a "
```
GiudGiud:
```suggestion
      "Generates a polyline mesh that is based on an input 2D-XY mesh. The 2D-XY mesh needs to be a "
```

modules/reactor/doc/content/source/meshgenerators/GapLineMeshGenerator.md
GiudGiud: nevermind you have it in the description of the class already.

framework/include/utils/GeometryUtils.h
```diff
@@ -167,4 +167,23 @@ bool pointOnEdge(const libMesh::Point & point,
...
                                        const libMesh::Real factor);
+
+/**
+ * Check if three points are collinear.
+ * @param p1 First point
+ * @param p2 Second point
+ * @param p3 Third point
+ * @return true if the three points are collinear, false otherwise
+ */
+bool isPointsColinear(const Point & p1, const Point & p2, const Point & p3);
```
GiudGiud:
```suggestion
bool arePointsColinear(const Point & p1, const Point & p2, const Point & p3);
```

framework/src/utils/GeometryUtils.C
```diff
@@ -277,4 +277,45 @@ boxCorners(const libMesh::BoundingBox & box, const Real factor)
   return verts;
 }
 
+bool
+isPointsColinear(const Point & p1, const Point & p2, const Point & p3)
```
GiudGiud:
```suggestion
arePointsColinear(const Point & p1, const Point & p2, const Point & p3)
```

modules/reactor/src/meshgenerators/GapLineMeshGenerator.C
```diff
@@ -0,0 +1,197 @@
...
+
+  std::set<std::size_t> mesh_bdry_ids(_boundary_ids.begin(), _boundary_ids.end());
+  // MeshedHole is a good tool to extract and sort boundary points
+  TriangulatorInterface::MeshedHole bdry_mh(*mesh, mesh_bdry_ids);
+
+  // Reduce the point list to only contain vertices
+  std::vector<Point> reduced_pts_list;
+  for (const auto i : make_range(bdry_mh.n_points()))
+  {
+    if (!geom_utils::isPointsColinear(
```
GiudGiud:
```suggestion
    if (!geom_utils::arePointsColinear(
```

GiudGiud (commented):
looks great!
some nitpicky comments. next is broken rn anyway

framework/include/utils/GeometryUtils.h
miaoyinb: Will leave it to the caller to decide what value to return for the parallel case.

miaoyinb: Actually, it's not too much work to detect that. Done.

GiudGiud:
dbg failure should be fixed now

GiudGiud (approved):

GiudGiud:
Test failure unrelated

GiudGiud:
Thanks for the PR!

loganharbour:
Reminder to add sweeps and valgrind to these significant mesh generator tests.

This broke devel: https://civet.inl.gov/event/274645/

GiudGiud:
Patch up in a few minutes

PR #32122, opened 2025-12-18
Add "not cuda" to Kokkos function testing

### LINKED ISSUE #32121: Kokkos function tests error when cuda backend is enabled
### Bug Description

All the kokkos function tests error

### Steps to Reproduce

Do `./run_tests -j28 --re "kokkos.*functions"` with a PETSc which has been built `--with-cuda`

### Impact

Erroring tests which could alarm a user

### [Optional] Diagnostics

_No response_

lindsayad:
Closes #32121

### DIFF
--- a/test/tests/kokkos/functions/constant_function/tests
+++ b/test/tests/kokkos/functions/constant_function/tests
@@ -6,7 +6,7 @@
     input = 'kokkos_constant_function.i'
     exodiff = 'kokkos_constant_function_out.e'
     requirement = 'The Kokkos function system shall include a constant function and the ability to retrieve it as an abstract type.'
-    capabilities = 'kokkos'
+    capabilities = 'kokkos & !cuda'
     compute_devices = 'cpu'
   []
 []
--- a/test/tests/kokkos/functions/default_function/tests
+++ b/test/tests/kokkos/functions/default_function/tests
@@ -6,7 +6,7 @@
     input = 'kokkos_default_function.i'
     exodiff = 'kokkos_default_function_out.e'
     requirement = 'The Kokkos function system shall include the ability to set default values for input parameters expecting a function name.'
-    capabilities = 'kokkos'
+    capabilities = 'kokkos & !cuda'
     compute_devices = 'cpu'
   []
 []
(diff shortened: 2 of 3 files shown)


NamjaeChoi (approved):

loganharbour (approved):

loganharbour:
Gonna hijack the draft and assume this is done with @NamjaeChoi's review

loganharbour:
@lindsayad @NamjaeChoi is this a configuration that we need to test?

NamjaeChoi:
@loganharbour It's a feature that's expected to work with CUDA when we can turn on relocatable device code configuration, which is not ready yet. So I'm only testing it on CPUs currently. 

The issue here is `compute_devices` doesn't know which backend Kokkos was actually configured with. It's merely a filter. So what @lindsayad encountered is when he configures Kokkos with CUDA and run CPU tests, those tests still run, but with CUDA.

It would be ideal if there is a way `capabilities` can be set differently depending on compute device type.

loganharbour:
> @loganharbour It's a feature that's expected to work with CUDA when we can turn on relocatable device code configuration, which is not ready yet. So I'm only testing it on CPUs currently. 
> 
> 
> 
> The issue here is `compute_devices` doesn't know which backend Kokkos was actually configured with. It's merely a filter. So what @lindsayad encountered is when he configures Kokkos with CUDA and run CPU tests, those tests still run, but with CUDA.
> 
> 
> 
> It would be ideal if there is a way `capabilities` can be set differently depending on compute device type.

We already dynamically modify capabilities before checking them in the test harness. We could easily query the app for something and assign it a capability string that we could check. What do you have in mind?

NamjaeChoi:
Something like `capabilities = if (compute_device = cpu, kokkos & !cuda)
                                                   else if (compute_device = cuda, kokkos & cuda)`

loganharbour:
> Something like `capabilities = if (compute_device = cpu, kokkos & !cuda)
> 
>                                                    else if (compute_device = cuda, kokkos & cuda)`

I don't know if we have the conditionals in place, but we can easily scrape compute_device from the harness args.

This would work right now (once we add in the device):

kokkos & ((compute_device=cpu & !cuda) | (compute_device=cuda & cuda))

NamjaeChoi:
That would be great. How about when someone specifies multiple `compute_devices` (not sure if it's allowed?) like `--compute-devices=cpu,cuda`?

NamjaeChoi:
Or I wonder if it's more desired to have test harness always force insert `cuda` or `!cuda` into `capabilities` depending on `compute_devices`.

Maybe not because Torch and MFEM still works on CPU even with CUDA enabled.

Issue #32116, opened 2025-12-17
Tidy mass scaling in solid mechanics


WilkAndy:
### Motivation

In explicit solid-mechanics simulations, to ensure that stiff/small/light regions of the mesh don't force the critical time-step to be inordinately small, inertial mass scaling is standard practice.  Tidy this existing functionality.

### Design

- This appears to be done by the `DensityScaling` Material.  It needs a few tweaks.  Eg:
  -  Perhaps it is suboptimally named, as it actually provdes a density **addition**, and does not actually scale the original density.  So, perhaps rename it to `DensityAddition`.
  - Or, change it so that it outputs an actual scaled density that can be used in explicit time stepping
- There appears to be no input file that illustrates how to use mass scaling in practice

### Impact

Hopefully better naming means it's easier for users to understand things

WilkAndy:
I believe this might already be done by the `DensityScaling` Material , https://mooseframework.inl.gov/source/materials/DensityScaling.html .   If that is true, then this issue will modify to:

- provide at least one more example input file
- perhaps tidy up code a bit, eg, "effective stiffness" is actually the square-root of effective stiffness.

WilkAndy:
I want to:

- rename `DensityScaling` to `DensityAddition`, as currently `DensityScaling` does not scale density.  Instead, it provides an addition to the density that can be summed to the original density if the user wants to construct a new density object.
- rename "effective stiffness"

WilkAndy:
Closed by #32123

PR #32123, opened 2025-12-18
DensityScaling Material modifications and associated

WilkAndy:
DensityScaling Material:

- algorithm cleaned up
- uses a material property with user-defined name instead of one hard-coded as "density"
- better .h and .C documentation, according to me at least!
- change _effective_stiffness to _sqrt_effective_stiffness, because that's what the parameter represents
- ability to output the actual scaled density, not just the additive contribution
- ability to name the computed quantity, so it's not always the hard-coded "density_scaling"

Added input file and example demonstrating how to use DensityScaling to achieve a given time-step size

DensityScaling.md modifications:
- more descriptive text, according to me at least!
- inclusion of example based on the input file

CriticalTimeStep Postprocessor:
- density param is a MaterialName, and not hard-coded as "density", but defaulting to "density" for backwards compatibility
- change_effective_stiffness to _sqrt_effective_stiffness, because that's what the parameter represents

Refs #32116


WilkAndy:
@bwspenc  - any idea why "VTB" recipe is failing?   I don't know what "VTB" is...

GiudGiud:
It's the virtual test bed. And don't worry about it, it's related to advanced reactor work and it's unrelated to this

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
```diff
@@ -4,17 +4,38 @@
...
 ## Description
 
-The material `DensityScaling` computes the extra element density required
-in every element to allow explicit dynamics solves to be solved with a time
-step prescribed by a user (see [CriticalTimeStep](/CriticalTimeStep.md)).
-The addition of mass can alter the dynamics of the system, so usage of mass
-scaling is particularly recommended when the finite element mesh contains a
-handful of small finite elements or when system fundamental frequencies do not
-have a key influence on the simulation output metrics.
+This Material computes the inertial density needed to enable stable explicit time-stepping in solid-mechanics problems (see [CriticalTimeStep](/CriticalTimeStep.md)).  It outputs:
```
bwspenc: A couple of comments on this:
1) It seems a little odd to refer to it as "inertial density" (which shows up in several places). Would it make sense to just call it "density"?
2) The first sentence should probably say something about the fact that the density is scaled to enable taking user-determined time steps that are larger than those determined by stability conditions.

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
```diff
@@ -4,17 +4,38 @@
...
 
-## Example Input File
+- when `additive_contribution = false`: the inertial density needed to enable time-stepping with given `desired_time_step`; 
+- when `additive_contribution = true`: the portion of intertial density that should be added to the true density to enable the desired time stepping, that is max(0, inertial density needed minus true inertial density).
```
bwspenc: Check spelling of "intertial", although as noted above I would prefer to remove that word.

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
```diff
@@ -4,17 +4,38 @@
...
+
+When used in the [MassMatrix](MassMatrix.md), time-steps of size `desired_time_step` will be stable.  However, note that the addition of mass will alter the dynamics of the system.  In particular, high-frequency oscillations will largely be eliminated in elements that are small and/or stiff and/or light.  Hence, using mass scaling is particularly recommended when the finite element mesh contains a handful of small/stiff/light elements, or when high-frequency dynamics are unimportant.
```
bwspenc: The last phrase about "or when high-frequency dynamics are unimportant" is maybe a bit misleading, because it can definitely affect low-frequency dynamics.

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
```diff
@@ -4,17 +4,38 @@
...
+
+## Example
+
+To use `DensityScaling` effectively, two steps are needed.  The following is a worked example.
+
+#### Step 1
+
+To ensure that a time step of $\Delta t = 4$ is stable, use a `DensityScaling` Material to compute the required density in each element:
```
bwspenc:
```suggestion
To ensure that a user-defined time step ($\Delta t = 4$ in this case) is stable, use a `DensityScaling` Material to compute the required density in each element:
```

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
```diff
@@ -4,17 +4,38 @@
...
+
+!listing modules/solid_mechanics/test/tests/dynamics/time_integration/mass_scaling.i start=[density_true] end=[Elasticity_tensor]
+
+The important features here are:
+
+- A Material Property called `density_true` is calculated by some Material (in this case, the `density_true` Material) and fed into the `DensityScaling` Material
+- The `desired_time_step` is defined in the `DensityScaling` Material
+- The `safety_factor` is set.  What happens is that the `DensityScaling` Material computes the density required for stable time-stepping, then increases it using 1/`safety_factor`.  This means that, in this case, $\Delta t = 4 / 0.8 = 5$ is theoretically stable, but for safety, the input file only uses $\Delta t = 4$.
```
bwspenc:
```suggestion
- The `DensityScaling` Material computes the density required for stable time-stepping, then increases it by multiplying it by 1/`safety_factor`.  For example, in this case, $\Delta t = 4 / 0.8 = 5$ is theoretically stable, but for safety, the input file only uses $\Delta t = 4$.
```

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
```diff
@@ -4,17 +4,38 @@
...
+- The `DensityScaling` Material stores the computed density in `density_scaled` (or whatever is specified by the `scaled_density` parameter).
+
+The other features, such as `implicit = false`, `output_properties`, `outputs`, are unimportant here (eg, the outputs are just so the result is stored in the exodus file).
```
bwspenc:
```suggestion
The other specified parameters, such as `implicit = false`, `output_properties`, `outputs`, are unimportant here (eg, the outputs are just so the result is stored in the exodus file).
```

bwspenc: Doesn't `implicit=false` affect the computation?

modules/solid_mechanics/include/materials/DensityScaling.h
```diff
@@ -13,7 +13,8 @@
 #include "RankTwoTensor.h"
 
 /**
- * This material computes the mass required to fulfill a prescribed critical time step.
+ * This material computes the density required to enable explicit time-stepping in
+ * solid-mechanics problems.
```
bwspenc:
```suggestion
 * This material computes a density that is scaled to enable stable explicit time stepping 
 * with steps larger than those governed by stability conditions in solid-mechanics problems.
```

modules/solid_mechanics/src/materials/DensityScaling.C
```diff
@@ -17,49 +17,54 @@ InputParameters
...
-      "Factor to multiply to the critical time step. This factor is typically less than one to be "
-      "on the conservative side due to two types of approximation: Time step lagging and the "
-      "approximation of the critical time step formula.");
+      "true_density",
+      "Name of Material Property defining the true inertial density of the material.");
+  params.addRequiredParam<MaterialPropertyName>(
+      "scaled_density", "Name of the Property that this Material will compute.");
+  params.addRequiredParam<Real>("desired_time_step", "The desired time step.");
+  params.addParam<bool>(
+      "additive_contribution",
```
bwspenc: This is a bit confusing to me. It seems like the idea of this is that you wouldn't actually use the computed scaled density if `additive_contribution=true`, is that right? Like it's mostly for diagnostic purposes? Maybe it's best to just compute two separate properties all the time, so that the property that is named `scaled_density` is always what its name claims it is.

modules/solid_mechanics/src/materials/DensityScaling.C
```diff
@@ -17,49 +17,54 @@ InputParameters
...
+      false,
+      "When false, this Material computes the inertial density needed to enable time-stepping with "
+      "given desired_time_step.  When true, the Material computes max(inertial_density_needed - "
+      "true_density, 0), ie, just the portion that needs to be added to the true density to enable "
+      "the desired time stepping.");
+  params.addRangeCheckedParam<Real>(
+      "safety_factor",
```
bwspenc, line 37: I could see some confusion from the name of this parameter about whether this is a factor that the scaled density is multiplied or divided by. Maybe it would be better to call it something like `density_scaling_factor`, and then multiply by that, just to make it more clear from its name what it does. That's still not perfect either -- I'll think about this and see if I come up with something better.

modules/solid_mechanics/test/tests/dynamics/time_integration/tests
```diff
@@ -110,4 +110,14 @@
     design = "ExplicitMixedOrder.md"
     issues = "#27833"
   []
+  [mass_scaling]
+    type = 'Exodiff'
+    input = 'mass_scaling.i'
+    exodiff = 'mass_scaling_out.e'
+    requirement = "The solid mechanics app shall support mass scaling to support large time steps when
```
bwspenc:
```suggestion
    requirement = "The solid mechanics app shall support mass scaling to enable taking time steps larger than those permitted by the stability condition when
```

modules/solid_mechanics/src/materials/DensityScaling.C
bwspenc: Actually, I think I'm liking `safety_factor` better the more I think about it because it makes it a little more clear that it's being applied in the computation of the needed scaled density, but only affects the scaled density if it is greater than the true density.

bwspenc (changes requested):

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
WilkAndy: Thanks for all your suggestions, @bwspenc  - good to have fresh eyes on this.  I'll probably implement all of your suggestions.  Regarding the "inertial" density - i just wanted to distinguish this current thing from the "gravitational" density that you probably don't want to change.  As it surprised you, i'll be clearer.

modules/solid_mechanics/src/materials/DensityScaling.C
WilkAndy: You are correct about the `additive_contribution=true` confusion: with `additive_contribution=true` this Material would compute something that seemingly wouldn't be particularly useful.  The reason i have this `additive_contribution` parameter is that the original code always computed the `additive_contribution=true` version, and i'm not sure why, but it must have been useful for someone.  I've made it into an "advanced parameter".

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
WilkAndy: Removed "intertial" (and its misspelling) everywhere, and added a sentence "Note that the density computed acts as your model's *inertial* density (see example below) and you should not use it as your *gravitational* density."

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
WilkAndy: Removed, "inertial" and misspellings, as mentioned above.

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
WilkAndy: Added this sentence: "Mass scaling has a smaller, yet noticeable, impact on low-frequency dynamics"

modules/solid_mechanics/doc/content/source/materials/DensityScaling.md
WilkAndy: Changed the first sentence to read "This Material computes the density required to achieve stable explicit time stepping with a user-defined time step (see [CriticalTimeStep](/CriticalTimeStep.md) for the stability condition)"

WilkAndy:
Hi @bwspenc - thank you for all your comments.  I think i've addressed them all.

modules/solid_mechanics/src/materials/DensityScaling.C
bwspenc: I think it is primarily useful for quantifying how much mass this has ended up adding to the system (which would need to be computed by integrating that added density over the volume). If I remember correctly, that's a pretty standard output from explicit codes that do mass scaling.

What do you think about storing that in a separate material property, and maybe having a flag that controls whether that is computed? That would be my preferred way of handling that.

modules/solid_mechanics/src/materials/DensityScaling.C
bwspenc: Let's just leave this as `safety_factor`

modules/solid_mechanics/src/materials/DensityScaling.C
bwspenc: I just talked with @dschwen about this and he and I agree that adding another material property for the additional density is the way to go. I would suggest naming it `additional_density`. There's very little overhead involved in computing that, so we should just compute that all the time. That way the property names will always reflect what they contain, which is not the case now. This is the only thing that I see that needs to be addressed in this PR now.

bwspenc (changes requested):

modules/solid_mechanics/src/materials/DensityScaling.C
WilkAndy: Thanks @bwspenc - i'll implement that `additional_density` idea.

modules/solid_mechanics/src/materials/DensityScaling.C
```diff
@@ -17,49 +17,51 @@ InputParameters
...
-  params.addParam<Real>(
-      "factor",
-      1.0,
-      "Factor to multiply to the critical time step. This factor is typically less than one to be "
-      "on the conservative side due to two types of approximation: Time step lagging and the "
-      "approximation of the critical time step formula.");
+      "true_density",
+      "Name of Material Property defining the true inertial density of the material.");
+  params.addRequiredParam<MaterialPropertyName>(
+      "scaled_density", "Name of the density that this Material will compute.");
```
bwspenc:
```suggestion
      "scaled_density", "Name of the scaled density property that this Material computes.");
```

modules/solid_mechanics/src/materials/DensityScaling.C
```diff
@@ -17,49 +17,51 @@ InputParameters
...
+  params.addParam<MaterialPropertyName>(
+      "additional_density",
+      "additional_density",
+      "Scaled_density - true_density will be put into a Material Property with this name.  This is "
+      "typically used to determine where density has been added.");
```
bwspenc:
```suggestion
      "Name of the additional density property that this Material computes");
```

modules/solid_mechanics/include/materials/DensityScaling.h
```diff
@@ -30,18 +31,21 @@ class DensityScaling : public Material
...
   const MaterialProperty<Real> & _material_density;
 
-  /// Effective stiffness of element: function of material properties and element size
-  const MaterialProperty<Real> & _effective_stiffness;
+  /// Square root of effective stiffness of element
+  const MaterialProperty<Real> & _sqrt_effective_stiffness;
 
   /// User defined factor to be multiplied to the critical time step
-  const Real & _factor;
+  const Real & _safety_factor;
```
bwspenc, line 50: The comment above this should be modified to something like this:
```
/// User-defined safety factor that the minimum density required for stability is divided by.

File: modules/solid_mechanics/src/materials/DensityScaling.C [hunk 14]
```diff
@@ -17,49 +17,51 @@ InputParameters
 DensityScaling::validParams()
 {
   InputParameters params = Material::validParams();
-  params.addClassDescription("Automatically scale the material density to achieve the desired time "
-                             "step size to satisfy CFL conditions.");
+  params.addClassDescription(
+      "Computes the inertial density needed to enable stable explicit time-stepping using the "
```
bwspenc: It might be helpful to put "scaled" in here somewhere. Like "Computes the scaled inertial density ..."

bwspenc (changes requested):

bwspenc (approved):

WilkAndy:
Woohoo, thank you, @bwspenc , for all your input with this.

PR #32127, opened 2025-12-18
Do not add non-existent material in Closures1PhaseTHM

### LINKED ISSUE #29651: `Closures1PhaseTHM` uses non-existent class in corner case
### Bug Description

When you use `Closures1PhaseTHM` with multiple heat transfers on a flow channel that are of the heat flux type, then it attempts to create an object of the non-existent class `TemperatureWallFromHeatFlux3EqnTHMMaterial`.

### Steps to Reproduce

See above.

### Impact

This requires users to work around this bug if they have multiple heat fluxes on a flow channel.

### [Optional] Diagnostics

_No response_

joshuahansel:
Whenever a heat flux type heat transfer was connected to a flow channel when using `Closures1PhaseTHM`, the closures tried to add a non-existent material. The prior behavior was to always try to create material properties for the wall heat transfer coefficient and wall temperature for each connected heat transfer, even if the heat transfer was a heat flux. Now, these material properties are not created. If the user wants them, they will need to come up with their own way to determine them, since in general, wall heat transfer coefficient depends on wall temperature (so it would be iterative).

Closes #29651

### DIFF
--- a/modules/thermal_hydraulics/doc/content/source/closures/Closures1PhaseTHM.md
+++ b/modules/thermal_hydraulics/doc/content/source/closures/Closures1PhaseTHM.md
@@ -1,11 +1,9 @@
 # Closures1PhaseTHM
 
-!syntax description /Closures/Closures1PhaseTHM
-
 The closures added are:
 
 - a wall friction factor for the pressure drop,
-- a wall heat transfer coefficient for heat transfer.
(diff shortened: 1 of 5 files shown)


GiudGiud (approved):

joshuahansel:
VTB failure is unrelated; shows on other PRs as well.

PR #32130, opened 2025-12-18
Restrict algebraic ranges to elements with active objects

### LINKED ISSUE #32066: ExplicitMixedOrder: setCurrentAlgebraicRanges appropriately
### Motivation

At the moment `ExplicitMixedOrder` does not use `_fe_problem.setCurrentAlgebraicElementRange` and `_fe_problem.setCurrentAlgebraicNodeRange` : it presumably lets MOOSE do it somewhere, or maybe the AlgebraicElementRange is just set to "all elements" by default.  Anyway, in the unusual case of:

1. the variables being not defined on a particular set of blocks, and
2. the elements in those blocks invert due to the dynamics, and
3. there is Contact (even if it is block restricted, as per #32054).

then MOOSE crashes with a negative Jacobian problem.   I have no idea why Contact matters here, but without it, MOOSE doesn't crash with a negative Jacobian.

### Design

Modify `ExplicitMixedOrder` so it checks the blocks that the variables are defined on, and then sets the appropriate Algebraic Ranges.

### Impact

Help my simulations not crash.

WilkAndy:
## Impact

Closes #32066


WilkAndy:
Hi @dschwen  and @bwspenc and @loganharbour , i have been puzzling over this for a couple of days, and have given up, so i'm hoping you can help.  My ultimate aim here is to restrict MOOSE to only considering elements in blocks of the mesh upon which variables are defined.   The reason for doing this is that in my simulations elements in blocks with no variables defined on them can invert, causing libmesh/moose to crash if moose causes libmesh to realise there's a negative jacobian.

I believe that I've constructed the correct element ranges, and used `setcurrentAlgebraicElementRange` correctly, for the 1-processor case, but i get a segfault when input files are run in parallel.  I think my segfault is due to a ghosting problem, resulting from incorrectly restricting the elements that each processor can see.  I've created a 2-element example of the problem: `two_eles_segfault`.  If you `mpirun -np 2 solid_mechanics-opt -i two_eles_segfault.i` you'll see what i mean.

loganharbour:
I think we might need an eye from @lindsayad or @roystgnr here

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
```diff
@@ -359,6 +432,44 @@ ExplicitMixedOrder::updateDOFIndices()
...
+  auto & mesh = _sys.system().get_mesh();
+
+  // ---- Build union element range across all relevant blocks ----
+  _elem_buffer.clear();
+  _elem_buffer.reserve(mesh.n_active_elem()); // heuristic
+
+  for (const auto id : _relevant_blocks)
+  {
+    // Per-subdomain active elements
+    for (const auto * e : mesh.active_subdomain_element_ptr_range(id))
```
dschwen, line 465: Are those _local_ elements?

WilkAndy: @dschwen , I believe these are *not* just local elements.  This is based on me introducing this code into the above loop:

```
      const int my_rank = mesh.comm().rank();
      const dof_id_type eid = e->id();
      const processor_id_type owner = e->processor_id();
      std::cerr << "rank=" << my_rank << " elem_id=" << eid << " owner=" << owner << std::endl;
```

and receiving this output from two_eles_segfault.i:

```
rank=0 elem_id=0 owner=0
rank=1 elem_id=0 owner=0
```

Note that rank=1 sees element zero, which is owned by rank=0

dschwen: You could either just skip elements that are not owned, or figure out if there is a range in libmesh that is purely local. Is `active_subdomain_local_element_ptr_range` or `active_local_subdomain_element_ptr_range` a thing? These are not findable in the code because they are built using macros. I also haven't seen docs on those ranges (nothing in the repo, nothing in Doxygen).

WilkAndy: Thanks @dschwen , adding 

```
      if (owner == my_rank)
        _elem_buffer.push_back(e);
```

results in no segfault in the two_eles_segfault situation.  I don't understand why, but ....  i'll keep testing on other cases.

roystgnr: > Is `active_subdomain_local_element_ptr_range` or `active_local_subdomain_element_ptr_range` a thing

They aren't, but they'll be easy to add.  In the meantime manually skipping the `owner != my_rank` case is a fine workaround.

roystgnr:
My first thought was to wonder whether MOOSE is prerequesting things it doesn't need to, which would be very easy to do (even `get_JxW()` would scream upon seeing an inverted element) and very easy not to notice (there'd be no performance hit except on variable-free elements) and relatively easy to fix if it was the problem ... but "I have no idea why Contact matters here, but without it, MOOSE doesn't crash with a negative Jacobian." suggests that we actually are doing the right thing with even the most foundational pre-requests.  Kudos, MOOSE!

I'd guess that Contact (or the PointLocatorTree it builds?) is at some point checking elements' geometry in a way that calls `inverse_map()` to examine point locations on the master element, and `inverse_map()` is screaming when it sees a non-invertable map, but I'd need a stack trace to be sure.

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
dschwen: >  I don't understand why, but ...

@WilkAndy, that's because you are forcing MOOSE to evaluate residual on elements it doesn't own (i.e. ghosted elements from the neighboring processor). Those elements do not have sufficient data to evaluate the shape functions, and you are producing residual vector entries that are not supposed to exist on the processor.

roystgnr: Technically we can evaluate shape functions on any element, we just don't have all the coefficients to multiply them by unless we're on a local or algebraically ghosted element.

dschwen: I knew there was a correction incoming... I'm curious where exactly the segfault happened

WilkAndy: Below is lldb output:

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x000000010d279424 libmpi.12.dylib`MPI_Abort
    frame #1: 0x0000000129b1f350 libmesh_dbg.0.dylib`libMesh::libmesh_terminate_handler() + 372
    frame #2: 0x0000000185bb0698 libc++abi.dylib`std::__terminate(void (*)()) + 16
    frame #3: 0x0000000185bb3ee4 libc++abi.dylib`__cxa_rethrow + 188
    frame #4: 0x0000000117833b38 libmoose-dbg.0.dylib`ThreadedElementLoopBase<libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*>>::operator()(this=0x000000016db6c6a8, range=0x0000600001c27ac0, bypass_threading=false) at ThreadedElementLoopBase.h:323:9
    frame #5: 0x0000000117fb22e8 libmoose-dbg.0.dylib`void libMesh::Threads::parallel_reduce<libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*>, ComputeMaterialsObjectThread>(range=0x0000600001c27ac0, body=0x000000016db6c6a8) at threads_pthread.h:367:5
    frame #6: 0x0000000117ed046c libmoose-dbg.0.dylib`FEProblemBase::initElementStatefulProps(this=0x0000000132852c18, elem_range=0x0000600001c27ac0, threaded=true) at FEProblemBase.C:8527:5
    frame #7: 0x0000000117d0a22c libmoose-dbg.0.dylib`FEProblemBase::initialSetup(this=0x0000000132852c18) at FEProblemBase.C:1482:7
    frame #8: 0x0000000116c3e33c libmoose-dbg.0.dylib`TransientBase::init(this=0x0000000132851c18) at TransientBase.C:252:12
    frame #9: 0x0000000116c41eac libmoose-dbg.0.dylib`Transient::init(this=0x0000000132851c18) at Transient.C:39:18
    frame #10: 0x0000000119958454 libmoose-dbg.0.dylib`MooseApp::executeExecutioner(this=0x0000000131816c00) at MooseApp.C:1858:19
    frame #11: 0x000000011994fab8 libmoose-dbg.0.dylib`MooseApp::run(this=0x0000000131816c00) at MooseApp.C:2263:5
    frame #12: 0x000000010229b550 solid_mechanics-dbg`int Moose::main<SolidMechanicsTestApp>(argc=4, argv=0x000000016db758c8) at MooseMain.h:41:8
    frame #13: 0x000000010229b4c8 solid_mechanics-dbg`main(argc=4, argv=0x000000016db758c8) at main.C:17:10
    frame #14: 0x000000018585ab98 dyld`start + 6076
```

```
frame #6: 0x0000000117ed046c libmoose-dbg.0.dylib`FEProblemBase::initElementStatefulProps(this=0x0000000132852c18, elem_range=0x0000600001c27ac0, threaded=true) at FEProblemBase.C:8527:5
   8524    ComputeMaterialsObjectThread cmt(
   8525        *this, _material_props, _bnd_material_props, _neighbor_material_props, _assembly);
   8526    if (threaded)
-> 8527      Threads::parallel_reduce(elem_range, cmt);
   8528    else
   8529      cmt(elem_range, true);
   8530  
```

dschwen: This is where you need to break when the exception is thrown (`breakpoint set -E c++`)

WilkAndy:
Probably the latest push will mean all tests pass (fingers crossed!) and this PR has the functionality i need.  However, I feel a bit hesitant about this PR, @dschwen , as it adds a lot of code to address an issue that might be quite niche.  (Of course for me, the issue is not niche but mainstream.).  Are you OK with all this extra code, or can you see a better way of achieving my goal.

dschwen:
@WilkAndy I need to think about this a bit. In #32198 I also manipulate the algebraic range and we need to make sure those two PRs work well together.

WilkAndy:
> @WilkAndy I need to think about this a bit. In #32198 I also manipulate the algebraic range and we need to make sure those two PRs work well together.

OK, thanks @dschwen .   What i'll do is tidy this PR up (eg, squash commits, add some more explanation, and add some more tests) so it's "ready to go", and then let's see how you can improve it :-)

modules/solid_mechanics/src/timeintegrators/ExplicitMixedOrder.C
WilkAndy: Aha, thanks @dschwen .   I *think* this, is it, but i need to go to bed now.

```
 * frame #0: 0x0000000185bb3e28 libc++abi.dylib`__cxa_rethrow
    frame #1: 0x00000001184e21c8 libmoose-dbg.0.dylib`ThreadedElementLoopBase<libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*>>::operator()(this=0x000000016d42bd38, range=0x000060000354a620, bypass_threading=false) at ThreadedElementLoopBase.h:323:9
    frame #2: 0x0000000118c61cac libmoose-dbg.0.dylib`void libMesh::Threads::parallel_reduce<libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*>, ComputeMaterialsObjectThread>(range=0x000060000354a620, body=0x000000016d42bd38) at threads_pthread.h:367:5
    frame #3: 0x0000000118b7ec00 libmoose-dbg.0.dylib`FEProblemBase::initElementStatefulProps(this=0x0000000136212a18, elem_range=0x000060000354a620, threaded=true) at FEProblemBase.C:8527:5
    frame #4: 0x00000001189b89c0 libmoose-dbg.0.dylib`FEProblemBase::initialSetup(this=0x0000000136212a18) at FEProblemBase.C:1482:7
    frame #5: 0x00000001178efb14 libmoose-dbg.0.dylib`TransientBase::init(this=0x000000013620fa18) at TransientBase.C:252:12
    frame #6: 0x00000001178f3684 libmoose-dbg.0.dylib`Transient::init(this=0x000000013620fa18) at Transient.C:39:18
    frame #7: 0x000000011a6e0618 libmoose-dbg.0.dylib`MooseApp::executeExecutioner(this=0x0000000136206200) at MooseApp.C:1858:19
    frame #8: 0x000000011a6d7c74 libmoose-dbg.0.dylib`MooseApp::run(this=0x0000000136206200) at MooseApp.C:2263:5
    frame #9: 0x00000001029df550 solid_mechanics-dbg`int Moose::main<SolidMechanicsTestApp>(argc=4, argv=0x000000016d434f50) at MooseMain.h:41:8
    frame #10: 0x00000001029df4c4 solid_mechanics-dbg`main(argc=4, argv=0x000000016d434f50) at main.C:17:10
    frame #11: 0x000000018585ab98 dyld`start + 6076
(lldb) f 2
frame #2: 0x0000000118c61cac libmoose-dbg.0.dylib`void libMesh::Threads::parallel_reduce<libMesh::StoredRange<libMesh::MeshBase::const_element_iterator, libMesh::Elem const*>, ComputeMaterialsObjectThread>(range=0x000060000354a620, body=0x000000016d42bd38) at threads_pthread.h:367:5
   364    // If we're running in serial - just run!
   365    if (libMesh::n_threads() == 1)
   366    {
-> 367      body(range);
   368      return;
   369    }
   370 
(lldb) f 3
frame #3: 0x0000000118b7ec00 libmoose-dbg.0.dylib`FEProblemBase::initElementStatefulProps(this=0x0000000136212a18, elem_range=0x000060000354a620, threaded=true) at FEProblemBase.C:8527:5
   8524   ComputeMaterialsObjectThread cmt(
   8525       *this, _material_props, _bnd_material_props, _neighbor_material_props, _assembly);
   8526   if (threaded)
-> 8527     Threads::parallel_reduce(elem_range, cmt);
   8528   else
   8529     cmt(elem_range, true);
   8530
(lldb) f 4
frame #4: 0x00000001189b89c0 libmoose-dbg.0.dylib`FEProblemBase::initialSetup(this=0x0000000136212a18) at FEProblemBase.C:1482:7
   1479     {
   1480       TIME_SECTION("computeMaterials", 2, "Computing Initial Material Properties");
   1481
-> 1482       initElementStatefulProps(getCurrentAlgebraicElementRange(), true);
   1483     }
   1484 #ifdef MOOSE_KOKKOS_ENABLED
   1485     if (_kokkos_material_props.hasStatefulProperties() ||
(lldb) 
```

WilkAndy:
Hi @dschwen , i'll leave this with you for a bit.  I think my git commit message adequately explains what goes on here, and my worries.

WilkAndy:
Any thoughts about this algebraic range stuff, @dschwen ?   My stuff still doesn't work in parallel - it doesn't segfault, but it doesn't do the right thing with ghosting.

WilkAndy:
Shall we somehow merge our "algebraic range" PRs together, @dschwen ?

WilkAndy:
Boohoo, @dschwen  - this looks like a failed PR.   Do you have any comments?

PR #32132, opened 2025-12-19
Expand moose code standard

### LINKED ISSUE #32131: Adding more requirements to the code standard
In my opinion, the current code standard is missing some important requirements, making some PRs difficult to review. We should expand (and keep expanding) the requirements.

Even if a contributor doesn't strictly follow the requirements that is fine, at least we have a reference to point at.

hugary1995:
Add more hints, requirements, and design guidelines to the SQA moose code standard page.

close #32131

### DIFF
--- a/framework/doc/content/sqa/framework_scs.md
+++ b/framework/doc/content/sqa/framework_scs.md
@@ -121,15 +121,15 @@ In addition, a few convenience methods are available to help generate integer ra
 ```C++
 #include "libmesh/int_range.h"
 
-// Looping through n1, n1+1, n1+2, ..., n2-1, n2
+// Looping through n1, n1+1, n1+2, ..., n2-1
 for (auto i : make_range(n1, n2))
   // ... Do something with index i
 
-// Looping through 0, 1, 2, ..., n
+// Looping through 0, 1, 2, ..., n-1
 for (auto i : make_range(n))
   // ... Do something with index i
 
-// Looping through 0, 1, 2, ..., v.size()
+// Looping through 0, 1, 2, ..., v.size()-1
 for (auto i : index_range(v))
   // ... Do something with index i
 ```
@@ -369,9 +369,9 @@ If a member is not intended to be used or overridden by derived classes, it shou
 
 In summary:
 
-- Default to private
-- Promote to public only when there is a clear, justified use case
-- Treat protected as part of the class’s internal design contract, not its public API
+- Default to `private`
+- Promote to `public` only when there is a clear, justified use case
+- Treat `protected` as part of the class’s internal design contract, not its public API
 - Prefer composition over inheritance when possible
 
 Code reviews may request access-level reductions if members are more visible than necessary.
@@ -398,7 +398,7 @@ Prefer compile-time constants when values are known at compile time and semantic
 
 Use:
 
-- constexpr or consteval for values and functions known at compile time
+- `constexpr` or `consteval` for values and functions known at compile time
 - Non-type template parameters (e.g., dimensions, tensor ranks, fixed sizes) when they affect layout, correctness, or optimization
 - `std::integral_constant` or `constexpr` variables instead of macros
 
--- a/framework/doc/content/sqa/framework_scs.md
+++ b/framework/doc/content/sqa/framework_scs.md
@@ -83,6 +83,7 @@ Const-correctness is not cosmetic. It:
 - Prevents accidental state mutation, especially in generic and templated code
 
 Code that is logically const but not marked const will generally be treated as a design issue during review.
+
 ## Index Variables in Looping Constructs
 
 In keeping with good `auto` practices, it would be really nice if we could use `auto` for
@@ -185,7 +186,7 @@ using ElemSide = std::tuple<dof_id_type, unsigned int, BoundaryID>;
 std::vector<ElemSide> my_vec2;
 ```
 
-Remember `using` also works with template.
+Remember that `using` also works with template.
 
 ## Lambdas
 
@@ -285,7 +286,7 @@ Define a member when the operation:
   tensor.transpose();
   ```
 
-  Ask: “Would users naturally discover this by typing obj.?” If so, make it a member function.
+  Ask yourself: “Would users naturally discover this by typing obj.?” If so, make it a member function.
 
 - Needs privileged access: If the function would otherwise require: friendship, access to private data, or bypassing encapsulation, make it a member (or a carefully chosen friend).
 
@@ -298,6 +299,7 @@ Use a free function for
   distance(p, q);
   misorientation(g1, g2);
   ```
+
   If no operand is conceptually dominant: Avoid `a.dot(b)` and prefer `dot(a, b)`.
 
 - Algorithms over data
@@ -317,7 +319,7 @@ Use a free function for
   auto norm(const T & x);
   ```
 
-  Free functions allow generic programming, ADL-based customization, extension to foreign types. Member functions cannot be added retroactively.
+  Free functions allow generic programming, customization via argument-dependent lookup, extension to foreign types. Member functions cannot be added retroactively.
 
 - Operations that conceptually sit “between” types
 
--- a/framework/doc/content/sqa/framework_scs.md
+++ b/framework/doc/content/sqa/framework_scs.md
@@ -354,7 +354,7 @@ When adding a new function, carefully decide whether it should be
 
 ### Decision checklist
 
-1.  Does it mutate the object or enforce consistency of internal state? Yes → non-static member
+1.  Does the function need access to object's internal data/state? Yes → non-static member
 2.  Does it conceptually belong to the type, but not an instance? Yes → static member
 3.  Is it symmetric, algorithmic, or cross-cutting? Yes → free function
 4.  Should users be able to extend it for new types via overloading/ADL? Yes → free function
--- a/framework/doc/content/sqa/framework_scs.md
+++ b/framework/doc/content/sqa/framework_scs.md
@@ -372,8 +372,8 @@ In summary:
 
 - Default to `private`
 - Promote to `public` only when there is a clear, justified use case
-- Treat `protected` as part of the class’s internal design contract, not its public API
-- Prefer composition over inheritance when possible
+- Treat `protected` members as implementation details meant only for subclasses, not as part of the class’s public interface
+- Consider composition as an alternative to inheritance, especially when behavior can be delegated rather than extended.
 
 Code reviews may request access-level reductions if members are more visible than necessary.
 
--- a/framework/doc/content/sqa/framework_scs.md
+++ b/framework/doc/content/sqa/framework_scs.md
@@ -65,6 +65,24 @@ Make sure your variables have +good+ names when using auto!
     doSomething();
 ```
 
+## Const-correctness
+
+We expect contributions to follow strict const-correctness.
+
+- Use const to accurately express intent, invariants, and ownership:
+- Mark function parameters const whenever they are not modified.
+- Mark member functions const whenever they do not modify observable object state.
+- Prefer returning const references when mutation is not intended.
+- Avoid casting away const (e.g., `const_cast`) except in narrowly justified, well-documented cases.
+
+Const-correctness is not cosmetic. It:
+
+- Improves API clarity and self-documentation
+- Enables safer refactoring and compiler diagnostics
+- Allows the compiler to reason more effectively about aliasing and optimization
+- Prevents accidental state mutation, especially in generic and templated code
+- Code that is logically const but not marked const will generally be treated as a design issue during review.
+
 ## Index Variables in Looping Constructs
 
 In keeping with good `auto` practices, it would be really nice if we could use `auto` for
@@ -91,12 +109,83 @@ you with `MooseIndex`:
 ```C++
   // Looping to a scalar index
   for (MooseIndex(n_nodes) i = 0; i < nodes; ++i)
-    ... Do something with index i
+    // ... Do something with index i
 
   // Looping to a container size (where range-for isn't sufficient)
   for (MooseIndex(vector) i = 0; i < vector.size(); ++i)
-    ... Do something with index i
+    // ... Do something with index i
 ```
+
+In addition, a few convenience methods are available to help generate integer ranges:
+
+```C++
+#include "libmesh/int_range.h"
+
+// Looping through n1, n1+1, n1+2, ..., n2-1, n2
+for (auto i : make_range(n1, n2))
+  // ... Do something with index i
+
+// Looping through 0, 1, 2, ..., n
+for (auto i : make_range(n))
+  // ... Do something with index i
+
+// Looping through 0, 1, 2, ..., v.size()
+for (auto i : index_range(v))
+  // ... Do something with index i
+```
+
+## Structured binding
+
+Prefer structured binding when unpacking aggregate types and tuple-like objects.
+
+An example for aggregate type structured binding:
+
+```C++
+struct S
+{
+  int x;
+  double y;
+};
+
+S s{1, 2.0};
+
+// Bind by copying
+auto [a, b] = s;
+
+// Bind by reference
+const auto & [a_ref, b_ref] = s;
+```
+
+An example for tuple-like object binding,
+
+```C++
+std::tuple<int, Real, Point> t = {1, 0.5, Point(0.1, 0.2, 0.3)};
+const auto & [a, b, c] = t;
+```
+
+Tuple-like object includes `std::pair`, `std::tuple`, `std::array` etc.
+
+Using structured binding for unpacking sub-objects in a loop makes the intention very clear, i.e.,
+
+```C++
+std::vector<std::tuple<dof_id_type, unsigned int, BoundaryID>> my_vec;
+for (const auto & [elem_id, side, bnd] : my_vec)
+  // ...
+```
+
+## Aliasing
+
+Consider using `using` (over `typedef`) to define type aliases to improve code readability:
+
+```C++
+std::vector<std::tuple<dof_id_type, unsigned int, BoundaryID>> my_vec;
+
+// Simplify code and clarify intention by defining an alias
+using ElemSide = std::tuple<dof_id_type, unsigned int, BoundaryID>;
+std::vector<ElemSide> my_vec2;
+```
+
+Remember `using` also works with template.
 
 ## Lambdas
 
@@ -110,7 +199,7 @@ List captured variables (by value or reference) in the capture list explicitly w
 
 ## Other C++ Notes
 
-- Use the `override` keyword on overridden `virtual` methods
+- Use the `override` keyword on overridden `virtual` methods; do not mark the overriding method as `virtual` as it is redundant
 - Use `std::make_shared<T>()` when allocating new memory for shared pointers
 - Use `std::make_unique<T>()` when allocating new memory for unique pointers
 - Make use of std::move() for efficiency when possible
@@ -134,6 +223,20 @@ following one-liner from the root directory of your repository will format the c
 find . -name '*.[Chi]' -or -name '*.py' | xargs perl -pli -e 's/\s+$//'
 ```
 
+## Prefer C++ constructs over C constructs
+
+Use modern C++ language features and standard library facilities in preference to legacy C-style constructs. C++ provides stronger type safety, better abstraction, and clearer intent. New code should reflect these advantages.
+
+Prefer:
+
+- `nullptr` over `NULL` or `0`
+- `static_cast`, `const_cast`, `reinterpret_cast` over C-style casts
+- `constexpr` and `const` over preprocessor macros
+- `enum class` over unscoped `enum`
+- `std::array`, `std::vector`, `std::span` over raw arrays and pointer arithmetic
+- RAII types (e.g., `std::unique_ptr`, `std::lock_guard`) over manual resource management
+- `<algorithm>` and `<numeric>` over hand-written loops where appropriate
+
 ## C++ Includes
 
 Only include things that are absolutely necessary in header files. Please use forward declarations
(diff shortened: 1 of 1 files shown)


hugary1995:
I feel some of them are pretty biased (but IMHO important). @idaholab/moose-ccb please review and advise. Please also feel free to add to the list.

It would be nice to let moosebot automatically leave a comment with a link to this page in every new PR. At least for first-time contributors I think that's going to be important.

In the future, we might as well let moosebot post a checklist to each PR for the reviewers to go through.

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +109,84 @@ you with `MooseIndex`:
...
-    ... Do something with index i
+    // ... Do something with index i
+```
+
+In addition, a few convenience methods are available to help generate integer ranges:
+
+```C++
+#include "libmesh/int_range.h"
+
+// Looping through n1, n1+1, n1+2, ..., n2-1, n2
```
nmnobre:
```suggestion
// Looping through n1, n1+1, n1+2, ..., n2-1
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +109,84 @@ you with `MooseIndex`:
...
+for (auto i : make_range(n1, n2))
+  // ... Do something with index i
+
+// Looping through 0, 1, 2, ..., n
```
nmnobre:
```suggestion
// Looping through 0, 1, 2, ..., n-1
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +109,84 @@ you with `MooseIndex`:
...
+for (auto i : make_range(n))
+  // ... Do something with index i
+
+// Looping through 0, 1, 2, ..., v.size()
```
nmnobre:
```suggestion
// Looping through 0, 1, 2, ..., v.size()-1
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,164 @@ All non-system includes should use quotes with a single space between `include`
...
+
+Public interfaces should be minimal, const-correct, and clearly documented. Adding something to `public` is a commitment.
+
+Prefer `private` by default. Use `private` for:
+
+- Representation details and internal state
+- Helper functions and implementation mechanics
+- Anything that should not be relied upon by derived classes or external code
+
+Private members allow internal refactoring without breaking users and are the strongest tool for enforcing invariants.
```
nmnobre: User assumptions? Sounds like a word is missing.

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,164 @@ All non-system includes should use quotes with a single space between `include`
...
+- Deep or obscure template metaprogramming that obscures intent
+- Encoding runtime variability as template parameters without a clear benefit
+
+Templates should make code more general and clearer, not harder to understand.
+
+Prefer compile-time constants when values are known at compile time and semantically fixed.
+
+Use:
+
+- constexpr or consteval for values and functions known at compile time
```
nmnobre:
```suggestion
- `constexpr` or `consteval` for values and functions known at compile time
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,164 @@ All non-system includes should use quotes with a single space between `include`
...
+- The invariant contract between base and derived classes is clear
+- Avoid exposing data members as `protected`. Prefer `protected` virtual functions or `protected` helper methods instead.
+
+If a member is not intended to be used or overridden by derived classes, it should **not** be `protected`.
+
+In summary:
+
+- Default to private
+- Promote to public only when there is a clear, justified use case
+- Treat protected as part of the class’s internal design contract, not its public API
```
nmnobre:
```suggestion
- Default to `private`
- Promote to `public` only when there is a clear, justified use case
- Treat `protected` as part of the class’s internal design contract, not its public API
```

nmnobre (changes requested):

framework/doc/content/sqa/framework_scs.md
hugary1995:
```suggestion
Private members allow internal refactoring without breaking user assumptions and are the strongest tool for enforcing invariants.
```

nmnobre (approved):

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +109,84 @@ you with `MooseIndex`:
...
+// Looping through n1, n1+1, n1+2, ..., n2-1
+for (auto i : make_range(n1, n2))
+  // ... Do something with index i
+
+// Looping through 0, 1, 2, ..., n-1
+for (auto i : make_range(n))
+  // ... Do something with index i
+
+// Looping through 0, 1, 2, ..., v.size()-1
+for (auto i : index_range(v))
```
joshuahansel:
```suggestion
for (const auto i : make_range(n1, n2))
  // ... Do something with index i

// Looping through 0, 1, 2, ..., n-1
for (const auto i : make_range(n))
  // ... Do something with index i

// Looping through 0, 1, 2, ..., v.size()-1
for (const auto i : index_range(v))
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -65,6 +65,24 @@ Make sure your variables have +good+ names when using auto!
...
+- Avoid casting away const (e.g., `const_cast`) except in narrowly justified, well-documented cases.
+
+Const-correctness is not cosmetic. It:
+
+- Improves API clarity and self-documentation
+- Enables safer refactoring and compiler diagnostics
+- Allows the compiler to reason more effectively about aliasing and optimization
+- Prevents accidental state mutation, especially in generic and templated code
+- Code that is logically const but not marked const will generally be treated as a design issue during review.
+
```
joshuahansel:
```suggestion

Code that is logically const but not marked const will generally be treated as a design issue during review.
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -65,6 +65,24 @@ Make sure your variables have +good+ names when using auto!
     doSomething();
 ```
 
+## Const-correctness
+
+We expect contributions to follow strict const-correctness.
+
+- Use const to accurately express intent, invariants, and ownership:
```
joshuahansel:
```suggestion
- Use const to accurately express intent, invariants, and ownership.
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,164 @@ All non-system includes should use quotes with a single space between `include`
...
+- The invariant contract between base and derived classes is clear
+- Avoid exposing data members as `protected`. Prefer `protected` virtual functions or `protected` helper methods instead.
+
+If a member is not intended to be used or overridden by derived classes, it should **not** be `protected`.
+
+In summary:
+
+- Default to `private`
+- Promote to `public` only when there is a clear, justified use case
+- Treat `protected` as part of the class’s internal design contract, not its public API
```
joshuahansel: Most people, including me, don't know what "internal design contract" means

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,164 @@ All non-system includes should use quotes with a single space between `include`
...
+- Prefer composition over inheritance when possible
```
joshuahansel: I'd prefer a weaker statement like "Consider composition as an alternative to inheritance when it could ..."

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,164 @@ All non-system includes should use quotes with a single space between `include`
...
+
+Private members allow internal refactoring without breaking user assumptions and are the strongest tool for enforcing invariants.
+
+`protected` exists to support controlled customization via inheritance, not general access. Use `protected` only when:
+
+- A derived class must override or extend behavior
+- The base class explicitly documents a customization or extension point
+- The invariant contract between base and derived classes is clear
+- Avoid exposing data members as `protected`. Prefer `protected` virtual functions or `protected` helper methods instead.
+
```
joshuahansel:
```suggestion

Avoid exposing data members as `protected`. Prefer `protected` virtual functions or `protected` helper methods instead.

```

joshuahansel (changes requested):

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,165 @@ All non-system includes should use quotes with a single space between `include`
...
+Avoid exposing data members as `protected`. Prefer `protected` virtual functions or `protected` helper methods instead.
+
+If a member is not intended to be used or overridden by derived classes, it should **not** be `protected`.
+
+In summary:
+
+- Default to `private`
+- Promote to `public` only when there is a clear, justified use case
+- Treat `protected` as part of the class’s internal design contract, not its public API
```
hugary1995:
```suggestion
- Treat `protected` members as implementation details meant only for subclasses, not as part of the class’s public interface
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,165 @@ All non-system includes should use quotes with a single space between `include`
...
+- Prefer composition over inheritance when possible
```
hugary1995:
```suggestion
- Consider composition as an alternative to inheritance, especially when behavior can be delegated rather than extended.
```

framework/doc/content/sqa/framework_scs.md
hugary1995: Yeah, I rephrased it.

framework/doc/content/sqa/framework_scs.md
hugary1995: Sure, that's a fair ask in a framework like moose.

framework/doc/content/sqa/framework_scs.md
```diff
@@ -65,6 +65,24 @@ Make sure your variables have +good+ names when using auto!
...
+
+Const-correctness is not cosmetic. It:
+
+- Improves API clarity and self-documentation
+- Enables safer refactoring and compiler diagnostics
+- Allows the compiler to reason more effectively about aliasing and optimization
+- Prevents accidental state mutation, especially in generic and templated code
+
+Code that is logically const but not marked const will generally be treated as a design issue during review.
 ## Index Variables in Looping Constructs
```
GiudGiud, line 87:
```suggestion

## Index Variables in Looping Constructs
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +109,84 @@ you with `MooseIndex`:
...
+
+```C++
+std::vector<std::tuple<dof_id_type, unsigned int, BoundaryID>> my_vec;
+
+// Simplify code and clarify intention by defining an alias
+using ElemSide = std::tuple<dof_id_type, unsigned int, BoundaryID>;
+std::vector<ElemSide> my_vec2;
+```
+
+Remember `using` also works with template.
```
GiudGiud:
```suggestion
Remember that `using` also works with template.
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,165 @@ All non-system includes should use quotes with a single space between `include`
...
+  If the function mutates internal state, must preserve class invariants, or requires intimate knowledge of representation, make it a member function.
+
+- Is conceptually “owned” by the type
+
+  ```C++
+  particle.update(dt);
+  tensor.transpose();
+  ```
+
+  Ask: “Would users naturally discover this by typing obj.?” If so, make it a member function.
```
GiudGiud:
```suggestion
  Ask yourself: “Would users naturally discover this by typing obj.?” If so, make it a member function.
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,165 @@ All non-system includes should use quotes with a single space between `include`
...
+Use a free function for
+
+- Symmetric operations
+
+  ```C++
+  dot(a, b);
+  distance(p, q);
+  misorientation(g1, g2);
+  ```
+  If no operand is conceptually dominant: Avoid `a.dot(b)` and prefer `dot(a, b)`.
```
GiudGiud:
```suggestion

  If no operand is conceptually dominant: Avoid `a.dot(b)` and prefer `dot(a, b)`.
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +256,165 @@ All non-system includes should use quotes with a single space between `include`
...
+  If it takes an object as input, produces a value, and does not modify the object, make it a free function. This aligns with the STL design: containers own storage; algorithms are free.
+
+- Extensibility across types
+
+  ```C++
+  template<class T>
+  auto norm(const T & x);
+  ```
+
+  Free functions allow generic programming, ADL-based customization, extension to foreign types. Member functions cannot be added retroactively.
```
GiudGiud: define ADL please

hugary1995:
```suggestion
  Free functions allow generic programming, customization via argument-dependent lookup, extension to foreign types. Member functions cannot be added retroactively.
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +110,84 @@ you with `MooseIndex`:
...
-    ... Do something with index i
+    // ... Do something with index i
 
   // Looping to a container size (where range-for isn't sufficient)
   for (MooseIndex(vector) i = 0; i < vector.size(); ++i)
-    ... Do something with index i
+    // ... Do something with index i
+```
+
+In addition, a few convenience methods are available to help generate integer ranges:
```
lindsayad: I think we should delete all the content between the section header and this line or at least move it behind the `int_range` example. The only reason to even mention the `MooseIndex` code at all is for the case where the index is unused in the loop body

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +110,84 @@ you with `MooseIndex`:
...
 
+Tuple-like object includes `std::pair`, `std::tuple`, `std::array` etc.
+
+Using structured binding for unpacking sub-objects in a loop makes the intention very clear, i.e.,
+
+```C++
+std::vector<std::tuple<dof_id_type, unsigned int, BoundaryID>> my_vec;
+for (const auto & [elem_id, side, bnd] : my_vec)
+  // ...
+```
```
lindsayad, line 161: You could add that you can use `_` if one of the bindings would go unused

framework/doc/content/sqa/framework_scs.md
```diff
@@ -91,13 +110,84 @@ you with `MooseIndex`:
...
+
+```C++
+std::vector<std::tuple<dof_id_type, unsigned int, BoundaryID>> my_vec;
+
+// Simplify code and clarify intention by defining an alias
+using ElemSide = std::tuple<dof_id_type, unsigned int, BoundaryID>;
+std::vector<ElemSide> my_vec2;
+```
+
+Remember that `using` also works with template.
```
lindsayad:
```suggestion
Remember that `using` also works with templates.
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -110,7 +200,7 @@ List captured variables (by value or reference) in the capture list explicitly w
 
 ## Other C++ Notes
 
-- Use the `override` keyword on overridden `virtual` methods
+- Use the `override` keyword on overridden `virtual` methods; do not mark the overriding method as `virtual` as it is redundant
```
lindsayad, line 189: I believe that recent clang warns about this? I personally have asked people in the past to use `virtual` even with override because otherwise `virtual` doesn't appear on the doxygen page, and it's convenient to search for just one word (`virtual`) when determining the `virtual` interface for a class

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +257,166 @@ All non-system includes should use quotes with a single space between `include`
 #include <system_library>
 ```
 
+## Free function v.s. member function
```
lindsayad:
```suggestion
## Free function vs. member function
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +257,166 @@ All non-system includes should use quotes with a single space between `include`
...
+
+When adding a new function, please carefully decide whether to make it a free function (i.e., outside the scope of the class) or a member function.
```
lindsayad: I think this discussion misses out on another valuable option IMO, which is static method members. We use these quite frequently in both libMesh and MOOSE when the concepts fit a class well but no object data is used. I suppose you could argue for nested namespaces as an alternative. @roystgnr do you have thoughts on this?

lindsayad: ADL to me is the best reason you put forward here for preferring a free function over a static method member

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +257,166 @@ All non-system includes should use quotes with a single space between `include`
...
+
+1. Does this operation mutate the object or enforce invariants? Yes → member
+2. Is it conceptually “owned” by the object? Yes → member
+3. Is the operation symmetric or algorithmic? Yes → free function
+4. Do I want/expect others to extend this behavior for new types? Yes → free function
+5. If you’re unsure → default to free function.
+
+As a reminder, helper free functions defined only in the translation unit (i.e., `.C` files) should be marked as `static` or encapsulated in an anonymous namespace to force static linking.
+
+## Access control: public, protected, and private
```
lindsayad, line 367: If you could add that access order should be, whenever possible
```
public:
protected:
private:
```
in the class definition for expedited API review

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +257,166 @@ All non-system includes should use quotes with a single space between `include`
...
+Prefer:
+
+- Function and class templates for algorithms that operate uniformly across types
+- `auto` and template parameter deduction to reduce redundancy
+- Concepts (not available yet in C++17) or `static_assert` to document and enforce template requirements
+
+Avoid:
+
+- Templates used only to avoid writing a small amount of code
+- Deep or obscure template metaprogramming that obscures intent
```
lindsayad:
```suggestion
- Deep or obscure template metaprogramming that obscures intent if it can be avoided
```

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +257,166 @@ All non-system includes should use quotes with a single space between `include`
...
+- Encoding runtime variability as template parameters without a clear benefit
+
+Templates should make code more general and clearer, not harder to understand.
```
lindsayad: I don't entirely agree here. The STL library is hideous to look at. But it sure makes downstream programming much nicer when you understand it. I think that even if a template is difficult to understand but it adds significant functionality, then it's useful. MetaPhysicL is probably hard for a lot of people to understand, but it's useful. We should certainly strive to make templates as clear to understand as possible but I don't think we should *require* that templates be both general/useful *and* clear
```suggestion
Templates should make code more general/flexible and ideally more clear, not harder to understand.
```

lindsayad (changes requested):
There are great additions, thanks for working on this @hugary1995

framework/doc/content/sqa/framework_scs.md
roystgnr: We definitely want some mention of static member functions here.  IMHO ADL is the key in a conceptual sense, not just a technical one: if a function is general enough that we could imagine implementing the exact same semantics for multiple different classes, then it doesn't make sense to put it in the "namespace" of one class; if it is, then it does.

framework/doc/content/sqa/framework_scs.md
dschwen: avoid if it can be avoided...

dschwen:
```suggestion
- Deep or obscure template metaprogramming that obscures intent, unless there is a strong technical justification (e.g., the required functionality is only achievable via template metaprogramming).
```

framework/doc/content/sqa/framework_scs.md
dschwen: https://github.com/doxygen/doxygen/issues/4763 ☹️

dschwen: Oh, actually the function gets the `virtual` label on the right hand side (confirmed locally) so searching for the word virtual should still surface it.

framework/doc/content/sqa/framework_scs.md
dschwen: IMO we should do away with `MooseIndex` entirely. Why not turn

```
for (MooseIndex(vector) i = 0; i < vector.size(); ++i)
```

into

```
for ([[maybe_unused]] const auto & i: index_range(vector))
```

(and promote the C++ feature for unused variables)? 

P.S.: It would be nice to have the C++26 `_` (especially for partially used structured bindings).

framework/doc/content/sqa/framework_scs.md
lindsayad: Interesting. Could you find an example of this on the MOOSE doxygen pages? If that's true, then I will remove my complaint here

framework/doc/content/sqa/framework_scs.md
lindsayad: I like where your head's at

framework/doc/content/sqa/framework_scs.md
lindsayad: > P.S.: It would be nice to have the C++26 _ (especially for partially used structured bindings).

@dschwen it seems like even c++17 compilers have chosen to support this? Because we already have this idiom used in a decent number of places in the framework

framework/doc/content/sqa/framework_scs.md
lindsayad: resolving in favor of @dschwen's later suggestion

framework/doc/content/sqa/framework_scs.md
hugary1995: I did some search on this. In C++17, `_` is just a normal identifier instead of a placeholder. The compilers we currently support probably don’t warn per-binding if any binding is used. For that reason, I guess I will not mention the use `_` in the guidelines for now.

framework/doc/content/sqa/framework_scs.md
hugary1995: https://mooseframework.inl.gov/docs/doxygen/moose/classNEML2ModelExecutor.html#ac17c64d6031114f251c86c6f61f87cbb

These functions are not marked virtual, but indeed there's a virtual label on the far right.

hugary1995:
I let chatgpt rewrite the sections on "free function vs. member function vs. static member function" and "generic programming". See if it is able to satisfy you. I manually addressed the other comments.

framework/doc/content/sqa/framework_scs.md
```diff
@@ -257,88 +243,129 @@ All non-system includes should use quotes with a single space between `include`
...
+When adding a new function, carefully decide whether it should be
 
-The core rule is
+- a **non-static member function** (operates on an object),
+- a **static member function** (conceptually belongs to a type, but not to an instance), or
+- a **free function** (algorithm or operation external to any one type).
 
-> Make it a member function only if it fundamentally depends on the object’s invariants or representation.
+!alert note title=Core rule
+Make it a non-static member function only if it fundamentally depends on the object's invariants or representation.
```
lindsayad: I would not use such big works in this sentence. I had to ask ChatGPT what invariants means. Maybe something like "fundamentally depends on object instance data"

framework/doc/content/sqa/framework_scs.md
```diff
@@ -257,88 +243,129 @@ All non-system includes should use quotes with a single space between `include`
...
 
-Define a member when the operation:
+### Use a **non-static member function** when the operation:
 
 - Maintains or relies on class invariants
```
lindsayad: again I would not use the word invariants

framework/doc/content/sqa/framework_scs.md
```diff
@@ -257,88 +243,129 @@ All non-system includes should use quotes with a single space between `include`
...
+  ``` cpp
   class Matrix
   {
   public:
     void resize(int m, int n); // must keep internal consistency
   };
   ```
 
-  If the function mutates internal state, must preserve class invariants, or requires intimate knowledge of representation, make it a member function.
+  If the function mutates internal state, must preserve invariants, or requires intimate knowledge of representation, it should be a non-static member.
```
lindsayad: get rid of invariants. Alternatively instead of replacing invariants everywhere you could give a very clear definition of invariants at the top of this section

framework/doc/content/sqa/framework_scs.md
```diff
@@ -257,88 +243,129 @@ All non-system includes should use quotes with a single space between `include`
...
 
-  ```C++
+  ``` cpp
   dot(a, b);
   distance(p, q);
   misorientation(g1, g2);
   ```
 
-  If no operand is conceptually dominant: Avoid `a.dot(b)` and prefer `dot(a, b)`.
+  If no operand is conceptually dominant, avoid `a.dot(b)`.
```
lindsayad: If no operand is conceptually dominant, e.g. avoid `a.dot(b)`.

framework/doc/content/sqa/framework_scs.md
```diff
@@ -257,88 +243,129 @@ All non-system includes should use quotes with a single space between `include`
...
 
-  ```C++
+  ``` cpp
   norm(v);
   det(A);
   project(u, v);
   ```
 
-  If it takes an object as input, produces a value, and does not modify the object, make it a free function. This aligns with the STL design: containers own storage; algorithms are free.
+  If it takes objects as inputs, produces a value, and does not manage invariants, prefer a free function (STL style).
```
lindsayad: ~invariants~

framework/doc/content/sqa/framework_scs.md
```diff
@@ -257,88 +243,129 @@ All non-system includes should use quotes with a single space between `include`
...
 
-  ```C++
-  template<class T>
+  ``` cpp
+  template <class T>
   auto norm(const T & x);
   ```
 
-  Free functions allow generic programming, customization via argument-dependent lookup, extension to foreign types. Member functions cannot be added retroactively.
+  Free functions enable generic programming, ADL customization, and retroactive extension to foreign types.
```
lindsayad: Do define acronym ADL if it wasn't earlier

framework/doc/content/sqa/framework_scs.md
```diff
@@ -257,88 +243,129 @@ All non-system includes should use quotes with a single space between `include`
...
 
-In summary, when making a decision, ask these questions in order:
+### Decision checklist
 
-1. Does this operation mutate the object or enforce invariants? Yes → member
-2. Is it conceptually “owned” by the object? Yes → member
-3. Is the operation symmetric or algorithmic? Yes → free function
-4. Do I want/expect others to extend this behavior for new types? Yes → free function
-5. If you’re unsure → default to free function.
+1.  Does it mutate the object or enforce invariants? Yes → non-static member
```
lindsayad: ~invariants~

framework/doc/content/sqa/framework_scs.md
```diff
@@ -377,25 +404,36 @@ In summary:
...
+- **`auto` and template deduction** to reduce redundancy
+- **Explicit documentation of template requirements**: Use `static_assert`, type traits, and clear comments to state what a template expects and guarantees.
+- **Encapsulation of template complexity**: Isolate nontrivial metaprogramming behind small, stable interfaces.
+
+Be cautious about
 
-- Templates used only to avoid writing a small amount of code
-- Deep or obscure template metaprogramming that obscures intent
-- Encoding runtime variability as template parameters without a clear benefit
+- **Templates used only to avoid small amounts of duplication**, without a clear gain
```
lindsayad: and with decreased readability

lindsayad (changes requested):

hugary1995:
When I view the page rendering, I feel like the level-2 heading (i.e. `##`) and level-3 heading (i.e. `###`) are hardly distinguishable, at least in my browser. Though that's a separate issue...

framework/doc/content/sqa/framework_scs.md
```diff
@@ -153,12 +243,215 @@ All non-system includes should use quotes with a single space between `include`
...
+  ``` cpp
+  intersect(mesh, plane);
+  assemble(system, element);
+  ```
+
+  These don't belong to either type alone.
+
+### Decision checklist
+
+1.  Does it mutate the object or enforce consistency of internal state? Yes → non-static member
```
lindsayad: Final nit-pick: "enforce consistency of internal state" confuses me for a couple reasons. The language kind of sounds like state (data) stays consistent, remains unchanged; consequently it almost implies it's a `const` member function. But obviously in our `Matrix::resize` example we are going to be *changing* the internal state most likely (changing stuff like `_m, _n` potentially).

The seeming converse of this language is that a free-function would not enforce consistency of an object's internal state, but a free-function generally is not going to have any direct access to internal state, and so would not have any capability to make an object's internal data inconsistent. And again the (transient/temporal) consistency of data is more determined by things like `const` or non-`const`.

Simpler language like "does the function need access to object's internal data/state" would be better IMO. Does that fit what you're going for? I think the main issue is that consistent can have two meanings here: consistent wrt program execution (arrow in "time") or consistent wrt other object data

hugary1995: Yeah, I agree the current language could be misleading.

lindsayad (approved):

hugary1995:
@MengnanLi91 An interesting next step would be to set up agent skills parsing this document and let it review PRs on conformity to our coding standard.

MengnanLi91:
> @MengnanLi91 An interesting next step would be to set up agent skills parsing this document and let it review PRs on conformity to our coding standard.

That's a really great idea! I can give it a try locally to see how it performs and update here

joshuahansel (approved):

MengnanLi91:
@hugary1995 I converted this into an agent skill and tested it with PR review. It works but the test PR is simple so it doesn't catch a lot. @dschwen also tried agent skill writing object documentation in claude code and it also works. I think we should start accumulating a MOOSE skill set and let the community contribute as well. Any suggestions on the approach to do this?
Should we do this in MOOSE or a separate repo? There is concern about skills safety. I prefer using a separate repo and add skill template, some safety check and guidance

hugary1995:
> Should we do this in MOOSE or a separate repo?

If the skills are only for MOOSE-related tasks, I'd say they should live inside the main MOOSE repo.

> There is concern about skills safety. I prefer using a separate repo and add skill template, some safety check and guidance

I wonder what safety issue one might run into. The markdown files are not executables nor scripts, so there's no risk of autorun, apparently. If a user uses some external service (in this context claude, codex, etc.) to parse the markdown files, then when running the external services they must have already been warned about and assumed the risks. Also I think most skill servers nowadays ask for permission when they discover a new skill for the first time, so that's an additional layer of safety measure.

Past all that, the skill markdowns are just normal markdowns, I think. I could be missing something though, I'll keep thinking.

hugary1995:
> I prefer using a separate repo and add skill template, some safety check and guidance

I'm not against that. What you proposed isn't a bad idea at all. If we maintain this as a submodule, then people can opt-in.

MengnanLi91:
> > Should we do this in MOOSE or a separate repo?
> 
> If the skills are only for MOOSE-related tasks, I'd say they should live inside the main MOOSE repo.
> 
> > There is concern about skills safety. I prefer using a separate repo and add skill template, some safety check and guidance
> 
> I wonder what safety issue one might run into. The markdown files are not executables nor scripts, so there's no risk of autorun, apparently. If a user uses some external service (in this context claude, codex, etc.) to parse the markdown files, then when running the external services they must have already been warned about and assumed the risks. Also I think most skill servers nowadays ask for permission when they discover a new skill for the first time, so that's an additional layer of safety measure.
> 
> Past all that, the skill markdowns are just normal markdowns, I think. I could be missing something though, I'll keep thinking.

There maybe be some "code" injections hidden in the markdown files targeting LLM based on the tech blogs I read. Some of them are quite long and some skills come with reference documents. There is some risk if user just install the skills blindly. If we do this, we need to make sure all content we release have nothing like that. There are some script to check things.

hugary1995:
Oh yea, we will definitely have to scrutinize PRs on skills.

PR #31619, opened 2025-09-26
Tidy up MooseParsed(Vector)Function

nmnobre:
## Reason
Remove deprecated functionality and fix flawed logic when determining valid inputs.

## Design
Removes deprecated `ADParsedFunction` alias and `value`, `vars` and `vals` parameters.
Simplifies logic to determine if a variable is scalar.

## Impact
Might break people's input files.

Patches

- [x] Sabertooth https://github.inl.gov/ncrc/sabertooth/pull/472
- [x] SAM - re-running
- [x] Marmot https://github.inl.gov/ncrc/marmot/pull/952
- [x] Pronghorn https://github.inl.gov/ncrc/pronghorn/pull/1316
- [x] Mastodon https://github.com/idaholab/mastodon/pull/467
- [x] Falcon https://github.com/idaholab/falcon/pull/164

PR #32137, opened 2025-12-20
Universally adopt ParsedFunction's expression, symbol_names and symbol_values params

nmnobre:
Assuming all tests pass, this should be a no-brainer @GiudGiud.

## Reason
Spin-off of #31619 that does _not_ remove deprecated functionality, but simply drops its usage.
This covers the vast majority of the commits and file changes in #31619 and is not dependent on changes to 3rd party apps.

## Design
Replaced usages of `ParsedFunction`'s `value`, `vars` and `vals` parameters with `expression`, `symbol_names` and `symbol_values`, respectively, throughout the test suite and documentation.

## Impact
Hopefully stops the propagation of deprecated functionality that sometimes happens when people copy old test input files.
Relieves #31619 of most of its weight, leaving only the breaking changes there.


GiudGiud (approved):
yeah this is fine for sure.
Sorry I can't make progress on the other one as I don't have access to the closed apps while out of the country

GiudGiud:
Thanks for the cleanup!

PR #32141, opened 2025-12-20
Add support for extruding polygons

### LINKED ISSUE #32158: Enable PNG output to plot 3D regions
### Motivation

PNG output is the only one that can handle polys without turning them into tets first
As such it's convenient to output a field on the poly mesh

But for polyhedra, being limited to the XY plane means translating and rotating the mesh at initialization, which is just too much. It would be easier to get PNG output to work better in 3D

### Design

Add more parameters for sampling on an arbitrary rectangular region with PNG output

### Impact

Easier to work with PNG output
Easier to output polyhedral results until exodus is supported directly

GiudGiud:
Four PRs in one:
Extruding polygons
closes #32140
Improving PNG output
closes #32158
Adding capability to split polys into tets following their tetrahedralization (useful for looking at polys in 3D in exodus)
refs #32159
Adding capability to create polygons and polyhedra in a mesh generator
refs #32159

early PR in case anyone is looking for this now. 
Going to make a PR to libmesh to get a MG producing polygons 
EDIT: done & approved
Made another PR in libmesh for:
- polygon prism tetrahedralization done
- polyhedra side-to-tetrahedralization routine done

Re-based on top of the curve extrusion PR, so no need to review that one until the other gets in


framework/src/outputs/png/PNGOutput.C
```diff
@@ -56,17 +89,19 @@ PNGOutput::PNGOutput(const InputParameters & parameters)
     _transparency(getParam<Real>("transparency")),
     _nl_sys_num(libMesh::invalid_uint),
     _variable(getParam<VariableName>("variable")),
-    _max(getParam<Real>("max")),
-    _min(getParam<Real>("min")),
+    _max(isParamValid("max") ? getParam<Real>("max") : 1),
+    _min(isParamValid("min") ? getParam<Real>("min") : 0),
```
roystgnr: Should we be using NaN or Inf or something instead of "1" and "0" as our "automatic" flags here?  Otherwise one of these days we're going to find someone who really does want to scale their output from 0.0 to 1.0 and is confused to find they can't.

framework/src/outputs/png/PNGOutput.C
```diff
@@ -56,17 +89,19 @@ PNGOutput::PNGOutput(const InputParameters & parameters)
...
     _out_bounds_shade(getParam<Real>("out_bounds_shade"))
 {
+  if (!MooseUtils::absoluteFuzzyEqual(
+          getParam<Point>("first_axis") * getParam<Point>("second_axis"), 0))
+    paramWarning("first_axis", "The two axis are not orthogonal");
 }
 
 // Funtion for making the _mesh_function object.
```
roystgnr:
```suggestion
// Function for making the _mesh_function object.
```

Long as we're here

framework/src/outputs/png/PNGOutput.C
roystgnr: No, wait, we're testing `isParamValid` later.  So why bother with `1` vs `0` if neither will get used?  I do like the thrust of this change, getting rid of "defaults" that leak into documentation but don't really get used.

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -1326,6 +1336,62 @@ AdvancedExtruderGenerator::generate()
 
             break;
           }
+          case libMesh::C0POLYGON:
+          {
+            const auto num_sides = elem->n_sides();
+            std::vector<std::shared_ptr<libMesh::Polygon>> sides;
```
roystgnr, line 1354: I don't think this works the way it's intended to.

Or if it does, that's because the libMesh polyhedra code is working more flexibly than *it* was intended to.

The reason we use `shared_ptr` for the sides input to polyhedra constructors, despite preferring `unique_ptr` basically everywhere else in libMesh, was that the shared pointers were intended to be *shared* in general, with one link from each of the two polyhedra sharing a single polygonal side.  Likewise the `bool` in `Polyhedron::_sidelinks_data` was meant to make sharing possible, by recording the orientation of each Polygon relative to the Polyhedron; if we were going to just store two `Polygon` objects for each shared polygon, we could just flip each of them as necessary to get a consistent orientation, and omit the `bool`.

I'm actually not sure what failing to share `Polygon` sides will break, though.  If they're linking to the same nodes then that's probably enough for us to construct neighbor links correctly.  There may actually be nothing breaking in the current code, just a bit of extra cost in an already-somewhat-costly element type.

I'd rather not paint ourselves into a corner where we have to deliberately support non-shared sides, though, just because now we've got a bunch of test coverage that relies on that accidental support.

Instead of a vector of sides for each element, could we put (additional shared_ptrs to) new vertical sides into something like an unordered map indexed on the current extrusion layer and a (sorted) pair of nodes, and maybe put horizontal sides into a map indexed on current extrusion layer and polygon?

roystgnr (changes requested):

framework/src/outputs/png/PNGOutput.C
GiudGiud: changed to nan

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: done

> The reason we use shared_ptr for the sides input to polyhedra constructors, despite preferring unique_ptr basically everywhere else in libMesh,

I had noticed. I should have questioned it

GiudGiud: umm that's not going to work for distributed meshes.

As in it will still keep some non-shared sides next to each other.

Let's do an error with distributed meshes for now?

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -480,7 +482,20 @@ AdvancedExtruderGenerator::generate()
   // If we're using a distributed mesh... then make sure we don't have any remote elements hanging
   // around
   if (!input->is_serial())
+  {
+    input->delete_remote_elements();
+    // This should be a no-op, and yet, it's needed for two THM tests
     mesh->delete_remote_elements();
```
GiudGiud, line 488: this is a bug fix, but deleting this line unexplainably screws things

framework/src/meshgenerators/AdvancedExtruderGenerator.C
roystgnr: That should be fine for now.  We don't yet have `Packing` specializations properly set up for polygons or polyhedra, so we can't load balance on distributed meshes yet either, which reduces their utility already.

GiudGiud: ok adding the error

GiudGiud: done

framework/src/meshgenerators/AdvancedExtruderGenerator.C
```diff
@@ -1591,6 +1593,9 @@ AdvancedExtruderGenerator::generate()
     }
   }
 
+  if (has_polygons && !input->is_serial())
+    mooseError("Distributed meshes are not supported when extruding polygons at this time.");
```
roystgnr, line 1597: We've got to come up with some vocabulary to

roystgnr (approved):

framework/src/meshgenerators/AdvancedExtruderGenerator.C
GiudGiud: hehe user does not know what a serial mesh is yet I cant put it in the error. 

and then I just did in that last commit in that other generator :)

loganharbour: Should be a documented error with an associated issue

GiudGiud: good idea I ll do this in the inevitable follow-on when i deploy this at scale on the blankie mesh

loganharbour: Sounds good. It's a good place to also describe why it's not supported before you forget

GiudGiud:
Recover failure unrelated

Valgrind testing will run later

PR #32149, opened 2025-12-23
Added passive transport capability to flow channel

joshuahansel:
This PR allows the user to specify any number of passive transport variables to `FlowChannel1Phase` and use any of the following BC components:

- `SolidWall1Phase`
- `InletMassFlowRateTemperature1Phase`
- `Outlet1Phase`

Junctions and boundary components by default will throw an error if a connected flow channel has passive transport variables; as the capability is implemented, components are white-listed.

A follow-up PR will implement the capability for `VolumeJunction1Phase`.

Refs #31847


modules/thermal_hydraulics/include/base/FlowModelSinglePhase.h
```diff
@@ -21,13 +21,21 @@ class FlowModelSinglePhase : public FlowModel1PhaseBase
...
 
+  virtual void addVariables() override;
+  virtual void addInitialConditions() override;
+  virtual void addKernels() override;
+
+  virtual std::vector<VariableName> solutionVariableNames() const override;
+  std::vector<VariableName> passiveTransportSolutionVariableNames() const
+  {
+    return _passives_times_area_names;
+  }
```
grmnptr, line 32:
```suggestion
  virtual const std::vector<VariableName> & solutionVariableNames() const override;
  const std::vector<VariableName> & passiveTransportSolutionVariableNames() const
  {
    return _passives_times_area_names;
  }
```

modules/thermal_hydraulics/src/base/FlowModelSinglePhase.C
```diff
@@ -79,10 +84,64 @@ FlowModelSinglePhase::getScalingFactorRhoEA() const
...
+void
+FlowModelSinglePhase::addVariables()
+{
+  FlowModel1PhaseBase::addVariables();
+
+  // Add passive transport variables
+  const auto scaling_factor_passives =
+      _flow_channel.isParamSetByUser("scaling_factor_passives")
+          ? _flow_channel.getParam<std::vector<Real>>("scaling_factor_passives")
+          : std::vector<Real>(_passives_times_area_names.size(), 1.0);
```
grmnptr, line 96: A check would be nice to error if the scalng factor vector size is not the same as the number of scalars.

modules/thermal_hydraulics/include/base/FlowModelSinglePhase.h
grmnptr: The second can definitely be a reference.

The first I see is assembled on call. Can we assemble it in the constructor/initial setup and then just return a reference?

modules/thermal_hydraulics/src/bcs/ADBoundaryFlux3EqnBC.C
```diff
@@ -85,6 +55,9 @@ ADBoundaryFlux3EqnBC::getIndexMapping() const
   jmap.insert(std::pair<unsigned int, unsigned int>(_rhoA_var, THMVACE1D::MASS));
   jmap.insert(std::pair<unsigned int, unsigned int>(_rhouA_var, THMVACE1D::MOMENTUM));
   jmap.insert(std::pair<unsigned int, unsigned int>(_rhoEA_var, THMVACE1D::ENERGY));
+  for (const auto i : make_range(_n_passives))
```
grmnptr, line 58: Can this index mapping be done once in the setup and then just returned by reference?
EDIT: Just noticed that this is actually called in the intial setup.

modules/thermal_hydraulics/src/components/FlowJunction1Phase.C
```diff
@@ -24,15 +24,23 @@ FlowJunction1Phase::init()
...
 
       _numerical_flux_names.push_back(comp.getNumericalFluxUserObjectName());
+
+      if (i == 0)
+        _passives_names = comp.getParam<std::vector<VariableName>>("passives_names");
+      else
+      { // check that other connections have the same passive transport variables
+        if (comp.getParam<std::vector<VariableName>>("passives_names") != _passives_names)
+          logError("All connected channels must have the same 'passives_names' parameter.");
+      }
```
grmnptr:
```suggestion
        // check that other connections have the same passive transport variables
        if (comp.getParam<std::vector<VariableName>>("passives_names") != _passives_names)
          logError("All connected channels must have the same 'passives_names' parameter.");
```

grmnptr (changes requested):
Nice!

modules/thermal_hydraulics/src/base/FlowModelSinglePhase.C
joshuahansel: There's a check in `FlowChannel1Phase` for this, but I'll add an assert here.

modules/thermal_hydraulics/include/base/FlowModelSinglePhase.h
joshuahansel: Changing the second.

For the first, it's possible but maybe a little risky. The best place would be `FlowModel::init()`, but I'd have to make sure that `solutionVariableNames()` doesn't get called before that. In this case, it's not really a performance hit since it just gets used at the component level.

grmnptr: Sounds good!

modules/thermal_hydraulics/src/base/FlowModelSinglePhase.C
grmnptr: Ahhh, okay I didn't see it.

grmnptr (approved):
Looks good to me!

PR #32153, opened 2025-12-24
Make sure we can disable different parts of the segregated solve after restart.

### LINKED ISSUE #32151: Freezing TH solutions to allow for focused scalar transport in linear PIMPLE solver
### Motivation

This would allow for users to load in an existing steady TH solution (pressure, energy, velocity fields) and then perform scalar transport only simplifying the and accelerating species only transport calculations.

### Design

One method this could be implemented is to solve for the TH fields from the exodus restart for time less than 1. Afterwards these TH variables would be skipped and the passive scalars can be solved using the 'frozen' TH fields. 

### Impact

This would allow for users to load in an existing steady TH solution (pressure, energy, velocity fields) and then perform scalar transport only simplifying the and accelerating species only transport calculations.

grmnptr:
Closes #32151


modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
         getParam<std::vector<Real>>("active_scalar_absolute_tolerance")),
     _cht(ex.parameters())
 {
+  if (!_solve_momentum && _solve_pressure)
+    paramError("solve_pressure", "Pressure correction requires solving the momentum equations.");
+  if (_solve_momentum && !_solve_pressure)
+    paramError("solve_pressure", "Solving momentum without a pressure corrector is not supported.");
```
lindsayad: Shouldn't you switch the `paramError` arguments here? Because we know the user set `solve_momentum = false` in the first case and `solve_pressure = false` in the second case whereas `solve_pressure` and `solve_momentum` could have just been from the default values respectively.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
   unsigned int simple_iteration_counter = 0;
 
   // Assign residuals to general residual vector
-  const unsigned int no_systems = _momentum_systems.size() + 1 + _has_energy_system +
-                                  _has_solid_energy_system + _active_scalar_systems.size() +
-                                  _turbulence_systems.size();
+  std::vector<std::pair<unsigned int, Real>> ns_residuals;
```
lindsayad: Need to document what the first member of this pair is

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+  std::vector<Real> ns_abs_tols;
+
+  std::vector<unsigned int> momentum_indices;
```
lindsayad: Can you use `std::size_t` and avoid casts?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+    {
+      momentum_indices.push_back(static_cast<unsigned int>(ns_residuals.size()));
+      ns_residuals.push_back(std::make_pair(0, 1.0));
+      ns_abs_tols.push_back(_momentum_absolute_tolerance);
+    }
 
-  std::vector<std::pair<unsigned int, Real>> ns_residuals(no_systems, std::make_pair(0, 1.0));
-  std::vector<Real> ns_abs_tols(_momentum_systems.size(), _momentum_absolute_tolerance);
-  ns_abs_tols.push_back(_pressure_absolute_tolerance);
+  auto pressure_index = libMesh::invalid_uint;
```
lindsayad: If the static casts and use of `unsigned int` is so that you can leverage this guy, I'd say we should define a `Moose::invalid_size_t` instead

lindsayad (approved):

modules/navier_stokes/include/executioners/LinearAssemblySegregatedSolve.h
```diff
@@ -114,6 +114,15 @@ class LinearAssemblySegregatedSolve : public SIMPLESolveBase
...
   std::vector<LinearSystem *> _systems_to_solve;
 
+  /// Flags controlling which systems are actively solved (can be used with restart to freeze flow)
+  const bool _solve_momentum;
+  const bool _solve_pressure;
+  const bool _solve_energy;
+  const bool _solve_solid_energy;
+  const bool _solve_turbulence;
+  const bool _solve_passive_scalars;
+  const bool _solve_active_scalars;
```
tanoret:
```suggestion
  /// Flags controlling which systems are solved (can be used with restart to freeze the flow)
  const bool _bool_solve_momentum;
  const bool _bool_solve_pressure;
  const bool _bool_solve_energy;
  const bool _bool_solve_solid_energy;
  const bool _bool_solve_turbulence;
  const bool _bool_solve_passive_scalars;
  const bool _bool_solve_active_scalars;
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -70,6 +70,21 @@ LinearAssemblySegregatedSolve::validParams()
...
+  params.addParam<bool>("solve_momentum", true, "Solve the momentum predictor/corrector.");
+  params.addParam<bool>("solve_pressure", true, "Solve the pressure corrector.");
+  params.addParam<bool>("solve_energy", true, "Solve the fluid energy equation.");
+  params.addParam<bool>("solve_solid_energy", true, "Solve the solid energy equation.");
+  params.addParam<bool>("solve_turbulence", true, "Solve the turbulence surrogate equations.");
+  params.addParam<bool>("solve_passive_scalars", true, "Solve passive scalar equations.");
+  params.addParam<bool>("solve_active_scalars", true, "Solve active scalar equations.");
+  params.addParamNamesToGroup("solve_momentum solve_pressure solve_energy solve_solid_energy "
+                              "solve_turbulence solve_passive_scalars solve_active_scalars",
+                              "Solve control");
```
tanoret:
```suggestion
  params.addParam<bool>("solve_momentum", true, "Solve the momentum predictor/corrector?");
  params.addParam<bool>("solve_pressure", true, "Solve the pressure corrector?");
  params.addParam<bool>("solve_energy", true, "Solve the fluid energy equation?");
  params.addParam<bool>("solve_solid_energy", true, "Solve the solid energy equation?");
  params.addParam<bool>("solve_turbulence", true, "Solve the turbulence surrogate equations?");
  params.addParam<bool>("solve_passive_scalars", true, "Solve passive scalar equations?");
  params.addParam<bool>("solve_active_scalars", true, "Solve active scalar equations?");
  params.addParamNamesToGroup("solve_momentum solve_pressure solve_energy solve_solid_energy "
                              "solve_turbulence solve_passive_scalars solve_active_scalars",
                              "Solve control");
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -92,6 +107,13 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
             : libMesh::invalid_uint),
     _solid_energy_system(
         _has_solid_energy_system ? &_problem.getLinearSystem(_solid_energy_sys_number) : nullptr),
+    _solve_momentum(getParam<bool>("solve_momentum")),
+    _solve_pressure(getParam<bool>("solve_pressure")),
+    _solve_energy(getParam<bool>("solve_energy")),
+    _solve_solid_energy(getParam<bool>("solve_solid_energy")),
+    _solve_turbulence(getParam<bool>("solve_turbulence")),
+    _solve_passive_scalars(getParam<bool>("solve_passive_scalars")),
+    _solve_active_scalars(getParam<bool>("solve_active_scalars")),
```
tanoret:
```suggestion
    _bool_solve_momentum(getParam<bool>("solve_momentum")),
    _bool_solve_pressure(getParam<bool>("solve_pressure")),
    _bool_solve_energy(getParam<bool>("solve_energy")),
    _bool_solve_solid_energy(getParam<bool>("solve_solid_energy")),
    _bool_solve_turbulence(getParam<bool>("solve_turbulence")),
    _bool_solve_passive_scalars(getParam<bool>("solve_passive_scalars")),
    _bool_solve_active_scalars(getParam<bool>("solve_active_scalars")),
```

modules/navier_stokes/include/executioners/LinearAssemblySegregatedSolve.h
tanoret: If these are booleans, let's add boolean to the name. Otherwise we don't know if this is actually the 'solve' coming out of a solver

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
         getParam<std::vector<Real>>("active_scalar_absolute_tolerance")),
     _cht(ex.parameters())
 {
+  if (!_solve_momentum && _solve_pressure)
+    paramError("solve_pressure", "Pressure correction requires solving the momentum equations.");
+  if (_solve_momentum && !_solve_pressure)
+    paramError("solve_pressure", "Solving momentum without a pressure corrector is not supported.");
+  if (_has_solid_energy_system && !_solve_energy && _solve_solid_energy)
+    paramError("solve_solid_energy",
+               "Solid energy solve cannot be enabled when the fluid energy solve is disabled.");
```
tanoret: Why not?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
tanoret:
```suggestion
  if (!_bool_solve_momentum && _bool_solve_pressure)
    paramError("solve_pressure", "Pressure correction requires solving the momentum equations.");
  if (_bool_solve_momentum && !_bool_solve_pressure)
    paramError("solve_pressure", "Solving momentum without a pressure corrector is not supported.");
```

tanoret: Also, if we want to study the convergence of pressure in flow with variable properties, it could be useful to "freeze" the flow. Not sure these checks make sense

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
...
+
   // We fetch the systems and their numbers for the momentum equations.
   for (auto system_i : index_range(_momentum_system_names))
   {
     _momentum_system_numbers.push_back(_problem.linearSysNum(_momentum_system_names[system_i]));
     _momentum_systems.push_back(&_problem.getLinearSystem(_momentum_system_numbers[system_i]));
```
tanoret: Why are we still doing this if we are not solving the momentum system?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
...
-    _systems_to_solve.push_back(_momentum_systems.back());
+    if (_solve_momentum)
```
tanoret:
```suggestion
    if (_bool_solve_momentum)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
...
+      _systems_to_solve.push_back(_momentum_systems.back());
   }
 
-  _systems_to_solve.push_back(&_pressure_system);
+  if (_solve_pressure)
```
tanoret:
```suggestion
  if (_bool_solve_pressure)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
...
+    _systems_to_solve.push_back(&_pressure_system);
 
-  if (_has_energy_system)
+  if (_has_energy_system && _solve_energy)
```
tanoret:
```suggestion
  if (_has_energy_system && _bool_solve_energy)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
...
     _systems_to_solve.push_back(_energy_system);
 
-  if (_has_solid_energy_system)
+  if (_has_solid_energy_system && _solve_solid_energy)
```
tanoret:
```suggestion
  if (_has_solid_energy_system && _bool_solve_solid_energy)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -101,20 +123,30 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
...
     _systems_to_solve.push_back(_solid_energy_system);
   // and for the turbulence surrogate equations
   if (_has_turbulence_systems)
```
tanoret:
```suggestion
  if (_has_turbulence_systems && _bool_solve_turbulence)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -134,7 +166,8 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
           _problem.linearSysNum(_passive_scalar_system_names[system_i]));
```
tanoret, line 177: Why we are doing this and the ones below when we are not solving for passive scalars?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -145,7 +178,8 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
           _problem.linearSysNum(_active_scalar_system_names[system_i]));
       _active_scalar_systems.push_back(
           &_problem.getLinearSystem(_active_scalar_system_numbers[system_i]));
```
tanoret, line 191: Why are we doing this when we are not solving for passive scalars?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -145,7 +178,8 @@ LinearAssemblySegregatedSolve::LinearAssemblySegregatedSolve(Executioner & ex)
...
-      _systems_to_solve.push_back(_active_scalar_systems.back());
+      if (_solve_active_scalars)
```
tanoret:
```suggestion
      if (_bool_solve_active_scalars)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
tanoret: Also, all the logic on how we are storing the residuals and what residual corresponds to what needs to be documented somewhere

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+  std::vector<Real> ns_abs_tols;
+
+  std::vector<unsigned int> momentum_indices;
+  if (_solve_momentum)
+    for ([[maybe_unused]] const auto system_i : index_range(_momentum_systems))
+    {
+      momentum_indices.push_back(static_cast<unsigned int>(ns_residuals.size()));
+      ns_residuals.push_back(std::make_pair(0, 1.0));
+      ns_abs_tols.push_back(_momentum_absolute_tolerance);
+    }
```
tanoret: Why are we doing this when the solve_momentum boolean is off?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
 
-  std::vector<std::pair<unsigned int, Real>> ns_residuals(no_systems, std::make_pair(0, 1.0));
-  std::vector<Real> ns_abs_tols(_momentum_systems.size(), _momentum_absolute_tolerance);
-  ns_abs_tols.push_back(_pressure_absolute_tolerance);
+  auto pressure_index = libMesh::invalid_uint;
+  if (_solve_pressure)
```
tanoret:
```suggestion
  if (_bool_solve_pressure)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+  {
+    pressure_index = static_cast<unsigned int>(ns_residuals.size());
+    ns_residuals.push_back(std::make_pair(0, 1.0));
+    ns_abs_tols.push_back(_pressure_absolute_tolerance);
+  }
 
-  // Push back energy tolerances
-  if (_has_energy_system)
+  auto energy_index = libMesh::invalid_uint;
+  if (_has_energy_system && _solve_energy)
```
tanoret:
```suggestion
  if (_has_energy_system && _bool_solve_energy)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+  {
+    energy_index = static_cast<unsigned int>(ns_residuals.size());
+    ns_residuals.push_back(std::make_pair(0, 1.0));
     ns_abs_tols.push_back(_energy_absolute_tolerance);
-  if (_has_solid_energy_system)
+  }
+
+  auto solid_energy_index = libMesh::invalid_uint;
+  if (_has_solid_energy_system && _solve_solid_energy)
```
tanoret:
```suggestion
  if (_has_solid_energy_system && _bool_solve_solid_energy)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
-      ns_abs_tols.push_back(scalar_tol);
+  }
 
-  // Push back turbulence tolerances
-  if (_has_turbulence_systems)
-    for (const auto turbulence_tol : _turbulence_absolute_tolerance)
-      ns_abs_tols.push_back(turbulence_tol);
+  std::vector<unsigned int> active_scalar_indices;
+  if (_has_active_scalar_systems && _solve_active_scalars)
+    for ([[maybe_unused]] const auto i : index_range(_active_scalar_system_names))
```
tanoret: Why would the index be unused if _solve_active_scalars boolean is on?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
-    for (const auto scalar_tol : _active_scalar_absolute_tolerance)
-      ns_abs_tols.push_back(scalar_tol);
+  }
 
-  // Push back turbulence tolerances
-  if (_has_turbulence_systems)
-    for (const auto turbulence_tol : _turbulence_absolute_tolerance)
-      ns_abs_tols.push_back(turbulence_tol);
+  std::vector<unsigned int> active_scalar_indices;
+  if (_has_active_scalar_systems && _solve_active_scalars)
```
tanoret:
```suggestion
  if (_has_active_scalar_systems && _bool_solve_active_scalars)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+    for ([[maybe_unused]] const auto i : index_range(_active_scalar_system_names))
+    {
+      active_scalar_indices.push_back(static_cast<unsigned int>(ns_residuals.size()));
+      ns_residuals.push_back(std::make_pair(0, 1.0));
+      ns_abs_tols.push_back(_active_scalar_absolute_tolerance[i]);
+    }
 
-  bool converged = false;
+  std::vector<unsigned int> turbulence_indices;
+  if (_has_turbulence_systems && _solve_turbulence)
```
tanoret:
```suggestion
  if (_has_turbulence_systems && _bool_solve_turbulence)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+    for ([[maybe_unused]] const auto i : index_range(_turbulence_system_names))
```
tanoret: Here again, why would the index be unused when solve turbulence systems is on?

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -571,31 +613,63 @@ LinearAssemblySegregatedSolve::solve()
...
+    {
+      turbulence_indices.push_back(static_cast<unsigned int>(ns_residuals.size()));
+      ns_residuals.push_back(std::make_pair(0, 1.0));
+      ns_abs_tols.push_back(_turbulence_absolute_tolerance[i]);
+    }
+
+  bool converged = ns_residuals.empty();
   // Loop until converged or hit the maximum allowed iteration number
-  if (_cht.enabled())
+  if (_cht.enabled() && _solve_energy)
```
tanoret:
```suggestion
  if (_cht.enabled() && _bool_solve_energy)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -605,28 +679,33 @@ LinearAssemblySegregatedSolve::solve()
     // We set the preconditioner/controllable parameters through petsc options. Linear
     // tolerances will be overridden within the solver. In case of a segregated momentum
     // solver, we assume that every velocity component uses the same preconditioner
-    Moose::PetscSupport::petscSetOptions(_momentum_petsc_options, solver_params);
+    if (_solve_momentum)
```
tanoret:
```suggestion
    if (_bool_solve_momentum)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -605,28 +679,33 @@ LinearAssemblySegregatedSolve::solve()
...
+      Moose::PetscSupport::petscSetOptions(_momentum_petsc_options, solver_params);
 
     // Initialize pressure gradients, after this we just reuse the last ones from each
     // iteration
-    if (simple_iteration_counter == 1)
+    if (_solve_pressure && simple_iteration_counter == 1)
```
tanoret:
```suggestion
    if (_bool_solve_pressure && simple_iteration_counter == 1)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -605,28 +679,33 @@ LinearAssemblySegregatedSolve::solve()
...
       _pressure_system.computeGradients();
 
     _console << "Iteration " << simple_iteration_counter << " Initial residual norms:" << std::endl;
 
     // Solve the momentum predictor step
-    auto momentum_residual = solveMomentumPredictor();
-    for (const auto system_i : index_range(momentum_residual))
-      ns_residuals[system_i] = momentum_residual[system_i];
+    if (_solve_momentum)
```
tanoret:
```suggestion
    if (_bool_solve_momentum)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -605,28 +679,33 @@ LinearAssemblySegregatedSolve::solve()
...
+    {
+      auto momentum_residual = solveMomentumPredictor();
+      for (const auto system_i : index_range(momentum_residual))
+        ns_residuals[momentum_indices[system_i]] = momentum_residual[system_i];
+    }
 
     // Now we correct the velocity, this function depends on the method, it differs for
     // SIMPLE/PIMPLE, this returns the pressure errors
-    ns_residuals[momentum_residual.size()] = correctVelocity(true, true, solver_params);
+    if (_solve_pressure)
```
tanoret:
```suggestion
    if (_bool_solve_pressure)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -605,28 +679,33 @@ LinearAssemblySegregatedSolve::solve()
...
+      ns_residuals[pressure_index] = correctVelocity(true, true, solver_params);
 
     // If we have an energy equation, solve it here.We assume the material properties in the
     // Navier-Stokes equations depend on temperature, therefore we can not solve for temperature
     // outside of the velocity-pressure loop
-    if (_has_energy_system)
+    if (_has_energy_system && _solve_energy)
```
tanoret:
```suggestion
    if (_has_energy_system && _bool_solve_energy)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -638,14 +717,13 @@ LinearAssemblySegregatedSolve::solve()
...
-                                _energy_l_abs_tol);
-
-        if (_has_solid_energy_system)
+        ns_residuals[energy_index] = solveAdvectedSystem(_energy_sys_number,
+                                                         *_energy_system,
+                                                         _energy_equation_relaxation,
+                                                         _energy_linear_control,
+                                                         _energy_l_abs_tol);
+
+        if (_has_solid_energy_system && _solve_solid_energy)
```
tanoret:
```suggestion
        if (_has_solid_energy_system && _bool_solve_solid_energy)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -680,33 +757,32 @@ LinearAssemblySegregatedSolve::solve()
     // If we have active scalar equations, solve them here in case they depend on temperature
     // or they affect the fluid properties such that they must be solved concurrently with
     // pressure and velocity
-    if (_has_active_scalar_systems)
+    if (_has_active_scalar_systems && _solve_active_scalars)
```
tanoret:
```suggestion
    if (_has_active_scalar_systems && _bool_solve_active_scalars)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -680,33 +757,32 @@ LinearAssemblySegregatedSolve::solve()
...
+                                *_active_scalar_systems[i],
+                                _active_scalar_equation_relaxation[i],
+                                _active_scalar_linear_control,
+                                _active_scalar_l_abs_tol);
     }
 
     // If we have turbulence equations, solve them here.
     // The turbulent viscosity depends on the value of the turbulence surrogate variables
-    if (_has_turbulence_systems)
+    if (_has_turbulence_systems && _solve_turbulence)
```
tanoret:
```suggestion
    if (_has_turbulence_systems && _bool_solve_turbulence)
```

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
```diff
@@ -725,7 +801,7 @@ LinearAssemblySegregatedSolve::solve()
   // If we have passive scalar equations, solve them here. We assume the material properties in
   // the Navier-Stokes equations do not depend on passive scalars, as they are passive, therefore
   // we solve outside of the velocity-pressure loop
-  if (_has_passive_scalar_systems && (converged || _continue_on_max_its))
+  if (_has_passive_scalar_systems && _solve_passive_scalars && (converged || _continue_on_max_its))
```
tanoret:
```suggestion
  if (_has_passive_scalar_systems && _bool_solve_passive_scalars && (converged || _continue_on_max_its))
```

modules/navier_stokes/src/executioners/SIMPLESolve.C
```diff
@@ -37,23 +37,25 @@ void
...
+  if (_solve_momentum)
+    for (const auto system : _momentum_systems)
+      checkTimeKernels(*system);
 
-  checkTimeKernels(_pressure_system);
+  if (_solve_pressure)
+    checkTimeKernels(_pressure_system);
 
-  if (_has_energy_system)
+  if (_has_energy_system && _solve_energy)
```
tanoret:
```suggestion
  if (_has_energy_system && _bool_solve_energy)
```

modules/navier_stokes/src/executioners/SIMPLESolve.C
```diff
@@ -37,23 +37,25 @@ void
...
     checkTimeKernels(*_energy_system);
 
-  if (_has_turbulence_systems)
+  if (_has_turbulence_systems && _solve_turbulence)
```
tanoret:
```suggestion
  if (_has_turbulence_systems && _bool_solve_turbulence)
```

modules/navier_stokes/src/executioners/SIMPLESolve.C
```diff
@@ -37,23 +37,25 @@ void
...
     for (const auto system : _turbulence_systems)
       checkTimeKernels(*system);
 
-  if (_has_passive_scalar_systems)
+  if (_has_passive_scalar_systems && _solve_passive_scalars)
```
tanoret:
```suggestion
  if (_has_passive_scalar_systems && _bool_solve_passive_scalars)
```

modules/navier_stokes/src/executioners/SIMPLESolve.C
```diff
@@ -37,23 +37,25 @@ void
...
     for (const auto system : _passive_scalar_systems)
       checkTimeKernels(*system);
 
-  if (_has_active_scalar_systems)
+  if (_has_active_scalar_systems && _solve_active_scalars)
```
tanoret:
```suggestion
  if (_has_active_scalar_systems && _bool_solve_active_scalars)
```

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/linear-segregated/restart-freeze/tests
```diff
@@ -0,0 +1,24 @@
+[Tests]
```
tanoret, line 1: In this test, in `base_th`, `Executioner/solve_passive_scalars=false` deactivates the scalar solve for `base_th`. If we are deactivating the solution, why is `scalar` still included in the `line_sampler`?

tanoret: Ideally, in the first test, we should solve for everything, including passive scalars. Then, in the restart test, we should be freezing everything and solving only for scalars (this is fine as we are doing it now). We should demonstrate that we reach the same solution with and without restart. So, both cases should actually test against the same gold file.

tanoret (changes requested):
Nice work! This capability is very much needed.

Main comments:
- Please use more descriptive names for the variables so that any future developer can easily understand the code
- Consider the modifications suggested in the testing system
- [Optional] The current solve logic is a bit of spaghetti code and quite cumbersome to get around. Consider breaking it up in smaller functions, where each of the solve systems, residuals, booleans, etc. are stored into separate structures. This will allow for extensibility later on

modules/navier_stokes/include/executioners/LinearAssemblySegregatedSolve.h
lindsayad: If the current names are unclear, I think we should modify them in some other way than adding a `bool` prefix. What about "should" instead of "bool"?

grmnptr: Added the should prefix! (Will push with the next commit)

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: This is a great catch, copy-paste mistake from the momentum solve by me.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: This is a great catch, copy-paste mistake from the momentum solve by me.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: This block is not hit if `_solve_momentum = false`.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: Great question! Because I forgot to protect this block with the switch.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: Because we will still load back the momentum system to get the solution. This is a restart/recover-based reload to make sure we also get back the rhie chow face fluxes.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: Because I forgot to protect these

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: Oh never mind, the system will be loaded we just dont care about it here.

grmnptr: NExt commit will fix these.

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/linear-segregated/restart-freeze/tests
grmnptr: I am testing in the first case that it is skipped. So it should stay at the initial value. Then in the second I know the analytic solution (should be constant with the inlet value) and I am testing against that.

grmnptr: I can add a third test that solves everything, but I though having the analytic value was enough.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: I thought about this. I would like to believe that SIMPLE/PIMPLE are fluid solvers and we should only use them to solve solid energy equation if there is CHT/volumetric heat exchange involved. If we have specific examples where we are only curious about the solid temperature without having a fluid temperature I am happy to change it, but I am not sure I could come up with any. If we only want to solve for solid temperature without the flow we can actually use a regular Steady/Transient solver in a multiapp setting.

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
grmnptr: For now I would like to keep momentum and pressure solves disabled or enabled together. 

If one changes the other changes as well, they are closely linked. Even when the material properties are changing. 
To see a change in pressure we will actually need to compute the momentum systems anyways with updated properties.

In the future, in another PR, I would like to add a method to disable the momentum solve and just operate only on pressure correctors. But I would like to experiment on that a little before we roll it out. This is just for loading solutions.

grmnptr: But then you could ask, why dont we just add a single parameter to load peressure and momentum togehter? And to that I would say: that is a great question, mostly because I haven't thought about that. But if this is something better I can add it.

lindsayad:
I see I was re-requested for review. I already approved 😄

grmnptr:
Ooops

modules/navier_stokes/src/executioners/LinearAssemblySegregatedSolve.C
tanoret: Yes. That is a fair point. One example could be where we just want to compute a heat exchange coefficient from CFD that goes to the solid side without needing to solve the temperature in the fluid. It is a bit far fetched though. Agree to keep this check

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/linear-segregated/restart-freeze/tests
tanoret: Somehow, yes, but the analytical solution is a constant, which is not really testing the advection/diffusion fields in detail. Maybe let's add one more test where we have a scalar profile and compare with and without restarting? That way we know that the scalar transport solution is not messed up by the restart

tanoret:
Almost all good. Check the comment on adding the third test. If you don't want to do it, that is fine with me

modules/navier_stokes/test/tests/finite_volume/ins/channel-flow/linear-segregated/restart-freeze/tests
grmnptr: Added!

grmnptr:
I thought about it and I think it was a great idea to add the other test!

tanoret (approved):
Nice work!

PR #32164, opened 2026-01-03
Allow interchangeability of int names and values in MooseEnum

nmnobre:
## Reason
It'd be beneficial if users could specify the order of variables/spaces using plain digits, e.g. `order = 3`, instead of spelled-out ordinals, e.g. `order = THIRD`.

## Design
When assigning a name string to a `MooseEnum` (like the `order` parameter) we now:
a) check its existence amongst the names of the enumerated items (just as before);
b) if that search is unsuccessful, and if the name string is plainly parseable as an integer, we check its existence amongst the values of the enumerated items;
c) if that fails and we don't allow out of range extensions, we error (as before), if do allow extensions we enumerate a new item, with the passed in name (as before), but whose value is either i) the next valid id (as before) or ii) if the name string is plainly parseable as an integer, that very integer.

~This would be enough for the intents here, but for completeness and symmetry, I changed the way we assign an integer value too. We now:
a) check its existence amongst the values of the enumerated items (just as before);
b) if that search is unsuccessful, we check the existence of a string holding that integer value amongst the names of the enumerated items;
c) if that fails and we don't allow out of range extensions, we error (as we would unconditionally before), if we do allow extensions we enumerate a new item with the given value and whose name is a string holding the integer value.~

## Impact
Users can now specify `order = 18` instead of `order = EIGHTTEENTH`. For MFEM-backed problems, the user can continue to use the spelled-out ordinals till `order = FORTYTHIRD` but is now free to ask for `order = 77`.
If the user is trying to maintain a `MooseEnum` whose names are strings holding plain integer values, things may get confusing in weird corner cases allowing out of range extensions. In MOOSE, of this kind, I can only see trivial enumerations that do not allow out of range extensions, e.g. `MooseEnum dims("1=1 2 3");`, which are unaffected.

### DIFF
--- a/framework/src/utils/MooseEnum.C
+++ b/framework/src/utils/MooseEnum.C
@@ -59,14 +59,21 @@ MooseEnum::operator=(const MooseEnumItem & item)
 void
 MooseEnum::assign(const std::string & name, const std::optional<std::string> & context)
 {
-  if (name == "")
+  if (MooseUtils::trim(name) == "")
   {
     _current = MooseEnumItem("", MooseEnumItem::INVALID_ID);
     return;
   }
 
-  std::set<MooseEnumItem>::const_iterator iter = find(name);
-  if (iter == _items.end())
+  // Check if name can be parsed cleanly into an integer
+  int value = 0;
+  bool integer = (std::istringstream(MooseUtils::trim(name)) >> value).eof();
+
+  if (auto iter = find(name); iter != _items.end())
+    _current = *iter;
+  else if (auto subs = find(value); subs != _items.end() && integer)
+    _current = *subs;
+  else
   {
     if (!_allow_out_of_range) // Are out of range values allowed?
       mooseError(context ? (*context + ":\n\n") : std::string(""),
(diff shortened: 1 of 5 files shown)


framework/src/utils/MooseEnum.C
```diff
@@ -96,15 +99,22 @@ MooseEnum::assign(int value)
...
-    mooseError("Invalid id \"",
-               value,
-               "\" in MooseEnum. Valid ids are \"",
-               Moose::stringify(getIDs()),
-               "\".");
-  else
+  if (auto iter = find(value); iter != _items.end())
     _current = *iter;
+  else if (auto subs = find(std::to_string(value)); subs != _items.end())
+    _current = *subs;
```
lindsayad: I don't know about this addition. I understand that you did this for symmetry, but the point of the `MooseEnum` is for giving a parameter named options. If the underlying names are actually just integers, then I don't think the developer should be using a `MooseEnum` as the parameter type

nmnobre: I agree, discarded the change and updated the description above.

lindsayad (approved):

PR #32165, opened 2026-01-04
Add convexity diagnostics and repair for polygons

GiudGiud:
based on #32141 , should likely wait for that one to be merged before reviewing to avoid double reviewing the same content
EDIT: previous PR merged

Both the check and the split could be in libmesh utils instead. This is a recurring situation with mesh utils tbh

### DIFF
--- a/framework/src/meshgenerators/MeshDiagnosticsGenerator.C
+++ b/framework/src/meshgenerators/MeshDiagnosticsGenerator.C
@@ -133,7 +133,7 @@ MeshDiagnosticsGenerator::MeshDiagnosticsGenerator(const InputParameters & param
       _check_element_types == "NO_CHECK" && _check_element_overlap == "NO_CHECK" &&
       _check_non_planar_sides == "NO_CHECK" && _check_non_conformal_mesh == "NO_CHECK" &&
       _check_adaptivity_non_conformality == "NO_CHECK" && _check_local_jacobian == "NO_CHECK" &&
-      _check_non_matching_edges == "NO_CHECK" && _check_polygons_convex == "NO_CHECK")
+      _check_non_matching_edges == "NO_CHECK" && _check_polygons == "NO_CHECK")
     mooseError("You need to turn on at least one diagnostic. Did you misspell a parameter?");
 }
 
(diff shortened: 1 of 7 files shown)


framework/src/meshgenerators/MeshDiagnosticsGenerator.C
```diff
@@ -1585,6 +1591,81 @@ MeshDiagnosticsGenerator::checkNonMatchingEdges(const std::unique_ptr<MeshBase>
...
+      Point base_top_dir(0, 0, 0);
+      bool nonconvex = false;
+      bool nonplanar = false;
+      for (const auto & i : make_range(n_nodes))
+      {
+        const auto n1 = elem->point(i);
+        const auto n2 = elem->point((i + 1) % n_nodes);
+        const auto n3 = elem->point((i + 2) % n_nodes);
+        // can't be const with unit
+        Point top_dir = (n2 - n1).cross(n3 - n2);
```
roystgnr, line 1616: What happens if we have near-collinearity here?  That might be something users want (to get technically-conforming versions of "non-conforming" meshes, then getting hit by roundoff error), but we'll recognize it as a non-convex polygon, which seems perhaps too harsh.  (or perhaps just harsh enough, if libMesh isn't handling almost-convex polygons appropriately, but maybe we should consider that a libMesh bug).

roystgnr: Wait, actually, libMesh *should* even be supporting really non-convex polygons.  It's polyhedra where I wimped out and added a simple algorithm that assumed convexity.  It's still good for MOOSE to test and fix non-convex polygons, for several other reasons, but IMHO non-convex-within-tolerance should be fine.

framework/src/meshgenerators/MeshRepairGenerator.C
```diff
@@ -223,3 +231,232 @@ MeshRepairGenerator::separateSubdomainsByElementType(std::unique_ptr<MeshBase> &
...
+
+      // Lambda function to determine if an element is convex
+      auto is_convex = [](const Elem * elem,
+                          std::vector<std::pair<int, Real>> & obtuse_angle_nodes) -> bool
+      {
+        mooseAssert(elem, "Should have an elem");
+        mooseAssert(obtuse_angle_nodes.empty(), "Should not be empty");
+
+        // Find the normal to the element plane
+        const auto elem_centroid = elem->true_centroid();
```
roystgnr: Should we just use `vertex_average()` here?  It's faster whenever it's good enough.

framework/src/meshgenerators/MeshRepairGenerator.C
```diff
@@ -223,3 +231,232 @@ MeshRepairGenerator::separateSubdomainsByElementType(std::unique_ptr<MeshBase> &
...
+        const auto cut2 = std::max(node_i_cut1, node_i_cut2);
+
+        // Count the number of sides on each side of the cut
+        const auto ns1 = cut2 - cut1 + 1;
+        const auto ns2 = elem->n_nodes() - ns1 + 2;
+        mooseAssert(ns1 >= 3, "Should cut at least a triangle");
+        mooseAssert(ns2 >= 3, "Should cut at least a triangle");
+
+        // Create the children and add their nodes
+        auto child1 = std::make_unique<libMesh::C0Polygon>(ns1);
```
roystgnr: Should we use `child1->inherit_data_from(*elem)` after we build it?  Make sure we get processor_id and subdomain_id copied, even if p_level and mapping_type might be irrelevant here?

roystgnr (changes requested):

framework/src/meshgenerators/MeshRepairGenerator.C
GiudGiud: didnt know about this one. lemme try

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
GiudGiud: if we support non-convex polygons we might not need this code much for now then!

framework/src/meshgenerators/MeshRepairGenerator.C
GiudGiud: I think we're fine either way, this isnt in a residual loop

framework/src/meshgenerators/MeshRepairGenerator.C
GiudGiud: done

framework/src/meshgenerators/MeshDiagnosticsGenerator.C
GiudGiud: nevermind we do need this. I am putting this ahread of an extrusion step to get convex polyhedra (polygonal prisms)

roystgnr (approved):
I still think we ought to be supporting collinear-within-tolerance "convexity" too, but it's fair to start by requiring proper convexity and expand later rather than start with borderline cases and see what that breaks elsewhere right away...

GiudGiud:
test failures unrelated

GiudGiud:
Thanks for the review!

Issue #32166, opened 2026-01-04
Add a mesh generator to project a sideset onto a surface


GiudGiud:
### Motivation

In the blanket project I'd like to impress the inner tube mesh onto the top surface, to remesh the top surface then extrude back down to fill the entire sector

### Design

New mesh generator object. Probably named ProjectSideSetOnto...Generator, to follow the usual naming convention

For the surface let's start with a level set for now. but I could see a surface mesh being of interest as well

### Impact

INL project need

GiudGiud:
Closing as addressed by https://github.com/idaholab/moose/pull/32167

PR #32167, opened 2026-01-04
Add a mesh generator to project a sideset onto a surface and mesh the projection

GiudGiud:
refs #32166


GiudGiud:
Pretty simple one here but needed for meshing the blankie too

GiudGiud:
Test failure unrelated

test/tests/meshgenerators/project_sideset_onto_levelset/cube_sides_onto_plane.i
```diff
@@ -0,0 +1,22 @@
...
+    input = 'cube'
+    old_boundary = 'top front right'
+    new_boundary = 'corner corner corner'
+  []
+  [project_onto_plane]
+    type = ProjectSideSetOntoLevelSetGenerator
+    input = 'combine_3_sides'
+    direction = '1 1 1'
+    sideset = 'corner'
+    level_set = 'x + y + z -10'
```
lindsayad:
```suggestion
    level_set = 'x + y + z - 10'
```

framework/include/meshgenerators/ProjectSideSetOntoLevelSetGenerator.h
```diff
@@ -0,0 +1,57 @@
...
+  /// The analytic level set function in the form of a string that can be parsed by FParser
+  const std::string _level_set;
+  /// Maximum distance to search for from starting nodeset to projection point
+  const Real _max_search_distance;
+  /// function parser object describing the level set
+  SymFunctionPtr _func_level_set;
+
+  /**
+   * Calculate the intersection point of a level set and a line segment defined by two
+   * points separated by the level set.
```
lindsayad, line 44: what happens if the line segment doesn't intersect the level set surface?

test/tests/meshgenerators/project_sideset_onto_levelset/cube_sides_onto_plane.i
```diff
@@ -0,0 +1,22 @@
...
+  [combine_3_sides]
+    type = RenameBoundaryGenerator
+    input = 'cube'
+    old_boundary = 'top front right'
+    new_boundary = 'corner corner corner'
+  []
+  [project_onto_plane]
+    type = ProjectSideSetOntoLevelSetGenerator
+    input = 'combine_3_sides'
+    direction = '1 1 1'
```
lindsayad, line 18: Is the projection direction generally going to be the level set surface normal?

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
```diff
@@ -0,0 +1,185 @@
...
+
+InputParameters
+ProjectSideSetOntoLevelSetGenerator::validParams()
+{
+  InputParameters params = MeshGenerator::validParams();
+  params += FunctionParserUtils<false>::validParams();
+
+  params.addRequiredParam<Point>("direction", "Projection direction");
+  params.addParam<SubdomainName>(
+      "subdomain_name", "projected", "Name of the subdomain on the projected mesh");
```
lindsayad:
```suggestion
      "subdomain_name", "projected", "Name of the subdomain of the projected surface");
```

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
```diff
@@ -0,0 +1,185 @@
...
+                        "the projection direction.");
+
+  // Source domain parameters
+  params.addRequiredParam<MeshGeneratorName>(
+      "input", "The input mesh generator in which the sideset exists.");
+  params.addRequiredParam<BoundaryName>("sideset", "The name of the sideset to project");
+
+  // Surface definition
+  params.addRequiredParam<std::string>(
+      "level_set", "Level set used to cut the mesh as a function of x, y, and z.");
```
lindsayad: The level set doesn't generally have the cut the mesh does it? For example in your test case would it be a problem if your initial mesh was a unit cube cube centered at the origin? Then there would be no intersection of that mesh with your level set plane

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
```diff
@@ -0,0 +1,185 @@
...
+      const auto end = start + _max_search_distance * _proj_dir;
+
+      new_nodes[i] = new_mesh->add_point(pointPairLevelSetInterception(start, end));
+    }
+
+    // Build the same element as the side, on the surface, using the projected nodes
+    if (side_type == libMesh::C0POLYGON)
+      mooseError("Projection of polygonal sides is not supported at this time");
+    std::unique_ptr<Elem> new_elem = Elem::build(Elem::first_order_equivalent_type(side_type));
+    for (const auto i : make_range(n_nodes))
```
lindsayad: Why the `first_order_equivalent_type`? Also when you do this you could potentially produce a bunch of dangling nodes because you added to the new mesh *all* the nodes of the original side element which could presumably be greater than first order

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
```diff
@@ -0,0 +1,185 @@
...
+    new_mesh->add_elem(std::move(new_elem));
+  }
+
+  new_mesh->subdomain_name(projection_block_id) = getParam<SubdomainName>("subdomain_name");
+  new_mesh->set_isnt_prepared();
+  return dynamic_pointer_cast<MeshBase>(new_mesh);
+}
+
+Point
+ProjectSideSetOntoLevelSetGenerator::pointPairLevelSetInterception(const Point & point1,
```
lindsayad, line 143: I can imagine a closed level set surface such as the unit sphere and then our mesh is the unit cube displaced down a few meters. And then if I decide I want to project the -z normal surface of the mesh and I pick a projection direction of -z, then my two points will always have the same level set sign and this algorithm will recurse forever not finding the intersection point

test/tests/meshgenerators/project_sideset_onto_levelset/cube_sides_onto_plane.i
GiudGiud: no does not need to be

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
GiudGiud: no i'd imagine in most cases it does not since the projection direction currently is a fixed vector

GiudGiud: with more options for how to project, I could see allowing cutting

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
GiudGiud: good point. will remove

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
GiudGiud: I ll add a counter with a failure mechanism. I just needed a non-intersecting simple plane projection but thought level set was a little more useful for the next person

framework/include/meshgenerators/ProjectSideSetOntoLevelSetGenerator.h
GiudGiud: fails
There is now an error message

framework/src/meshgenerators/ProjectSideSetOntoLevelSetGenerator.C
GiudGiud: done

test/tests/meshgenerators/project_sideset_onto_levelset/cube_sides_onto_plane.i
GiudGiud: but yeah it is coincidentally here

GiudGiud:
invalidated but dont think failures are related

lindsayad (approved):

GiudGiud:
added a minor commit to address distributed mesh failures (in the simplest way)

GiudGiud:
Thanks for the review!

PR #32168, opened 2026-01-05
Update user tutorial

loganharbour:
## Reason
User tutorial needs updates.

## Design
Update the user tutorial.

## Impact
New content.

### DIFF
--- a/tutorials/darcy_thermo_mech/doc/content/workshop/intro/inl_background.md
+++ b/tutorials/darcy_thermo_mech/doc/content/workshop/intro/inl_background.md
@@ -1,17 +1,17 @@
 # Idaho National Laboratory
 
-!style halign=center
-www.inl.gov
+### [www.inl.gov](https://inl.gov)
 
 !---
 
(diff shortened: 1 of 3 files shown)


GiudGiud (approved):
force pushed an issue number.
Let's merge this so I can capture the rest on a follow up PR

note: I ll re-add the alt text manually in that follow up

loganharbour:
Looks like we need the timeline picture.

We can get it from here https://github.com/idaholab/moose/blob/e14963d87d7c3ddbb0e186b67582a75ffda0b03b/tutorials/ictp/doc/content/ictp/media/inl_reactor_timeline.png and add to large media

GiudGiud:
ugh this still has eVinci on it
if we include this pic we have to update it every 6 months when the timeline changes. I dont mind that much, just something to keep in mind

GiudGiud:
Abdalla does not have one without eVinci. i think we should do without for now

GiudGiud:
good to go @loganharbour ?

loganharbour:
I approve of @GiudGiud's changes

PR #32171, opened 2026-01-05
Add a VPP to combine multiple VPP output

GiudGiud:
@jessecarterMOOSE if you could please take a look to see if this suffices for your need.
Code-wise it ended up being reasonable.


jessecarterMOOSE:
Ref #32063

jessecarterMOOSE:
Thanks @GiudGiud. This looks promising. 

Initial thoughts - could the fill value be user specified? Now one can't tell a true zero from an empty zero. 

Also looks like vectors will hold their original order - so for VPP that sample spatial values, node/element/qp ids need to be consistent, but I don't suppose there are any internal checks for that.

GiudGiud:
> Initial thoughts - could the fill value be user specified? Now one can't tell a true zero from an empty zero.

ok should be easy I ll do that

> Also looks like vectors will hold their original order - so for VPP that sample spatial values, node/element/qp ids need to be consistent, but I don't suppose there are any internal checks for that.

As long as you include the x, y ,z / id vectors for each VPP, the ordering is always provided as well.
Checking the ordering in a general manner seems difficult. I kind of expect the use of this VPP to aggregate data to either include VPPs of the same type OR include all the vectors so the ordering is not lost

also the element samplers use two vectors for ordering (id of the element, qp index). Seems like a lot of logic to re-order a second vector with both

jessecarterMOOSE:
> As long as you include the x, y ,z / id vectors for each VPP, the ordering is always provided as well. Checking the ordering in a general manner seems difficult. I kind of expect the use of this VPP to aggregate data to either include VPPs of the same type OR include all the vectors so the ordering is not lost

I suppose including the id's is the best way to verify the different VPP's are consistent in their ordering. Otherwise you would have to get into database merging territory (like [pandas merge](https://pandas.pydata.org/docs/reference/api/pandas.merge.html)) and that's more effort than it's worth.

GiudGiud:
maybe loading the CSV with all the index vectors we might be able to get pandas to do the re-ordering? (if needed)

jessecarterMOOSE:
> maybe loading the CSV with all the index vectors we might be able to get pandas to do the re-ordering? (if needed)

The whole point is to do this in MOOSE so I can stop using pandas :smile: 

Let's not worry about the ordering for now. We can address it later if it turns out to be an issue.

GiudGiud:
I m thinking of adding more sorting abilities to individual VPPs. If you use the same flexible ordering for each VPP, we don't need to re-sort in the combined one. 
Might be a bit of a delay there, and another PR

GiudGiud:
This should be enough for now. i ll work on adding more sorting in the samplers which should be enough to achieve sorted & combined output in another PR

framework/include/vectorpostprocessors/CombinedVectorPostprocessor.h
```diff
@@ -0,0 +1,49 @@
...
+
+#pragma once
+
+#include "GeneralVectorPostprocessor.h"
+
+/**
+ * CombinedVectorPostprocessor is a type of VectorPostprocessor that outputs the
+ * values of multiple vectors from other vectorpostprocessors.
+ */
+
```
lindsayad:
```suggestion
```

lindsayad (commented):
Aren't we supposed to be using the reporter system instead of vector postprocessors?

GiudGiud:
way more difficult of a PR with reporters imo.
With VPPs I know im dealing with `std::vector<Real>`
With reporters it could be any T or vector of T or anything

GiudGiud:
also do we have CSV output for reporters? I dont know

lindsayad:
@loganharbour @zachmprince you two are the reporter codeowners. Do you have any opinions on this?

loganharbour:
No concerns here

GiudGiud:
@lindsayad thoughts on this?
I'll refresh the tests now

framework/src/vectorpostprocessors/CombinedVectorPostprocessor.C
```diff
@@ -0,0 +1,104 @@
...
+
+  return params;
+}
+
+CombinedVectorPostprocessor::CombinedVectorPostprocessor(const InputParameters & parameters)
+  : GeneralVectorPostprocessor(parameters), _filler_value(getParam<Real>("vector_filler_value"))
+{
+  // Get all the names of the vectors to combine them
+  std::vector<VectorPostprocessorName> vpps_names(
+      getParam<std::vector<VectorPostprocessorName>>("vectorpostprocessors"));
```
lindsayad:
```suggestion
  const auto &  vpps_names =
      getParam<std::vector<VectorPostprocessorName>>("vectorpostprocessors"));
```

framework/src/vectorpostprocessors/CombinedVectorPostprocessor.C
```diff
@@ -0,0 +1,104 @@
...
+  bool vecs_from_param = isParamValid("vectors");
+  if (!vecs_from_param)
+    for (const auto i : index_range(vpps_names))
+    {
+      const auto & vpp_name = vpps_names[i];
+      const auto & vpp = _vpp_fe_problem.getVectorPostprocessorObjectByName(vpp_name);
+      for (const auto & vec_name : vpp.getVectorNames())
+        vpp_vectors[i].push_back(vec_name);
+    }
+  if (vecs_from_param)
```
lindsayad:
```suggestion
  else
```

framework/src/vectorpostprocessors/CombinedVectorPostprocessor.C
```diff
@@ -0,0 +1,104 @@
...
+CombinedVectorPostprocessor::initialize()
+{
+  for (auto & vec : _vpp_vecs)
+    vec->clear();
+}
+
+void
+CombinedVectorPostprocessor::execute()
+{
+  unsigned long max_size = 0;
```
lindsayad:
```suggestion
  std::size_t max_size = 0;
```

test/tests/vectorpostprocessors/combined_vpp/test.i
lindsayad, line 1: you can apply formatting to this file 😄

GiudGiud: done

lindsayad (approved):

GiudGiud:
HPC test failure unrelated.
Thanks for the review!

Issue #32178, opened 2026-01-06
Reduce heap allocation in solid-mechanics plasticity


WilkAndy:
### Motivation

Profiling simulations reveals that about 20% of compute time is spent on allocating and deallocating heap memory when `ComputeMultipleInelasticStress` is used for the constitutive model.  I want to speed up MOOSE. 

### Design

Use class variables instead of heap variables in `ComputeMultipleInelasticStress` , and potentially in related classes.  Possibly the culprits are `CappedMohrCoulombStressUpdate` and `CappedWeakPlaneStressUpdate`, or maybe even the associated UserObjects, but i haven't checked this yet.

### Impact

Speed up MOOSE in these solid-mechanics simulations.  Perhaps as much as 20% speedup.

PR #32180, opened 2026-01-07
Pre-allocate computation arrays for plasticity

WilkAndy:
Created some stack variables to replace heap variables in plasticity algorithms.  A number of other heap variables should be put on stack to close the issue: this is a first-pass only.  Refs #32178

### DIFF
--- a/modules/solid_mechanics/include/materials/MultiParameterPlasticityStressUpdate.h
+++ b/modules/solid_mechanics/include/materials/MultiParameterPlasticityStressUpdate.h
@@ -121,7 +121,7 @@ protected:
     return TangentCalculationMethod::FULL;
   }
 
-  /// Internal dimensionality of tensors (currently this is 3 throughout tensor_mechanics)
+  /// Internal dimensionality of tensors (currently this is 3 throughout solid mechanics)
   constexpr static unsigned _tensor_dimensionality = 3;
 
   /// Number of stress parameters
@@ -226,6 +226,19 @@ protected:
         d2g(num_var, std::vector<Real>(num_var, 0.0)),
         d2g_di(num_var, std::vector<Real>(num_intnl, 0.0))
     {
+    }
+
+    // Zero everything without resizing/reallocating
+    void reset()
+    {
+      f = 0.0;
+      std::fill(df.begin(), df.end(), 0.0);
+      std::fill(df_di.begin(), df_di.end(), 0.0);
+      std::fill(dg.begin(), dg.end(), 0.0);
+      for (auto & row : d2g)
+        std::fill(row.begin(), row.end(), 0.0);
+      for (auto & row : d2g_di)
+        std::fill(row.begin(), row.end(), 0.0);
     }
 
     // this may be involved in the smoothing of a group of yield functions
@@ -566,13 +579,12 @@ protected:
    * @param inelastic_strain_increment[out] The inelastic strain increment resulting from this
    * return-map
    */
-  virtual void
-  setInelasticStrainIncrementAfterReturn(const RankTwoTensor & stress_trial,
-                                         Real gaE,
-                                         const yieldAndFlow & smoothed_q,
-                                         const RankFourTensor & elasticity_tensor,
-                                         const RankTwoTensor & returned_stress,
-                                         RankTwoTensor & inelastic_strain_increment) const;
+  virtual void setInelasticStrainIncrementAfterReturn(const RankTwoTensor & stress_trial,
+                                                      Real gaE,
+                                                      const yieldAndFlow & smoothed_q,
+                                                      const RankFourTensor & elasticity_tensor,
+                                                      const RankTwoTensor & returned_stress,
+                                                      RankTwoTensor & inelastic_strain_increment);
 
   /**
    * Calculates the consistent tangent operator.
@@ -608,18 +620,39 @@ protected:
 
   /**
    * d(stress_param[i])/d(stress) at given stress
+   * TODO: remove this function when Blackbear and others have implemented the void version
    * @param stress stress tensor
    * @return d(stress_param[:])/d(stress)
    */
-  virtual std::vector<RankTwoTensor> dstress_param_dstress(const RankTwoTensor & stress) const = 0;
+  virtual std::vector<RankTwoTensor> dstress_param_dstress(const RankTwoTensor & stress) const;
+
+  /**
+   * d(stress_param[i])/d(stress) at given stress.  When overriding, ensure to assert that dsp has
+   * size _num_sp
+   * TODO: when Blackbear and others have this implemented, make this pure virtual
+   * @param stress[in] stress tensor
+   * @param dsp[out] d(stress_param[:])/d(stress)
+   */
+  virtual void dstressparam_dstress(const RankTwoTensor & stress,
+                                    std::vector<RankTwoTensor> & dsp) const;
 
   /**
    * d2(stress_param[i])/d(stress)/d(stress) at given stress
+   * TODO: remove this function when Blackbear and others have implemented the void version
    * @param stress stress tensor
    * @return d2(stress_param[:])/d(stress)/d(stress)
    */
-  virtual std::vector<RankFourTensor>
-  d2stress_param_dstress(const RankTwoTensor & stress) const = 0;
+  virtual std::vector<RankFourTensor> d2stress_param_dstress(const RankTwoTensor & stress) const;
+
+  /**
+   * d2(stress_param[i])/d(stress)/d(stress) at given stress.  When overriding, ensure to assert
+   * that d2sp has size _num_sp
+   * TODO: when Blackbear and others have this implemented, make this pure virtual
+   * @param stress[in] stress tensor
+   * @param d2sp[out] d2(stress_param[:])/d(stress)/d(stress)
+   */
+  virtual void d2stressparam_dstress(const RankTwoTensor & stress,
+                                     std::vector<RankFourTensor> & d2sp) const;
 
   /// Sets _Eij and _En and _Cij
   virtual void setEffectiveElasticity(const RankFourTensor & Eijkl) = 0;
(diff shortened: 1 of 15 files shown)


WilkAndy:
Hi @loganharbour .  While at INL, we did some profiling and found that heap allocation/deallocation was taking about 20% of the compute time in plasticity models.  This is my attempt at addressing a bit of this problem (by making class variables).  Could you please look and comment?   There are probably better ways of doing this.  I'm worried about:

- any problems with threading and these mutable variables?
- i have no idea how to change mastodon/blackbear at the same time (it fails to compile with this change)
- there are obviously >=1 Internal app that also fails to compile because i've changed the signature of `dstress_param_dstress`

modules/solid_mechanics/include/materials/MultiParameterPlasticityStressUpdate.h
```diff
@@ -609,9 +609,10 @@ class MultiParameterPlasticityStressUpdate : public StressUpdateBase
   /**
    * d(stress_param[i])/d(stress) at given stress
    * @param stress stress tensor
-   * @return d(stress_param[:])/d(stress)
+   * @param dsp d(stress_param[:])/d(stress).  Caller must ensure this has size _num_sp
```
loganharbour: You should add this size check as an assertion in the function definition.

loganharbour (commented):
The mutable is mostly okay.

Do the functions that call it need to be const? If not, it's probably to just make the functions non-const so that it's clear that they change state. And add in the docstring that they changed that shared state.

For mastodon - you can actually re-add the old methods, and just have them call your new function (returning a copy of tensor). We can merge this, fix mastodon, and then remove the old methods.

WilkAndy:
Hi @loganharbour , thank you for your input so far.  Can you have another look and tell me whether this approach is good?   If so, i'll make a bunch of other similar changes.  I suspect not only mastodon/blackbear will need modification - there might be closed-source apps that need small modification too.

Perhaps mutable+const is actually better, as "const" tells the reader "this isn't changing class members" which is reasonable conceptually for the dstressparam_dstress function, which doesn't change important things like stress.   But then mutable is needed, and i don't see many mutable things in MOOSE.

loganharbour:
>  I suspect not only mastodon/blackbear will need modification - there might be closed-source apps that need small modification too.

Yeah, because of this I think it's best that we keep the old functions around (in their non-optimized state) so that we can cleanly patch them and eventually remove the old functions in MOOSE once the apps are patched.

WilkAndy:
OK, thanks @loganharbour .  Today at some stage, i'm going to go ahead and change more similar things in a similar way, assuming you think this approach (putting the "out" result as a function argument instead of a function return) is optimal.

WilkAndy:
Hi @loganharbour - could you please confirm that these mutables won't cause problems with multi-threading?

modules/solid_mechanics/include/materials/MultiParameterPlasticityStressUpdate.h
```diff
@@ -328,7 +341,7 @@ class MultiParameterPlasticityStressUpdate : public StressUpdateBase
                  const std::vector<Real> & intnl_ok,
                  std::vector<Real> & intnl,
                  std::vector<Real> & rhs,
-                 Real & linesearch_needed) const;
+                 Real & linesearch_needed);
```
GiudGiud: so the point of these mutables is usually to still mark the routine as const.
If the routine is not marked const, the attributes dont need to be mutable.

GiudGiud (commented):
I dont think these are stack variables. They are just pre-allocated, and still on the heap.

I would rename the PR to "Pre-allocate computation arrays for plasticity"

modules/solid_mechanics/include/materials/MultiParameterPlasticityStressUpdate.h
WilkAndy: Thanks @GiudGiud .   I'm going to reinstate the const here.  Do you have any other comments about this PR ?   I'm not an expert at this sort of thing, so am relying on the experience of the MOOSE team to make good code.

WilkAndy:
By the way, @loganharbour and @GiudGiud , with this PR, the walltime of residual calculations (which are the bulk of compute time in explicit time stepping) is about 0.7x what it was before this PR.  This is in realistic models.  I'm super pleased with that speedup.

GiudGiud:
Code looks good

There's quite a bit of not-covered code, so we can't really know if the changes break things
https://mooseframework.inl.gov/docs/PRs/32180/coverage_diff/solid_mechanics/

Are these routines tested in combined or in another module than solid mechanics? Or in an app?
The two main parts uncovered are TwoParameterPlasticityStressUpdate
and MultiParameterPlasticityStressUpdate the dstress_dparam parts

GiudGiud:
> with this PR, the walltime of residual calculations (which are the bulk of compute time in explicit time stepping) is about 0.7x what it was before this PR

this is really great. We should take another look at any code allocating tensors on the fly

loganharbour (approved):

WilkAndy:
> Code looks good
> 
> There's quite a bit of not-covered code, so we can't really know if the changes break things https://mooseframework.inl.gov/docs/PRs/32180/coverage_diff/solid_mechanics/
> 
> Are these routines tested in combined or in another module than solid mechanics? Or in an app? The two main parts uncovered are TwoParameterPlasticityStressUpdate and MultiParameterPlasticityStressUpdate the dstress_dparam parts

Hi @GiudGiud - thank you...  i want to address the reduced coverage.  (But not all of it, as i note that about half the "misses" are on the functions that are now depreciated, so none of solid mechanics should hit those functions.). Could you please:

1. confirm that `PetscJacobianTester` is covered by "coverage"?  I'm surprised that those misses occur, because i thought i wrote lots of tests for the d2stressparam_dstress stuff.
2. I can't remember how to build the coverage on my own machine, and couldn't find instructions on mooseframework or in discussions - could you please instruct me how to do it?

GiudGiud:
> confirm that PetscJacobianTester is covered by "coverage"? I'm surprised that those misses occur, because i thought i wrote lots of tests for the d2stressparam_dstress stuff.

Yes unless they are marked "heavy"

> I can't remember how to build the coverage on my own machine, and couldn't find instructions on mooseframework or in discussions - could you please instruct me how to do it?

I never do it that way, I just push to civet and civet takes care of generating the coverage report

GiudGiud:
Rebased to get an updated coverage report

GiudGiud:
so it looks like all the derived classes override
```
void
MultiParameterPlasticityStressUpdate::consistentTangentOperatorV(
    const RankTwoTensor & stress_trial,
    const std::vector<Real> & /*trial_stress_params*/,
    const RankTwoTensor & stress,
    const std::vector<Real> & /*stress_params*/,
    Real gaE,
    const yieldAndFlow & smoothed_q,
    const RankFourTensor & elasticity_tensor,
    bool compute_full_tangent_operator,
    const std::vector<std::vector<Real>> & dvar_dtrial,
    RankFourTensor & cto)
{
```
so that's why it is never covered by tests.
You can see the derived classes here:
https://mooseframework.inl.gov/docs/doxygen/modules/classMultiParameterPlasticityStressUpdate.html
It is the same situation for:
```
MultiParameterPlasticityStressUpdate::d2stressparam_dstress(
    const RankTwoTensor & stress, std::vector<RankFourTensor> & d2sp) const
```
but this one is ok, it's just a shim to call the old definition in the old materials that don't have this  new one

So my questions are:
- do you need the first routine `consistentTangentOperatorV` in the base class ?
- The base class does not make sense on its own right? Since it is missing a few implementations.
- Is there a derived class in the apps that derives directly from the base class (instead of the intermediate derived classes) and does not override this routine? (if you don't know I can check, I just don't know which apps I should start with)

GiudGiud:
the other uncovered is 
```
TwoParameterPlasticityStressUpdate::d2stressparam_dstress(const RankTwoTensor & stress,
                                                          std::vector<RankFourTensor> & d2sp) const
```
because `TwoParameterPlasticityStressUpdate::consistentTangentOperatorV` does not call it.
`:consistentTangentOperatorV` is the only routine calling this in the other materials.

- Do you need `TwoParameterPlasticityStressUpdate::d2stressparam_dstress` implemented then?
- Is this a miss to not call it from `TwoParameterPlasticityStressUpdate::consistentTangentOperatorV` ? (I don't know)
- None of the derived classes call this routine either? Is that tied to having only 2 parameters? Or is it coincidental?

WilkAndy:
> the other uncovered is
> 
> ```
> TwoParameterPlasticityStressUpdate::d2stressparam_dstress(const RankTwoTensor & stress,
>                                                           std::vector<RankFourTensor> & d2sp) const
> ```
> 
> because `TwoParameterPlasticityStressUpdate::consistentTangentOperatorV` does not call it. `:consistentTangentOperatorV` is the only routine calling this in the other materials.
> 
>     * Do you need `TwoParameterPlasticityStressUpdate::d2stressparam_dstress` implemented then?
> 
>     * Is this a miss to not call it from `TwoParameterPlasticityStressUpdate::consistentTangentOperatorV` ? (I don't know)
> 
>     * None of the derived classes call this routine either? Is that tied to having only 2 parameters? Or is it coincidental?

Thanks @GiudGiud for keeping this "on the boil".   Yes, it looks like i could remove `TwoParameterPlasticityStressUpdate::d2stressparam_dstress` and `TwoParameterPlasticityStressUpdate::dstressparam_dstress` based on what is currently in modules, since nothing uses those functions.  However, at the moment, i'd like to maintain that capability for someone to derive from `TwoParameterPlasticityStressUpdate` and use those functions.  If i stick with this, i clearly need to add some tests for these functions.

WilkAndy:
> so it looks like all the derived classes override
> 
> ```
> void
> MultiParameterPlasticityStressUpdate::consistentTangentOperatorV(
>     const RankTwoTensor & stress_trial,
>     const std::vector<Real> & /*trial_stress_params*/,
>     const RankTwoTensor & stress,
>     const std::vector<Real> & /*stress_params*/,
>     Real gaE,
>     const yieldAndFlow & smoothed_q,
>     const RankFourTensor & elasticity_tensor,
>     bool compute_full_tangent_operator,
>     const std::vector<std::vector<Real>> & dvar_dtrial,
>     RankFourTensor & cto)
> {
> ```
> 
> so that's why it is never covered by tests. You can see the derived classes here: https://mooseframework.inl.gov/docs/doxygen/modules/classMultiParameterPlasticityStressUpdate.html It is the same situation for:
> 
> ```
> MultiParameterPlasticityStressUpdate::d2stressparam_dstress(
>     const RankTwoTensor & stress, std::vector<RankFourTensor> & d2sp) const
> ```
> 
> but this one is ok, it's just a shim to call the old definition in the old materials that don't have this new one
> 
> So my questions are:
> 
>     * do you need the first routine `consistentTangentOperatorV` in the base class ?
> 
>     * The base class does not make sense on its own right? Since it is missing a few implementations.
> 
>     * Is there a derived class in the apps that derives directly from the base class (instead of the intermediate derived classes) and does not override this routine? (if you don't know I can check, I just don't know which apps I should start with)

Thanks, @GiudGiud , it's great to have your expertise supporting non-experts like me.  At the moment, I want to keep `MultiParameterPlasticityStressUpdate::consistentTangentOperatorV`, since someone might not like to calculate their own form for their derived class, and this allows "derivers" to just use my generic implementation and get the correct Jacobian.  If i stick with this line, I need to write some tests for this function.

GiudGiud:
This makes sense to me.
I think the easiest way to test these routines would be to create a derived test object (in solid_mechanics/test/src) that calls:
```
MultiParameterPlasticityStressUpdate::consistentTangentOperatorV(...)
```
in its implementation of `TestObject::consistentTangentOperatorV(...)`

You could for example derive from one of the existing `TwoParameterPlasticityStressUpdate`-derived classes (CappedWeakPlane or CappedDruckerPrager), and compare the results of `MultiParameterPlasticityStressUpdate::consistentTangentOperatorV`and the derived `DerivedClass::consistentTangentOperatorV` and compare the two results. 

If they are the same, the test object could return either and the gold file for the test can be the same as is used for the derived class (can even be a symbolic link to it).

Test objects have to be allowed in the `tests` file with `allow_test_objects = true`

WilkAndy:
Hi @GiudGiud - i'm happy with the coverage now.  If you're unhappy, please say so.

GiudGiud:
coverage looks great. will merge once next is unlocked. Thanks!

GiudGiud (approved):

PR #32188, opened 2026-01-07
Kokkos Array Layout Option

NamjaeChoi:
Refs #30655.

### DIFF
--- a/framework/include/kokkos/base/KokkosArray.h
+++ b/framework/include/kokkos/base/KokkosArray.h
@@ -75,19 +75,19 @@ class Array;
 template <typename T>
 struct ArrayDeepCopy
 {
-  static const bool value = false;
+  static constexpr bool value = false;
 };
 
 template <typename T, unsigned int dimension>
 struct ArrayDeepCopy<Array<T, dimension, LayoutType::LEFT>>
 {
-  static const bool value = ArrayDeepCopy<T>::value;
+  static constexpr bool value = ArrayDeepCopy<T>::value;
 };
 
 template <typename T, unsigned int dimension>
 struct ArrayDeepCopy<Array<T, dimension, LayoutType::RIGHT>>
 {
-  static const bool value = ArrayDeepCopy<T>::value;
+  static constexpr bool value = ArrayDeepCopy<T>::value;
 };
 ///@}
 
@@ -837,9 +837,9 @@ copyToDeviceInner(T & /* data */)
 {
 }
 
-template <typename T, unsigned int dimension>
+template <typename T, unsigned int dimension, LayoutType layout>
 void
-copyToDeviceInner(Array<T, dimension> & data)
+copyToDeviceInner(Array<T, dimension, layout> & data)
 {
   data.copyToDeviceNested();
 }
@@ -920,9 +920,9 @@ ArrayBase<T, dimension>::operator=(const T & scalar)
   return *this;
 }
 
-template <typename T, unsigned int dimension>
+template <typename T, unsigned int dimension, LayoutType layout>
 void
-dataStore(std::ostream & stream, Array<T, dimension> & array, void * context)
+dataStore(std::ostream & stream, Array<T, dimension, layout> & array, void * context)
 {
   using ::dataStore;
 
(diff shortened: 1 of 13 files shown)


framework/include/kokkos/base/KokkosArray.h
```diff
@@ -70,7 +79,13 @@ struct ArrayDeepCopy
 };
 
 template <typename T, unsigned int dimension>
-struct ArrayDeepCopy<Array<T, dimension>>
+struct ArrayDeepCopy<Array<T, dimension, LayoutType::LEFT>>
+{
+  static const bool value = ArrayDeepCopy<T>::value;
```
lindsayad:
```suggestion
  static constexpr bool value = ArrayDeepCopy<T>::value;
```

framework/include/kokkos/base/KokkosArray.h
```diff
@@ -70,7 +79,13 @@ struct ArrayDeepCopy
...
+};
+
+template <typename T, unsigned int dimension>
+struct ArrayDeepCopy<Array<T, dimension, LayoutType::RIGHT>>
 {
   static const bool value = ArrayDeepCopy<T>::value;
```
lindsayad:
```suggestion
  static constexpr bool value = ArrayDeepCopy<T>::value;
```

framework/include/kokkos/base/KokkosArray.h
```diff
@@ -1004,10 +1042,11 @@ dataLoad(std::istream & stream, Array<T, dimension> & array, void * context)
  * Kokkos object if it is used on device, because the reference refers to a host object and
  * therefore is not accessible on device. If storing it as a reference is required, see
  * ReferenceWrapper.
+ * @tparam layout The memory layout type
```
lindsayad, line 1045: This is not an available template parameter for this specialization

lindsayad: It's hard-coded to `LEFT`

lindsayad: Do you want this to be hard-coded?

framework/include/kokkos/base/KokkosArray.h
```diff
@@ -1186,8 +1226,8 @@ class Array<T, 1> : public ArrayBase<T, 1>
 #endif
 };
 
-template <typename T>
-class Array<T, 2> : public ArrayBase<T, 2>
+template <typename T, LayoutType layout>
+class Array<T, 2, layout> : public ArrayBase<T, 2>
```
lindsayad, line 1230: Yea here you are not hard-coding it (and you don't have any doxygen to be inconsistent with 😄 (which is fine for a specialization))

framework/include/kokkos/base/KokkosJaggedArray.h
```diff
@@ -350,15 +360,25 @@ JaggedArrayBase<T, inner, outer>::reserve(const std::array<dof_id_type, outer> &
...
-    stride *= dimension[i];
-
     _dims[idx].dim[i] = dimension[i];
-    _dims[idx].stride[i] = stride;
-  }
+
+  stride = 1;
+
+  if (_layout == LayoutType::LEFT)
+    for (unsigned int i = 0; i < inner; ++i)
```
lindsayad, line 369: Not directly related to this function, It would be nice to make `make_range` device capable so that we can use it on device.

Because this isn't a device capable function, consider using `make_range` in places where you can. Not a hard requirement; you don't have to go and edit this in all the places where you use the classic idiom

framework/include/kokkos/base/KokkosArray.h
NamjaeChoi: I'd like to not allow RIGHT layout specialization for 1D array.

framework/include/kokkos/base/KokkosJaggedArray.h
NamjaeChoi: What is the advantage?

framework/include/kokkos/base/KokkosArray.h
lindsayad: Ah, that makes sense. 🤦

framework/include/kokkos/base/KokkosJaggedArray.h
lindsayad: I think it's more idiomatic post c++11 to use ranges. Additionally it shows clear intent by giving the ability to mark the indexing variable as const

lindsayad: Finally it removes the possibility of implicit conversions when comparing the index variable in the loop condition. E.g. the type of `i` will always clearly match the type of `inner`

lindsayad (approved):

PR #32192, opened 2026-01-08
Fix BMBB distributed mesh segmentation fault and unique ID

hugary1995:
* Fixed the segmentation fault caused by empty boundary ID on a processor
* Ensured unique ID consistency for the newly added nodes

ref #21161
ref #32091

### DIFF
--- a/framework/src/meshgenerators/BreakMeshByBlockGenerator.C
+++ b/framework/src/meshgenerators/BreakMeshByBlockGenerator.C
@@ -177,7 +177,17 @@ std::unique_ptr<MeshBase>
 BreakMeshByBlockGenerator::generate()
 {
   std::unique_ptr<MeshBase> mesh = std::move(_input);
-  const auto max_node_id = mesh->max_node_id();
+
+  // Max node id is used later to generate new unique node IDs
+  // We need to make sure that max_node_id is the same on all processors so that unique IDs are
+  // consistent across processors
+  auto max_node_id = mesh->max_node_id();
+  if (!mesh->is_replicated())
+    mesh->comm().max(max_node_id);
+#if LIBMESH_ENABLE_UNIQUE_ID
+  const auto max_unique_id = mesh->parallel_max_unique_id();
+#endif
+
   BoundaryInfo & boundary_info = mesh->get_boundary_info();
 
   // Handle block restrictions
@@ -336,6 +346,10 @@ BreakMeshByBlockGenerator::generate()
                                            : (current_elem->subdomain_id() + 1) * max_node_id +
                                                  current_node->id())
                                .release();
+#if LIBMESH_ENABLE_UNIQUE_ID
+                new_node->set_unique_id((current_elem->subdomain_id() + 1) * max_unique_id +
+                                        current_node->unique_id());
+#endif
                 // We're duplicating nodes so that each subdomain elem has its own copy, so it
                 // seems natural to assign this new node the same proc id as corresponding
                 // subdomain elem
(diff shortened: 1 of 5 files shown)


GiudGiud (approved):
fix looks good

added parallel distributed testing to the PR testing

hugary1995:
Don't merge yet. I have only fixed the failure related to BMBB distributed mesh. There's still one failure related to EqualValueBoundaryConstraint which @ChengHauYang will take a look at.

roystgnr:
There's a bug in https://github.com/libMesh/libmesh/pull/4283 that's causing some distributed tests to fail whenever we have enough processors that one can end up without any elements.  I'll have a fix in a libMesh PR shortly, but we may want to revert the BMBB changes temporarily rather than let distributed sweeps stay broken until the next submodule update.

hugary1995:
@roystgnr 's comment above covers the distributed mesh prepare_for_use bug. There's another potential bug related to the call
```
Partitioner::set_node_processor_ids(*mesh);
```
inside BMBB which could leave some nodes with invalid processor ID. I think there's a fair chance that we will also need some libmesh patches to fix this one.

Given the two issues needing libmesh update, I'm closing this PR now. I will create a PR reverting the BMBB commits from #32091 momentarily.

roystgnr:
> I think there's a fair chance that we will also need some libmesh patches to fix this one.

Nope.  Just need a quick prep beforehand:

```
  // We want to reset node processor ids, which may need to change
  // because of elements being disconnected, but we may have
  // completely disconnected some ghosted nodes on some processors
  // from all ghosted elements that should own them, in which case
  // libMesh on those processors may not be able to tell who to query
  // for an authoritative node processor id.  Let's just delete
  // orphaned nodes so we don't end up with invalid_processor_id on
  // some processors (confusing consistency checks).
  if (!mesh->is_serial())
    mesh->remove_orphaned_nodes();
  Partitioner::set_node_processor_ids(*mesh);
```

With this plus the new libMesh ghosting functor fix, I'm seeing that test pass on 1 through 64 MPI ranks.

ChengHauYang:
Thank you, @roystgnr! I will be sure to apply your suggested fixing to BMBB. Thanks!

hugary1995:
@ChengHauYang Can you do some investigation to figure out why we have orphaned nodes in the first place?

hugary1995:
In theory, we could be given an initial mesh with orphaned nodes to begin with. We shouldn't bluntly remove orphaned nodes assuming it will be fine.

roystgnr:
> Can you do some investigation to figure out why we have orphaned nodes in the first place?

It's probably unavoidable here!

Start with `Edge2` elements {A,B,C}, with nodes 1-4: `1--A--2--B--3--C--4`, and partition it so that {A,B} are on proc 0 and C on proc 1, and ghost only the usual one layer of elements.  So proc 1 makes element A and node 1 remote.  Now add a disjoint boundary between A and B such that node 2 remains connected to A and a new node 5 takes its place in B.  Then proc 1 will still see node 2, but no longer sees any elements connected to 2; 2 has been orphaned.

That's my theory, anyway.  I wouldn't object to someone investigating to confirm or disprove it, but there's not necessarily a bug here in *need* of investigation.

> In theory, we could be given an initial mesh with orphaned nodes to begin with.

Possibly?  But you shouldn't be expected to support it.

> We shouldn't bluntly remove orphaned nodes assuming it will be fine.

If you don't, libMesh will.  Mesh preparation will drop any nodes that aren't attached to *some* element, if only a `NodeElem`.

hugary1995:
So if your theory is correct, we ought to be able to systematically collect all the orphaned remote nodes introduced by BMBB and manually delete them.

> If you don't, libMesh will. Mesh preparation will drop any nodes that aren't attached to some element, if only a NodeElem.

Hmm, that's interesting -- I didn't know that. A potential use case would be to have one mesh plus one orphaned node, and the variables defined on the mesh is somehow coupled to another variable defined on that node. Perhaps I have been thinking about it the wrong way -- in this example that floating node is connected to the `NodeElem` and so mesh preparation would not delete that node? If so, then I would agree with you on removing orphan nodes altogether in BMBB.

hugary1995:
@ChengHauYang so I guess we don't need to waste too much time on investigating this (though it would still be nice to confirm @roystgnr 's hypothesis above). That simple fix which removes all orphaned nodes should be sufficient.

roystgnr:
> and the variables defined on the mesh is somehow coupled to another variable defined on that node.

There are no variables defined on that node.  Variables are defined on elements (that's how the subdomain restriction option works, and that's how we distinguish between things like vertex nodes and mid-edge nodes, which in general have different DoF indexing), so if a node has no elements attached to it it will never get assigned a DoF index for any variable.

> in this example that floating node is connected to the `NodeElem` and so mesh preparation would not delete that node?

Precisely.

PR #32198, opened 2026-01-09
Recover boundary elements in zero algebraic range explicit NEML2

dschwen:
## Reason
In #30853 we introduced a capability to offload residual calculations to NEML2 for fast explicit dynamics simulations (`NEML2CentralDifference`). This works by clearing the algebraic range (the list of elements initialized and processed by MOOSE). However this effectively disables calculation of integrated boundary conditions.

Also, for stateful materials we should advance the state from current -> old on device, without the roundtrip via MOOSE materials (which we added to ensure correct restepping / failed timestep rollback). We're not going to evaluate the materials with a cleared out algebraic range, and this would add substantial overhead.

## Design
Use the integrated boundary condition warehouse to compile a list of all boundaries which have BCs applied. Then filter the boundary element range to get all elements touching one of those boundaries and add them back to the algebraic range.

Add an option to retain state on device to minimize communication.

## Impact
Integrated BCs are available again with NEML2 residual calculation, stateful property support for `NEML2CentralDifference`.

### DIFF
--- a/framework/src/neml2/userobjects/NEML2ModelExecutor.C
+++ b/framework/src/neml2/userobjects/NEML2ModelExecutor.C
@@ -11,11 +11,14 @@
 #include "MOOSEToNEML2.h"
 #include "NEML2Utils.h"
 #include <string>
+#include <sstream>
 
 #ifdef NEML2_ENABLED
+#include <ATen/ATen.h>
 #include "libmesh/id_types.h"
 #include "neml2/tensors/functions/jacrev.h"
 #include "neml2/dispatchers/ValueMapLoader.h"
+#include "neml2/misc/string_utils.h"
 #endif
 
 registerMooseObject("MooseApp", NEML2ModelExecutor);
@@ -31,6 +34,17 @@ NEML2ModelExecutor::actionParams()
       "List of NEML2 variables to skip error checking when setting up the model input. If an "
       "input variable is skipped, its value will stay zero. If a required input variable is "
       "not skipped, an error will be raised.");
+  params.addParam<bool>(
+      "advance_step_on_device",
+      false,
+      "Keep state and forces on the device and advance it to old_state and old_forces without a "
+      "roundtrip through MOOSE materials. This is only recommended for explicit time integration "
+      "or when absolutely no restepping occurs (e.g. failed timesteps).");
+  params.addParam<bool>(
+      "debug_inputs_on_failure",
+      false,
+      "When a NEML2 solve fails, append a detailed dump of input tensors (defined/missing, "
+      "shapes, and devices) to the error message.");
   return params;
 }
 
@@ -57,7 +71,7 @@ NEML2ModelExecutor::validParams()
   // want to execute this user object only at execute_on = LINEAR (i.e. during residual evaluation).
   // The NONLINEAR exec flag below is for computing Jacobian during automatic scaling.
   ExecFlagEnum execute_options = MooseUtils::getDefaultExecFlagEnum();
-  execute_options = {EXEC_INITIAL, EXEC_LINEAR, EXEC_NONLINEAR};
+  execute_options = {EXEC_INITIAL, EXEC_LINEAR, EXEC_NONLINEAR, EXEC_TIMESTEP_END};
   params.set<ExecFlagEnum>("execute_on") = execute_options;
 
   return params;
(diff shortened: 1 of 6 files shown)


framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -31,6 +34,17 @@ NEML2ModelExecutor::actionParams()
       "List of NEML2 variables to skip error checking when setting up the model input. If an "
       "input variable is skipped, its value will stay zero. If a required input variable is "
       "not skipped, an error will be raised.");
+  params.addParam<bool>(
+      "advance_step_on_device",
```
hugary1995: perhaps this parameter would be better named something like `keep_tensors_on_device`. I suppose you are not just storing states on devices, but also forces. Also, we need to advance the state regardless of the value of this parameter (it's just a matter who's doing it --- moose or the neml2 executor), so perhaps this could be misleading.

framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -236,6 +265,13 @@ NEML2ModelExecutor::execute()
   if (!NEML2Utils::shouldCompute(_fe_problem))
     return;
 
+  if (_current_execute_flag == EXEC_TIMESTEP_END)
```
hugary1995, line 261: if exec flag is not timestep_end, are we doing an extraneous neml2 model update?

framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -259,6 +295,48 @@ NEML2ModelExecutor::fillInputs()
...
+          continue;
+
+        if (!_skip_vars.count(var) && !var.is_state() &&
+            !(_advance_step_on_device && (var.is_old_state() || var.is_old_force())))
+          continue;
+
+        const auto & v = model().input_variable(var);
+        const auto & intmd_shape = v.intmd_sizes();
+        const auto & base_shape = v.base_sizes();
+        _in[var] = neml2::Tensor::zeros(dynamic_shape, intmd_shape, base_shape, options);
```
hugary1995: I am not sure if we need to figure out the dynamic and intermediate shapes of the zeros tensor. I think you can find the variable by querying the neml2 model with its name, and call zero() on it

hugary1995 (changes requested):

framework/src/neml2/userobjects/NEML2ModelExecutor.C
dschwen: I'm afraid I don't understand. If exec flag is not timestep_end we do a normal solve like we always did.

hugary1995: uhh nvmd... apparently reviewing on phone screen limits my ability to comprehend.

framework/src/neml2/userobjects/NEML2ModelExecutor.C
hugary1995: Correction: zeros() not zero().

framework/src/neml2/userobjects/NEML2ModelExecutor.C
dschwen: changed

framework/src/neml2/userobjects/NEML2ModelExecutor.C
dschwen: Hm I tried [this](https://github.com/idaholab/moose/compare/65a83dfeafbb3f20d36851509128501b0281b739..01fac26922f5db73618cabb3faf4643a20330102), but now I get 

```
too many indices for tensor of dimension 1
```

hugary1995: I've checked out your branch. How do I reproduce this error?

hugary1995: Alright, I can reproduce it using the test you added. I'm looking into it now.

hugary1995: I pushed a fix.

framework/src/neml2/userobjects/NEML2ModelExecutor.C
```diff
@@ -176,6 +194,13 @@ NEML2ModelExecutor::initialSetup()
   }
 }
 
+void
+NEML2ModelExecutor::timestepSetup()
+{
+  if (_keep_tensors_on_device)
+    return;
+}
```
hugary1995: Why do we need to skip timestepSetup if keep_tensors_on_device=true?

dschwen: Ugh, refactoring artifact.

hugary1995 (approved):

PR #32199, opened 2026-01-09
Revert "Merge pull request #32091 from ChengHauYang/break_mesh"

hugary1995:
This reverts commit e350e4252f49e4a0015ee336254c9e8eba470407, reversing changes made to 2c1be390f5ea10cbad2cb74e2518abb47afd7292.

See also #32192 for the reason of the revert


hugary1995:
@GiudGiud

GiudGiud (approved):

GiudGiud:
Thanks for the quick turnaround on this!

mlesueur:
Hi @hugary1995,

Following this PR, we got a problem in REDBACK where our interface kernel implementing darcy's law for pressure discontinuity across the interface (https://github.com/RedbackProject/redback/blob/master/src/interfacekernels/InterfaceDarcy.C), which was running normally, is now failing its test against an analytical solution.(https://github.com/RedbackProject/redback/blob/master/tests/interface_flow/interface_darcy_test.i).

I can see that there was a lot of changes to BreakMeshByBlockGenerator in this PR, but I still couldn't figure out what is the problem. Our test is running but gives a result as if a different value of permeability is applied. The difference between the two outputs seems not to be systematic like a simple ratio. I don't know if this is the right platform to raise this issue but I thought this could have been relevant if this is a problem stemming from the PR.

Thanks in advance for your help,

PR #32201, opened 2026-01-09
Generalize StochasticControl to work with vectorpostprocessor return values

### LINKED ISSUE #32196: Allow StochasticControl to process vector qoi's and generalize cache tolerance
### Motivation

This allows direct comparison between experimental data and simulation output in global optimization and similar problems. Additionally, cache tolerance should be user-configurable on a per-parameter basis to allow parameters to vary on different scales.

### Design

Extend the internal configuration of StochasticControl.py to add capability to process return parameters with more than two dimensions in the __call__ function. Also allow the tolerance parameter of `configCache` to be a  list-like so that tolerances can be specified on a per-parameter basis. 

Tests to implement
1. Does the runner successfully run and interpret a simulation that returns a vector of reals?
2. Can the runner successfully run a simulation set that return vectors of reals of different length (internal qoi dimensions differ)?
3. Can ConfigCache handle list-like tol input?
4. Caching of a qoi vector set (array type)
5. Caching of a qoi vector set (list type) (jagged array)

### Impact

Existing API functionality should be unaffected. New functionality will sit on top of the current API.

ke7kto:
Closes #32196

## Reason
This functionality allows direct comparison of models with concentration profiles and time-series data via stochasticControl.

## Design
extend the N-th dimensionality behavior of the StochasticControl interface. Allow object format for arrays

## Impact
No existing functionality is expected to be affected.


modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
```diff
@@ -271,7 +289,15 @@ def getOutput(self) -> np.ndarray:
...
-        return np.array(y).T
+        yarr = np.array(y,dtype=object)
+        if len(yarr.shape)>3:
+            return np.transpose(yarr,axes=(1,0,2,3))
+        elif len(yarr.shape)==3:
+            return np.transpose(yarr,axes=(1,0,2))
+        elif len(yarr.shape)==2:
+            return np.transpose(yarr)
+        else:
+            raise ValueError("Quantities of interest returned unexpected shape: "+"({:s})".format(','.join(['{:f}'.format(x) for x in y.shape])))
```
zachmprince:
```suggestion
        if yarr.ndim <= 2:
            return yarr.T
        else:
            # 3-D: (1, 0, 2), 4-D: (1, 0, 2, 3)
            axes = tuple([1, 0] + list(range(2, yarr.ndim)))
            return np.transpose(yarr, axes=axes)
```

modules/stochastic_tools/python/moose_stochastic_tools/StochasticControl.py
```diff
@@ -491,7 +517,10 @@ def __init__(self, maxsize: int = 10000, tol: float = 1e-14):
                 Must be > 0.
         """
         assert maxsize > 0
-        assert tol > 0.0
+        if type(tol) is list:
```
zachmprince:
```suggestion
        if isinstance(tol, list):
```

ke7kto:
Is there a status on this @zachmprince? I can't remember whether you'd asked for more changes.

zachmprince (approved):

PR #32206, opened 2026-01-10
Add more sorting capabilities to the samplers

GiudGiud:
#31913

- sort by an arbitrary column (so you can make any sorting decision and store that in a variable, added to the list of sampled variables)
- sort by multiple columns in a nested fashion (less potent but easier to set up) [not completely functional yet]

@jessecarterMOOSE if you are mostly using samplers this should be enough for sorting before grouping the VPPs?

EDIT: removed the second option for now, it was becoming too much code. see https://github.com/idaholab/moose/commit/2e25c3ff9ec43215caaf5b8be7275ceb54dac12f for the first part, but needed more

patches:
- [x] SAM https://git-nse.egs.anl.gov/SAM/SAM/-/merge_requests/1184
- [x] Sockeye https://github.inl.gov/ncrc/sockeye/pull/891
- [x] Bison https://github.inl.gov/ncrc/bison/pull/6692
- [x] Grizzly https://github.inl.gov/ncrc/grizzly/pull/822


test/tests/vectorpostprocessors/element_value_sampler/gold/sort_xy_element_value_sampler_0000.csv
```diff
@@ -0,0 +1,5 @@
+id,u,v,x,y,z
+1.25,0.5,2.2025656884517e-314,0.25,0,0
```
GiudGiud: this is because the two-column sampling is not working yet

GiudGiud: removed

framework/src/vectorpostprocessors/SamplerBase.C
```diff
@@ -173,9 +175,31 @@ SamplerBase::finalize()
...
+  else if (_sort_by == "y")
+    sort_by_i = 1;
+  else if (_sort_by == "z")
+    sort_by_i = 2;
+  else if (_sort_by == "id")
+    sort_by_i = 3;
+  else
+  {
+    // Find in 'variable_names'
+    const auto & it = std::find(_variable_names.begin(), _variable_names.end(), _sort_by);
```
lindsayad:
```suggestion
    const auto it = std::find(_variable_names.begin(), _variable_names.end(), _sort_by);
```

framework/include/vectorpostprocessors/SamplerBase.h
```diff
@@ -116,7 +116,7 @@ class SamplerBase
   std::vector<std::string> _variable_names;
 
   /// What to sort by
-  const unsigned int _sort_by;
+  const std::string _sort_by;
```
lindsayad: I would feel better if this remained an integer so that downstream user code is not doing O(N) string comparisons for possibly every `getValue` call. I understand we cannot have an `enum` with a fixed number of enumerators, but the integer data member allows us an arbitrary number of enumerator values

GiudGiud: ok we can do that

framework/src/vectorpostprocessors/SamplerBase.C
```diff
@@ -25,8 +25,10 @@ SamplerBase::validParams()
 {
   InputParameters params = emptyInputParameters();
 
-  MooseEnum sort_options("x y z id");
-  params.addRequiredParam<MooseEnum>("sort_by", sort_options, "What to sort the samples by");
+  params.addRequiredParam<std::string>(
+      "sort_by",
+      "What to sort the samples by. Options are 'x y z id' and the name of any of the sampled "
+      "quantities (= name of the vector created by the vectorpostprocessor).");
```
joshuahansel:
```suggestion
      "What to sort the samples by. Options include 'x', 'y', 'z', 'id', and the name of any of the sampled quantities (which each create a vector of the same name).");
```

joshuahansel: You'll need to clang-format this suggestion if you take it (sorry!)

framework/src/vectorpostprocessors/SamplerBase.C
```diff
@@ -173,9 +175,31 @@ SamplerBase::finalize()
...
+  else
+  {
+    // Find in 'variable_names'
+    const auto & it = std::find(_variable_names.begin(), _variable_names.end(), _sort_by);
+    if (it != _variable_names.end())
+      sort_by_i = 4 + it - _variable_names.begin();
+    else
+      mooseError("Sorting index '" + _sort_by +
+                 "' was not found in x/y/z/id or sampled variable names: " +
+                 Moose::stringify(_variable_names));
```
joshuahansel:
```suggestion
      mooseError("The 'sort_by' parameter must be one of x/y/z/id or one of the sampled variable names: " +  Moose::stringify(_variable_names));
```

joshuahansel (changes requested):
Just minor stuff

framework/src/vectorpostprocessors/SamplerBase.C
GiudGiud: no worries I have to clang-format my own commits all the time (got sick of lag from vscode format)

framework/include/vectorpostprocessors/SamplerBase.h
GiudGiud: ok done

joshuahansel (approved):

framework/include/vectorpostprocessors/SamplerBase.h
lindsayad: It doesn't look done

GiudGiud: See the second attribute below

lindsayad: If we keep `_sort_by` the same type, there is no need to change downstream code. Either way can you make the `std::string _sort_by*` member private?

GiudGiud: downstream code is failing because of the param type change not because of the attribute change. I ll make private

GiudGiud: done

lindsayad (approved):

framework/include/vectorpostprocessors/SamplerBase.h
```diff
@@ -115,8 +115,8 @@ class SamplerBase
   /// The variable names
   std::vector<std::string> _variable_names;
 
-  /// What to sort by
-  const unsigned int _sort_by;
+  /// The index for what to sort by: x=0, y=1, z=2, then sampled variables in ordered specified in the parameter
```
bwspenc, line 118: order?
```suggestion
  /// The index for what to sort by: x=0, y=1, z=2, then sampled variables in order specified in the parameter
```

GiudGiud: order!
I added a little more explanation

framework/src/vectorpostprocessors/SamplerBase.C
```diff
@@ -25,8 +25,10 @@ SamplerBase::validParams()
 {
   InputParameters params = emptyInputParameters();
 
-  MooseEnum sort_options("x y z id");
-  params.addRequiredParam<MooseEnum>("sort_by", sort_options, "What to sort the samples by");
+  params.addRequiredParam<std::string>(
```
loganharbour, line 28: Why not keep a moose enum, with:

```
x=0 y=1 z=2 id=3
```

and then just get the ID from the current value for the index? Would rather not get rid of the enum option and have to deal with errors when it's not a valid value.

This would also then make it unnecessary to patch the apps

loganharbour: You can then get the corresponding ID (which will become the index with):

```
_sort_by_index(static_cast<int>(<the enum>))
```

to call `operator int` to get the ID

GiudGiud:
Resolved the app failures a different way. Thanks for reviews!

PR #32213, opened 2026-01-13
Support index type selection and optimize index calculation in Kokkos arrays

NamjaeChoi:
Refs #30655.

Users can choose the index type instead of being stuck at 8-byte integers. 4-byte indices have significant performance advantage over 8-byte indices on GPUs. Sorry for a messy change, but this seems to improve the performance of our application by ~25%.

### DIFF
--- a/framework/include/kokkos/base/KokkosArray.h
+++ b/framework/include/kokkos/base/KokkosArray.h
@@ -16,18 +16,20 @@
 #include "Conversion.h"
 #include "DataIO.h"
 
-#define usingKokkosArrayBaseMembers(T, dimension)                                                  \
+#define usingKokkosArrayBaseMembers(T, dimension, index_type)                                      \
 private:                                                                                           \
-  using ArrayBase<T, dimension>::_n;                                                               \
-  using ArrayBase<T, dimension>::_s;                                                               \
-  using ArrayBase<T, dimension>::_d;                                                               \
+  using ArrayBase<T, dimension, index_type>::_n;                                                   \
+  using ArrayBase<T, dimension, index_type>::_s;                                                   \
+  using ArrayBase<T, dimension, index_type>::_d;                                                   \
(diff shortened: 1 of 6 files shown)


framework/include/kokkos/base/KokkosArray.h
```diff
@@ -367,15 +372,19 @@ class ArrayBase
...
+  index_type _s[dimension] = {0};
   /**
    * Offset of each dimension
    */
-  dof_id_signed_type _d[dimension] = {0};
+  signed_index_type _d[dimension] = {0};
+  /**
+   * Flag whether the array indices are offset
+   */
+  bool _is_offset = false;
```
lindsayad, line 387: why this new member?

framework/include/kokkos/base/KokkosJaggedArray.h
```diff
@@ -26,19 +26,19 @@ struct JaggedArrayInnerDim
   /**
    * Size of each dimension
    */
-  dof_id_type dim[inner] = {0};
+  unsigned int dim[inner] = {0};
```
lindsayad, line 29: If you're templating the other one I feel like you might as well template this one? It seems incongruent to template one and not the other?

NamjaeChoi: I think inner arrays are typically small. If inner arrays have sizes larger than the unsigned int limit, the user is likely using the jagged array incorrectly. The outer array index type is templated.

framework/include/kokkos/base/KokkosArray.h
NamjaeChoi: If the array indices are not offset, a few integer operations can be reduced.

lindsayad (approved):

loganharbour:
~~This broke parallel sweeps: https://civet.inl.gov/event/275435/~~

Disregard, wrong PR

PR #32230, opened 2026-01-14
Intra-variable Jacobian

dschwen:
Closes #32212

### DIFF
--- a/framework/include/nodalkernels/ArrayNodalKernel.h
+++ b/framework/include/nodalkernels/ArrayNodalKernel.h
@@ -65,20 +65,21 @@ protected:
   virtual void computeQpResidual(RealEigenVector & residual) = 0;
 
   /**
-   * The user can override this function to compute the "on-diagonal"
-   * Jacobian contribution.  If not overriden,
-   * returns an array of 1s.
-   * @return On-diagonal Jacobian contribution
+   * The user can override this function to compute the intra-variable
+   * off-diagonal Jacobian contribution (the coupling between array
+   * components of the variable itself). Use setJacobian in this function.
    */
-  virtual RealEigenVector computeQpJacobian();
+  virtual void computeQpJacobian() {}
 
   /**
-   * This is the virtual that derived classes should override for
-   * computing an off-diagonal jacobian component.
-   * @param Coupled variable for the off-diagonal Jacobian contribution
-   * @return Off-diagonal Jacobian contribution
+   * The user can override this function to compute the inter-variable
+   * off-diagonal Jacobian contribution (the coupling between array
+   * components of of the kernel variable and the coupled variable jvar).
+   * Use setJacobian in this function.
    */
-  virtual RealEigenMatrix computeQpOffDiagJacobian(const MooseVariableFieldBase & jvar);
+  virtual void computeQpOffDiagJacobian(unsigned int /*jvar*/) {}
+
+  void setJacobian(unsigned int i, unsigned int j, Real value);
 
   /// variable this works on
   MooseVariableFE<RealEigenVector> & _var;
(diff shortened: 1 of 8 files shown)


dschwen:
VTB failure is unrelated

lynnmunday:
@dschwen Will you drop the second commit and get the first commit merged?  I have my generic cluster dynamics code built off your first commit and it is doing everything I want when I use 
```
[Problem]
 use_hash_table_matrix_assembly = true
[]
```

Would you add an error in the array nodal kernel or maybe its intraJacobian method if use_hash_table is false?

lynnmunday:
@dschwen will you remove the draft so this can be reviewed and merged?

lynnmunday:
Thanks @dschwen I need this for https://github.com/idaholab/blackbear/pull/434

lynnmunday:
@lindsayad Will you review this again?

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -65,20 +65,21 @@ class ArrayNodalKernel : public NodalKernelBase, public MooseVariableInterface<R
...
    */
-  virtual RealEigenVector computeQpJacobian();
+  virtual void computeQpJacobian() {}
 
   /**
-   * This is the virtual that derived classes should override for
-   * computing an off-diagonal jacobian component.
-   * @param Coupled variable for the off-diagonal Jacobian contribution
-   * @return Off-diagonal Jacobian contribution
+   * The user can override this function to compute the inter-variable
```
lindsayad, line 75:
```suggestion
   * The user can override this function to compute the inter-(libMesh)variable
```

lindsayad: Coupling within array components would be *intra*-(Moose)variable

framework/include/nodalkernels/ArrayNodalKernel.h
```diff
@@ -65,20 +65,21 @@ class ArrayNodalKernel : public NodalKernelBase, public MooseVariableInterface<R
...
+   * off-diagonal Jacobian contribution (the coupling between array
+   * components of of the kernel variable and the coupled variable jvar).
+   * Use setJacobian in this function.
    */
-  virtual RealEigenMatrix computeQpOffDiagJacobian(const MooseVariableFieldBase & jvar);
+  virtual void computeQpOffDiagJacobian(unsigned int /*jvar*/) {}
+
+  void setJacobian(unsigned int i, unsigned int j, Real value);
```
lindsayad, line 88: Add doxygen for this as it appears to be critical for users to call this in their derived class hand-coded jacobian implementations

framework/include/nodalkernels/ArrayNodalKernel.h
dschwen: right, but this is inter... it's the off diagonal which reference two different MOOSE variables.

dschwen: I find it confusing to start talking about libmesh variables here. This requires to explain that extra leap, that MOOSE array variables are composed from multiple libmesh variables. And again, that is confusing, because the methods only reference MOOSE variable IDs.

lindsayad: Ah I misunderstood what you were saying here. I understand now that with your changes the conceptually matrix-sized coupling between intra-variable components is handled within `computeQpJacobian` as opposed to it formerly being only for $\partial R_c/\partial u_c$, e.g. something that could be represented as a vector

lindsayad (approved):

lynnmunday:
@dschwen will you merge.  The richards module was deleted so the coverage is wrong.

dschwen:
image

dschwen:
I've invalidated the Modules and Coverage targets. Hopefully that's enough.

lindsayad:
That message just means that merging is blocked for everyone. @loganharbour does that from time to time for various good reasons

Issue #32109, opened 2025-12-16
UserObjects don't seem to play nice with SolutionInvalidity


vincentlaboure:
### Bug Description

`flagSolutionWarning/flagInvalidSolution` seem to work fine inside a Material but if I try to put it in a UserObject, they appear to be ignored.

Changing the `execute_on` to TIMESTEP_BEGIN, TIMESTEP_END, LINEAR or NONLINEAR does not seem to fix it.

Tag @GiudGiud 

### Steps to Reproduce

In the MOOSE test suite, if I change the source code to:

```
diff --git a/test/src/userobjects/RandomElementalUserObject.C b/test/src/userobjects/RandomElementalUserObject.C
index 5a1e686a3d..e49e5fe8db 100644
--- a/test/src/userobjects/RandomElementalUserObject.C
+++ b/test/src/userobjects/RandomElementalUserObject.C
@@ -47,6 +47,9 @@ void
 RandomElementalUserObject::finalize()
 {
   _communicator.set_union(_random_data);
+  flagSolutionWarning("working??");
+  flagInvalidSolution("invalid!!!");
+  std::cout << "here" << std::endl;
 }
```

I see the std::cout printout once per time step but no SolutionInvalidity  table. Changing the execute_on doesn't help.

### Impact

Limits the generalization of SolutionInvalidity

lindsayad:
@vincentlaboure can you remind me where you work now? That will help me determine what project to file this in

PR #32231, opened 2026-01-15
Use solution invalid to tally warnings from a variety of sources

GiudGiud:
solution invalid was mainly thought for the solves, so all that code was resetting & tallying entirely in the systems. 
I moved the reset on iterations earlier to avoid resetting right after various objects on linear/nonlinear
Kept the tally on iterations where it was
Moved the tally iteration->time step after timestep_end
Moved the reset on time steps before timestep_begin

refs #32109

### DIFF
--- a/framework/include/utils/SolutionInvalidity.h
+++ b/framework/include/utils/SolutionInvalidity.h
@@ -68,6 +68,13 @@ public:
    */
   bool hasInvalidSolution() const;
 
+  /**
+   * Whether or not any warning or invalid solution has ever been encountered
+   *
+   * This must be called after a sync.
+   */
+  bool hasEverHadSolutionIssue() const;
+
   /// Reset the number of solution invalid occurrences back to zero for the current time step
   void resetSolutionInvalidTimeStep();
 
@@ -161,6 +168,8 @@ private:
   bool _has_solution_warning;
   /// Whether or not we have an invalid solution (only after a sync)
   bool _has_solution_error;
+  /// Whether or not we have ever had any warning or solution issue during the simulation
+  bool _has_recorded_issue;
 };
 
 // datastore and dataload for recover
--- a/framework/include/utils/SolutionInvalidity.h
+++ b/framework/include/utils/SolutionInvalidity.h
@@ -69,7 +69,7 @@ public:
   bool hasInvalidSolution() const;
 
   /**
-   * Whether or not any warning or invalid solution has ever been encountered
+   * Whether or not any warning or invalid solution has ever been encountered during the simulation
    *
    * This must be called after a sync.
    */
@@ -84,7 +84,7 @@ public:
   /// Pass the number of solution invalid occurrences from current iteration to cumulative counters
   void accumulateIterationIntoTimeStepOccurences();
 
-  /// Pass the number of solution invalid occurrences from current iteration to cumulative time iteration counters
+  /// Pass the number of solution invalid occurrences from current timestep to cumulative timestep counter (e.g. the total)
   void accumulateTimeStepIntoTotalOccurences(const unsigned int timestep_index);
 
   /// Struct used in InvalidCounts for storing the time history of invalid occurrences
@@ -96,12 +96,16 @@ public:
     unsigned int counts = 0;
   };
 
-  /// Struct used in _counts for storing invalid occurrences
+  /// Struct used in _counts for storing warning and invalid-solution occurrences
   struct InvalidCounts
   {
+    /// Counts for the current iteration (depends on the count, but usually linear iteration)
     unsigned int current_counts = 0;
+    /// Counts for the current time step
     unsigned int current_timestep_counts = 0;
+    /// Total counts across the entire simulation
     unsigned int total_counts = 0;
+    /// Keep track of the occurences across all time steps
     std::vector<TimestepCounts> timestep_counts;
   };
 
@@ -127,10 +131,11 @@ public:
 
   /**
    * Sync iteration counts to main processor
+   * Sum across all processors
    */
   void syncIteration();
 
-  /// Whether the solution invalidity has synchronized iteration counts
+  /// Whether the solution invalidity has synchronized iteration counts across MPI processes
   bool hasSynced() const { return _has_synced; }
 
   friend void dataStore(std::ostream &, SolutionInvalidity &, void *);
--- a/framework/include/utils/SolutionInvalidity.h
+++ b/framework/include/utils/SolutionInvalidity.h
@@ -130,6 +130,9 @@ public:
    */
   void syncIteration();
 
+  /// Whether the solution invalidity has synchronized iteration counts
+  bool hasSynced() const { return _has_synced; }
+
   friend void dataStore(std::ostream &, SolutionInvalidity &, void *);
   friend void dataLoad(std::istream &, SolutionInvalidity &, void *);
 
--- a/framework/include/utils/SolutionInvalidity.h
+++ b/framework/include/utils/SolutionInvalidity.h
@@ -87,9 +87,6 @@ public:
   /// Pass the number of solution invalid occurrences from current iteration to cumulative time iteration counters
   void solutionInvalidAccumulationTimeStep(const unsigned int timestep_index);
 
-  /// Compute the total number of solution invalid occurrences
-  void computeTotalCounts();
-
   /// Struct used in InvalidCounts for storing the time history of invalid occurrences
   struct TimestepCounts
   {
--- a/framework/include/utils/SolutionInvalidity.h
+++ b/framework/include/utils/SolutionInvalidity.h
@@ -76,16 +76,16 @@ public:
   bool hasEverHadSolutionIssue() const;
 
   /// Reset the number of solution invalid occurrences back to zero for the current time step
-  void resetSolutionInvalidTimeStep();
+  void resetTimeStepOccurences();
 
   /// Reset the number of solution invalid occurrences back to zero
-  void resetSolutionInvalidCurrentIteration();
+  void resetIterationOccurences();
 
   /// Pass the number of solution invalid occurrences from current iteration to cumulative counters
-  void solutionInvalidAccumulation();
+  void accumulateIterationIntoTimeStepOccurences();
 
   /// Pass the number of solution invalid occurrences from current iteration to cumulative time iteration counters
-  void solutionInvalidAccumulationTimeStep(const unsigned int timestep_index);
+  void accumulateTimeStepIntoTotalOccurences(const unsigned int timestep_index);
 
   /// Struct used in InvalidCounts for storing the time history of invalid occurrences
   struct TimestepCounts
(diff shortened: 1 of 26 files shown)


test/tests/misc/solution_invalid/gold/solution_invalid_checkpoint.json
```diff
@@ -50,14 +50,14 @@
...
-                        "converged_counts": 16,
+                        "converged_counts": 0,
                         "message": "The diffusivity is greater than the threshold value!",
                         "object_type": "NonsafeMaterial",
                         "timestep_counts": 48,
                         "total_counts": 96
                     },
                     {
-                        "converged_counts": 16,
+                        "converged_counts": 0,
```
GiudGiud: we lost the ability to output that one since we need to reset the counter to count the timestep_end warnings. 

Maybe I can delete this item from the JSON output?

GiudGiud: or we could figure out how to:
- dont accumulate to time step on last (converged) iteration
- instead, only accumulate to time step after the time step ends (right before TIMESTEP_END output)
- then no need to reset anymore since we only accumulating once

It's not easy. Feel free to take a look though

test/tests/misc/solution_invalid/gold/solution_invalid_outside_solve_out.json
```diff
@@ -0,0 +1,64 @@
...
+                        "message": "tend: warning",
+                        "object_type": "WarningUserObject",
+                        "timestep_counts": 2,
+                        "total_counts": 6
+                    },
+                    {
+                        "converged_counts": 2,
+                        "message": "final: warning",
+                        "object_type": "WarningUserObject",
+                        "timestep_counts": 0,
```
GiudGiud: right now there is no accumulation from the iteration count into the time step count on FINAL. 
I think this is fine since final is not really a time step. 
However, for other outputs (exodus, csv), FINAL is effectively its own time step. 

Since the calls to perform stuff on FINAL are in the xecutioners (all of them) rather than the shared fixed point solve, we would need to modify all of the executioners if we wanted to have this timestep_counts represent the final step.

if we do that, we then also lose the timestep_counts for all the other categories in this case (because their time step count for FINAL is 0, it's not 0 for the step before that).

So imo it's best the way it is now.

framework/include/utils/SolutionInvalidity.h
```diff
@@ -68,20 +68,24 @@ class SolutionInvalidity : protected ConsoleStreamInterface, public libMesh::Par
    */
   bool hasInvalidSolution() const;
 
+  /**
+   * Whether or not any warning or invalid solution has ever been encountered
+   *
+   * This must be called after a sync.
+   */
+  bool hasEverHadSolutionIssue() const;
```
lindsayad, line 76: For the entirety of the simulation?

framework/include/utils/SolutionInvalidity.h
```diff
@@ -68,20 +68,24 @@ class SolutionInvalidity : protected ConsoleStreamInterface, public libMesh::Par
...
 
   /// Reset the number of solution invalid occurrences back to zero
-  void resetSolutionInvalidCurrentIteration();
+  void resetIterationOccurences();
 
   /// Pass the number of solution invalid occurrences from current iteration to cumulative counters
-  void solutionInvalidAccumulation();
+  void accumulateIterationIntoTimeStepOccurences();
 
   /// Pass the number of solution invalid occurrences from current iteration to cumulative time iteration counters
```
lindsayad:
```suggestion
  /// Pass the number of solution invalid occurrences from current timestep to cumulative timestep counter (e.g. the total)
```
If this isn't correct, then I think the API name could use a little massaging

framework/include/utils/SolutionInvalidity.h
```diff
@@ -126,6 +130,9 @@ class SolutionInvalidity : protected ConsoleStreamInterface, public libMesh::Par
    */
   void syncIteration();
 
+  /// Whether the solution invalidity has synchronized iteration counts
```
lindsayad: synchronized across what? Threads? MPI ranks?

framework/include/utils/SolutionInvalidity.h
```diff
@@ -161,6 +168,8 @@ class SolutionInvalidity : protected ConsoleStreamInterface, public libMesh::Par
   bool _has_solution_warning;
   /// Whether or not we have an invalid solution (only after a sync)
   bool _has_solution_error;
+  /// Whether or not we have ever had any warning or solution issue during the simulation
```
lindsayad, line 176: I think this likely answers my very first comment question. Can you add "during the simulation" to the API doxygen?

framework/src/outputs/SolutionInvalidityOutput.C
```diff
@@ -22,14 +22,15 @@ SolutionInvalidityOutput::validParams()
...
+  params.set<ExecFlagEnum>("execute_on") = {EXEC_FINAL};
 
   params.addParam<unsigned int>("solution_invalidity_timestep_interval",
                                 1,
                                 "The number of time steps to group together in the table reporting "
-                                "the solution invalidity occurrences.");
+                                "the warnings and solution invalidity occurrences.");
 
-  params.addClassDescription("Controls output of the time history of solution invalidity object");
+  params.addClassDescription("Controls output of the time history of regular warnings and solution "
```
lindsayad, line 32: Like `mooseWarning`?

test/include/userobjects/WarningUserObject.h
```diff
@@ -0,0 +1,29 @@
...
+ */
+class WarningUserObject : public ThreadedGeneralUserObject
+{
+public:
+  static InputParameters validParams();
+
+  WarningUserObject(const InputParameters & params);
+
+  virtual void initialSetup() override;
+  virtual void initialize() override {};
```
lindsayad:
```suggestion
  virtual void initialize() override {}
```

test/include/userobjects/WarningUserObject.h
```diff
@@ -0,0 +1,29 @@
...
+  virtual void execute() override;
+  virtual void finalize() override;
+  virtual void threadJoin(const UserObject &) override {};
```
lindsayad:
```suggestion
  virtual void threadJoin(const UserObject &) override {}
```

test/tests/misc/solution_invalid/gold/solution_invalid_checkpoint.json
lindsayad: it would be nice for this number not to become trivial

lindsayad: Although to be honest I don't even know what this number represents. The `_counts` data member in `SolutionInvalidity` could use some more doxygen. Like how many elements should we expect in that vector; what do the elements represent?

framework/include/utils/SolutionInvalidity.h
GiudGiud: will update

framework/include/utils/SolutionInvalidity.h
GiudGiud: probs both

framework/src/outputs/SolutionInvalidityOutput.C
GiudGiud: yeah! that s the big change

test/tests/misc/solution_invalid/gold/solution_invalid_checkpoint.json
GiudGiud: per item, there's the iteration count (the one that gets reset every iteration)
the time step count (kept)
and the total count (sum of the timestep counts essentially, but also can tally non-timesteps as well like FINAL)

GiudGiud: I ll add to the doygen though

framework/include/utils/SolutionInvalidity.h
```diff
@@ -68,20 +68,24 @@ class SolutionInvalidity : protected ConsoleStreamInterface, public libMesh::Par
    */
   bool hasInvalidSolution() const;
 
+  /**
+   * Whether or not any warning or invalid solution has ever been encountered
```
GiudGiud:
```suggestion
   * Whether or not any warning or invalid solution has ever been encountered during the simulation
```

framework/include/utils/SolutionInvalidity.h
GiudGiud: MPI
there's a lock, the count keeping is not replicated across threads

test/tests/misc/solution_invalid/gold/solution_invalid_checkpoint.json
MengnanLi91: I am not against removing `converged_counts`. Do we know anyone actually rely on that counter? I think per time step is good enough. I also couldn't see a better way to keep the `converged_counts` and make it works for both solve and non-solve condition. 
How much changes we need to make if we accumulate to time step after the time step ends (right before TIMESTEP_END output)?

test/tests/misc/solution_invalid/gold/solution_invalid_outside_solve_out.json
MengnanLi91: I agree

test/tests/misc/solution_invalid/gold/solution_invalid_checkpoint.json
GiudGiud: > How much changes we need to make if we accumulate to time step after the time step ends (right before TIMESTEP_END output)?

It's a couple lines but right now if I do that I get double counts

MengnanLi91: double counts for iteration count? Does it affect the time step count and total counts?

GiudGiud: Yes they are all affected as the iteration counts are correctly accumulated so if they double count on the last iteration, the time step count and total count also get a double contribution from the last iteration (which is only part of the time step and total counts)

MengnanLi91: I vote to remove it then. It's hard to change without touching our execution order

MengnanLi91 (approved):
I agree to remove the iteration count. All rest looks good to me

test/tests/misc/solution_invalid/gold/solution_invalid_checkpoint.json
GiudGiud: ok that's what the last commit does. Thanks for the review!

lindsayad:
what's up with this phase field test

lindsayad:
Oh I see it on https://github.com/idaholab/moose/issues/27555

lindsayad (approved):

GiudGiud:
Thanks for the reviews!

PR #32233, opened 2026-01-15
Showcasing hypre AMS and ADS

nmnobre:
## Reason
Scalable preconditioners appropriate for H(curl) and H(div) discretisations.

## Design
Switched existing example to hypre AMS.
Added a new div-div kernel and implemented a new grad-div example showcasing hypre ADS.
Also took the chance to a) make sure all MFEM-backed examples using AMS now run w/ CUDA too and b) to remove a workaround while we waited for support for stage state solves in MFEM.

## Impact
Faster solves, specially for big enough problems for which direct solvers would struggle.


lindsayad (approved):

loganharbour:
This is failing sweeps on next. civet.inl.gov/job/3543186/ and https://civet.inl.gov/job/3543188/ for example

PR #32235, opened 2026-01-15
Give the option to not copy end-of-FP solution to old solution keep_solution_during_restore in FullSolveMultiApp

### LINKED ISSUE #32234: keep_solution_during_restore=true issue in FullSolveMultiApp of a transient with solution-state dependent material properties
## Description
We keep solution of the app in the restore, but then advanceState() copies the unrestored solution onto the old state, changing the old state forever.
Changing the old state changes the solution due to the material properties dependencies

## How to reproduce
Current MOOSE
inputs in discussion post

## Impact
Wrong results!

### Discussed in https://github.com/idaholab/moose/discussions/32229

<sup>Originally posted by **lynnmunday** January 14, 2026</sup>
### Check these boxes if you have followed the posting rules.

- [X] Q&A General is the most appropriate section for my question
- [X] I have consulted the posting Guidelines on the Discussions front page
- [X] I have searched the Discussions forum and my question has not been asked before
- [X] I have searched the MOOSE website and the documentation does not answer my question
- [X] I have formatted my post following the posting guidelines (screenshots as a last resort, triple back quotes around pasted text)

### Question

I made the below input files as a minimum working example to show the problem I'm having with a multiapp and ```keep_solution_during_restore=true``` with picard iterations.   My main app is a transient simulation that calls a full solve multiapp that is doing a thermal mechanical transient solve.  The thermomechanical solve has history dependent material properties (incremental deformation).  I run a transient for the thermomechanical solve until it reaches equilibrium.  During the picard solves, I would like to ```use keep_solution_during_restore=true``` so I will be closer to the equilibration of the subapp.   However, the displacement keeps growing by the same amount for every picard solve, see the below main app output for the postprocessor.  
```
Main app solve:
 0 Nonlinear |R| = 0.000000e+00
 Solve Converged!
 1 fixed point disp_x_max = 5.27730465e-03
 2 fixed point disp_x_max = 1.05686819e-02
 3 fixed point disp_x_max = 1.58743749e-02
 4 fixed point disp_x_max = 2.11946292e-02
```

.  ```use keep_solution_during_restore=false``` gives me the correct behavior.  I didn't really look into the issue and using the new mechanics kernel system solves the problem but I wanted to see if anyone knows the problem or if I should create an issue.  Maybe it should error out if you have history dependent material properties.  

main.i:

```
[Mesh]
  [generated]
    type = GeneratedMeshGenerator
    dim = 2
    nx = 10
    ny = 10
    xmax = 2
    ymax = 1
  []
  [pin]
    type = ExtraNodesetGenerator
    input = generated
    new_boundary = pin
    coord = '0 0 0'
  []
[]

[Variables]
  [diffused]
    order = FIRST
    family = LAGRANGE
  []
[]
[Kernels]
  [diff]
    type = Diffusion
    variable = diffused
  []
[]
[AuxVariables]
  [disp_x]
    order = FIRST
    family = LAGRANGE
    initial_condition = 0.0
  []
[]
[Executioner]
  type = Steady
  solve_type = 'PJFNK'
  fixed_point_algorithm = picard
  custom_pp = disp_x_max
  custom_abs_tol = 1e-6
  fixed_point_min_its = 4
  fixed_point_max_its = 5
[]

[MultiApps]
  [sub_app]
    type = FullSolveMultiApp
    input_files = 'subapp.i'
    execute_on = 'timestep_begin'
    keep_solution_during_restore = true # uses previous iterate as initial guess for the next iteration
  []
[]

[Transfers]
  [disp_z_avg_clad]
    type = MultiAppGeneralFieldNearestLocationTransfer
    from_multi_app = sub_app
    source_variable = 'disp_x'
    variable = disp_x
    displaced_source_mesh = false
    displaced_target_mesh = false
    search_value_conflicts = false
    greedy_search = true
  []
[]
[Postprocessors]
  [disp_x_max]
    type = ElementExtremeValue
    variable = 'disp_x'
    value_type = max
    execute_on = 'INITIAL TIMESTEP_END'
  []
[]
[Outputs]
  csv = true
[]
```

subapp.i:
```
[GlobalParams]
  displacements = 'disp_x disp_y'
[]

[Mesh]
  [generated]
    type = GeneratedMeshGenerator
    dim = 2
    nx = 10
    ny = 10
    xmax = 2
    ymax = 1
  []
  [pin]
    type = ExtraNodesetGenerator
    input = generated
    new_boundary = pin
    coord = '0 0 0'
  []
[]

[Physics/SolidMechanics/QuasiStatic]
  [all]
    add_variables = true
    strain = FINITE
    incremental=true
  []
[]

[Materials]
  [elasticity]
    type = ComputeIsotropicElasticityTensor
    youngs_modulus = 1e9
    poissons_ratio = 0.3
  []
  [stress]
  []
[]

[BCs]
  [pull_x]
    type = NeumannBC
    variable = disp_x
    boundary = right
    value = 1e6
  []
  [pin_x]
    type = DirichletBC
    variable = disp_x
    boundary = pin
    value = 0
  []
  [bottom_y]
    type = DirichletBC
    variable = disp_y
    boundary = bottom
    value = 0
  []
[]

[Executioner]
  type = Transient
  solve_type = NEWTON
  petsc_options_value = 'lu'
  num_steps =1
  dt = 1
[]

[Outputs]
[]
```
</div>

GiudGiud:
Prevents a bug where the unrestored solution gets copied onto the old state 
Which for transients is totally fine, but for quasi-static with old displacement values used in calculations is not. 

closes #32234

I think there will be test failures. Question is how many
EDIT:
looks like only 1 in the framework
EDIT:
should be fine now

Patches:
- moltres https://github.com/arfc/moltres/pull/335
- blue_crab https://github.inl.gov/ncrc/blue_crab/pull/354
- griffin https://github.inl.gov/ncrc/griffin/pull/2921

### DIFF
--- a/framework/src/multiapps/FullSolveMultiApp.C
+++ b/framework/src/multiapps/FullSolveMultiApp.C
@@ -30,17 +30,42 @@ FullSolveMultiApp::validParams()
   params.addParam<bool>("ignore_solve_not_converge",
                         false,
                         "True to continue main app even if a sub app's solve does not converge.");
+  params.addParam<bool>("update_old_solution_when_keeping_solution_during_restore",
+                        true,
+                        "Whether to update the old solution vector (to the previous fixed point "
+                        "iteration solution) when keeping the solution during restore.");
   return params;
 }
 
 FullSolveMultiApp::FullSolveMultiApp(const InputParameters & parameters)
-  : MultiApp(parameters), _ignore_diverge(getParam<bool>("ignore_solve_not_converge"))
+  : MultiApp(parameters),
+    _ignore_diverge(getParam<bool>("ignore_solve_not_converge")),
+    _update_old_state_when_keeping_solution_during_restore(
+        getParam<bool>("update_old_solution_when_keeping_solution_during_restore"))
 {
   // You could end up with some dirty hidden behavior if you do this. We could remove this check,
   // but I don't think that it's sane to do so.
   if (_no_restore && (_app.isRecovering() || _app.isRestarting()))
     paramError("no_restore",
                "The parent app is restarting or recovering, restoration cannot be disabled");
+  if (_keep_solution_during_restore &&
+      !isParamSetByUser("update_old_solution_when_keeping_solution_during_restore"))
+    paramError("update_old_solution_when_keeping_solution_during_restore",
+               "Due to 'keep_solution_during_restore' parameter being true, which is an "
+               "optimization for fixed point iterations, the "
+               "unrestored solution will be kept as the starting solution for the next solve "
+               "of the MultiApp. You must set this parameter to decide if this solution should "
+               "be copied as the old solution at the beginning of the next time step, "
+               "or not. If the MultiApp is running a transient, you likely want to set this to "
+               "true. If the MultiApp is a quasi-static simulation, you likely want to set "
+               "this to false. If you don't know what this error message means, please set "
+               "'keep_solution_during_restore' to false and no need to set "
+               "'update_old_solution_when_keeping_solution_during_restore'.");
+  if (isParamSetByUser("update_old_solution_when_keeping_solution_during_restore") &&
+      !_keep_solution_during_restore)
+    paramError("update_old_solution_when_keeping_solution_during_restore",
+               "Should not be set if not keeping the solution during restore "
+               "(keep_solution_during_restore=false)");
 }
 
 void
@@ -110,6 +135,9 @@ FullSolveMultiApp::solveStep(Real /*dt*/, Real /*target_time*/, bool auto_advanc
     // reset output system if desired
     if (!getParam<bool>("keep_full_output_history"))
       _apps[i]->getOutputWarehouse().reset();
+    // Prevent the copy of the post-FP iteration solution state onto the old vector
+    if (!_update_old_state_when_keeping_solution_during_restore)
+      appProblemBase(_first_local_app + i).skipNextForwardSolutionCopyToOld();
 
     Executioner * ex = _executioners[i];
     ex->execute();
--- a/framework/src/multiapps/FullSolveMultiApp.C
+++ b/framework/src/multiapps/FullSolveMultiApp.C
@@ -50,6 +50,7 @@ FullSolveMultiApp::FullSolveMultiApp(const InputParameters & parameters)
   if (_no_restore && (_app.isRecovering() || _app.isRestarting()))
     paramError("no_restore",
                "The parent app is restarting or recovering, restoration cannot be disabled");
+  // Force the user to make a decision on updating or not the old state of variables
   if (_keep_solution_during_restore &&
       !isParamSetByUser("update_old_solution_when_keeping_solution_during_restore"))
     paramError("update_old_solution_when_keeping_solution_during_restore",
@@ -109,12 +110,34 @@ FullSolveMultiApp::initialSetup()
       ex->init();
 
       if (!_update_old_state_when_keeping_solution_during_restore &&
-          appProblemBase(_first_local_app + i).getMaterialPropertyStorage().hasStatefulProperties())
-        paramError("update_old_solution_when_keeping_solution_during_restore",
-                   "While we are updating old solutions using the solution from the previous fixed "
-                   "point iteration, we are not updating the old stateful material properties as "
-                   "well. This is not consistent. We recommend you consider using the 'no_restore' "
-                   "parameter instead of 'keep_solution_during_restore'.");
+          (appProblemBase(_first_local_app + i).getMaterialPropertyStorage().hasStatefulProperties()
+#ifdef KOKKOS_ENABLED
+           || appProblemBase(_first_local_app + i)
+                  .getKokkosMaterialPropertyStorage()
+                  .hasStatefulProperties()
+#endif
+               ))
+        paramError(
+            "update_old_solution_when_keeping_solution_during_restore",
+            "While we are updating old solutions using the solution from the previous fixed "
+            "point iteration, we are not updating the old stateful material properties as "
+            "well. This is not consistent. We recommend you consider using the 'no_restore' "
+            "parameter instead of 'keep_solution_during_restore', or stop using the latter.");
+      if (_keep_solution_during_restore &&
+          appProblemBase(_first_local_app + i)
+              .hasSolutionState(2, Moose::SolutionIterationType::Time))
+        mooseDoOnce(paramWarning(
+            "keep_solution_during_restore",
+            "This FullSolveMultiApp simulation(s) uses older time step variable states (notably "
+            "from two time steps prior in transients). Due to 'keep_solution_during_restore' "
+            "parameter being true, which is an optimization for fixed point iterations, the "
+            "unrestored solution will be kept as the starting solution. It would normally be "
+            "copied onto the 'old' state at the beginning of the first time step. This copy can be "
+            "skipped using 'update_old_solution_when_keeping_solution_during_restore', while the "
+            "copy of the old state onto the 'older' state and the stateful material properties "
+            "state updates do not have such an option at this time. This warning relates to this "
+            "inconsistency. If you suspect this is a problem, please set "
+            "'keep_solution_during_restore' to false"));
 
       _executioners[i] = ex;
     }
--- a/framework/src/multiapps/FullSolveMultiApp.C
+++ b/framework/src/multiapps/FullSolveMultiApp.C
@@ -11,6 +11,8 @@
 #include "LayeredSideDiffusiveFluxAverage.h"
 #include "Executioner.h"
 #include "TransientBase.h"
+#include "FEProblemBase.h"
+#include "MaterialPropertyStorage.h"
 #include "Console.h"
 
 // libMesh
@@ -105,6 +107,14 @@ FullSolveMultiApp::initialSetup()
       }
 
       ex->init();
+
+      if (!_update_old_state_when_keeping_solution_during_restore &&
+          appProblemBase(_first_local_app + i).getMaterialPropertyStorage().hasStatefulProperties())
+        paramError("update_old_solution_when_keeping_solution_during_restore",
+                   "While we are updating old solutions using the solution from the previous fixed "
+                   "point iteration, we are not updating the old stateful material properties as "
+                   "well. This is not consistent. We recommend you consider using the 'no_restore' "
+                   "parameter instead of 'keep_solution_during_restore'.");
 
       _executioners[i] = ex;
     }
--- a/test/tests/multiapps/picard_multilevel/2level_picard/mutilevel_app.i
+++ b/test/tests/multiapps/picard_multilevel/2level_picard/mutilevel_app.i
@@ -79,6 +79,8 @@
     input_files = sub_level1.i
     execute_on = 'timestep_end'
     keep_solution_during_restore = true
+    # We update the old solution to be able to compute the time derivative when we start the child app's
+    # transient again at the beginning of the next fixed point iteration
     update_old_solution_when_keeping_solution_during_restore = true
   []
 []
--- a/test/tests/multiapps/picard_multilevel/2level_picard/mutilevel_app.i
+++ b/test/tests/multiapps/picard_multilevel/2level_picard/mutilevel_app.i
@@ -79,6 +79,7 @@
     input_files = sub_level1.i
     execute_on = 'timestep_end'
     keep_solution_during_restore = true
+    update_old_solution_when_keeping_solution_during_restore = true
   []
 []
 
--- a/framework/include/multiapps/FullSolveMultiApp.h
+++ b/framework/include/multiapps/FullSolveMultiApp.h
@@ -51,6 +51,8 @@ protected:
 private:
   /// Switch to tell executioner to keep going despite app solve not converging
   const bool _ignore_diverge;
+  /// Switch to tell the systems or not to update the old solution using the unrestored solution (the one we 'kept during restore')
+  const bool _update_old_state_when_keeping_solution_during_restore;
 
   std::vector<Executioner *> _executioners;
 };
--- a/framework/include/problems/FEProblemBase.h
+++ b/framework/include/problems/FEProblemBase.h
@@ -570,6 +570,11 @@ public:
 
   virtual void copySolutionsBackwards();
 
+  /// Prevents the copy of the solution vector to the old solution vector in each system.
+  /// Old -> Older is still performed
+  /// This is useful for MultiApps fixed point iterations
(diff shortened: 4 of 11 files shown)


framework/src/multiapps/FullSolveMultiApp.C
```diff
@@ -30,17 +30,42 @@ FullSolveMultiApp::validParams()
...
+               "Due to 'keep_solution_during_restore' parameter being true, which is an "
+               "optimization for fixed point iterations, the "
+               "unrestored solution will be kept as the starting solution for the next solve "
+               "of the MultiApp. You must set this parameter to decide if this solution should "
+               "be copied as the old solution at the beginning of the next time step, "
+               "or not. If the MultiApp is running a transient, you likely want to set this to "
+               "true. If the MultiApp is a quasi-static simulation, you likely want to set "
+               "this to false. If you don't know what this error message means, please set "
+               "'keep_solution_during_restore' to false and no need to set "
+               "'update_old_solution_when_keeping_solution_during_restore'.");
```
hugary1995, line 66: I must be fundamentally misunderstanding how fixed point iteration works. For a given time step, say we have the following transient simulation with FP iterations:
```
step n:
  FPI 0: main, sub, ...
  FPI 1: main, sub, ...
  FPI 2: main, sub, ...
  ...

step n+1
```
With `keep_solution_during_restore = true` and `update_old_solution_when_keeping_solution_during_restore = true` (what you suggested), are you saying the old solution vector could be different across different fixed point iterations? That is no longer a fixed point.

hugary1995 (changes requested):

framework/src/multiapps/FullSolveMultiApp.C
GiudGiud: exactly. It is no longer a 'true' fixed point. 
It's advancing a subapp transient in time, starting from where we finished in the previous fixed point step. 
It's what we want because we are actually looking for a steady state when we do this (considering this choice of FullSolveMultiApp, we want a full solve e.g. often a steady state at the end)

GiudGiud: It's useful though, because if we were nearing the steady state at the end of FP#1, we are now real close to steady state on FP#2

hugary1995: Alright, thanks for explaining the use case. I think this is a valid use case we could support. Here are my two problems with the current implementation:
1. This sounds to me like a special type of fixed point iteration with some custom logic, which you also acknowledged when saying "It is no longer a 'true' fixed point". In fact, in the optimization literature people stop calling this fixed point iteration. What you will commonly find in literature is words like "successive substitution with updated state". I am totally fine with supporting such a parameter like `fixed_point_with_updated_state = true/false` (note I still keep the `fixed_point_` prefix to keep it syntactically tied to fixed point iteration), but what I don't like is making `fixed_point_with_updated_state = true` the default (which is the case when we currently set `keep_solution_during_restore` with `FullSolveMultiApp`).
2. Changing the old solution vector across FP iterations is fine. That's not totally uncommon in the literature. As I mentioned above people typically call it "successive substitution with updated state" when assemblying the operator. However, what's definitely uncommon (and IMHO unacceptable) is changing the state in an inconsistent manner -- for example changing the old solution vector while keeping the old stateful material properties from the last FP iteration -- and that's what we are currently doing. The knob we are trying to add here is `update_old_solution_when_keeping_solution_during_restore`, which implies exactly that: it sounds like we allow separate (and very likely inconsistent) controls of old solution and old material data. Bottom line is we should always update solution vector and material data together.

So putting my thoughts together, I envision three groups of parameters controlling the FP behavior:
```
# first group
fixed_point_algorithm = ...
fixed_point_min_its = ...
fixed_point_max_its = ...
fixed_point_....

# second group
fixed_point_with_updated_state = true/false

# third group
keep_solution_during_restore = true/false
```
The three groups serve different and _independent_ purposes. And now that I think about it more carefully, `keep_solution_during_restore` is just a different relaxation technique, because all it effectively does is setting a different initial guess to the equation system. Since we already support some relaxation schemes with `fixed_point_algorithm`, why don't we unify the two parameters? This also leads me to wonder how `fixed_point_algorithm = secant/steffensen` interact with `keep_solution_during_restore = true`...

hugary1995: I guess I can compromise a little. I am okay with merging this additional flag for the time being, but at some point in the future we need to do things right.

GiudGiud: > ), but what I don't like is making fixed_point_with_updated_state = true the default (which is the case when we currently set keep_solution_during_restore with FullSolveMultiApp

There are no defaults anymore (only for FullSolveMultiApp), the new boolean is mandatory.

> it sounds like we allow separate (and very likely inconsistent) controls of old solution and old material data. Bottom line is we should always update solution vector and material data together.

We should but we don't have the machinery. And the 'keep_solution' option is so simple in its implementation it won't do material properties. We'd need to move that logic to the problem. Maybe something like `FEProblemBase::cacheState( & some_cache) ` and `FEProblemBase::restoreFromCache( & the cache)` to keep the same design. OR better logic in "restoreMultiApps" to mostly prevent the restore. Which actually, there is a `no_restore` option one can use too. Maybe that's what people with stateful properties and running pseudo-transients should look into instead

> This also leads me to wonder how fixed_point_algorithm = secant/steffensen interact with keep_solution_during_restore = true..

This will likely depend on what is being "transformed" by those algorithms. For the main apps, what is happening to the child app trying to reach steady state is likely not a problem either way as long as they ran until steady state. So applying the secant transformation on main app PPs and variables should work just fine.

hugary1995: Can we check if the problem has stateful properties? When `update_old_solution_when_keeping_solution_during_restore = true`, we should error if there's any stateful properties.

GiudGiud: There isn't an API right now but yes I can try that
EDIT: we do, on the MaterialPropertyStorage

GiudGiud: done

test/tests/multiapps/picard_multilevel/2level_picard/mutilevel_app.i
```diff
@@ -79,6 +79,9 @@
     input_files = sub_level1.i
     execute_on = 'timestep_end'
     keep_solution_during_restore = true
+    # We update the old solution to be able to compute the time derivative when we start the child app's
+    # transient again at the beginning of the next fixed point iteration
+    update_old_solution_when_keeping_solution_during_restore = true
```
hugary1995, line 84: I wonder if we can come up with a more intuitive parameter name... also the current parameter name is a bit too long. How about something like `continue_transient` (or its complement `restart_transient`)? That probably makes the intent clearer, in line with your comment in another input file:
```python
    # The choice does not matter here, but conceptually there is no need to override
    # the old solution when the subapp is a Steady solve
    update_old_solution_when_keeping_solution_during_restore = false
```
Imagine if the parameter were `continue_transient`/`restart_transient`, your comment is somewhat implied.

framework/src/multiapps/FullSolveMultiApp.C
```diff
@@ -30,17 +32,42 @@ FullSolveMultiApp::validParams()
   params.addParam<bool>("ignore_solve_not_converge",
                         false,
                         "True to continue main app even if a sub app's solve does not converge.");
+  params.addParam<bool>("update_old_solution_when_keeping_solution_during_restore",
+                        true,
+                        "Whether to update the old solution vector (to the previous fixed point "
+                        "iteration solution) when keeping the solution during restore.");
```
hugary1995, line 38: Can we not provide a default here?

If there's a way to query the executioner type of the subapp, we can do a better error checking:
- `isParamValid` --> error/warn if the subapp is steady, just say this parameter takes no effect
- `!isParamValid` -> paramError if the subapp is transient

hugary1995 (changes requested):
Per our discussion above, I am onboard with supporting this behavior. Just have two minor comments.

test/tests/multiapps/picard_multilevel/2level_picard/mutilevel_app.i
GiudGiud: this is not specific enough. If you use higher order time integration (which uses the older state), this `continue_transient` option does not suffice to make it work.
Same issue for stateful material properties.
(same problem with current name though)

Also if you were doing an eigenvalue subapp using the old state for something, a "continue_transient" would not be very helpful.

And finally, if you are not keeping_solution_during_restore, my parameter is obviously irrelevant whereas `continue_transient` could be relevant, but it only triggers an error message. 

Short and to the point might make it "intuitive" but it's less correct. This parameter turns on a very specific behavior, and it is named after that very specific behavior.

framework/src/multiapps/FullSolveMultiApp.C
GiudGiud: no default, both options make sense to different simulations

I will improve on the error message with this advice.

GiudGiud: > If there's a way to query the executioner type of the subapp, we can do a better error checking:

nevermind, we indeed don't have access to the executioner on construction of the FullSolveMultiapp app so we'd need to do this later. 

> error/warn if the subapp is steady, just say this parameter takes no effect

I don't know that is has no effect. I dont see in Coupleable anything preventing you from using an old state (which you could be loading from file) without a time integrator
Also saving a solution vector copy is likely a "0.5%" optimization

test/tests/multiapps/picard_multilevel/2level_picard/mutilevel_app.i
GiudGiud: I think 'continue_transient' / 'restart_transient' sort of implies using BOTH of these parameters (keep solution during restore and update old solution when keeping solution during restore).
But when creating the first parameter they intended to be more specific, possibly because they knew it was hacky and not really continuing the transient. 

I think my new parameter name is well aligned with the original intent, and the new code is great at picking out two unsupported cases. 
I recommend approving my PR :D

hugary1995: But isn't that the use case we intended to support? So it's like our ultimate goal is to support the use of FullSolveMultiApp to continue a transient during fixed point iterations, but our current implementation only works under certain assumptions. From that perspective, I'd suggest using a parameter that explains our intent (which is `continue_transient`) but perform careful error checks to rule out the cases we currently do not support. In the future, we could relax the constraints one by one.

> this is not specific enough. If you use higher order time integration (which uses the older state), this continue_transient option does not suffice to make it work.
> Same issue for stateful material properties.

That's why we are doing the error check for stateful properties now, isn't it? So technically we should also error if the subapp uses a higher order time integrator.

> Also if you were doing an eigenvalue subapp using the old state for something, a "continue_transient" would not be very helpful.

Then why do an eigenvalue subapp want to continue from old state? I just want to make sure I understand the use case.

> And finally, if you are not keeping_solution_during_restore, my parameter is obviously irrelevant whereas continue_transient could be relevant, but it only triggers an error message.

Yeah, that's good point. So we should make sure if `continue_transient = true`, `keeping_solution_during_restore` must also be true.

> Short and to the point might make it "intuitive" but it's less correct. This parameter turns on a very specific behavior, and it is named after that very specific behavior.

All we are doing here is to prevent Griffin users from shooting their foot. Using a parameter name like `update_old_solution_when_keeping_solution_during_restore` sounds more like saying "this is what it does and I don't care if it doesn't do what you wanted it to do -- you are on your own". That's not a very friendly message to send in my opinion. Also, similar to my response to your eigenvalue problem comment: if not for continuing a transient, why would someone want to toggle `update_old_solution_when_keeping_solution_during_restore`?

GiudGiud: > But isn't that the use case we intended to support?

Not necessarily which is why 'keep_solution_during_restore' was not called 'continue_transient' either. This stuff is meant to be quite general, it pertains to internal details of the fixed point solve. It's not ideal I agree but it's all over MOOSE like that. There is no terminology that just naturally solves everything for everyone. All we can do is be precise about what any parameter does, and do our best to keep people within the bounds of what works. If anyone wants better names that are fitted to the application at hand, they need to create an app and make an Action that set these simulations up exactly as intended. We do this in BlueCrab notably. 

> That's why we are doing the error check for stateful properties now, isn't it? So technically we should also error if the subapp uses a higher order time integrator.

yes I just added this new check in the last commit

> Then why do an eigenvalue subapp want to continue from old state? I just want to make sure I understand the use case.

no idea but also I can't know what everyone is doing or trying to do. If any eigenvalue subapp person ends up having to set `continue_transient = true`, the parameter naming will have failed.
Whereas my beautiful name works either way.

> this is what it does and I don't care if it doesn't do what you wanted it to do -- you are on your own

Better than saying "this will continue your transient" but actually it does not because there is something we did not think about (like recoverable data that should have gotten changed during the transient for example, or 'Control' states that got reset by restarting the transient). 
And for this last point, neither 'keep_solution.." or 'update ..." make any promise about that.

hugary1995:
Alright, I surrender. I hope the long parameter name can scare people away from using it... Just leaving my final comment here to clarify that this new, beautiful parameter named `update_old_solution_when_keeping_solution_during_restore` is to temporarily patch an ugly bug which we all agreed to address eventually.

hugary1995 (approved):
Approving now and leaving to @GiudGiud to confirm downstream patches are ready before merging.

GiudGiud:
Thanks! Yup I ll handle the patches and only merge when green

GiudGiud:
Thanks for the review!

framework/src/multiapps/FullSolveMultiApp.C
```diff
@@ -110,6 +168,9 @@ FullSolveMultiApp::solveStep(Real /*dt*/, Real /*target_time*/, bool auto_advanc
     // reset output system if desired
     if (!getParam<bool>("keep_full_output_history"))
       _apps[i]->getOutputWarehouse().reset();
+    // Prevent the copy of the post-FP iteration solution state onto the old vector
+    if (!_update_old_state_when_keeping_solution_during_restore)
+      appProblemBase(_first_local_app + i).skipNextForwardSolutionCopyToOld();
```
lindsayad, line 173: Where does the "when_keeping_solution" portion of this factor in since this is applied regardless of the value of "keep_solution_during_restore"?

GiudGiud: I fixed that in a later PR?

GiudGiud: oh I think the fix is in the transfers PR

PR #32242, opened 2026-01-16
Allowed `ChangeOverTimePostprocessor` to divide by time step size

### LINKED ISSUE #32241: Allow `ChangeOverTimePostprocessor` to divide by time step size
### Motivation

The time derivative of a post-processor is a useful measure of how steady the value is, which can be used in steady-state detection criteria.

### Design

Add a `divide_by_dt` parameter to `ChangeOverTimePostprocessor`.

### Impact

New optional parameter in `ChangeOverTimePostprocessor` defaulting to the existing behavior.

### LINKED ISSUE #32246: Create Convergence to check post-processors for steady-state convergence
### Motivation

Users should be able to check the value of a post-processor for steady-state convergence. A post-processor might compute the time derivative of a quantity, for example, which should be near zero at steady-state conditions. While `PostprocessorConvergence` can be used for this purpose, it requires the user to set `max_iterations` to something other than the default of 50 to avoid false divergence. Also, in most cases, the tested post-processor, like a time derivative, is falsely 0 at the initial time, resulting in false convergence, so the first iteration should return `ITERATING`.

### Design

Create `PostprocessorSteadyStateConvergence`, similar to `PostprocessorConvergence` but not inheriting from `IterationCountConvergence`.

### Impact

New class.

joshuahansel:
- Added `divide_by_dt` parameter to `ChangeOverTimePostprocessor` (closes #32241).
- Added verbose output to `PostprocessorConvergence` when its still iterating.
- Added `PostprocessorSteadyStateConvergence` (closes #32246).


framework/src/postprocessors/ChangeOverTimePostprocessor.C
```diff
@@ -99,6 +113,9 @@ ChangeOverTimePostprocessor::finalize()
     _value = std::fabs(change);
   else
     _value = change;
+
+  if (_divide_by_dt)
+    _value /= std::max(_dt, 1e-15); // Guard against division by zero; _dt=0 on INITIAL
```
GiudGiud:
```suggestion
    _value /= std::max(_dt, 1); // Guard against division by zero; _dt=0 on INITIAL
```
if executing on initial it would be very surprising to see 1e15 come up

framework/src/postprocessors/ChangeOverTimePostprocessor.C
```diff
@@ -23,6 +23,11 @@ ChangeOverTimePostprocessor::validParams()
   params.addParam<bool>(
       "compute_relative_change", false, "Compute magnitude of relative change instead of change");
   params.addParam<bool>("take_absolute_value", false, "Option to take absolute value of change");
+  params.addParam<bool>("divide_by_dt",
+                        false,
+                        "If true, divide by the time step size. This may only be enabled if "
+                        "'change_with_respect_to_initial' is 'false'. approximates the time "
```
GiudGiud:
```suggestion
                        "'change_with_respect_to_initial' is 'false'. Approximates the time "
```

framework/src/postprocessors/ChangeOverTimePostprocessor.C
```diff
@@ -23,6 +23,11 @@ ChangeOverTimePostprocessor::validParams()
...
+                        "derivative of the post-processor. This may only be enabled");
```
GiudGiud: sentence end missing?

GiudGiud (commented):
thoughts about SteadyStatePostprocessorConvergence (grouping convergence object by application in the index) vs PostprocessorSteadyStateConvergence (grouping by operation, and the user looking for steady state picks the appropriate one)

framework/src/postprocessors/ChangeOverTimePostprocessor.C
joshuahansel: I'm confused - with your change, time step size could never be less than 1.

joshuahansel:
> thoughts about SteadyStatePostprocessorConvergence (grouping convergence object by application in the index) vs PostprocessorSteadyStateConvergence (grouping by operation, and the user looking for steady state picks the appropriate one)

My preference is to keep it consistent with `DefaultSteadyStateConvergence`. And I think it might be a little weird since some of the generic ones might be options for a steady-state check, so looking for ones starting with "SteadyState" would miss some.

framework/src/postprocessors/ChangeOverTimePostprocessor.C
joshuahansel: Resolved offline; with the suggestion made, need to additional check that `_dt` is zero. Agreed on
```
if (_divide_by_dt && !MooseUtils::absoluteFuzzyEquals(_dt, 0.0))
    _value /= _dt;
```

GiudGiud:
Makes sense though I'd want to rename `DefaultSteadyStateConvergence` to something to do with the nonlinear residual at some point. 
`DefaultSteadyStateConvergence` could still stay as an alias though

GiudGiud (approved):

GiudGiud:
test failure (gallium solidification test) is in my todo list

PR #32248, opened 2026-01-19
Integrate stateless random number generators into stochastic tools samplers.

pbehne:
Ref #32194.


modules/stochastic_tools/src/samplers/LatinHypercubeSampler.C
```diff
@@ -98,3 +101,134 @@ LatinHypercubeSampler::computeSample(dof_id_type row_index, dof_id_type col_inde
...
+  if (!comm_ptr || comm_ptr->size() == 1)
+  {
+    const std::size_t n_global = data.size();
+    if (n_global < 2)
+      return;
+
+    std::size_t rn_ind = 0;
+    for (std::size_t i = n_global - 1; i > 0; --i)
+    {
+      const auto j = getRandlStateless(rn_ind++, 0, i, seed_index);
```
zachmprince: If you look at the implementation on how the stateless RNG draws ranged integers, the memory footprint will blow up with this call. Basically, it will eventually create `n_global - 1` objects.  As such, I would recommend adding a new method in MooseRandomStateless specifically for grabbing indices for shuffling. I'm thinking the signature would look like this:
```c++
std::size_t randlShuffle(std::size_t n, std::size_t size)
```
Which will effectively be the same as calling:
```c++
randl(n, 0, size - n - 1)
```
but it will create a new generator object based on `size` instead of `size - n  - 1`.

modules/stochastic_tools/include/samplers/PMCMCBase.h
```diff
@@ -12,6 +12,7 @@
 #include "Sampler.h"
 #include "TransientInterface.h"
 #include "Distribution.h"
+#include <cstddef>
```
zachmprince: Are these includes necessary?

modules/stochastic_tools/src/samplers/LatinHypercubeSampler.C
```diff
@@ -98,3 +101,134 @@ LatinHypercubeSampler::computeSample(dof_id_type row_index, dof_id_type col_inde
...
+    return;
+
+  // Advance the per-column bin RNG and shuffle RNG to match sampleSetUp usage.
+  const auto shuffle_count = n_rows > 0 ? n_rows - 1 : 0;
+  for (const auto col : make_range(n_cols))
+  {
+    advanceStatelessGenerator(col, n_rows);
+    if (shuffle_count > 0)
+      advanceStatelessGenerator(col + n_cols, shuffle_count);
+  }
```
zachmprince: Is there any way we can rely on the advancement step in `Sampler::execute` instead of having a specialized one here? Perhaps we could add a virtual protected member in `Sampler` to get the advance count for a particular seed index. I'd like to see as little customization of these `execute*` routines as possible.

zachmprince (changes requested):
This is a great first start! This definitely doesn't close the issue yet, but I doubt it would in a single PR.

The biggest thing I'd like to see from this PR is the removal of `sampleSetUp` and `sampleTearDown` methods. This will pave the way to replacing `getNextLocalRow` logic throughout the module.

modules/stochastic_tools/include/samplers/PMCMCBase.h
pbehne: Nope, my editor inserted them. Will remove.

PR #32253, opened 2026-01-20
Build hwloc topology up front when possible

### LINKED ISSUE #32252: Don't use mpich hwloc in testing
### Motivation

Now that MPICH is configured with hwloc in many places, it looks like the initialization time with hwloc is quite expensive. For small tests, this could be upwards of 0.5s+. We shouldn't need hwloc for most testing, as we usually have an affinity for all processors for the individual tests that are ran. That is - we don't use any CPU binding, we just let the kernel do its thing.

### Design

Disable hwloc with mpich when it is detected. This can be done with `HWLOC_COMPONENTS=stop` in the environment. Add an option `--disable-mpi-options` to the test harness if the user doesn't want these variables to be set.

### Impact

Will likely speed up testing based on the evidence we have so far.

loganharbour:
Closes #32252

This decreases runtime for framework tests with `-j 24` from ~400 sec to ~330 sec on rod.

### DIFF
--- a/python/TestHarness/TestHarness.py
+++ b/python/TestHarness/TestHarness.py
@@ -30,7 +30,8 @@ import pyhit
 from FactorySystem.Factory import Factory
 from FactorySystem.Parser import Parser
 from FactorySystem.Warehouse import Warehouse
-from mooseutils import colorText
+
+from .mpi_config import get_mpi_config
 
 if TYPE_CHECKING:
     from pycapabilities import Capabilities
@@ -449,6 +450,9 @@ class TestHarness:
         self.options._checks = checks
         # So that testers can see if we have an application
         self.options._app_name = self.app_name
+
+        # Store the mpi configuration we have discovered
+        self.options._mpi_config = get_mpi_config()
 
         # Initialize the scheduler
         self.initialize()
@@ -1684,6 +1688,16 @@ class TestHarness:
             dest="method",
             const="oprof",
             help="Test the <app_name>-oprof binary",
+        )
+
+        envgroup = parser.add_argument_group(
+            "Environment Options", "Control the runtime environment"
+        )
+        envgroup.add_argument(
+            "--disable-mpi-options",
+            action="store_true",
+            dest="disable_mpi_options",
+            help="Disable automated MPI environment options",
         )
 
         screengroup = parser.add_argument_group(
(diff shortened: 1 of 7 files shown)


python/TestHarness/TestHarness.py
```diff
@@ -450,6 +451,9 @@ def __init__(
         # So that testers can see if we have an application
         self.options._app_name = self.app_name
 
+        # Store the mpi configuration we have discovered
+        self.options._mpi_config = get_mpi_config()
```
zachmprince: This probably shouldn't be a private member since it is being accessed outside of this class. If you don't want other objects to modify it, then add a `@property` to make its value readable.

python/TestHarness/mpi_config.py
```diff
@@ -0,0 +1,90 @@
...
+
+    if (exe := which("lstopo-no-graphics")) is None and (
+        exe := which("lstopo")
+    ) is None:
+        return None
+
+    try:
+        store_dir = os.path.join(Path.home(), ".local", "share", "moose", "testharness")
+        if not os.path.exists(store_dir):
+            os.makedirs(store_dir, exist_ok=True)
```
zachmprince:
```suggestion
        os.makedirs(store_dir, exist_ok=True)
```

python/TestHarness/testers/RunApp.py
```diff
@@ -378,7 +383,7 @@ def getCommand(self, options):
             )
 
             # Need to run mpiexec with containerized openmpi
-            if options.hpc and self.hasOpenMPI():
+            if options.hpc and options._mpi_config == MPIConfig.OPENMPI:
```
zachmprince: `MPIConfig` doesn't have a member `OPENMPI`. Should it be `MPIType`?

python/TestHarness/mpi_config.py
```diff
@@ -0,0 +1,89 @@
...
+        run(cmd, check=True)
+        return store_file
+    except Exception:
+        print("WARNING: Failed to build hwloc topology file")
+        print(format_exc())
+
+    return None
+
+
+@staticmethod
```
zachmprince:
```suggestion
```

zachmprince (approved):

loganharbour:
Failure is unrelated; due to a test box out of resources.

PR #32254, opened 2026-01-20
TestHarness: Expect explicit exit codes, remove "crash" verbaige

### LINKED ISSUE #32251: Revamp exit code checking in the TestHarness

loganharbour:
Closes https://github.com/idaholab/moose/issues/32251

### DIFF
--- a/python/TestHarness/testers/RunApp.py
+++ b/python/TestHarness/testers/RunApp.py
@@ -291,6 +291,10 @@ class RunApp(Tester):
         if options.scaling and specs['scale_refine'] > 0:
             cli_args.insert(0, ' -r ' + str(specs['scale_refine']))
 
+        # WIP REMOVE ONCE LIBMESH IS FIXED
+        if specs["expect_exit_code"] != 0:
+            cli_args.append("--disable-thread-safe-output")
+
         # Get the number of processors and threads the Tester requires
         ncpus = self.getProcs(options)
         nthreads = self.getThreads(options)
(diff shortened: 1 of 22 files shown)


python/TestHarness/testers/RunApp.py
```diff
@@ -387,35 +392,31 @@ def testFileOutput(self, moose_dir, options, runner_output):
...
-            reason = 'CRASH'
+        reason = None
+
+        if specs["expect_exit_code"] != exit_code:
+            reason = f'EXIT CODE {exit_code} != {specs["expect_exit_code"]}'
+        elif (
+            options.valgrind_mode == ""
+            and not specs.isValid("expect_err")
+            and len([x for x in filter(lambda x: x in runner_output, specs["errors"])])
+            > 0
```
zachmprince, line 403:
```suggestion
            and len(list(filter(lambda x: x in runner_output, specs["errors"])))
```

zachmprince: Or just do list comprehension:
```suggestion
            and len([x for x in specs["errors"] if x in runner_output])
```

zachmprince (commented):
Small suggestion

loganharbour:
Added to https://github.com/idaholab/moose/pull/31935

PR #32256, opened 2026-01-20
MFEM Non-linear support for steady state

SohailSTFC:
## Reason
Need to solve non-linear problems generated by using non-linear/function
coefficients  in the MFEM backend

## Design
Various small changes including: 
Adding non-linear actions to MFEM-kernels (like linear-forms but are integrated every EquationSystem Mult)
Adding Re-assembly of linear, non-linear and Jacobian Forms during the equation system Mult.
Does not yet support complex or time varying systems but may be extendable in the coming branches.
Patches to allow the ComplexEquationSystem to function normally given the non-linear changes 

## Impact
Will add the capacity to solve weakly non-linear steady state problems, and gives kernel extensions that allow for the solutions of strongly non-linear problems using the MFEM backend. 
@alexanderianblair, @karthichockalingam, @nmnobre


karthichockalingam:
The PR closes https://github.com/idaholab/moose/issues/31772

loganharbour:
@SohailSTFC - you have been sent a request on github to join the repository. Once you accept this, your testing on future pushes will be automatically activated.

framework/src/mfem/kernels/MFEMNLDiffusionKernel.C
```diff
@@ -0,0 +1,65 @@
...
+}
+
+mfem::LinearFormIntegrator *
+MFEMNLDiffusionKernel::createNLAIntegrator()
+{
+  mfem::VectorCoefficient & vec_coef =
+      getMFEMProblem().getCoefficients().getVectorCoefficient(name());
+  _product_coef_res = new mfem::ScalarVectorProductCoefficient(_coef, vec_coef);
+  return new mfem::DomainLFGradIntegrator(*_product_coef_res);
+}
```
lindsayad: This is a bit surprising to me. I assumed that we would be adding a `NonlinearFormIntegrator`

lindsayad (changes requested):

framework/include/mfem/equation_systems/EquationSystem.h
```diff
@@ -85,17 +85,21 @@ class EquationSystem : public mfem::Operator
...
+  /// Build non-linear action forms
+  virtual void BuildNonLinearActionForms();
   /// Build bilinear forms (diagonal Jacobian contributions)
   virtual void BuildBilinearForms();
   /// Build mixed bilinear forms (off-diagonal Jacobian contributions)
   virtual void BuildMixedBilinearForms();
   /// Build all forms comprising this EquationSystem
   virtual void BuildEquationSystem();
+  // Reassemble the Jacobian matrix/Operator (Assuming Gridfunction dependant Operator)
+  virtual void ReassembleJacobian(mfem::BlockVector & x, mfem::BlockVector & rhs);
```
lindsayad: Why does this take a RHS? Can `x` be passed as `const`?

framework/doc/content/source/mfem/kernels/MFEMKernel.md
```diff
@@ -22,7 +22,9 @@ variable names is the same as the set of trial variable names for a square syste
 
 `MFEMKernel` is a purely virtual base class. Derived classes should override the `createBFIntegrator`
 and/or the `createLFIntegrator` methods to return a `BilinearFormIntegrator` and/or a
-`LinearFormIntegrator` (respectively) to add to the `EquationSystem`.
+`LinearFormIntegrator` (respectively) to add to the `EquationSystem`. Derived classes could also 
+override `createNLAIntegrator` to solve non-linear problems using
```
lindsayad: Need to define acronym within this markdown file

framework/doc/content/source/mfem/equation_systems/EquationSystem.md
```diff
@@ -32,6 +32,18 @@ form. [`mfem::BilinearFormIntegrators`](https://mfem.org/bilininteg/) add contri
 $A_{ij}(\varphi_i, \phi_j)$ and [`mfem::LinearFormIntegrators`](https://mfem.org/lininteg/) add
 contributions to $b_i(\varphi_i)$ when assembled.
 
+[MFEMKernels](source/mfem/kernels/MFEMKernel.md) can also contribute to domain integrators for 
+non-linear actions. This allows to form the residual $\mathcal{L}(u)$ for non-linear Newton's 
+methood as shown below
```
lindsayad:
```suggestion
method as shown below
```

framework/include/mfem/bcs/MFEMIntegratedBC.h
```diff
@@ -24,6 +24,9 @@ class MFEMIntegratedBC : public MFEMBoundaryCondition
   /// Create MFEM integrator to apply to the RHS of the weak form. Ownership managed by the caller.
   virtual mfem::LinearFormIntegrator * createLFIntegrator() = 0;
 
+  /// Create MFEM integrator to apply non-linear residual form. Ownership managed by the caller.
+  virtual mfem::LinearFormIntegrator * createNLAIntegrator() { return nullptr; };
```
lindsayad: I don't understand why a nonlinear action integrator API is returning a `LinearFormIntegrator` and not a `NonlinearFormIntegrator`. Need to see concrete design documentation motivating this choice for us to go forward with it. We are using nonlinear forms in https://github.com/mfem/mfem/pull/5142 and I had planned for that miniapp to be wrapped by MOOSE-MFEM framework code as a foundation for a new solid mechanics module implementation

lindsayad (changes requested):

framework/include/mfem/bcs/MFEMIntegratedBC.h
karthichockalingam: Using ```LinearFormIntegrator``` is primarily advantageous because it gives you access to the full range of MFEM’s linear integrators, whereas ```NonlinearFormIntegrator``` supports only a much smaller and more restrictive set. In addition, ```LinearFormIntegrator``` lets you supply a ```GridFunction``` as a nonlinear coefficient and update it inside each nonlinear iteration. That ability to refresh coefficients dynamically makes the overall formulation far more flexible and better suited to problems where material properties, source terms, or other coefficients evolve during the solve.

framework/src/mfem/kernels/MFEMNLDiffusionKernel.C
karthichockalingam: Answered below.

framework/doc/content/source/mfem/equation_systems/EquationSystem.md
karthichockalingam: resolved.

framework/include/mfem/equation_systems/EquationSystem.h
karthichockalingam: Because the MFEM’s FormLinearSystem() which is called within the function to assemble the system matrix modifies both the solution vector x and the right‑hand side rhs, so the routine must take a non‑const x and a non‑const rhs.

framework/include/mfem/bcs/MFEMIntegratedBC.h
lindsayad: > Using `LinearFormIntegrator` is primarily advantageous because it gives you access to the full range of MFEM’s linear integrators, whereas `NonlinearFormIntegrator` supports only a much smaller and more restrictive set.

The set of nonlinear form integrators could be increased with time. It probably wouldn't be hard to ask something like Codex to perform a mapping

> In addition, `LinearFormIntegrator` lets you supply a `GridFunction` as a nonlinear coefficient and update it inside each nonlinear iteration.

How does `LinearFormIntegrator` innately let you do this any better than the `NonlinearFormIntegrator`? There are no references to `GridFunction` within the `LinearFormIntegrator` class definition itself. 

I'd feel a lot more comfortable with an argument based on math as opposed to current breadth of implementation. Presumably the MFEM team has created the `NonlinearFormIntegrator` for a reason. It just still feels innately wrong to me to use a `LinearFormIntegrator` for integrating nonlinear forms when there is a `NonlinearFormIntegrator`. @nmnobre @alexanderianblair you are the other code-owners for these files. What are your thoughts?

karthichockalingam: > The set of nonlinear form integrators could be increased with time. It probably wouldn't be hard to ask 
> something like Codex to perform a mapping

If the range of ```NonlinearFormIntegrator``` options in MFEM could be expanded, that could offer a meaningful advantage over relying on ```LinearFormIntegrator```, especially when the operator or coefficients depend on the solution itself. ```NonlinearFormIntegrator``` is already designed to participate fully in MFEM’s nonlinear solve pipeline: it updates coefficients at each nonlinear iteration, incorporates constraint matrices directly (without requiring the management of prolongation, restriction, or essential boundary conditions), and is structured to operate across multiple FiniteElementSpace objects when assembling coupled systems.

lindsayad: We are also adding additional modern assembly level support in https://github.com/mfem/mfem/pull/5142 to the nonlinear forms as that is very helpful in doing things like p-multigrid where you only want partial assembly on fine levels but may want full assembly on the coarse level (e.g if doing AMG)

lindsayad: Would you want to open an issue in MFEM for the `LinearFormIntegrators` that you'd want to see present as `NonlinearFormIntegrators`? I'd be happy to work on that ticket if you'd like the support. Adding nonlinear support to MOOSE-MFEM is critical to our U.S. DOE-NE program  as we are planning to build a new solid mechanics module implementation on top of MFEM-NEML2. So I definitely want to assist you in your efforts here

karthichockalingam: I like the overall direction you’re proposing, though I think there’s an alternative way to look at it.

Right now, ```BilinearFormIntegrator``` inherits from ```NonlinearFormIntegrator```, but ```LinearFormIntegrator``` does not. My view is that neither ```BilinearFormIntegrator``` nor ```LinearFormIntegrator``` should inherit from ```NonlinearFormIntegrator```.

Instead, ```NonlinearFormIntegrator``` could be constructed using instances of both ```BilinearFormIntegrator``` and ```LinearFormIntegrator```, which it would then use to implement ```NonlinearFormIntegrator::AssembleElementGrad``` and ```NonlinearFormIntegrator::AssembleElementVector```.  The same idea would apply to face integrators. 

If this approach proves workable, a similar pattern could be applied to ```BlockNonlinearFormIntegrator``` for coupled or mixed systems.

lindsayad: I disagree with that because in general a nonlinear operator is not representable by a matrix whereas it is presupposed in a bilinear form that you can represent your operator as such. It is only by assuming some kind of linearization that at each nonlinear iteration you can represent the nonlinear operator as a matrix. To me the current class hierarchy in MFEM makes a lot of sense as a bilinear term is a nonlinear term with more specific structure/constraints. I was actually thinking that by analogy it wouldn't be so bad for `LinearFormIntegrator` to be a subclass of `NonlinearFormIntegrator`. Tagging @tzanio in case he has any thoughts on this discussion thread or wants to pay attention; feel free to unsubscribe otherwise 😄

alexanderianblair: Just adding a couple of my thoughts here...

> I don't understand why a nonlinear action integrator API is returning a LinearFormIntegrator and not a NonlinearFormIntegrator

I think this is primarily to avoid a more extensive overhaul of `EquationSystem` to include `NonlinearForm`s. It doesn't look straightforward to replace current (mixed) `BilinearForm` objects with `NonlinearForm` in `EquationSystem` - not least because mixed form contributions would require use of a `BlockNonlinearForm` instead, which does not support the addition of `NonlinearFormIntegrator` contributions anyway.

That said, there would be nothing to stop the inclusion of `NonlinearForm`s alongside existing bilinear and linear forms; kernels could return any (or all) of a `NonlinearFormIntegrator`, `(Mixed)BilinearFormIntegrator`, and `LinearFormIntegrator`, such that `EquationSystem::Mult` adds contributions from the actions of `NonlinearForm` and `LinearForm` components, and  `EquationSystem::GetGradient` returns an operator with contributions from both `NonlinearForm::GetGradient` and `(Mixed)BilinearForm` components. We will want to support addition of `NonlinearFormIntegrator` contributions regardless, and this route should avoid unnecessary bifurcation of `EquationSystem`.

In terms of larger `EquationSystem` overhauls - we will certainly need a new `DifferentiableEquationSystem` to support the addition of dFEM-style nonlinear integrators down the line, which will be necessary for AD (and, I think, are syntactically simpler long-term for MOOSE users to specify when writing their own kernels/integrators). But this is effectively a parallel system, with MFEM ADKernels.

> We are also adding additional modern assembly level support in https://github.com/mfem/mfem/pull/5142 to the nonlinear forms as that is very helpful in doing things like p-multigrid where you only want partial assembly on fine levels but may want full assembly on the coarse level (e.g if doing AMG)

Very interested in this - the extensions to `NonlinearForm`'s assembly levels also reduce the gap between supported features between `BilinearForm` and `NonlinearForm`. (More generally, I'd like to be able to replace a `BilinearForm` entirely with a `NonlinearForm` with only `BilinearFormIntegrators` added, since it would make our assembly much more readable...)

> Right now, BilinearFormIntegrator inherits from NonlinearFormIntegrator, but LinearFormIntegrator does not. My view is that neither BilinearFormIntegrator nor LinearFormIntegrator should inherit from NonlinearFormIntegrator.

I also disagree with this for the reasons Alex has said - a `BilinearFormIntegrator` (and a `LinearFormIntegrator`) is a `NonlinearFormIntegrator`, and makes sense to be a subclass. Could be possible to create a specialised (derived class from) `NonlinearFormIntegrator` that is formed of a `BilinearFormIntegrator` and `LinearFormIntegrator`, if that is what you mean, with a reduction of generality?

karthichockalingam: > Could be possible to create a specialised (derived class from) NonlinearFormIntegrator that is formed of a
 > BilinearFormIntegrator and LinearFormIntegrator, if that is what you mean, with a reduction of generality?

Yes, this is what I meant to say, but I didn't word it properly. 

> I was actually thinking that by analogy it wouldn't be so bad for LinearFormIntegrator to be a subclass of
> NonlinearFormIntegrator. 

That sounds good.

SohailSTFC: My envisioning of nonlinear (and eventually DFEM) equation system was to use residual and Jacobian forms (As in Finite elements we usual try to minimise a norm of the residual and we don't usually use Halley's method often only going as far as Newtons method) as oppose to linear and bilinear forms, the residual can simply be thought of as a series of block-linear-form-integrators contracted against a set of possibly non-linear coefficients, the use of linear-forms was simply out of convenience cause it was the closest to what I wanted to achieve. 

The coefficient Is where the non-linearity lies, cause eventually from enough chain rule you just end up with a possibly tensor nonlinear coefficient contracted against some nth-order-derivative of the FE-basis function on a particular block of the residual component you are dealing with. As the basis functions (and derivatives) are already all in Linear forms, the difficulty only comes from the definition of the coefficients themselves. The only change I would possible make to the linear form class to turn it into a residual form class would be giving it a vector/block-vector argument and making it an operator.

It also makes forming residuals for mixed problems convenient, this is because the coefficient of a given block can contain functions of coupled sampled variables at the integration points.

karthichockalingam: > Would you want to open an issue in MFEM for the LinearFormIntegrators that you'd want to see present as NonlinearFormIntegrators? 

I think it would be a helpful addition. I am okay to open an issue in MFEM. What about forming the Jacobian matrix via ```NonlinearFormIntegrator::AssembleElementGrad```? Would it be made optional? 

I think it fits with our current design in ```EquationSystem```.  @alexanderianblair, any thoughts on this?

lindsayad: > What about forming the Jacobian matrix via `NonlinearFormIntegrator::AssembleElementGrad`? Would it be made optional?

For gradient-free nonlinear solvers, I would not expect that method to ever be called. For Newton's method, presumably you would want the method implemented or you would want the modern assembly methods `AssembleGradPA` (and after mfem/mfem#5142 `AssembleGradEA`) implemented

alexanderianblair: > That said, there would be nothing to stop the inclusion of `NonlinearForm`s alongside existing bilinear and linear forms; kernels could return any (or all) of a `NonlinearFormIntegrator`, `(Mixed)BilinearFormIntegrator`, and `LinearFormIntegrator`, such that `EquationSystem::Mult` adds contributions from the actions of `NonlinearForm` and `LinearForm` components, and `EquationSystem::GetGradient` returns an operator with contributions from both `NonlinearForm::GetGradient` and `(Mixed)BilinearForm` components. We will want to support addition of `NonlinearFormIntegrator` contributions regardless, and this route should avoid unnecessary bifurcation of `EquationSystem`.

Just updating this thread following discussions with @karthichockalingam and @SohailSTFC - I've added a couple of commits rewriting `EquationSystem` and `MFEMKernel` to use `NonlinearForm` and `NonlinearFormIntegrator, and updated the newly added kernel to use a `NonlinearFormIntegrator`. 

Current concerns:
- The Jacobian is currently rebuilt in `EquationSystem::Mult`, rather than `EquationSystem::GetGradient` - this is confusing, since there should be no need to build/rebuild the jacobian unless the gradient is requested.
- The current nonlinear kernel is inconsistent with the documentation (and fragile), since the documentation claims it is associated with the domain integral (k(u) grad u, grad v), whereas the kernel itself implements (u grad u, grad v). In the test input file, k(u) = u is specified, but any other choice will result in an inconsistent Jacobian.
- This currently lacks coverage for a) nonlinear boundary integrators, and b) running the nonlinear tests in parallel. 

I will add a comment here once these issues are resolved

alexanderianblair: Updates:
- `EquationSystem` now supports `NonlinearFormIntegrator` contributions for both domain and boundary integrators, in addition to existing `BilinearFormIntegrator` and `LinearFormIntegrator` contributions; the latter two are assumed to remain independent of trial variables and are not re-assembled every call to `EquationSystem::Mult`. The main limitation of this design comes from limitations in `NonlinearForm`, in that `NonlinearForm::GetGradient` will only return on-diagonal Jacobian contributions. This is a problem for the use of gradient-based solvers for problems with nonlinear terms of more than one (trial) variable, but otherwise should support mixed systems with nonlinear on-diagonal terms.

- Similarly, `MFEMKernel` and `MFEMIntegratedBC` now support returning `NonlinearFormIntegrator` contributions. Two new `NonlinearFormIntegrator` classes have been added in this PR, with associated `MFEMKernel` classes wrapping them: 
  - `Moose::MFEM::NLBoundaryConvectiveHeatFluxIntegrator`, to enable the solution of eg. thermal problems including boundary conditions involving a temperature-dependent heat transfer coefficient
  - `Moose::MFEM::NLDiffusionIntegrator`, which implements the original design of the domain integrator described in the documentation of this PR, for diffusion problems involving a non-linear diffusion coefficient `k(u)`. This should now take in coefficients for $k(u)$ and $\frac{\partial k}{\partial u}$, as per the original documentation, without being restricted to $k(u)=u$ as in the as-opened version of this PR. 

- The new integrators and `EquationSystem` updates have not been generalised for execution on GPU in this PR (and are not expected to be performant as-written); integrators are built as composites of existing MFEM integrators with non-linear coefficients, lack (non-legacy) assembly options, and `EquationSystem` currently uses them to populate a Jacobian MFEM operator in `HypreParMatrix` format rather than as a generalised `mfem::Operator`. 

- Transient nonlinear problems are now supported with relatively minimal extensions, thanks to earlier refactoring of `Moose::MFEM::TimeDependentEquationSystem` to use a state-state solve.

- The following tests have been added:
  - A heat transfer problem using `Moose::MFEM::NLBoundaryConvectiveHeatFluxIntegrator`, reproducing the results of [this](https://github.com/idaholab/moose/blob/next/modules/heat_transfer/test/tests/convective_heat_flux/coupled.i) MOOSE test from the heat transfer module including a BC with a temperature dependent heat transfer coefficient. Tested in serial and parallel.
  - A linearised version of the above problem using the same BC (which does indeed solve in one nonlinear iteration per timestep). Tested in serial and parallel.
  - A non-linear heat conduction problem based on [MFEM Example 16](https://mfem.org/examples/), testing support for transient problems where both bilinear and non-linear domain integrators are present. Tested in serial and parallel.
  - Some clean-up of the original non-linear diffusion problem test provided in the as-opened version of this PR, checking usage when linear forms are present in addition to non-linear domain integrators.

- Minor updates to generalise the type of MFEM `IterativeSolver` used to perform nonlinear iterations have been added; the only place now where this is specified to be an `mfem::NewtonSolver` is in `MFEMProblem::addMFEMNonlinearSolver`. Generalising this to allow this to be user-specified is to be the topic of a follow-up PR. 

Thanks to @karthichockalingam for helping with various sanity checks during this - I will recuse myself from further review, as I've now made substantive changes to this PR!

lindsayad: > * The main limitation of this design comes from limitations in `NonlinearForm`, in that `NonlinearForm::GetGradient` will only return on-diagonal Jacobian contributions.

Is there an open MFEM issue for this?

alexanderianblair: It's not an error, per se, since `NonlinearForm` is defined with respect to a single FESpace, and thus only can represent the contribution of an on-diagonal block.

For the general block case, MFEM has [`BlockNonlinearForm`](https://github.com/mfem/mfem/blob/dd931b2584f88a5ea905bd1532b24fdbe6401cde/fem/nonlinearform.hpp#L256), which is utilised by codes such as [Smith](https://serac.readthedocs.io/en/latest/) and Serac, which supports definition of the operator with respect to an ordered set of FESpaces. However, this mandates the use of `BlockNonlinearFormIntegrators`, which also are defined with respect to the owning ordering of FESpaces in the `BlockNonlinearForm`. It is not possible to add a `NonlinearFormIntegrator` to a `BlockNonlinearForm`.

As a result, I expect we will need a successor to `BlockNonlinearForm` to do this, which is probably outside the scope of this PR. It should be noted that the implementation in `EquationSystem`  in this PR will still support mixed systems though, when either a) all off-diagonal contributions are linear (expressed as `MixedBilinearFormIntegrators`), or b) when a nonlinear `IterativeSolver` is used that does not require gradient information.

lindsayad (approved):
I have some changes I want to make, but I think it's easier to work on those post this being merged

lindsayad:
follow-on at #32596

loganharbour:
Broke valgrind on devel: https://civet.inl.gov/job/3705079/

```
[44.08s] [1244MB]     FAIL mfem/kernels.MFEMNonLinearDiffusion FAILED (MEMORY ERROR)
[44.36s] [1248MB]     FAIL mfem/kernels.MFEMNonLinearDiffusionNonZeroRHS FAILED (MEMORY ERROR)
[48.71s] [1322MB]     FAIL mfem/kernels.MFEMNonLinearHeatTransfer/non-linear FAILED (MEMORY ERROR)
[67.35s] [1320MB]     FAIL mfem/kernels.MFEMNonLinearHeatConduction FAILED (MEMORY ERROR)
```

Please address ASAP

alexanderianblair:
> Broke valgrind on devel: https://civet.inl.gov/job/3705079/
> 
> ```
> [44.08s] [1244MB]     FAIL mfem/kernels.MFEMNonLinearDiffusion FAILED (MEMORY ERROR)
> [44.36s] [1248MB]     FAIL mfem/kernels.MFEMNonLinearDiffusionNonZeroRHS FAILED (MEMORY ERROR)
> [48.71s] [1322MB]     FAIL mfem/kernels.MFEMNonLinearHeatTransfer/non-linear FAILED (MEMORY ERROR)
> [67.35s] [1320MB]     FAIL mfem/kernels.MFEMNonLinearHeatConduction FAILED (MEMORY ERROR)
> ```
> 
> Please address ASAP

Fix in https://github.com/idaholab/moose/pull/32543

PR #32266, opened 2026-01-25
Fix heavy test with gallium solidification and physics that was timing out

GiudGiud:
made the physics test the same as the non-physics, removing differences that were either just numerical / time stepping parameters and an actual mistake

improved some setup messages and ordering in the flow physics

refs https://github.com/idaholab/moose/issues/30827 https://github.com/idaholab/moose/issues/27555

Note:
I got started on using a functor material to avoid lagging the liquid fraction but the convergence is really really bad. If we could get it to work we would then have a fully implicit solve. Since it's not working, I did not include this.


grmnptr (approved):
Looks good to me!

grmnptr:
And thanks for the fix!

PR #32270, opened 2026-01-27
Use `libmesh_terminate()` for `mooseError()`

loganharbour:
WIP

### DIFF
--- a/framework/include/problems/FEProblemBase.h
+++ b/framework/include/problems/FEProblemBase.h
@@ -510,6 +510,21 @@ public:
   virtual void setException(const std::string & message);
 
   /**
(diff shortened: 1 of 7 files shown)


dschwen:
Shouldn't this be a one line change ;-)

Can we make an option to opt out? Having the backtrace is sometimes nice to have.

loganharbour:
> Shouldn't this be a one line change ;-)
> 
> Can we make an option to opt out? Having the backtrace is sometimes nice to have.

You still get the backtrace

dschwen:
> You still get the backtrace

Ok, cool, I guess we don't need to make the segfault opt-in :-D

PR #32275, opened 2026-01-28
Kokkos ParsedFunction

NamjaeChoi:
Refs #30655.


dschwen:
I suggest using the PEG parser already shipped with MOOSE to parse the functions.

dschwen:
Alternatively you can use FParser (and get algebraic optimization) and access the tape and transform that into GPU operations.

dschwen:
Here is an example for a GPU function parser (that generates torch compute graphs):

https://github.com/idaholab/marlin/blob/devel/include/utils/MarlinExpressionParser.h
https://github.com/idaholab/marlin/blob/devel/src/utils/MarlinExpressionParser.C

and uses `peg::parser` (grammar is at the bottom of the .h file).

NamjaeChoi:
> access the tape

How do I access the tape? Using FParser will be probably better for consistency.

NamjaeChoi:
PEG parser was adopted.

framework/src/kokkos/functions/KokkosFunctionParser.K
```diff
@@ -0,0 +1,276 @@
...
+               # Identifiers
+               Identifier      <- [a-zA-Z_] [a-zA-Z0-9_]*
+
+               # Numbers (integer/real/scientific)
+               Number          <- (# .5 or .5e-2
+                                   '.' [0-9]+ ([eE] [+-]? [0-9]+)?
+                                   /
+                                   # 2. or 2.0 or 42 or with exponent
+                                   [0-9]+ ('.' [0-9]*)? ([eE] [+-]? [0-9]+)?)
+            )"),
```
NamjaeChoi: Used Copilot to generate this PEG grammar.

NamjaeChoi:
This is not urgent. Take your time

lindsayad:
@rochi00 for review

framework/src/kokkos/functions/KokkosParsedFunction.K
```diff
@@ -0,0 +1,86 @@
...
+  const auto & symbol_values = getParam<std::vector<std::string>>("symbol_values");
+
+  for (const auto i : make_range(symbol_names.size()))
+  {
+    if (hasPostprocessorByName(symbol_values[i]))
+      _builder->associateValue(symbol_names[i], &getPostprocessorValueByName(symbol_values[i]));
+    else if (MooseUtils::parsesToReal(symbol_values[i]))
+    {
+      _symbol_values.push_back(MooseUtils::convert<Real>(symbol_values[i]));
+      _builder->associateValue(symbol_names[i], &_symbol_values.back());
```
lindsayad, line 76: This is not a reliable pattern. If `push_back` exceeds the capacity of the container then the `vector` will reallocate and all your pointers will be invalidated

lindsayad: If you just `reserve` based on the size of `symbol_names` prior to the loop, everything should be fine

framework/include/kokkos/functions/KokkosFunctionParser.h
```diff
@@ -0,0 +1,528 @@
...
+  /**
+   * Get AST
+   * @returns The AST
+   */
+  auto ast() const { return _ast; }
+  /**
+   * Get input expression
+   * @returns The input expression
+   */
+  const auto & expression() const { return _expression; }
```
lindsayad: Please write out the return type to improve readability of the public APIs

framework/include/kokkos/functions/KokkosFunctionParser.h
```diff
@@ -0,0 +1,528 @@
...
+   */
+  peg::parser _parser;
+  /**
+   * Abstract Syntax Tree (AST)
+   */
+  std::shared_ptr<peg::Ast> _ast;
+  /**
+   * Input expression
+   */
+  const std::string & _expression;
```
lindsayad: I would just save this by-value so that we don't have to take it on faith that the reference will always be valid

framework/src/kokkos/functions/KokkosParsedFunction.K
```diff
@@ -0,0 +1,86 @@
...
+registerKokkosFunction("MooseApp", KokkosParsedFunction);
+
+InputParameters
+KokkosParsedFunction::validParams()
+{
+  InputParameters params = FunctionBase::validParams();
+  params += MooseParsedFunctionBase::validParams();
+  params.addCustomTypeParam<std::string>(
+      "expression", "FunctionExpression", "The user defined function.");
+  params.addParam<bool>("print_rpn", "Print parsed Reverse Polish Notation (RPN) for debugging.");
```
lindsayad: You don't have a default value here and you call `getParam<bool>("print_rpn")` on line 43 without checking `isParamValid`

framework/src/kokkos/functions/KokkosFunctionParser.K
```diff
@@ -0,0 +1,319 @@
...
+unsigned int
+RPNBuilder::addVariable(const std::string & name)
+{
+  finalizedError();
+
+  auto it = _variables.find(name);
+  if (it != _variables.end())
+    return it->second.idx;
+
+  unsigned int idx = _variables.size() + (_has_default_variables ? 0 : 4);
```
lindsayad, line 214: A reminder comment about what the `4` offset corresponds to would be good (x/y/z/t)

framework/include/kokkos/functions/KokkosFunctionParser.h
```diff
@@ -0,0 +1,528 @@
...
+  /**
+   * Print RPN sequence for debugging
+   * @param console The console object
+   */
+  void printRPN(const ConsoleStream & console) const;
+  /**
+   * Get RPN sequence
+   * @returns The RPN sequence
+   */
+  const auto & getRPN() const { return _rpn; }
```
lindsayad: Not `auto` please

framework/include/kokkos/functions/KokkosFunctionParser.h
```diff
@@ -0,0 +1,528 @@
...
+  /**
+   * Get numbers used in the expression
+   * @returns The numbers
+   */
+  const auto & getNumbers() const { return _numbers; }
+  /**
+   * Get variables used in the expression
+   * @returns The variables
+   */
+  const auto & getVariables() const { return _variables; }
```
lindsayad: No `auto`

framework/include/kokkos/userobjects/KokkosElementUserObject.h
```diff
@@ -11,6 +11,7 @@
 
 #include "KokkosUserObject.h"
 #include "KokkosElementReducer.h"
+#include "KokkosMaterialPropertyValue.h"
```
lindsayad: why is this needed?

framework/include/kokkos/userobjects/KokkosNodalUserObject.h
```diff
@@ -11,6 +11,7 @@
 
 #include "KokkosUserObject.h"
 #include "KokkosNodalReducer.h"
+#include "KokkosMaterialPropertyValue.h"
```
lindsayad: ?

framework/include/kokkos/userobjects/KokkosSideUserObject.h
```diff
@@ -11,6 +11,7 @@
 
 #include "KokkosUserObject.h"
 #include "KokkosSideReducer.h"
+#include "KokkosMaterialPropertyValue.h"
```
lindsayad: ?

framework/src/kokkos/functions/KokkosFunctionParser.K
```diff
@@ -0,0 +1,319 @@
...
+  }
+}
+
+void
+RPNBuilder::printRPN(const ConsoleStream & console) const
+{
+  console << std::endl;
+
+  for (const auto & instruction : _rpn)
+    console << instruction.text << " ";
```
lindsayad: `MooseUtils::join` likely could be used to remove the trailing whitespace

framework/src/kokkos/functions/KokkosFunctionParser.K
```diff
@@ -0,0 +1,319 @@
...
+RPNBuilder::builderError(const peg::Ast & ast, const std::string & message) const
+{
+  std::string arrow(ast.column, ' ');
+  arrow.back() = '^';
+
+  mooseError("Kokkos parsed function error: ", message, "\n", _parser.expression(), "\n", arrow);
+}
+
+void
+RPNBuilder::finalizedError()
```
lindsayad: I think `checkFinalized` would be a better name. Calling `finalizedError` at the top of every method sounds like I'm asking for an error every time

test/include/kokkos/kernels/KokkosParsedFuncCoefDiffusion.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+template <typename Derived>
+KOKKOS_FUNCTION Real
+KokkosParsedFuncCoefDiffusion::computeQpResidual(const unsigned int i,
+                                                 const unsigned int qp,
+                                                 AssemblyDatum & datum) const
+{
+  const auto & func = static_cast<const KokkosParsedFunction &>(_function);
+  Real k = func.value(_t, datum.q_point(qp));
```
lindsayad:
```suggestion
  const Real k = func.value(_t, datum.q_point(qp));
```

test/include/kokkos/kernels/KokkosParsedFuncCoefDiffusion.h
```diff
@@ -0,0 +1,57 @@
...
+
+template <typename Derived>
+KOKKOS_FUNCTION Real
+KokkosParsedFuncCoefDiffusion::computeQpJacobian(const unsigned int i,
+                                                 const unsigned int j,
+                                                 const unsigned int qp,
+                                                 AssemblyDatum & datum) const
+{
+  const auto & func = static_cast<const KokkosParsedFunction &>(_function);
+  Real k = func.value(_t, datum.q_point(qp));
```
lindsayad:
```suggestion
  const Real k = func.value(_t, datum.q_point(qp));
```

lindsayad:
@NamjaeChoi let me know your thoughts on the commit I just pushed

NamjaeChoi:
> @NamjaeChoi let me know your thoughts on the commit I just pushed

Let's see how the test goes

framework/include/kokkos/userobjects/KokkosElementUserObject.h
NamjaeChoi: I haven't figured out this yet, but NVCC complains on link step when I don't have this header and someone uses material property in their derived class without including it. Host compiler doesn't. Btw this commit was moved to another PR and will be dropped from here.

NamjaeChoi:
Claude seems clever. Might be time for me to retire

lindsayad:
> Claude seems clever. Might be time for me to retire

Haha it's very clever. But it also tries to do stupid things sometimes too. And if left to its own devices it drifts from clear high-level design. Maybe 5 years from now we'll be out of jobs but at present it still needs a human in the loop

lindsayad (approved):

PR #32279, opened 2026-01-28
Add `-march` and `-mtune` to container builds

loganharbour:
Closes #32277


GiudGiud:
A lot of these tests are already flagged as "not working on arm macs".
I think most will have a design issue we can fix. 

I can fix them here or make separate PRs, as you want. 
There is no option to "skip them and leave it for the next infrastructure change" this time right?

loganharbour:
Replaced with https://github.com/idaholab/moose/pull/33331

PR #32282, opened 2026-01-28
Represent regions in CSGBase using postfix notation

shikhar413:
refs #32281 and #32272


shikhar413:
@eshemon @kkiesling FYI

kkiesling:
@lefebvrera here is the RPN/postfix notation for the regions in the CSG output files. Can you check the test output gold files for the `region_postfix` and confirm this is what you wanted this to be to make the Shift connection easier?

shikhar413:
@GiudGiud this PR should be ready to review. We represent regions using postfix notation (reverse polish notation) now, and the output CSG JSON file will contain cell region representations using posfilx notation and infix notation (how regions were represented before). This was a feature request from ORNL Shift team

framework/doc/content/syntax/CSG/index.md
```diff
@@ -92,18 +92,20 @@ Below is an example [!ac](JSON) surface output for a model with a `CSG::CSGPlane
...
-| union              | `|`                   |
-| intersection       | `&`                   |
-| complement         | `~`                   |
+| Boolean Operator   | Infix String Representation | Postfix String Representation |
+|--------------------|-----------------------------|-------------------------------|
+| positive half-space | `+`                   |   `halfspace::POSITIVE`            |
+| negative half-space | `-`                   |   `halfspace::NEGATIVE`            |
+| union              | `|`                   |    `regionOperator::UNION`          |
+| intersection       | `&`                   |    `regionOperator::INTERSECTION`   |
+| complement         | `~`                   |    `regionOperator::COMPLEMENT`     |
```
loganharbour:
```suggestion
| Boolean Operator    | Infix String Representation | Postfix String Representation  |
|---------------------|-----------------------------|--------------------------------|
| positive half-space | `+`                         | `halfspace::POSITIVE`          |
| negative half-space | `-`                         | `halfspace::NEGATIVE`          |
| union               | `|`                         | `regionOperator::UNION`        |
| intersection        | `&`                         | `regionOperator::INTERSECTION` |
| complement          | `~`                         | `regionOperator::COMPLEMENT`   |
```

Why not be pretty? :)

framework/include/csg/CSGCell.h
```diff
@@ -118,11 +118,22 @@ class CSGCell
...
 
   /**
-   * @brief Get the string representation of the cell region
+   * @brief Get the infix string representation of the cell region
    *
-   * @return std::string string representation of the cell region
+   * @return std::string infix string representation of the cell region
    */
-  const std::string & getRegionAsString() const { return _region.toString(); }
+  const std::string getRegionAsInfixString() const { return _region.toInfixString(); }
```
loganharbour:
```suggestion
  std::string getRegionAsInfixString() const { return _region.toInfixString(); }
```

These consts don't mean anything because it's not a reference

framework/include/csg/CSGCell.h
```diff
@@ -118,11 +118,22 @@ class CSGCell
...
+
+  /**
+   * @brief Get the postfix string representation of the cell region
+   *
+   * @return std::vector<std::string> list of postfix tokens of the cell region in string
+   *                                  representation
+   */
+  const std::vector<std::string> getRegionAsPostfixStringList() const
```
loganharbour:
```suggestion
  std::vector<std::string> getRegionAsPostfixStringList() const
```

framework/include/csg/CSGRegion.h
```diff
@@ -70,11 +78,27 @@ class CSGRegion
   virtual ~CSGRegion() = default;
 
   /**
-   * @brief gets the string representation of the region
+   * @brief gets the infix string representation of the region, which involves converting
+   *        region representation from postfix to infix notation
+   *
+   * @return infix string representation of the region
+   */
+  const std::string toInfixString() const;
```
loganharbour:
```suggestion
  std::string toInfixString() const;
```

framework/include/csg/CSGRegion.h
```diff
@@ -70,11 +78,27 @@ class CSGRegion
...
+
+  /**
+   * @brief gets the list of postfix tokens of the region in string representation
+   *
+   * @return list of postfix tokens that represent the region
+   */
+  const std::vector<std::string> toPostfixStringList() const;
```
loganharbour:
```suggestion
  std::vector<std::string> toPostfixStringList() const;
```

framework/include/csg/CSGRegion.h
```diff
@@ -70,11 +78,27 @@ class CSGRegion
...
+
+  /**
+   * @brief converts postfix token from PostfixTokenVariant to string representation
    *
-   * @return string representation of the region
+   * @param token postfix token of type PostfixTokenVariant
+   * @return string representation of postfix token
    */
-  const std::string & toString() const { return _region_str; }
+  const std::string postfixTokenToString(const PostfixTokenVariant & token) const;
```
loganharbour:
```suggestion
  std::string postfixTokenToString(const PostfixTokenVariant & token) const;
```

framework/src/csg/CSGRegion.C
```diff
@@ -59,19 +66,145 @@ CSGRegion::CSGRegion(const CSGRegion & region, const std::string & region_type)
...
+  }
+
+  // Top of stack should now have region string we desire
+  return postfix_stack.top();
+}
+
+const std::vector<std::string>
+CSGRegion::toPostfixStringList() const
+{
+  std::vector<std::string> postfix_string_list;
```
loganharbour, line 182:
```suggestion
  std::vector<std::string> postfix_string_list;
  postfix_string_list.reserve(_postfix_tokens.size());
```

So that we don't end up with extra capacity in this vector

framework/src/csg/CSGRegion.C
```diff
@@ -59,19 +66,145 @@ CSGRegion::CSGRegion(const CSGRegion & region, const std::string & region_type)
...
+
+bool
+CSGRegion::nextRegionOpIsIdentical(const std::string & region_op_string,
+                                   unsigned int postfix_token_index) const
+{
+  for (unsigned int i = postfix_token_index; i < _postfix_tokens.size(); ++i)
+  {
+    const auto & postfix_token_string = postfixTokenToString(_postfix_tokens[i]);
+    if (postfix_token_string.find("regionOperator::") != std::string::npos)
+      return (postfix_token_string == region_op_string);
```
loganharbour:
```suggestion
  for (const auto & token : _postfix_tokens)
  {
    const auto postfix_token_string = postfixTokenToString(token);
    if (postfix_token_string.find("regionOperator::") != std::string::npos)
      return (postfix_token_string == region_op_string);
```

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
```diff
@@ -3,42 +3,240 @@
...
+      ]
     },
     "pin1_void_cell": {
       "fill": "",
       "filltype": "VOID",
-      "region":
-          "((((~(+pin1_radial_duct_2_surf_0 & +pin1_radial_duct_2_surf_1 & +pin1_radial_duct_2_surf_2 & +pin1_radial_duct_2_surf_3 & +pin1_radial_duct_2_surf_4 & +pin1_radial_duct_2_surf_5) & +pin1_axial_plane_0) & -pin1_axial_plane_2) | -pin1_axial_plane_0) | +pin1_axial_plane_2)"
+      "region_infix":
+          "((~(+pin1_radial_duct_2_surf_0 & +pin1_radial_duct_2_surf_1 & +pin1_radial_duct_2_surf_2 & +pin1_radial_duct_2_surf_3 & +pin1_radial_duct_2_surf_4 & +pin1_radial_duct_2_surf_5) & +pin1_axial_plane_0 & -pin1_axial_plane_2) | -pin1_axial_plane_0 | +pin1_axial_plane_2)",
+      "region_postfix": [
```
loganharbour, line 208: Is this a standard postfix notation that's used?

framework/src/csg/CSGRegion.C
```diff
@@ -59,19 +66,145 @@ CSGRegion::CSGRegion(const CSGRegion & region, const std::string & region_type)
...
+  // Define maps to conert halfspace and region operator names to their symbols
+  std::map<std::string, std::string> halfspace_to_symbol = {{"halfspace::POSITIVE", "+"},
+                                                            {"halfspace::NEGATIVE", "-"}};
+  std::map<std::string, std::string> region_op_to_symbol = {{"regionOperator::INTERSECTION", "&"},
+                                                            {"regionOperator::UNION", "|"},
+                                                            {"regionOperator::COMPLEMENT", "~"}};
+
+  // Define which region operators are unary. This will set how many pop operations are needed to
+  // build the region string
+  std::set<std::string> unary_region_ops = {"regionOperator::COMPLEMENT"};
```
loganharbour: These standardized strings (both the prefix and the whole value) should be static strings somewhere. I think that it also might be easier to be able to make this an enum that we can convert to a string, but I need to think about that one a bit more

shikhar413: I struggled with this one because ideally we convert the data type itself (CSGRegion::regionOperator and CSGSurface::halfspace) to strings, but these data types are defined as enum classes and I couldn't find a good way to convert this to a string. I'm open to changes to this logic so that we aren't explicitly defining string conversions for these

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
shikhar413: What you're highlighting is the infix string notation, which is the postfix notation converted to infix notation. I would say this format is standardized and should be understood by MC developers. The postfix notation is the list of strings in the output, and I don't believe there is much variability with how postfix is represented

framework/src/csg/CSGRegion.C
loganharbour: I'll clone this real quick to see if I can think of anything

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
```diff
@@ -3,42 +3,240 @@
...
+      ]
     },
     "pin1_cell_radial_1_axial_0": {
       "fill": "rgmb_region_12",
       "filltype": "CSG_MATERIAL",
-      "region":
-          "((+pin1_radial_duct_1_surf_0 & +pin1_radial_duct_1_surf_1 & +pin1_radial_duct_1_surf_2 & +pin1_radial_duct_1_surf_3 & +pin1_radial_duct_1_surf_4 & +pin1_radial_duct_1_surf_5 & ~(-pin1_radial_ring_0)) & +pin1_axial_plane_0 & -pin1_axial_plane_1)"
+      "region_infix":
+          "((+pin1_radial_duct_1_surf_0 & +pin1_radial_duct_1_surf_1 & +pin1_radial_duct_1_surf_2 & +pin1_radial_duct_1_surf_3 & +pin1_radial_duct_1_surf_4 & +pin1_radial_duct_1_surf_5) & ~(-pin1_radial_ring_0) & +pin1_axial_plane_0 & -pin1_axial_plane_1)",
+      "region_postfix": [
```
lefebvrera, line 38: Instead of using overly verbose 'halfspace::X' or 'regionOperator::X', why not just use the single character operators (&|~+-)? This will be a more succinct RPN notation. 

JSON CSG representation with `regionOperator::X` will make JSON's potentially huge in practice. Alternatively, the given JSON array could be a JSON string:

```json
"region_postfix":"+pin1_radial_duct_1_surf_0 +pin1_radial_duct_1_surf_1 & +pin1_radial_duct_1_surf_2 & +pin1_radial_duct_1_surf_3 & +pin1_radial_duct_1_surf_4 & +pin1_radial_duct_1_surf_5 & -pin1_radial_ring_0 ~ & +pin1_axial_plane_0 & -pin1_axial_plane_1 &"
```

kkiesling: FWIW I suggested to @shikhar413 to not make the postfix notation a single string, and instead make it the list of strings like this, because we don't actually restrict surface names from having spaces in them. So as a single string, it would be potentially hard to determine if a space in a name means a new item in the postfix list or if its just the continuation of the surface name. I also assume that if it's a single string, it would have to be split anyway to be able to process it.

shikhar413: We had discussed this in our meeting in December, but the concern here was that those symbols weren’t universally understood by all MC codes. That was why I opted for explicit strings here so that it wouldn’t be ambiguous at the MC code conversion step what operation these symbols are referring to

shikhar413: I do agree with you @lefebvrera your suggestion is more succinct and I do like this format. Maybe we need to bring this up with the rest of the shift team to see what they prefer? @kkiesling

framework/src/csg/CSGRegion.C
shikhar413: Actually @loganharbour this needs to iterate over a specified index. This is a special function that iterates from postfix_token_index to the end of _postfix_tokens. We don't assume this index starts at 0

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
shikhar413: @tghaddar any opinion on this? Based on our discussion in December, we are trying to represent regions in postfix notation (reverse polish notation). Rob had the suggestion of representing region/surface operators (intersections, half spaces, complements, unions) using symbols for succinctness. But you would rather have these operators represented as strings for easier parsing to convert to shift/titan geometry input correct?

loganharbour: I spoke Tarek some so we could touch base, and here's our combined thoughts:

- Let's just not allow spaces in surfaces; that's a dirty road to go down
- We should represent the infix with arrays so that we don't need to go back and re-parse the parenthesis; something like the following enables the most _general_ output
```
(foo & bar) & -baz -> [["foo", "&", "bar"], "&", "baz"]
```
- The infix won't be used for serialization anyway (save to JSON, load from JSON to CSGBase will just load directly from postfix)
- Restrict the use of operators (+,-,~,&,|) in surface names so we don't get ambiguity
- Store the postfix with the operators instead of the strings (Rob's suggestion)
- For the sake of unit testing, we don't need to store infix; the postfix is the real representation in code anyway so we should really only be storing infix when we explicitly want to test postfix -> infix

For cases of converting to other codes, going from this array representation of infix is as simple as string joining and wrapping nested arrays with `( ... )`

framework/src/csg/CSGRegion.C
```diff
@@ -147,7 +147,7 @@ CSGRegion::toInfixString() const
       else
       {
         const auto region = std::get<RegionType>(token);
-        const auto symbol = regionSymbol(region);
+        const std::string symbol{regionSymbol(region)};
```
shikhar413, line 148: FYI @loganharbour needed to convert char to string here because the addition operation in `region_string = symbol + "(" + region_string + ")";` was not valid and led to some wonky output results in the PinMeshGenerator CSG tests

framework/src/csg/CSGRegion.C
```diff
@@ -193,7 +193,7 @@ CSGRegion::postfixTokenToString(const PostfixTokenVariant & token) const
 {
   // Lambda function to return all variant types as strings
   return std::visit(
-      [this](auto && arg) -> std::string
+      [](auto && arg) -> std::string
```
shikhar413, line 194: @loganharbour reverted the change here because there was a compiler warning `lambda capture 'this' is not used [-Wunused-lambda-capture]`

framework/src/csg/CSGSurfaceList.C
```diff
@@ -48,12 +48,31 @@ CSGSurface &
...
     mooseError("Surface with name " + surf_name + " already exists in geometry.");
   return *it->second;
 }
 
+void
+CSGSurfaceList::checkValidSurfaceName(const std::string & name)
+{
+  // Check if invalid symbol is present in the surface name. These include whitespaces and symbols
+  // for halfspaces and region operators
+  const std::regex invalid_symbols(R"([\+\-~|& ])");
```
shikhar413: @loganharbour ended up using regex to check validity of surface name. Let me know if there's a better way to do this instead

framework/include/csg/CSGSurfaceList.h
```diff
@@ -92,6 +92,13 @@ class CSGSurfaceList
    */
   CSGSurface & addSurface(std::unique_ptr<CSGSurface> surf);
 
+  /**
+   * @brief check surface name for disallowed characters and symbols
+   *
+   * @param name name of surface to check
+   */
+  void checkValidSurfaceName(const std::string & name);
```
GiudGiud: const?

framework/src/csg/CSGRegion.C
```diff
@@ -59,19 +106,114 @@ CSGRegion::CSGRegion(const CSGRegion & region, const std::string & region_type)
...
   {
-    // no change to surfaces, but update string
-    if (region.toString()[0] == '(')
-      _region_str = "~" + region.toString();
-    else
-      _region_str = "~(" + region.toString() + ")";
     _surfaces = region.getSurfaces();
+    // (complement region) in postfix is represented as (region complement)
+    _postfix_tokens = region.getPostfixTokens();
+    _postfix_tokens.push_back(getRegionType());
```
GiudGiud, line 110: are we missing the complement operator in the tokens?

framework/src/csg/CSGRegion.C
```diff
@@ -167,7 +273,7 @@ bool
 CSGRegion::operator==(const CSGRegion & other) const
 {
   const bool region_type_eq = this->getRegionType() == other.getRegionType();
-  const bool region_str_eq = this->toString() == other.toString();
+  const bool region_str_eq = this->toInfixString() == other.toInfixString();
```
GiudGiud: which one is faster to compare? Building the infix string might be an extra cost no?

framework/src/csg/CSGSurfaceList.C
GiudGiud: no space then right? Is that the character between & and ] ?

shikhar413: Nah checking for whitespace in the surface name as well. It will be more difficult to parse for MC converter if surface names have spaces in them

framework/include/csg/CSGSurfaceList.h
shikhar413: const not needed here?

framework/src/csg/CSGRegion.C
shikhar413: getRegionType contains the complement operator so  _postfix_tokens.push_back(getRegionType()) will accomplish that

framework/src/csg/CSGRegion.C
shikhar413: Yeah you're right. Let me try updating this to postfix

shikhar413: Updated

framework/src/csg/CSGRegion.C
GiudGiud: but like why

framework/include/csg/CSGSurfaceList.h
GiudGiud: const needed everywhere

framework/src/csg/CSGRegion.C
GiudGiud: getRegionType does not really convey "complement"

shikhar413: Yeah initially we created `_region_type` to track what the "outermost" region operator was (halfspace, union, intersection, or empty). In this context we are now using that same variable to set the region operator symbol. They are functionally equivalent now, so maybe we should rename `_region_type` to `_region_operator`?

shikhar413: The `_region_type` variable gets used in other parts of the code to check that regions are being combined properly together (i.e. we aren't trying to apply a complement with two regions as input parameters or we aren't trying to define an intersection with a single region)

framework/include/csg/CSGSurfaceList.h
shikhar413: Oh I see you're talking about the function being const not the input parameter

shikhar413: Updated, let me know if I missed anything else

framework/src/csg/CSGSurfaceList.C
kkiesling: should we add the same check to all the CSG components just to prevent potential issues down the line? (cells, lattices, universes)

shikhar413: Yeah that's a good idea

shikhar413: Any opposition to defining `checkValidCSGName` in `CSGUtils` instead of in `CSGBase`? This allows any CSG[X}List class to call this function directly instead of having the check at the CSGBase level.

shikhar413: Ok this is done now

loganharbour: Fine with me. This looks good

framework/include/csg/CSGCell.h
```diff
@@ -118,11 +118,22 @@ class CSGCell
...
 
   /**
-   * @brief Get the string representation of the cell region
+   * @brief Get the infix string representation of the cell region
    *
-   * @return std::string string representation of the cell region
+   * @return std::string infix string representation of the cell region
    */
-  const std::string & getRegionAsString() const { return _region.toString(); }
+  std::string getRegionAsInfixString() const { return _region.toInfixString(); }
```
loganharbour: Is there a reason this is a separate getter? Can't we just get the region and call `toInfixString` in the serialization?

Same for `getRegionAsPostfixStringList`

framework/src/csg/CSGCellList.C
```diff
@@ -18,6 +19,7 @@ CSGCell &
 CSGCellList::addCell(std::unique_ptr<CSGCell> cell)
 {
   auto name = cell->getName();
+  CSGUtils::checkValidCSGName(name);
```
loganharbour: Can't this just be in the `CSGCell` constructor?

framework/src/csg/CSGLatticeList.C
```diff
@@ -37,6 +38,7 @@ CSGLattice &
 CSGLatticeList::addLattice(std::unique_ptr<CSGLattice> lattice)
 {
   auto name = lattice->getName();
+  CSGUtils::checkValidCSGName(name);
```
loganharbour: Same - in the lattice constructor (and for all other uses)

framework/src/csg/CSGRegion.C
```diff
@@ -9,21 +9,65 @@
 
 #include "CSGRegion.h"
 
+#include "libmesh/string_to_enum.h"
+
```
loganharbour: I think I included this and then didn't use it

```suggestion
```

loganharbour (changes requested):

framework/src/csg/CSGCellList.C
shikhar413: Yeah I think you're right. Initially I thought this check should happen once CSGBase got updated but we should be able to add this check one level earlier

framework/include/csg/CSGCell.h
shikhar413: Yeah, that should work as well. I can remove these methods

framework/src/csg/CSGLatticeList.C
shikhar413: These calls are now moved to constructor

framework/src/csg/CSGCellList.C
shikhar413: Updated

framework/include/csg/CSGCell.h
shikhar413: These methods are now removed

loganharbour (approved):
Good pending success. Waiting on @GiudGiud and @lefebvrera

framework/src/csg/CSGRegion.C
GiudGiud: ok I can see it now. resolving

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
```diff
@@ -3,42 +3,240 @@
...
-      "region": "(-pin1_radial_ring_0 & +pin1_axial_plane_0 & -pin1_axial_plane_1)"
+      "region_infix": "(-pin1_radial_ring_0 & +pin1_axial_plane_0 & -pin1_axial_plane_1)",
+      "region_postfix": [
+        "pin1_radial_ring_0",
+        "-",
+        "pin1_axial_plane_0",
+        "+",
+        "pin1_axial_plane_1",
+        "-",
+        "&",
```
GiudGiud, line 14:
```suggestion
        "pin1_radial_ring_0",
        "-",
        "pin1_axial_plane_0",
        "+",
        "&",
        "pin1_axial_plane_1",
        "-",
```
not sure why it's not following this pattern?

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
```diff
@@ -3,42 +3,240 @@
...
-      "region":
-          "((+pin1_radial_duct_1_surf_0 & +pin1_radial_duct_1_surf_1 & +pin1_radial_duct_1_surf_2 & +pin1_radial_duct_1_surf_3 & +pin1_radial_duct_1_surf_4 & +pin1_radial_duct_1_surf_5 & ~(-pin1_radial_ring_0)) & +pin1_axial_plane_0 & -pin1_axial_plane_1)"
+      "region_infix":
+          "((+pin1_radial_duct_1_surf_0 & +pin1_radial_duct_1_surf_1 & +pin1_radial_duct_1_surf_2 & +pin1_radial_duct_1_surf_3 & +pin1_radial_duct_1_surf_4 & +pin1_radial_duct_1_surf_5) & ~(-pin1_radial_ring_0) & +pin1_axial_plane_0 & -pin1_axial_plane_1)",
+      "region_postfix": [
+        "pin1_radial_duct_1_surf_0",
+        "+",
+        "pin1_radial_duct_1_surf_1",
+        "+",
+        "&",
```
GiudGiud, line 43: like here it's fine

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
```diff
@@ -3,42 +3,240 @@
...
+        "pin1_radial_ring_0",
+        "-",
+        "~",
+        "&",
+        "pin1_axial_plane_1",
+        "+",
+        "pin1_axial_plane_2",
+        "-",
+        "&",
```
GiudGiud, line 99: same here?

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
shikhar413: It’s likely due to how the CSGRegion is being built within PinMeshGenertator::generateCSG. It looks like the two postfix representations correspond to the same physical region, so I don’t believe this is an error. The same region can be represented multiple ways using postfix

GiudGiud: makes sense. Redid a couple manually and they're correct

modules/reactor/test/tests/meshgenerators/pin_mesh_generator/gold/pin_hex_3d_csg.json
GiudGiud: ok all in has been done except `We should represent the infix with arrays so that we don't need to go back and re-parse the parenthesis; `

Resolving and feel free to create an issue if needed

GiudGiud (approved):

shikhar413:
Thanks for the review @GiudGiud and @loganharbour. I created an issue for converting infix representation to list representation instead: https://github.com/idaholab/moose/issues/32299

I'm not familiar with arbitrarily nested lists in C++, so if either of you have insight on how to implement this in C++ on the issue there? Alternatively, if there is another way to represented arbitrarily nested lists in C++, let me know.

loganharbour:
> I'm not familiar with arbitrarily nested lists in C++, so if either of you have insight on how to implement this in C++ on the issue there?

I think we should just directly use `nlohmann::json` for this, since it's an output only thing anyway

